PR #276, opened 2024-03-20
Implement system caller and system txs

### LINKED ISSUE #228: EVM: impl. system transactions
In some rollups, first few transactions of a block is reserved for [system transactions](https://optimistic.etherscan.io/tx/0xc447acaaef5ee907edd813ed7c856fe3cdfb2d1f4f7d37c1341ba17c3d1ee51b). These transactions are exempt from paying gas fees. (But they consume gas)

Our first system contract is under development right now (#227)

We need to implement necessary handlers to apply these transactions.

TODOs:

- [ ] Add pretx handler to CitreaEVM handler.
  - [ ] Everything is the same with mainnet evm pretx handler. But for system contracts no gas fee check is applied and no balance is deduced. (How do i know a tx is a system tx? We can designate a specific address, (optimism uses 0xdeaddead....) which the handler and smart contract looks for. If there's another way, let's discuss)
- [ ] Track last da_root_hash - when it changes in a `begin_soft_confirmation_hook`, execute the system transaction and append to pending_transactions.

kpp:
# Description

This introduces a special signature and a special address (SYSTEM_SIGNATURE and SYSTEM_SIGNER) to handle special transactions to be exempt from paying gas fees.

TODO:
- [x] Handle `if system_caller` in EVM handler
- [x] Don't deduct system caller txs with gas fee
- [x] Make sure no one can send us a tx with `SYSTEM_SIGNATURE`
- [x] Tests

## Linked Issues
- Fixes #228

## Testing

Can't be properly tested until we preload system contracts in genesis.

## Doc

- SYSTEM_SIGNER (and caller) is 0xdeaddeaddeaddeaddeaddeaddeaddeaddeaddead.
- L1BlockHashList addr must be 0x3100000000000000000000000000000000000001

### DIFF
--- a/crates/evm/src/tests/call_tests.rs
+++ b/crates/evm/src/tests/call_tests.rs
@@ -55,6 +55,7 @@ fn call_multiple_test() {
         [42u8; 32],
         &[10u8; 32],
         l1_fee_rate,
+        0,
         &mut working_set,
     );
 
@@ -162,6 +163,7 @@ fn call_test() {
         [42u8; 32],
         &[10u8; 32],
         l1_fee_rate,
+        0,
         &mut working_set,
     );
 
@@ -238,6 +240,7 @@ fn failed_transaction_test() {
         [42u8; 32],
         &[10u8; 32],
         l1_fee_rate,
+        0,
         working_set,
     );
     {
@@ -294,6 +297,7 @@ fn self_destruct_test() {
         [42u8; 32],
         &[10u8; 32],
         l1_fee_rate,
+        0,
         &mut working_set,
     );
     {
@@ -349,6 +353,7 @@ fn self_destruct_test() {
         [42u8; 32],
         &[99u8; 32],
         l1_fee_rate,
+        0,
         &mut working_set,
     );
     {
@@ -425,6 +430,7 @@ fn test_block_hash_in_evm() {
         [42u8; 32],
         &[10u8; 32],
         l1_fee_rate,
+        0,
         &mut working_set,
     );
     {
@@ -455,6 +461,7 @@ fn test_block_hash_in_evm() {
             [42u8; 32],
             &[99u8; 32],
             l1_fee_rate,
+            0,
             &mut working_set,
         );
         evm.end_soft_confirmation_hook(&mut working_set);
@@ -537,6 +544,7 @@ fn test_block_gas_limit() {
         [42u8; 32],
         &[10u8; 32],
         l1_fee_rate,
+        0,
         &mut working_set,
     );
     {
@@ -803,6 +811,7 @@ fn test_l1_fee_success() {
             [42u8; 32],
             &[10u8; 32],
             l1_fee_rate,
+            0,
             &mut working_set,
         );
         {
@@ -877,6 +886,7 @@ fn test_l1_fee_not_enough_funds() {
         [42u8; 32],
         &[10u8; 32],
         l1_fee_rate,
+        0,
         &mut working_set,
     );
     {
--- a/crates/evm/src/tests/genesis_tests.rs
+++ b/crates/evm/src/tests/genesis_tests.rs
@@ -53,8 +53,6 @@ lazy_static! {
             .collect(),
         chain_id: 1000,
         block_gas_limit: reth_primitives::constants::ETHEREUM_BLOCK_GAS_LIMIT,
-        block_timestamp_delta: 2,
-        genesis_timestamp: 50,
         coinbase: Address::from([3u8; 20]),
         limit_contract_code_size: Some(5000),
         starting_base_fee: 1000000000,
@@ -62,10 +60,10 @@ lazy_static! {
     };
 
     pub(crate) static ref GENESIS_HASH: B256 = B256::from(hex!(
-        "5c3afd4e90c378a3807947cfbaab0485031a4a2f647e1c5323777997338775a0"
+        "5aa8e649f234ae2128a8eb8536f557fafd7b6a46f12cba8d89a73eb74ee0f0e3"
     ));
     pub(crate) static ref GENESIS_STATE_ROOT: B256 = B256::from(hex!(
-        "aad642e56d49fbdb0a7b7ad7da81490fde7cb22b3860c77ae92c4c1d37af98b4"
+        "62dcfcbbc789410f13b8ec77a27fa8fc1c804c5cedbf2c2498b886b7c5a2f35b"
     ));
     pub(crate) static ref GENESIS_DA_TXS_COMMITMENT: B256 = B256::from(hex!(
         "4a5e1e4baab89f3a32518a88c31bc87f618f76673e2cc77ab2127b7afdeda33b"
@@ -150,7 +148,6 @@ fn genesis_cfg() {
             spec: vec![(0, SpecId::BERLIN), (1, SpecId::SHANGHAI)],
             chain_id: 1000,
             block_gas_limit: reth_primitives::constants::ETHEREUM_BLOCK_GAS_LIMIT,
-            block_timestamp_delta: 2,
             coinbase: Address::from([3u8; 20]),
             limit_contract_code_size: Some(5000),
             base_fee_params: BaseFeeParams::ethereum(),
@@ -217,7 +214,7 @@ fn genesis_block() {
                     number: 0,
                     gas_limit: ETHEREUM_BLOCK_GAS_LIMIT,
                     gas_used: 0,
-                    timestamp: 50,
+                    timestamp: 0,
                     extra_data: Bytes::default(),
                     mix_hash: B256::default(),
                     nonce: 0,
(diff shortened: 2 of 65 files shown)


crates/evm/src/call.rs
```diff
@@ -28,14 +29,115 @@ pub struct CallMessage {
...
+            cfg_env,
+            &mut citrea_handler_ext,
+        );
+
+        let mut cumulative_gas_used = 0;
+        let mut log_index_start = 0;
+        if let Some(tx) = self.pending_transactions.last(working_set) {
+            cumulative_gas_used = tx.receipt.receipt.cumulative_gas_used;
+            log_index_start = tx.receipt.log_index_start + tx.receipt.receipt.logs.len() as u64;
+        }
```
eyusufatik, line 90: why do we check this?

crates/evm/src/call.rs
```diff
@@ -28,14 +29,115 @@ pub struct CallMessage {
...
+                    signer: tx.signer(),
+                    signed_transaction: tx.into(),
+                    block_number,
+                },
+                receipt,
+            };
+            pending_txs.push(pending_transaction);
+        }
+        for pending_tx in pending_txs {
+            self.pending_transactions.push(&pending_tx, working_set);
```
eyusufatik, line 127: can do this in line 124 right?

crates/evm/src/call.rs
```diff
@@ -28,14 +29,115 @@ pub struct CallMessage {
...
+        }
+
+        // block_env.gas_used += cumulative_gas_used;
```
eyusufatik, line 130: remove comment

crates/evm/src/call.rs
eyusufatik: let's double down here, 

if pending_transactions is not empty, panic with system transactions should be the first tx in block.

cum_gas_used and log_index_start = 0 always in this func.

crates/evm/src/evm/system_events.rs
```diff
@@ -0,0 +1,80 @@
...
+            input: RethBytes::from(
+                sys_block_hash
+                    .set_block_info(block_hash, txs_commitments)
+                    .to_vec(),
+            ),
+            nonce,
+            chain_id,
+            value: U256::ZERO,
+            gas_limit: 1_000_000u64,
+            max_fee_per_gas: u64::MAX as u128,
```
eyusufatik, line 50: what about max fee per gas 0, just to be clear to the reader.

crates/evm/src/tests/genesis_tests.rs
```diff
@@ -291,7 +291,7 @@ pub(crate) fn get_evm(config: &EvmConfig) -> (Evm<C>, WorkingSet<DefaultContext>
     let mut working_set: WorkingSet<DefaultContext> = WorkingSet::new(storage.clone());
     evm.finalize_hook(&root.into(), &mut working_set.accessory_state());
 
-    evm.begin_soft_confirmation_hook([1u8; 32], [2u8; 32], &root, 0, &mut working_set);
+    evm.begin_soft_confirmation_hook([1u8; 32], 1, [2u8; 32], &root, 0, &mut working_set);
     evm.end_soft_confirmation_hook(&mut working_set);
```
eyusufatik, line 295: first system tx happens in this block right?

that's why other tests aren't affected by system txs?

crates/evm/src/tests/call_tests.rs
```diff
@@ -586,7 +600,7 @@ pub fn create_contract_message<T: TestContract>(
         .unwrap()
 }
 
-fn create_contract_message_with_fee<T: TestContract>(
+pub fn create_contract_message_with_fee<T: TestContract>(
     dev_signer: &TestSigner,
     nonce: u64,
     contract: T,
```
eyusufatik, line 606: pub(crate)

crates/evm/src/tests/genesis_tests.rs
eyusufatik: ok, you didn't push system contract to the genesis, i see.

so they are not executing.

crates/evm/src/tests/sys_tx_tests.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    let l1_fee_rate = 1;
+
+    let deploy_message =
+        create_contract_message_with_fee(&dev_signer, 0, BlockHashContract::default(), 1);
+
+    let system_account = evm.accounts.get(&SYSTEM_SIGNER, &mut working_set).unwrap();
+    // The system caller balance is unchanged(if exists)/or should be 0
+    assert_eq!(system_account.info.balance, U256::from(0));
+    assert_eq!(system_account.info.nonce, 0);
```
eyusufatik, line 85: shouldn't nonce be 2 here?

executed two txs above

crates/evm/src/tests/sys_tx_tests.rs
```diff
@@ -0,0 +1,165 @@
...
+                diff_size: 412,
+                error: None
+            },
+        ]
+    );
+
+    let l1_fee_rate = 1;
+
+    let deploy_message =
+        create_contract_message_with_fee(&dev_signer, 0, BlockHashContract::default(), 1);
```
eyusufatik, line 80: either should be below where this is used or just call the function inside the vec

crates/evm/src/tests/sys_tx_tests.rs
```diff
@@ -0,0 +1,165 @@
...
+        .unwrap();
+    }
+    evm.end_soft_confirmation_hook(&mut working_set);
+    evm.finalize_hook(&[99u8; 32].into(), &mut working_set.accessory_state());
+
+    let system_account = evm.accounts.get(&SYSTEM_SIGNER, &mut working_set).unwrap();
+    // The system caller balance is unchanged(if exists)/or should be 0
+    assert_eq!(system_account.info.balance, U256::from(0));
+    assert_eq!(system_account.info.nonce, 0);
+
```
eyusufatik, line 117: again nonce

crates/evm/src/tests/sys_tx_tests.rs
```diff
@@ -0,0 +1,165 @@
...
+            },
+            Receipt {
+                receipt: reth_primitives::Receipt {
+                    tx_type: reth_primitives::TxType::Eip1559,
+                    success: true,
+                    cumulative_gas_used: 189900,
+                    logs: vec![]
+                },
+                gas_used: 114235,
+                log_index_start: 1,
```
eyusufatik, line 153: this has no nonce but log index start is 1

are we sure about this?

crates/evm/src/tests/sys_tx_tests.rs
```diff
@@ -0,0 +1,165 @@
...
+                error: None
+            }
+        ]
+    );
+
+    let coinbase_account = evm
+        .accounts
+        .get(&config.coinbase, &mut working_set)
+        .unwrap();
+    assert_eq!(coinbase_account.info.balance, U256::from(114235 + 477));
```
eyusufatik, line 164: try querying the system contract for blockhases and assert_eq

crates/evm/src/call.rs
```diff
@@ -28,14 +29,115 @@ pub struct CallMessage {
...
+            .expect("L1 fee rate must be set");
+
+        let l1_block_hash_exists = self
+            .accounts
+            .get(&L1BlockHashList::address(), working_set)
+            .is_some();
+        if !l1_block_hash_exists {
+            tracing::error!("System contract not found: L1BlockHashList");
+            return;
+        }
```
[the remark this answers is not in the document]
eyusufatik: ok I now realize you use this for tests that you didn't predeploy the system contract.

if we can, let's panic if we are not in tests and return if we are in tests.

Then I assume e2e and integration will fail due to this.

Ok, we will implement this in #342

eyusufatik (changes requested):

crates/evm/src/evm/system_contracts/out/L1BlockHashList.sol/L1BlockHashList.json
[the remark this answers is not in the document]
kpp: Yes, `forge build`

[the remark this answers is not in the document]
kpp: it would be nice to include it in build.rs to produce the artifact but...

crates/evm/src/evm/system_events.rs
kpp: Then I will get Transaction(GasPriceLessThanBasefee) error, because we check gas fees for system txs.

crates/evm/src/call.rs
kpp: I can init them with 0 but I am not quite sure about our l1 fee strategy, if we end up with exec a sys tx in the end of the block then 0 will be an incorrect init value

crates/evm/src/tests/sys_tx_tests.rs
kpp: Yes because they are in the same L2 block.

eyusufatik (approved):
please create an issue for 
```rust
// TODO check: Nonce for CREATE will be bumped in `handle_create`.
```

Explaining the code there, and what needs to be checked.

Everything else is looking good.

PR #338, opened 2024-04-03
Bug: Fix install risczero toolchain bug for x86-64-macos

clint419:
# Bug description
risc0/risc0#1507

## Fixed Issue
risc0/risc0#1512

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,11 @@ install-dev-tools:  ## Installs all necessary cargo helpers
 	cargo install cargo-nextest --locked
 	cargo install cargo-binstall
 	cargo binstall cargo-risczero
+ifeq ($(shell uname -ms), Darwin x86_64)
+	cargo risczero build-toolchain
+else
 	cargo risczero install
+endif
 	rustup target add thumbv6m-none-eabi
 
 lint:  ## cargo check and clippy. Skip clippy on guest code since it's not supported by risc0


Makefile
```diff
@@ -29,7 +29,11 @@ install-dev-tools:  ## Installs all necessary cargo helpers
 	cargo install cargo-nextest --locked
 	cargo install cargo-binstall
 	cargo binstall cargo-risczero
+ifeq ($(shell uname -mo), Darwin x86_64)
```
eyusufatik, line 32: indent right please

eyusufatik (changes requested):

Makefile
otaliptus, line 32: Thanks for the PR!

I believe -o is not directly POSIX for `uname`. Just tested on an x86_64 machine, -s seems to be the correct one. Could you check if -s works, and then convert -o to -s?

So it should be `uname -ms`

otaliptus (changes requested):

Makefile
clint419: > indent right please

https://stackoverflow.com/questions/4483313/make-error-for-ifeq-syntax-error-near-unexpected-token

clint419: > Thanks for the PR!
> 
> I believe -o is not directly POSIX for `uname`. Just tested on a Mac x86_64 machine, -o does not work | -s seems to be the correct one. Could you check if -s works, and then convert -o to -s?
> 
> So it should be `uname -ms`
> 
> Edit: [pystardust/ani-cli#416 (comment)](https://github.com/pystardust/ani-cli/issues/416#issuecomment-1023607443)

Yes, thank you a lot

otaliptus: Great

PR #340, opened 2024-04-03
Convert tx hash bytes into hex

### LINKED ISSUE #337: Use hex string instead of u8 vector in soft batch serialization
Right now in rpc responses for Soft batch receipts etc. sov-txs are serialized into u8 vector.

Let's run some numbers.

let's say a sov tx is 100 bytes 

that's 100 u8's in the response

but the response is JSON, so it's actually strings, so the sov-tx takes 201 to 401 bytes because it's represented as strings (2 bytes from [] and 99 from , + 1 digit num to 3 digit num in u8)

if we've used hex strings, 

the 100 byte sov-tx would be constant 202 (2 from 0x, and every two chars represent a byte)

TODOs:

- [ ] update SoftBatchResponse to serialize tx body's as hex
- [ ] update sequencer-client
- [ ] update rpc-tests

rakanalh:
# Description

This PR adds a serializer to the `txs` field of `SoftBatchResponse` to make sure the fields are converted from a vector of bytes into a hex representation of the value.

## Linked Issues
- Fixes #337 

## Testing
- Tests were updated accordingly.

## Docs
N/A


crates/sovereign-sdk/rollup-interface/src/node/rpc/mod.rs
```diff
@@ -178,7 +178,7 @@ pub struct SoftBatchResponse {
     #[serde(with = "hex::serde")]
     pub hash: [u8; 32],
     /// The transactions in this batch.
-    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(skip_serializing_if = "Option::is_none", with = "utils::tx_hex")]
     pub txs: Option<Vec<Vec<u8>>>,
```
kpp, line 182: So basically txs can be an Option<TxsVec> where TxsVec is a flatten proxy for serde::with hex. This can be reduce our code (not having to implement our custom ser/de visitors is a plus) but can lead to more patches in the code.

kpp (changes requested):

crates/sovereign-sdk/rollup-interface/src/node/rpc/mod.rs
eyusufatik: yeah I agree with @kpp here, 

if we can get away without needing custom serializers 

it's much better 😄

crates/sovereign-sdk/rollup-interface/src/node/rpc/mod.rs
```diff
@@ -476,6 +476,87 @@ pub mod utils {
...
+        where
+            S: Serializer,
+        {
+            if let Some(data) = data {
+                let data = data
+                    .iter()
+                    .map(|item| format!("0x{}", item.encode_hex::<String>()));
+                let mut seq = serializer.serialize_seq(Some(data.len()))?;
+                for e in data {
+                    seq.serialize_element(&e)?;
```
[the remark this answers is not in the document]
rakanalh: `seq` does not have `serialize_str`... we need to serialize an array element in this case.

crates/sovereign-sdk/rollup-interface/src/node/rpc/mod.rs
rakanalh: @kpp I do not get what you mean by "flatten proxy" here. Regardless of the option, using `#[serde(with = "hex::serde")]` would not work on the inner `Vec<Vec<u8>>` but would work on `Vec<u8>` as `hex::serde::serialize` expects a slice of u8s. 
If you can elaborate further, i will be happy to simplify the code.

kpp: I mean `#[serde(transparent)]`.

crates/sovereign-sdk/rollup-interface/src/node/rpc/mod.rs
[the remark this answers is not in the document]
kpp: My bad.

rakanalh:
@kpp @eyusufatik I messed up something during a merge from nightly so i had to force push. But PR is ready again for review.

crates/sequencer-client/src/lib.rs
```diff
@@ -99,7 +100,11 @@ impl From<GetSoftBatchResponse> for SignedSoftConfirmationBatch {
             val.da_slot_txs_commitment,
             val.pre_state_root,
             val.l1_fee_rate,
-            val.txs.unwrap_or_default(),
+            val.txs
```
eyusufatik, line 103: is there any way we can get rid of this?

if any other client queries that endpoint it will look like this:

```json
{
  ...
  txs: [
    { tx: "" },
    { tx: "" }
  ]
}
```

Maybe if we define HexTx like this it won't get serialized that way:
```rust
pub struct HexTx(Vec<u8>);
```

eyusufatik: anyone that sends an http request :)

crates/sequencer-client/src/lib.rs
```diff
@@ -77,7 +78,7 @@ pub struct GetSoftBatchResponse {
     #[serde(with = "hex::serde")]
     pub da_slot_txs_commitment: [u8; 32],
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub txs: Option<Vec<Vec<u8>>>,
+    pub txs: Option<Vec<HexTx>>,
```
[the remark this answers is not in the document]
eyusufatik: we need that, bc some full nodes prefer not to keep the txs field, because they already used those txs to update their state

crates/sequencer-client/src/lib.rs
eyusufatik: checked it now works as expected

eyusufatik (approved):

kpp (approved):
LGTM

Issue #303, opened 2024-03-27
Sequencer: Don't wait for sequencer commitment submisson


eyusufatik:
For adapters like Bitcoin, sending transactions takes some time.
Because we try random nonces to make the `txid` start with 2 zeros.

Since the block producing and sequncer commitment submission happen in the same function, while producing blocks, the sequencer first has to submit the sequencer commitment.

This is not ideal, because we want the sequencer to publish soft confirmations every x seconds.

Due to the nature of Bitcoin DA adapter (remember random nonces above), we don't know how much time the commitment submission will take. So that postpones the block creation.

`crates/sequencer/src/sequencer.rs`:
```rust
 // submit commitment
self.da_service
    .send_transaction(
        DaData::SequencerCommitment(commitment)
            .try_to_vec()
            .unwrap()
            .as_slice(),
    )
    .await
    .expect("Sequencer: Failed to send commitment");
```

This should happen in another thread or event loop (this already runs in `tokio::spawn`)

PR #343, opened 2024-04-04
Remove async recursion from BitcoinNode::call

kpp:
# Description
Preserving infinite retries

## Linked Issues
- Related to #303  

## Testing
`make test`

### DIFF
--- a/crates/bitcoin-da/src/rpc.rs
+++ b/crates/bitcoin-da/src/rpc.rs
@@ -2,7 +2,6 @@ use core::fmt::Display;
 use core::str::FromStr;
 
 use anyhow::anyhow;
-use async_recursion::async_recursion;
 use bitcoin::block::{Header, Version};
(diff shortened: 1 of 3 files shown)


crates/bitcoin-da/src/rpc.rs
```diff
@@ -93,27 +90,44 @@ impl BitcoinNode {
...
+    }
+
+    // TODO: add max retries
+    async fn call<T: serde::de::DeserializeOwned>(
+        &self,
+        method: &str,
+        params: Vec<serde_json::Value>,
+    ) -> Result<T, anyhow::Error> {
+        let mut attempt = 1;
+        loop {
```
eyusufatik, line 111: still looping?

I thought this pr was going to remove retrying and we'd do it in an outer scope?

kpp: It's still looping, but not recursing.

eyusufatik (approved):

PR #361, opened 2024-04-08
Don't wait on Sequencer::send_transaction

kpp:
# Description
Send tx in a tokio thread. That's actually not ideal, because it can block some io in tokio if the number of tokio threads = number of async sendings. This impl can be improved ofc.

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/303

## Testing
`make test`

### DIFF
--- a/crates/sovereign-sdk/rollup-interface/src/node/services/da.rs
+++ b/crates/sovereign-sdk/rollup-interface/src/node/services/da.rs
@@ -2,6 +2,8 @@
 
 use serde::de::DeserializeOwned;
 use serde::Serialize;
+#[cfg(feature = "native")]
+use tokio::sync::oneshot::{channel as oneshot_channel, Receiver as OneshotReceiver};
 
 use crate::da::BlockHeaderTrait;
 #[cfg(feature = "native")]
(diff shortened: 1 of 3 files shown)


crates/sovereign-sdk/rollup-interface/src/node/services/da.rs
```diff
@@ -113,6 +115,19 @@ pub trait DaService: Send + Sync + 'static {
     /// Returns nothing if the transaction was successfully sent.
     async fn send_transaction(&self, blob: &[u8]) -> Result<Self::TransactionId, Self::Error>;
 
+    /// Send a transaction directly to the DA layer.
+    /// The default impl of this method actually blocks on `self.send_transaction`.
+    /// But the implementors can redefine it to be non-blocking.
+    async fn send_tx_no_wait(
```
eyusufatik, line 121: if we implement the function at trait level, why do we reimplement it in MockDa service.rs?

kpp: 1. Because DaService in general is not Clone, I can't do that trick with .clone & tokio::spawn here.
2. If there is no default impl then I have to implement it in every Da
3. I could implement it as `unimplemented` but then I would have to define it at least for MockDa?
4. So I made the most innocent default implementation that works for all cases and can be redefined. Indeed it was redefined for BitcoinService.

eyusufatik (approved):

PR #373, opened 2024-04-11
Get rid of borsh::maybestd

kpp:
# Description

This is a part of upgrade to a new version of borsh.

Remove line `pub use borsh::maybestd::{borrow, boxed, collections, format, io, string, vec};` and use `alloc` crate whenever possible. Instead of `borsh::maybestd::hashmap` use `hashbrown`. Instead of `borsh::maybestd::io` use `no_std_io` crate.

With a new version of borsh it will be possible to use `borsh::io` module instead of `no_std_io` crate.

## Testing
`make test`


eyusufatik:
I think using `std::Vec` etc. might break zk vm compatability. Are you sure about your changes?

kpp:
Alloc::vec::Vec is designed for no_std. Borsh::maybestd reexports Vec from alloc, I did it transparently to reduce code dependency from borsh. In version 1 borsh doesnt have that maybestd module

kpp:
Fixed by https://github.com/chainwayxyz/citrea/pull/851

PR #374, opened 2024-04-12
Bump rustc version

kpp:
# Description

A new version of rustc is released.

See error: https://github.com/chainwayxyz/citrea/actions/runs/8653017088/job/23732699699#step:11:1759


eyusufatik (approved):

eyusufatik:
aproving for now, can we utilize macros to auto include the compiler version in that script?

PR #380, opened 2024-04-12
Use bytecode instead of bytes in state

ercecan:
# Description

## Linked Issues
- Fixes #
- Related to # (issue) 

## Testing

## Docs


crates/evm/src/evm/db.rs
```diff
@@ -44,7 +44,8 @@ impl<'a, C: sov_modules_api::Context> Database for EvmDb<'a, C> {
         let bytecode = Bytecode::new_raw(
             self.code
                 .get(&code_hash, self.working_set)
```
eyusufatik, line 46: shouldn't this whole function now be 
```rust
return self.code.get(&code_hash, self.working_set).unwrap_or_default()
```

if evm db wants us to commit bytecode type 

we should keep it as is and return it as is 

because revm does some work on the bytecode and wants to keep that work

crates/evm/src/genesis.rs
```diff
@@ -168,7 +168,7 @@ impl<C: sov_modules_api::Context> Evm<C> {
             );
 
             if acc.code.len() > 0 {
-                evm_db.insert_code(acc.code_hash, acc.code.clone());
+                evm_db.insert_code(acc.code_hash, Bytecode::new_raw(acc.code.clone()));
```
eyusufatik, line 171: Bytecode has function `hash_slow`

let's  use that in above lines instead of keccak256(acc.code)

eyusufatik (changes requested):

crates/evm/src/genesis.rs
```diff
@@ -158,17 +158,19 @@ impl<C: sov_modules_api::Context> Evm<C> {
         let mut evm_db = self.get_db(working_set);
 
         for acc in &config.data {
+            let code = Bytecode::new_raw(acc.code.clone());
```
eyusufatik, line 161: no clone here

eyusufatik (approved):
small nit aproved

PR #387, opened 2024-04-16
Fix access list gas

### LINKED ISSUE #272: create_access_list returns incorrect gas consumption
## Summary
#256 implements the Access List functionality, with parts mostly taken from Reth. 
The function works mostly correct, given that the tests of cross-contract calls & wrong storage calls using eth_estimateGas seems feasible.

However, the field of `gas_used` in the `AccessListWithGasUsed` struct returning from `create_access_list` function does not take the access list into account. We should check this.

## Additional Information
https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-eth#eth-createaccesslist

https://github.com/paradigmxyz/reth/blob/332e412a0f8d34ff2bbb7e07921f8cacdcf69d64/crates/rpc/rpc/src/eth/api/call.rs#L370

https://github.com/chainwayxyz/secret-sovereign-sdk/blob/3551e7ee4c56490afb53d6869c909dab4f1ba655/module-system/module-implementations/sov-evm/src/tests/queries/estimate_gas_tests.rs#L203-L257

otaliptus:
# Description
Closes #272 (finally). Also updated gas estimation from reth. [[1](https://github.com/paradigmxyz/reth/pull/7485)] [[2](https://github.com/paradigmxyz/reth/pull/7521)]

p.s. I still stick with my opinion of better env's though.

### DIFF
--- a/crates/evm/src/tests/queries/estimate_gas_tests.rs
+++ b/crates/evm/src/tests/queries/estimate_gas_tests.rs
@@ -198,7 +198,6 @@ fn test_tx_request_fields_gas() {
         ..tx_req_contract_call.clone()
     };
 
-    // TODO: Fix create_access_list returning wrong gas (i.e. returns as if access list is none, rather than the provided argument).
     let create_no_access_list_test = evm.create_access_list(
         no_access_list_req,
         Some(BlockNumberOrTag::Latest),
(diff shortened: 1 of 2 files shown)


crates/evm/src/tests/queries/estimate_gas_tests.rs
```diff
@@ -218,7 +218,7 @@ fn test_tx_request_fields_gas() {
             .into(),
             // This should actually be 0x6e66
             // See: https://github.com/chainwayxyz/citrea/issues/272
-            gas_used: Uint::from_str("0x6601").unwrap()
+            gas_used: Uint::from_str("0x6e66").unwrap()
```
eyusufatik, line 221: delete todo above

eyusufatik (changes requested):

eyusufatik (approved):

PR #400, opened 2024-04-18
Add sequencer mempool_conf

otaliptus:
# Description
The sequencer is the standalone block producer (as of now), and full nodes do not involve in this. Therefore, the machine can carry a bit more load in the mempool(s). This PR introduces configurable limits for this.

Must read:
https://github.com/ledgerwatch/erigon/wiki/Transaction-Pool-Design

https://github.com/paradigmxyz/reth/blob/027f920eb43f4d17c2eae9939dbfcc4d36cd2ce5/book/cli/reth/node.md?plain=1#L321

### DIFF
--- a/crates/sequencer/src/mempool.rs
+++ b/crates/sequencer/src/mempool.rs
@@ -1,16 +1,17 @@
 use std::sync::Arc;
 
 use citrea_evm::SYSTEM_SIGNER;
-use reth_primitives::{BaseFeeParamsKind, Chain, ChainSpec, TxHash};
+use reth_primitives::{Chain, ChainSpecBuilder, Genesis, TxHash};
 use reth_tasks::TokioTaskExecutor;
 use reth_transaction_pool::blobstore::NoopBlobStore;
 use reth_transaction_pool::error::PoolError;
 use reth_transaction_pool::{
     BestTransactions, BestTransactionsAttributes, CoinbaseTipOrdering, EthPooledTransaction,
-    EthTransactionValidator, Pool, PoolResult, TransactionPool, TransactionValidationTaskExecutor,
-    ValidPoolTransaction,
+    EthTransactionValidator, Pool, PoolConfig, PoolResult, SubPoolLimit, TransactionPool,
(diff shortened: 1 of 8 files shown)


crates/sequencer/src/mempool.rs
```diff
@@ -24,16 +25,56 @@ type Transaction<C> = <CitreaMempoolImpl<C> as TransactionPool>::Transaction;
...
+                    )),
+            )
+            .build();
+
+        // Default 10x'ed from standard limits
+        let pool_config = Default::default();
+        let pool_config = PoolConfig {
+            pending_limit: SubPoolLimit {
+                max_txs: mempool_conf.pending_tx_limit as usize,
+                max_size: TXPOOL_SUBPOOL_MAX_SIZE_MB_DEFAULT * 10 * 1024 * 1024,
```
kpp, line 60:
```
  88   │ impl Default for SubPoolLimit {
  89   │     fn default() -> Self {
  90   │         // either 10k transactions or 20MB
  91   │         Self {
  92   │             max_txs: TXPOOL_SUBPOOL_MAX_TXS_DEFAULT,
  93   │             max_size: TXPOOL_SUBPOOL_MAX_SIZE_MB_DEFAULT * 1024 * 1024,
  94   │         }
  95   │     }
  96   │ }
```

So better use syntax:

```
SubPoolLimit {
    max_txs: mempool_conf.pending_tx_limit as usize,
    ..Default::default()
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=97a6d8b64de9512439aa424852fc11bf

kpp (changes requested):
Fix SubPoolLimit options

crates/sequencer/src/mempool.rs
otaliptus: I see your point. That was my first attempt too but at the same time the Default seems to be wrong - it works as if it's KB rather than MB. 

I pointed this to Esad and tested on my machine, and will open a PR to Reth (if I am not completely mistaken, will check again)

eyusufatik (approved):

PR #408, opened 2024-04-19
Erce/sequencer test mode

### LINKED ISSUE #407: Use eth_publishBatch only for tests
Sequencer should produce blocks by itself without rpc, however since we have many tests depending on producing blocks we can create a config option and run the sequencer accordingly

ercecan:
# Description
Runs sequencer in test mode for tests and produces blocks itself without rpc

## Linked Issues
- Fixes #407
- Fixes https://github.com/chainwayxyz/citrea/issues/410


crates/sequencer/Cargo.toml
```diff
@@ -65,3 +65,4 @@ tempfile = { workspace = true }
 default = []
 local = []
 native = ["citrea-stf/native", "citrea-evm/native"]
+test_mode = []
```
kpp, line 68: Sorry for bikeshedding, I think this should be something like manual_publish_batch/block. Or even with an unsafe_ preffix in order to keep users away from it. Also cargo has `build --all-features` which can trigger it unwillingly. So maybe it would be wisely to put this as an option to sequencer config instead.

eyusufatik (approved):

crates/sequencer/Cargo.toml
eyusufatik: Hmm, we initially thought of putting this in sequencer config but then decided to go against it.

build --all-features will not cause problems bc everything is branched based on !cfg(test_mode) and cfg(test_mode)

the reason we reverted putting it to config was it looked ugly tbh :D

eyusufatik (changes requested):
let's go back to doing it in sequencer config

but let's not pass it through the stf blueprint


the sequencer also already has seqeuncer config

bin/citrea/tests/test_helpers.rs
```diff
@@ -37,6 +37,7 @@ pub async fn start_rollup(
     db_path: Option<&str>,
     min_soft_confirmations_per_commitment: u64,
     include_tx_body: bool,
+    test_mode: Option<bool>,
```
kpp, line 40: None and Some(false) are equal. So there is no need to make test_mode `Option<bool>`, make it bool with default=false if it's missing in toml config.

kpp:
LGTM but remove this line: https://github.com/chainwayxyz/citrea/blob/86fc5dbf92aa237e4e61a3df2efc2a3a5044f965/.github/workflows/checks.yml#L162

eyusufatik (approved):

eyusufatik:
> LGTM but remove this line:
> 
> https://github.com/chainwayxyz/citrea/blob/86fc5dbf92aa237e4e61a3df2efc2a3a5044f965/.github/workflows/checks.yml#L162

I guess we can remove that script directly

ercecan:
> > LGTM but remove this line:
> > https://github.com/chainwayxyz/citrea/blob/86fc5dbf92aa237e4e61a3df2efc2a3a5044f965/.github/workflows/checks.yml#L162
> 
> I guess we can remove that script directly

With that I have also updated hive Dockerfile and docs. Lmk if I missed anything.

kpp (approved):

docs/run.md
```diff
@@ -28,10 +28,8 @@ _Optional_: Run full node on Mock DA:
 
 Full node RPC is accessible at `127.0.0.1:12346`
 
-To publish blocks on Mock DA, run these two on seperate terminals:
+To publish blocks on Mock DA, run the sequencer with `test_mode` in sequencer config set to false and run this script:
```
eyusufatik, line 31: If `test_mode` is set to false in the sequencer config, the sequencer will publish blocks every 2 seconds. To also publish mock DA blocks, run this script:

eyusufatik (approved):

PR #411, opened 2024-04-19
Ignore system transaction execution

### LINKED ISSUE #406: EVM: discard sys txs. in execute_transactions
Currently, the sequencer can add txs coming from 0xdead... in its soft confirmations, and full nodes would replay those txs as free.

Where normal txs are executed, ignore txs coming from 0xdead...

otaliptus:
# Description
Closes #406.

### DIFF
--- a/crates/evm/src/evm/executor.rs
+++ b/crates/evm/src/evm/executor.rs
@@ -83,7 +83,7 @@ pub(crate) fn execute_multiple_tx<
             ))
         } else if tx.signer() == SYSTEM_SIGNER {
             Err(EVMError::Custom(format!(
-                "Ignored system transaction: {:?}",
+                "Invalid system transaction: {:?}",
                 hex::encode(tx.hash())
             )))
         } else {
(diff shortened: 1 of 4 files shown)


crates/evm/src/evm/executor.rs
```diff
@@ -80,6 +81,11 @@ pub(crate) fn execute_multiple_tx<
             Err(EVMError::Transaction(
                 InvalidTransaction::CallerGasLimitMoreThanBlock,
             ))
+        } else if tx.signer() == SYSTEM_SIGNER {
+            Err(EVMError::Custom(format!(
+                "Ignored system transaction: {:?}",
```
eyusufatik, line 86: call this invalid system transaction

eyusufatik (changes requested):

crates/evm/src/evm/executor.rs
kpp: Why not retain !sys txs from `users_txs` before even passing then into executor::execute_multiple_tx?

eyusufatik: > Why not retain !sys txs from users_txs before even passing then into executor::execute_multiple_tx?

I didn't understand this. 

But this is a fix for the following case:

Sequencer creates a system transaction and puts it along with other mempool transactions.

We execute them for free.

So where we process user txs, there shouldn't be any system txs.

eyusufatik (dismissed):

crates/evm/src/evm/executor.rs
otaliptus: Is it clear now @kpp ?

kpp: Discussed in slack

eyusufatik (approved):

PR #432, opened 2024-04-23
Fix EVM RPC's fee handling

### LINKED ISSUE #241: EVM: Add L1 fee to `eth_call` and other RPCs
RPCs that need updating:

- [x] eth_getBlockByHash and eth_getBlockByNumber (depends on #245)
- [x] eth_getTransactionReceipt
- [ ] debug_transaction (not quiet sure #244 probably introduced an implementation for this)

### LINKED ISSUE #247: EVM: decide on fees on `eth_call`
Currently, we:
- disable block gas fee
- disable base fee
- disable eip 3607
- don't track L1 fees

for eth_call and trace_transaction.

checked reth didn't see anything that disables base fees etc. for this. However it makes sense to do that too.

Will check geth and reth.

eyusufatik:
# Description
Read latest comments on issues

## Linked Issues
Closes #241 
Closes #247

### DIFF
--- a/crates/evm/src/query.rs
+++ b/crates/evm/src/query.rs
@@ -576,7 +576,7 @@ impl<C: sov_modules_api::Context> Evm<C> {
         working_set: &mut WorkingSet<C>,
     ) -> RpcResult<reth_primitives::Bytes> {
         info!("evm module: eth_call");
-        let block_env = match block_number {
+        let mut block_env = match block_number {
             None | Some(BlockNumberOrTag::Pending | BlockNumberOrTag::Latest) => {
                 // so we don't unnecessarily set archival version
                 self.block_env.get(working_set).unwrap_or_default().clone()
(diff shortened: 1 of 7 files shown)


otaliptus (changes requested):

crates/evm/src/query.rs
```diff
@@ -666,28 +682,38 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+                    .balance,
+            ),
+        )?;
 
         let from = request.from.unwrap_or_default();
         let to = if let Some(to) = request.to {
             to
         } else {
-            let account = evm_db.basic(from).unwrap_or_default();
+            let account = evm_db.basic(from).unwrap();
```
kpp, line 716: What if `from` does not exist in the db yet?

crates/evm/src/query.rs
```diff
@@ -666,28 +682,38 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
 
             let nonce = account.unwrap_or_default().nonce;
```
kpp, line 718: Okay, then we should check account.is_some here.

eyusufatik: if account doesnt exists, the default impl. applies

kpp (approved):

otaliptus (approved):

Issue #417, opened 2024-04-22
Add l1 fee to gas estimation


kpp:
Follow up of https://github.com/chainwayxyz/citrea/issues/398

TODO:
- [ ] Similar to Arbitrum, somehow make fee endpoints add the L1 fee rate to the estimations. (open another issue for this)

See https://github.com/OffchainLabs/nitro/blob/v2.2.5/nodeInterface/NodeInterface.go#L511 and https://docs.arbitrum.io/build-decentralized-apps/nodeinterface/reference

eyusufatik:
not to be confused with #420, this issue should somehow add in the L1 fee estimations into the normal gas estimation (or fee rate estimation) so that existing eth libraries and infra knows if a tx will fail due to L1 fees. We have #420 for operations Citrea specific (more precise).

PR #437, opened 2024-04-24
RPC: estimate L1 fee

kpp:
# Description

Change `eth_createAccessList` and `eth_estimateGas`.

TODO:

- [x] pass l1_fee_rate to `inspect_no_tracing`
- [x] extract l1_fee from `inspect_no_tracing` in Evm::eth_estimate_gas
- [x] apply l1_fee in binary search in Evm::eth_estimate_gas
- [x] make estimate_gas return U256 instead of U64 

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/417

## Testing
`make test`


crates/evm/src/tests/queries/estimate_gas_tests.rs
```diff
@@ -364,7 +365,7 @@ fn estimate_gas_with_varied_inputs_test() {
 
     assert_eq!(
         value_transfer_result.unwrap(),
-        Uint::from_str("0x5208").unwrap()
+        Uint::from(MIN_TRANSACTION_GAS + 156)
```
kpp, line 368: I wonder if a simple transfer fee must be MIN_TRANSACTION_GAS=21000 or MIN_TRANSACTION_GAS+l1 fee

bin/citrea/tests/evm/tracing.rs
```diff
@@ -112,14 +112,14 @@ async fn tracing_tests() -> Result<(), Box<dyn std::error::Error>> {
...
     let reth_json = serde_json::from_value::<CallFrame>(json![{
         "from": "0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266",
-        "gas": "0xbbdf",
+        "gas": "0xc477",
         "gasUsed": "0xba65",
         "to": "0xe7f1725e7734ce288f8367e1bb143e90bb3f0512",
         "input": "0xb7d5b6580000000000000000000000005fbdb2315678afecb367f032d93f642f64180aa30000000000000000000000000000000000000000000000000000000000000003",
         "calls": [
             {
                 "from": "0xe7f1725e7734ce288f8367e1bb143e90bb3f0512",
```
eyusufatik, line 121: debug trace gas changed?

weird

crates/evm/src/query.rs
```diff
@@ -36,14 +36,30 @@ use crate::rpc_helpers::*;
...
+/// The result of gas estimation.
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+pub(crate) struct EstimatedGas {
+    /// Evm gas used.
+    gas: U64,
+    /// L1 fee.
+    l1_fee: U256,
+}
+
+impl EstimatedGas {
```
eyusufatik, line 56: this is wrong

gas field is gas used, not gas fee.

l1 fee is diff size * l1 fee rate

we want this to be  (l1_fee / base_fee) + gas 

so that when a wallet etc. tries to estimate if its tx goes through, 

it can do the following:

1. call eth_estimateGas
2. call eth_maxFeePerGas
3. multiply returned values
4. check balance

eyusufatik (changes requested):

bin/citrea/tests/evm/tracing.rs
kpp: Because it does `deploy_contract` which calls `eth_estimate_gas` inside.

crates/evm/src/query.rs
```diff
@@ -35,14 +35,30 @@ use crate::rpc_helpers::*;
...
 
 /// https://github.com/paradigmxyz/reth/pull/7133/files
 /// Allowed error ratio for gas estimation
 /// Taken from Geth's implementation in order to pass the hive tests
 /// <https://github.com/ethereum/go-ethereum/blob/a5a4fa7032bb248f5a7c40f4e8df2b131c4186a4/internal/ethapi/api.go#L56>
 const ESTIMATE_GAS_ERROR_RATIO: f64 = 0.015;
 
+/// The result of gas estimation.
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+pub(crate) struct EstimatedGas {
```
eyusufatik, line 48: since it's not l1_fee but l1_fee / base_fee now, let's call it l1_fee_overhead or smth like that

eyusufatik (changes requested):

crates/evm/src/query.rs
```diff
@@ -48,14 +48,14 @@ const ESTIMATE_GAS_ERROR_RATIO: f64 = 0.015;
 pub(crate) struct EstimatedGas {
```
eyusufatik, line 48: maybe we can add an explanation comment on this struct.

let's explain why we add an l1_fee_overhead to the actual estimated gas

eyusufatik (approved):

PR #446, opened 2024-04-25
Make Citrea-STF provable

### LINKED ISSUE #106: Fix need for SKIP_GUEST_BUILD
Checked sovereign-sdk, they are using SKIP_GUEST_BUILD=1 in CI for lint but not for `make test`. And `make test` works in their repo on my machine while with our repo it fails to compile.

We need to solve what's wrong with our repo probably by examining Cargo.toml differences.

eyusufatik:
# Description
Makes Citrea provable 🔥 

Didn't remove SKIP_GUEST_BUILD=1 from lint commands, still doesn't work but we can fix that later.

## Linked Issues
Closes #106 :)

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -87,7 +87,7 @@ jobs:
         with:
           tool: cargo-risczero@0.19
       - name: Install risc0-zkvm toolchain # Use the risc0 cargo extension to install the risc0 std library for the current toolchain
-        run: cargo risczero install
+        run: cargo risczero install --version v2024-04-22.0
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: Run lint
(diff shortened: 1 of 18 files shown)


bin/citrea/src/bitcoin_rollup.rs
```diff
@@ -115,7 +115,8 @@ impl RollupBlueprint for BitcoinRollup {
         rollup_config: &RollupConfig<Self::DaConfig>,
         _da_service: &Self::DaService,
     ) -> Self::ProverService {
-        let vm = Risc0Host::new(risc0::ROLLUP_ELF);
+        // TODO: will be BITCOIN_ELF
+        let vm = Risc0Host::new(risc0::MOCK_DA_ELF);
```
kpp, line 119: TODO

.github/workflows/checks.yml
```diff
@@ -87,7 +87,7 @@ jobs:
         with:
           tool: cargo-risczero@0.19
       - name: Install risc0-zkvm toolchain # Use the risc0 cargo extension to install the risc0 std library for the current toolchain
-        run: cargo risczero install
+        run: cargo risczero install --version v2024-04-22.0
```
ercecan, line 90: Is this. the preview version?

PR #450, opened 2024-04-26
RPC point to head storage

kpp:
# Description

TODO:
- [x] RPC point to head storage
- [x] RPC send storage updates on block change

## Testing
`make test`

### DIFF
--- a/crates/sequencer/src/sequencer.rs
+++ b/crates/sequencer/src/sequencer.rs
@@ -604,11 +604,6 @@ where
         let new_rpc_storage = self
             .storage_manager
             .create_storage_on_l2_height(l2_height)?;
-        // `send_replace` is superior to `send` for our use case. It never fails
-        // because it doesn't need to notify all receivers, unlike `send`, which
-        // we don't need. It will also keep working even if there are no
-        // receivers currently alive, which makes it easier to reason about the
-        // code.
         self.rpc_storage_sender.send_replace(new_rpc_storage);
 
         Ok(())
--- a/crates/sequencer/src/sequencer.rs
+++ b/crates/sequencer/src/sequencer.rs
@@ -32,6 +32,7 @@ use sov_rollup_interface::stf::{SoftBatchReceipt, StateTransitionFunction};
 use sov_rollup_interface::storage::HierarchicalStorageManager;
 use sov_rollup_interface::zk::ZkvmHost;
 use sov_stf_runner::{InitVariant, RpcConfig, RunnerConfig};
+use tokio::sync::watch;
 use tokio::time::sleep;
 use tracing::{debug, info, warn};
 
@@ -58,11 +59,11 @@ where
     l2_force_block_tx: UnboundedSender<()>,
     l2_force_block_rx: UnboundedReceiver<()>,
     db_provider: DbProvider<C>,
-    storage: C::Storage,
     ledger_db: LedgerDB,
     config: SequencerConfig,
     stf: Stf,
     storage_manager: Sm,
+    rpc_storage_sender: watch::Sender<Sm::NativeStorage>,
     state_root: StateRoot<Stf, Vm, Da::Spec>,
     sequencer_pub_key: Vec<u8>,
     rpc_config: RpcConfig,
@@ -77,7 +78,7 @@ impl<C, Da, Sm, Vm, Stf> CitreaSequencer<C, Da, Sm, Vm, Stf>
 where
     C: Context,
     Da: DaService,
-    Sm: HierarchicalStorageManager<Da::Spec>,
+    Sm: HierarchicalStorageManager<Da::Spec, NativeStorage = C::Storage>,
     Vm: ZkvmHost,
     Stf: StateTransitionFunction<
             Vm,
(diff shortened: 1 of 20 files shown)


crates/sequencer/src/sequencer.rs
```diff
@@ -593,25 +596,28 @@ where
...
-
-        match accounts
-            .get_account(self.sov_tx_signer_priv_key.pub_key(), &mut working_set)
-            .expect("Sequencer: Failed to get sov-account")
-        {
-            AccountExists { addr: _, nonce } => nonce,
-            AccountEmpty => 0,
-        }
+        0
+        // FIXME
```
kpp, line 600: TODO

crates/sovereign-sdk/module-system/sov-modules-rollup-blueprint/src/lib.rs
```diff
@@ -183,14 +177,26 @@ pub trait RollupBlueprint: Sized + Send + Sync {
             },
         };
 
+        // TODO: Decide what to do with this.
+        let prover_storage_c = prover_storage.clone();
```
kpp, line 181: TODO?

crates/sequencer/src/sequencer.rs
```diff
@@ -595,10 +600,24 @@ where
...
 
+    fn update_rpc_storage(&mut self, l2_height: u64) -> Result<(), anyhow::Error> {
+        let new_rpc_storage = self
+            .storage_manager
+            .create_storage_on_l2_height(l2_height)?;
+        // `send_replace` is superior to `send` for our use case. It never fails
+        // because it doesn't need to notify all receivers, unlike `send`, which
+        // we don't need. It will also keep working even if there are no
+        // receivers currently alive, which makes it easier to reason about the
+        // code.
```
rakanalh, line 611: - What is "our use case" here? 
- "unlike `send` which we don't need" - Sentence not needed when we clarify what the use case is.

crates/sovereign-sdk/full-node/sov-stf-runner/src/runner.rs
```diff
@@ -900,6 +908,20 @@ where
...
 
+    fn update_rpc_storage(&mut self, l2_height: u64) -> Result<(), anyhow::Error> {
+        let new_rpc_storage = self
+            .storage_manager
+            .create_storage_on_l2_height(l2_height)?;
+        // `send_replace` is superior to `send` for our use case. It never fails
+        // because it doesn't need to notify all receivers, unlike `send`, which
+        // we don't need. It will also keep working even if there are no
+        // receivers currently alive, which makes it easier to reason about the
+        // code.
```
rakanalh, line 919: Same as above regarding the comment.

crates/sovereign-sdk/module-system/sov-modules-rollup-blueprint/src/lib.rs
```diff
@@ -183,14 +177,26 @@ pub trait RollupBlueprint: Sized + Send + Sync {
             },
         };
 
+        // TODO: Decide what to do with this.
```
rakanalh, line 180: Can you please clarify what's wrong with the following line? The comment is a bit vague as to why we need to do something at all.

rakanalh (commented):
Overall LGTM with some comments.

crates/sovereign-sdk/module-system/sov-modules-rollup-blueprint/src/lib.rs
kpp: Well. I need a storage to create: `DbProvider::new(storage.clone())`. The question where to build it - in sequencer.rs or here in in sov rollup blueprint. If inside sequencer then it already works. If inside sov rollup blueprint (which I would prefer) - we diverge from sov code more and more.

crates/sequencer/src/sequencer.rs
kpp: Our case means that `send` will fail if a recv part is dead - that recv part is rpc module.

crates/sovereign-sdk/module-system/sov-modules-rollup-blueprint/src/lib.rs
rakanalh: Understood, i suggest adding what you wrote to the code comment so that it's more clear to whoever is reading. Unless you intend to resolve this TODO in this PR?

crates/sequencer/src/sequencer.rs
rakanalh: Sure, i would add that to the comment replacing "our use case"

crates/sequencer/src/sequencer.rs
```diff
@@ -556,10 +561,10 @@ where
 
     /// Signs batch of messages with sovereign priv key turns them into a sov blob
     /// Returns a single sovereign transaction made up of multiple ethereum transactions
-    fn make_blob(&mut self, raw_message: Vec<u8>) -> Vec<u8> {
+    fn make_blob(&mut self, raw_message: Vec<u8>, storage: &Sm::NativeStorage) -> Vec<u8> {
```
eyusufatik, line 564: why are we passing down storage instead of using self.storage?

crates/sovereign-sdk/module-system/sov-modules-rollup-blueprint/src/lib.rs
kpp: I think yes, we need to resolve it here

crates/sequencer/src/sequencer.rs
kpp: Because there is no more self.storage

rakanalh (dismissed):
LGTM

PR #469, opened 2024-04-30
Add Deposit Mempool

otaliptus:
# Description
Implements a basic FIFO mempool mechanism for deposit transactions.

### DIFF
--- a/bin/citrea/configs/mock/sequencer_config.toml
+++ b/bin/citrea/configs/mock/sequencer_config.toml
@@ -10,4 +10,3 @@ queue_tx_size = 200
 base_fee_tx_limit = 100000
 base_fee_tx_size = 200
 max_account_slots = 16
-deposit_mempool_fetch_limit = 10
--- a/bin/citrea/configs/mock/sequencer_config.toml
+++ b/bin/citrea/configs/mock/sequencer_config.toml
@@ -1,5 +1,7 @@
 min_soft_confirmations_per_commitment = 10
 test_mode = false
+deposit_mempool_fetch_limit = 10
+
 [mempool_conf] # Mempool Configuration - https://github.com/ledgerwatch/erigon/wiki/Transaction-Pool-Design
 pending_tx_limit = 100000
 pending_tx_size = 200
@@ -8,3 +10,4 @@ queue_tx_size = 200
 base_fee_tx_limit = 100000
 base_fee_tx_size = 200
 max_account_slots = 16
+deposit_mempool_fetch_limit = 10
--- a/bin/citrea/publish_da_block.sh
+++ b/bin/citrea/publish_da_block.sh
@@ -1,4 +1,4 @@
-# script to auto send 'eth_publishBatch' requests every 2 seconds
+# script to auto send 'citrea_testPublishBlock' requests every 2 seconds
 # TODO: read sequencer url from .toml files
 
 SLEEP_DURATION=60
--- a/bin/citrea/publish_da_block.sh
+++ b/bin/citrea/publish_da_block.sh
@@ -1,4 +1,4 @@
-# script to auto send 'citrea_testPublishBlock' requests every 2 seconds
+# script to auto send 'da_publishBlock' requests every 2 seconds
 # TODO: read sequencer url from .toml files
 
 SLEEP_DURATION=60
--- a/crates/sequencer/src/deposit_data_mempool.rs
+++ b/crates/sequencer/src/deposit_data_mempool.rs
@@ -25,6 +25,11 @@ impl DepositDataMempool {
         }
     }
 
+    // Given the fact that sys txs are included in the block gas limit, we should be careful to not to block other transactions
+    // + a sys tx should not fail here, so we also want to limit it to not to fail any sys tx at the beginning of the block
+    // (i.e. if you have 500 dep tx, due to gas, they may not be included, so it panics - we don't want that)
+
+    // Considering the deposit amounts to be allowed, and the block count, a limit per block is convenient
     pub fn fetch_deposits(&mut self, limit_per_block: usize) -> Vec<Vec<u8>> {
         let mut deposits = Vec::new();
         for _ in 0..limit_per_block {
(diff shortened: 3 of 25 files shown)


bin/citrea/configs/mock/sequencer_config.toml
eyusufatik, line 1: why some confs have this deposit_mempoo_fetch_limit param twice?

crates/sequencer/src/rpc.rs
```diff
@@ -92,5 +96,38 @@ pub(crate) fn create_rpc_module<C: sov_modules_api::Context>(
             },
         }
     })?;
+
+    rpc.register_async_method("eth_depositTransaction", |parameters, ctx| async move {
```
eyusufatik, line 100: let's call this citrea_sendRawDepositTransaction

crates/sequencer/src/rpc.rs
```diff
@@ -92,5 +96,38 @@ pub(crate) fn create_rpc_module<C: sov_modules_api::Context>(
...
+        let mut params = parameters.sequence();
+        let deposits: Vec<Vec<u8>> = params.next().unwrap();
```
eyusufatik, line 102: we want to parse something like this

"0x213123123"

I don't think it'll work

crates/sequencer/src/rpc.rs
```diff
@@ -92,5 +96,38 @@ pub(crate) fn create_rpc_module<C: sov_modules_api::Context>(
...
+
+        info!("Sequencer: eth_depositTransaction");
+
+        let evm = Evm::<C>::default();
+        let mut working_set = WorkingSet::<C>::new(ctx.storage.clone());
+
+        for deposit in deposits {
```
eyusufatik, line 109: accept single deposit here

crates/sequencer/src/deposit_data_mempool.rs
```diff
@@ -0,0 +1,46 @@
+use std::collections::VecDeque;
+
+use citrea_evm::system_contracts::Bridge;
+use reth_primitives::{self, address};
+use reth_rpc_types::{TransactionInput, TransactionRequest};
+use sov_rollup_interface::rpc::HexTx;
+
+#[derive(Clone, Debug)]
+pub struct DepositDataMempool {
+    accepted_deposit_txs: VecDeque<HexTx>,
```
eyusufatik, line 10: no HexTx 
just plain Vec<u8>

crates/sequencer/src/deposit_data_mempool.rs
```diff
@@ -0,0 +1,46 @@
...
+    pub fn make_deposit_tx_from_data(&mut self, deposit_tx_data: HexTx) -> TransactionRequest {
+        TransactionRequest {
+            from: Some(address!("deaddeaddeaddeaddeaddeaddeaddeaddeaddead")),
+            to: Some(Bridge::address()),
+            input: TransactionInput::new(Bridge::deposit(deposit_tx_data.tx)),
+            ..Default::default()
+        }
+    }
+
+    pub fn fetch_deposits(&mut self) -> Vec<Vec<u8>> {
```
eyusufatik, line 31: if you get limit in this function instead of having it as a struct field, we'll have more flexibility

eyusufatik (changes requested):

bin/citrea/publish_da_block.sh
```diff
@@ -1,4 +1,4 @@
-# script to auto send 'eth_publishBatch' requests every 2 seconds
+# script to auto send 'citrea_testPublishBlock' requests every 2 seconds
```
eyusufatik, line 1: check below in this file and fix pls 

funny mistake I did before :)

crates/sequencer/src/deposit_data_mempool.rs
```diff
@@ -0,0 +1,45 @@
+use std::collections::VecDeque;
+
+use citrea_evm::system_contracts::Bridge;
+use reth_primitives::{self, address};
+use reth_rpc_types::{TransactionInput, TransactionRequest};
+
+#[derive(Clone, Debug)]
+pub struct DepositDataMempool {
+    accepted_deposit_txs: VecDeque<Vec<u8>>,
+    limit_per_block: usize,
```
kpp, line 10: Why do we need limit per block? Add comment plz

crates/sequencer/src/rpc.rs
```diff
@@ -92,5 +95,39 @@ pub(crate) fn create_rpc_module<C: sov_modules_api::Context>(
...
+            let evm = Evm::<C>::default();
+            let mut working_set = WorkingSet::<C>::new(ctx.storage.clone());
+
+            let dep_tx = ctx
+                .deposit_mempool
+                .lock()
+                .await
+                .make_deposit_tx_from_data(deposit.clone().into());
+
+            let tx_res = evm.get_call(dep_tx, None, None, None, &mut working_set);
```
kpp, line 116: Why do we need to check it? Please add comment

eyusufatik: so basically our bridge backend is going to send calls to the bridge contract to our sequencer to be run as system transactions. we want to check if its not working in order not to include a failing tx in our soft confirmations.

crates/sequencer/src/deposit_data_mempool.rs
otaliptus: Given the fact that sys txs are included in the block gas limit, we should be careful to not to block other transactions + a sys tx should not fail here, so we also want to limit it to not to fail any sys tx at the beginning of the block (i.e. if you have 500 dep tx, due to gas, they may not be included, so it panics - we don't want that)

Considering the deposit amounts to be allowed, and the block count, this is convenient

crates/sequencer/src/deposit_data_mempool.rs
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    // Given the fact that sys txs are included in the block gas limit, we should be careful to not to block other transactions
+    // + a sys tx should not fail here, so we also want to limit it to not to fail any sys tx at the beginning of the block
+    // (i.e. if you have 500 dep tx, due to gas, they may not be included, so it panics - we don't want that)
+
+    // Considering the deposit amounts to be allowed, and the block count, a limit per block is convenient
+    pub fn fetch_deposits(&mut self, limit_per_block: usize) -> Vec<Vec<u8>> {
+        let mut deposits = Vec::new();
+        for _ in 0..limit_per_block {
```
kpp, line 35: https://doc.rust-lang.org/stable/std/collections/struct.VecDeque.html#method.drain

eyusufatik (dismissed):

kpp (dismissed):

kpp (approved):

eyusufatik (approved):

PR #470, opened 2024-04-30
Improve sequencer error handling

### LINKED ISSUE #333: Improvie error handling in sequencer
## Summary

The sequencer crate uses `unwrap` all over the place, though `anyhow::Result` is an expected return value.

We should come up with a better error handling by clearing our all unwraps for actual errors, either by using `bail!` from `anyhow` or creating custom error enum/struct that provide detailed error information.

rakanalh:
# Description
Replace all `unwrap` and `expect` calls with proper error propagation using anyhow.

## Linked Issues
- Fixes #333


crates/sequencer/src/db_provider/mod.rs
```diff
@@ -32,27 +33,23 @@ impl<C: sov_modules_api::Context> DbProvider<C> {
...
-        let rich_block = self
-            .evm
-            .get_block_by_number(None, None, &mut working_set)
-            .unwrap()
-            .unwrap();
-        match rich_block.inner.transactions {
-            BlockTransactions::Hashes(hashes) => hashes,
-            _ => vec![],
+        let rich_block = self.evm.get_block_by_number(None, None, &mut working_set)?;
+        let hashes = rich_block.map(|b| b.inner.transactions);
```
kpp, line 39: No need to map if you have match over Some/None later. So either map or match IMO

rakanalh: - `rich_block` before map is `Option<Rich<Block>>`. 
- `hashes` becomes after map `Option<BlockTransactions>`
- The match checks that we only return hashes and an empty vector otherwise (for `None` and for `BlockTransaction` other variants)

rakanalh:
@eyusufatik since the sequencer run method:
https://github.com/chainwayxyz/citrea/blob/f28eda7d2f40c4ac64df01d6caf985f23e91cdfb/crates/sequencer/src/sequencer.rs#L493-L512

Is propagating errors to main, i suggest we also `tracing::error!` log the errors inside the loop. Otherwise, main would also propagate them upwards which means that they're never seen. what do you think?

eyusufatik:
> @eyusufatik since the sequencer run method:
> 
> https://github.com/chainwayxyz/citrea/blob/f28eda7d2f40c4ac64df01d6caf985f23e91cdfb/crates/sequencer/src/sequencer.rs#L493-L512
> 
> Is propagating errors to main, i suggest we also `tracing::error!` log the errors inside the loop. Otherwise, main would also propagate them upwards which means that they're never seen. what do you think?

I think we should not propagate errors to main, we should handle this differently.

Imagine something errors out in sequencer, would we like it to shutdown? 

No. 

What can we do?

Inspect the problem and do something differently or try it N times. etc.

eyusufatik (approved):

PR #472, opened 2024-05-01
Implement batch explorer api

ercecan:
# Description
Implements batch explorer api


batch-explorer/README.md
eyusufatik, line 1: adding a small readme explaining why we have it here would be nice

batch-explorer/app.py
```diff
@@ -0,0 +1,62 @@
...
+
+app = FastAPI(lifespan=lifespan)
+
+app.add_middleware(TrustedHostMiddleware, allowed_hosts=["*"])
+
+
+origins = [
+    "*",
+]
+
```
eyusufatik, line 39: should be settable by config

batch-explorer/services/l2_rpc.py
eyusufatik, line 1: can use ethers-py here

batch-explorer/dtos/commitment.py
```diff
@@ -0,0 +1,23 @@
+from pydantic import BaseModel
+
+
+class CommitmentResponse(BaseModel):
+    l1_tx_id: str
+    l1_start_hash_mempool_url: str
+    l1_end_hash_mempool_url: str
+    l2_start_height_block_exp_url: str
+    l2_end_height_block_exp_url: str
+    merkle_root: str
```
eyusufatik, line 10: return heights and hashes as well along with urls to explorers

eyusufatik (changes requested):

batch-explorer/services/l2_rpc.py
ercecan: that is not very well maintained I believe,
I can use web3py?

Issue #477, opened 2024-05-03
Remove failed transactions from mempool as well


ercecan:
## Summary
If transaction fails we are not removing it from mempool, we should also remove the failing transactions from the mempool.

### Example scenario where this is needed
User can dos mempool by sending a tx with enough gas*price balance but not enough l1 fee rate and this tx can stay in mempool forever and never land in a block

eyusufatik:
should have been closed in #487

PR #487, opened 2024-05-04
Esad/detect l1 fee fail txs

eyusufatik:
# Description
If a transaction fails due to L1 fee, we don't want to rerun it, or we try that every block which is DOS. So we remove them from the mempool after its first execution.

no merge yet. waiting on some confirmations on writing to accessory state in sov-module calls. 

## Linked Issues
#477

### DIFF
--- a/crates/evm/src/call.rs
+++ b/crates/evm/src/call.rs
@@ -222,6 +222,12 @@ impl<C: sov_modules_api::Context> Evm<C> {
                         continue;
                     }
                     EVMError::Custom(msg) => {
+                        #[cfg(feature = "native")]
+                        if !msg.starts_with("Gas used") {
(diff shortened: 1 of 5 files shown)


crates/evm/src/call.rs
```diff
@@ -222,6 +222,13 @@ impl<C: sov_modules_api::Context> Evm<C> {
                         continue;
                     }
                     EVMError::Custom(msg) => {
+                        #[cfg(feature = "native")]
+                        if !msg.starts_with("Gas used") {
+                            tracing::error!("native native");
```
ercecan, line 227: native antive

ercecan (approved):

otaliptus (approved):
Wait for CI

PR #495, opened 2024-05-06
Use eth_maxPriorityFeePerGas and eth_maxFeePerGas in test client

kpp:
# Description
A small improvement


kpp:
Seems like there is no interest in this PR =)

PR #497, opened 2024-05-06
Add docker compose for postgres

### LINKED ISSUE #496: Update sequencer read me and have a docker-compose for postgres
Since we've implemented PostgreSQL for backup and monitoring of our sequencer, it's important to update the README to include information about PostgreSQL. Additionally, we can deploy Postgres using a docker-compose.yml file as suggested by @rakanalh.

rakanalh:
# Description

Makes it easy to run postgres as storage for the node.

Fixes #496


Makefile
```diff
@@ -14,6 +14,7 @@ clean: ## Cleans compiled
 	@cargo clean
 
 clean-node: ## Cleans local dbs needed for sequencer and nodes
+	sudo rm -rf data/postgres
```
ercecan, line 17: We should also ignore data folder in hive/dockerignore

rakanalh: `.dockerignore` file added.

docs/run.md
```diff
@@ -11,6 +11,16 @@ Build citrea:
 make build
 ```
 
+### Prequisites
+
+In order for the sequencer node to run, you need to have Postgres up and running.
```
eyusufatik, line 16: this is not needed - if no db_config is given in the sequencer_config.toml, then the sequencer doesnt need postgres

eyusufatik (changes requested):

docs/run.md
eyusufatik: I think we can just skip db stuff in readmes, only we are gonna use that feature of the sequencer

docs/run.md
```diff
@@ -11,6 +11,18 @@ Build citrea:
 make build
 ```
 
+### Prequisites
+
+In order for the sequencer node to run with the storage DB is needed.
```
eyusufatik, line 16: For production use cases, we leverage PostgreSQL for a few extra features in the sequencer. These features are optional, if you don't want to run them, make sure the `sequencer_config.toml` file does not contain the `db_config` altogether in order to skip using a storage DB backend such as postgres.

eyusufatik (changes requested):

eyusufatik (approved):

PR #511, opened 2024-05-06
Erce/eventually consistent backup mempool

### LINKED ISSUE #478: Store mempool transactions in shared backup db
We should store mempool transactions in db

Let's discuss how we can store them in the most optimal way so that we can retrieve them if sequencer crashes and continue to publish the same block as if the crash never happened.

ercecan:
# Description
Implements mempool backup&restore functionality in case of sequencer crash

## Linked Issues
- Fixes #478


bin/citrea/src/lib.rs
```diff
@@ -22,7 +22,7 @@ pub fn initialize_logging() {
         .with(
             EnvFilter::from_str(
                 &env::var("RUST_LOG")
-                    .unwrap_or_else(|_| "debug,hyper=info,risc0_zkvm=info".to_string()),
+                    .unwrap_or_else(|_| "warn,hyper=info,risc0_zkvm=info".to_string()),
```
eyusufatik, line 25: debug

crates/shared-backup-db/src/tables.rs
```diff
@@ -87,8 +89,24 @@ CREATE TABLE IF NOT EXISTS sequencer_commitments (
...
+    "CREATE INDEX IF NOT EXISTS idx_l1_end_height ON sequencer_commitments(l1_end_height);";
 pub const INDEX_L1_END_HASH: &str =
-    "CREATE INDEX idx_l1_end_hash ON sequencer_commitments(l1_end_hash);";
+    "CREATE INDEX IF NOT EXISTS idx_l1_end_hash ON sequencer_commitments(l1_end_hash);";
+
+// tx is rlp encoded
+pub const MEMPOOL_TXS_TABLE_CREATE_QUERY: &str = "
+CREATE TABLE IF NOT EXISTS mempool_txs (
+    id          SERIAL PRIMARY KEY,
+    tx_hash     BYTEA NOT NULL,
```
eyusufatik, line 102: unique maybe ?

crates/sequencer/src/sequencer.rs
```diff
@@ -769,6 +824,39 @@ where
...
+                    let pooled_tx =
+                        EthPooledTransaction::new(signed_ec_recovered, length_without_header);
+
+                    // TODO Handle error
+                    let _ = self
+                        .mempool
+                        .add_external_transaction(pooled_tx)
+                        .await
+                        .unwrap();
+                }
```
eyusufatik, line 850: why so different from in rpc?

eyusufatik (changes requested):

crates/sequencer/src/sequencer.rs
ercecan: turns I was using the wrong encoding when I should have used encode_enveloped and same for decoding.
Fixed it

eyusufatik (dismissed):

crates/shared-backup-db/src/postgres_connector.rs
```diff
@@ -1,82 +1,94 @@
...
-            if let Err(e) = connection.await {
-                eprintln!("connection error: {}", e);
-            }
-        });
+    pub async fn new_test_client() -> Result<Self, PoolError> {
+        let mut cfg = PgConfig::new();
+        cfg.host("localhost")
+            .port(5432)
+            .user("postgres")
+            .password("password")
```
kpp, line 83: It used to be postgres

ercecan: Yeah it should be don't know how I did that :D Thanks

crates/sequencer/src/sequencer.rs
```diff
@@ -392,7 +395,23 @@ where
                 let mut txs_to_remove = self.db_provider.last_block_tx_hashes()?;
                 txs_to_remove.extend(l1_fee_failed_txs);
 
-                self.mempool.remove_transactions(txs_to_remove);
+                self.mempool.remove_transactions(txs_to_remove.clone());
+
+                match pg_pool.clone() {
+                    Some(pg_pool) => {
+                        // TODO: Is this okay? I'm not sure because we have a loop in this and I can't do async in spawn_blocking
```
kpp: Not okay because we will have inconsistency with rocksdb and pg if a process is killed while removing txs.

eyusufatik: we don't put mempool txs to rocksdb btw, only written to memory

ercecan: what would you suggest not to have inconsistent pg? Is there a way to handle it better in sequencer? 
Maybe we can have a separate listener outside of sequencer and every time a block is published it removes the txs of block from pg ?

kpp (dismissed):

crates/shared-backup-db/Cargo.toml
```diff
@@ -12,8 +12,8 @@ publish = false
 resolver = "2"
 
 [dependencies]
-postgres = "0.19.7"
-tokio-postgres = "0.7.10"
+deadpool = "0.12.0"
```
ercecan, line 15: remove

eyusufatik (approved):

PR #530, opened 2024-05-10
Fix running tests

rakanalh:
# Description

With #479, running a specific test would fail because of missing `reth-rpc` dependency missing and `sov-attester-incentives` and `sov-prover-incentives` running without the `native` feature in tests.

This fix would by default run both crates with default `native` feature. Outside tests, if these crates are needed, we use `default-features = false`.


eyusufatik:
> ok, if it also works without skip guest build

seems to work.

PR #531, opened 2024-05-10
Async rocksdb ops

### LINKED ISSUE #528: Make Rocksdb blocking I/O async
This is to make sure blocking operations do not affect other components such as JSONRPC,

rakanalh:
# Description

Makes calls to rocksdb asynchronous. 

`block_in_place`, when called, uses a thread that is specifically spawned for blocking operations. This basically helps non-blocking threads to continue to run async code without rocksdb starving those threads and preventing futures from progressing forward.

## Linked Issues
- Fixes #528


eyusufatik:
If these block_in_place functions create threads, then wouldn't this slow down execution of blocks?

I mean when RPC is involved I assume this will bring a lot of benefits, but I think we might be better off if we implement two different NativeStorages, one for RPC which has block_in_place, one for block execution without block_in_place.

How would that work?

eyusufatik:
> If these block_in_place functions create threads, then wouldn't this slow down execution of blocks?
> 
> I mean when RPC is involved I assume this will bring a lot of benefits, but I think we might be better off if we implement two different NativeStorages, one for RPC which has block_in_place, one for block execution without block_in_place.
> 
> How would that work?

Forgot about `crates/sovereign-sdk/module-system/sov-modules-macros/src/rpc/expose_rpc.rs`.

Maybe we can modify this module to use block_in_place, so the block production does not create threads but the RPC storage does

rakanalh:
> If these block_in_place functions create threads, then wouldn't this slow down execution of blocks?
> 
> I mean when RPC is involved I assume this will bring a lot of benefits, but I think we might be better off if we implement two different NativeStorages, one for RPC which has block_in_place, one for block execution without block_in_place.
> 
> How would that work?

The idea is that you run with a multi-threaded tokio runtime which has a threadpool. Based on the [block_in_place](https://docs.rs/tokio/latest/tokio/task/fn.block_in_place.html), it mentions that whenever `block_in_place` is called, other tasks in the current thread are moved to other threads so that the task we request to be blocked will not prevent other tasks from being driven forward.

rakanalh:
> > If these block_in_place functions create threads, then wouldn't this slow down execution of blocks?
> > I mean when RPC is involved I assume this will bring a lot of benefits, but I think we might be better off if we implement two different NativeStorages, one for RPC which has block_in_place, one for block execution without block_in_place.
> > How would that work?
> 
> Forgot about `crates/sovereign-sdk/module-system/sov-modules-macros/src/rpc/expose_rpc.rs`.
> 
> Maybe we can modify this module to use block_in_place, so the block production does not create threads but the RPC storage does

I'll check that out. However, we're not creating threads. By default, in `bin/citrea` we're using `tokio::main` for the main function which by default uses a multi-threaded runtime (number of threads = num of cpus). Which means that different tasks will run in different threads anyway. The threads themselves are created and added into the threadpool at the beginning which means that we won't incur the cost of spawning a thread every time we call `block_in_place`.

kpp:
Tokio spawns several OS worker threads for IO futures - lightweight "nonblocking" futures (here goes all tokio::spawn stuff) and several OS worker threads for CPU bound "blocking" things (here goes all spawn_blocking stuff).

kpp (changes requested):
A lot of code changes in this PR can be reduced by renaming old methods with a _blocking suffix and introducing new methods which call _blocking methods inside of block_in_place

rakanalh:
Closing in favor of #544

Issue #529, opened 2024-05-10
BitcoinDA force order of txs


kpp:
BitcoinDA enforcing order of transactions sent (by using outputs of the previous tx).

To make txs, the bitcoin da adapter select random utxos as inputs. But we’d actually want the adapter to send transactions using the output of the previous transaction so the sequencer commitments land in blocks in order.

The first tx can select a utxo same as is - the way it works now.

u1 u2 u3 u4 u5 u6tx1
 
u1 + u2 ==> u7tx2
u3 + u4 ==> u8

tx2
u7 + (other utxos if we need to cover fees) ===> u8

PR #535, opened 2024-05-10
Bitcoin DA tx order

kpp:
# Description

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/529

## Testing
`make test`


kpp:
6d14dc26f46 breaks tests =(

crates/bitcoin-da/src/service.rs
```diff
@@ -424,28 +485,20 @@ impl DaService for BitcoinService {
 
     async fn send_transaction(
         &self,
-        blob: &[u8],
+        _blob: &[u8],
     ) -> Result<<Self as DaService>::TransactionId, Self::Error> {
-        let fee_sat_per_vbyte = self.get_fee_rate().await?;
-        self.send_transaction_with_fee_rate(blob, fee_sat_per_vbyte)
-            .await
+        unimplemented!("Use send_tx_no_wait instead")
```
eyusufatik, line 490: why?


use send_tx_no_wait and wait on the receiver :D

crates/bitcoin-da/src/spec/utxo.rs
```diff
@@ -1,7 +1,7 @@
 use bitcoin::Txid;
 use serde::{Deserialize, Serialize};
 
-#[derive(Clone, Debug, PartialEq, Serialize)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Hash)]
```
eyusufatik, line 4: Why is derive Hash is added?

If we need it, we can use txid and vout as it's unique identifier.

eyusufatik (commented):
great PR

crates/bitcoin-da/src/spec/utxo.rs
kpp: Oh, I was playing with hashmaps. Will fix

crates/bitcoin-da/src/service.rs
kpp: Thats right, one can wait on the receiver anywhere in the code, not to .await at the call cite

eyusufatik (approved):

PR #536, opened 2024-05-10
Add `ethers.js` tests

okkothejawa:
# Description
Adds tests that checks the Citrea RPCs against the RPC methods of `ethers.js v6`. The following `ethers.js` methods are tested:

- `getBlockNumber`
- `getBalance`
- `getNetwork`
- `getFeeData`
- `estimateGas`
- `call`
- `broadcastTransaction`
- `getBlock`
- `getCode`
- `getStorage`
- `getLogs`
- `getTransaction`
- `getTransactionCount`
- `getTransactionReceipt`

Additionally `WETH9` is deployed through `ethers.js` and ether is deposited into it, and both the deployment and the interactions are verified through `ContractInteractionTests`.

Tests can be run executing `npm install` and `npx mocha test.js` in the `ethers.js_tests` folder.


eyusufatik:
Let's write a CI test for this

bin/citrea/tests/evm/ethers.js_tests/test.js
```diff
@@ -0,0 +1,340 @@
...
+        const bitcoinAddress = ethers.encodeBytes32String('bc1qa0a0a0a0a0a0a0a0a0a0a0a0');
+    
+        let tx = {
+            to: contractAddress,
+            value: ethers.parseEther('0.9'),
+            data: contract.interface.encodeFunctionData('withdraw', [bitcoinAddress]),
+            from: wallet.address
+        };
+
+        try {
```
eyusufatik, line 223: isnt there a better parsed ethers.js error object?

does it really just throw the erros like this?

have you tried the same thing on another rpc or reth?

eyusufatik (approved):

PR #544, opened 2024-05-13
Async RPC storage access

### LINKED ISSUE #528: Make Rocksdb blocking I/O async

### LINKED ISSUE #563: Create tempdir for tests
The following directories are created for specific tests:
- `bin/citrea/demo_data_sequencer_full_node/`
- `bin/citrea/demo_data_sequencer_restore_mempool/`
- `bin/citrea/demo_data_test_reopen_prover/`
- `bin/citrea/proptest-regressions/`

We should use `tempdir` crate to create temporary directories in `/tmp` so that we don't fill the project's directory with test ones.

rakanalh:
# Description

- Use `block_in_place` for RPC storage access
- Use `tempfile::tempdir` to generate temporary directories for tests
- Pass DB directory around for use by tests. `tempfile` cleans up the created tempdir after falling out of scope (after test is done).
- Add `wait_for_l2_batch` and `wait_for_l1_block` instead of calling `sleep`

## Linked Issues
- Fixes #528 
- Fixes #563 
- Related #593


crates/sequencer/src/rpc.rs
```diff
@@ -80,17 +82,21 @@ pub(crate) fn create_rpc_module<C: sov_modules_api::Context>(
...
-            let da = MockDaService::new(MockAddress::from([0; 32]));
-            da.publish_test_block().await.map_err(|e| {
-                ErrorObjectOwned::owned(
-                    INTERNAL_ERROR_CODE,
-                    INTERNAL_ERROR_MSG,
-                    Some(format!("Could not publish mock-da block: {e}")),
-                )
-            })?;
-            Ok::<(), ErrorObjectOwned>(())
+        let da_db_path = da_db_path.expect("DA DB should be set");
```
kpp, line 85: Put inside context?

rakanalh: We want to panic in the initialization phase if db path is not specified. IMO, better than panicing when the handler is called.

PR #565, opened 2024-05-14
fix l1 fee rate selection and error handling in sequencer

eyusufatik:
# Description
Sequencer now checks l1 fee rate is in the allowable range by SoftConfirmationRuleEnforcer.

Also we realized the produce_l2block does not return error on begin_soft_confirmation_hook error, which causes problems

### DIFF
--- a/crates/sequencer/src/sequencer.rs
+++ b/crates/sequencer/src/sequencer.rs
@@ -476,19 +476,13 @@ where
 
         let fee_rate_range = self.get_l1_fee_rate_range()?;
 
-        let mut l1_fee_rate = self
+        let l1_fee_rate = self
             .da_service
             .get_fee_rate()
             .await
             .map_err(|e| anyhow!(e))?;
 
-        l1_fee_rate = if l1_fee_rate < *fee_rate_range.start() {
-            *fee_rate_range.start()
-        } else if l1_fee_rate > *fee_rate_range.end() {
-            *fee_rate_range.end()
-        } else {
-            l1_fee_rate
-        };
+        let l1_fee_rate = l1_fee_rate.clamp(*fee_rate_range.start(), *fee_rate_range.end());
 
         let new_da_block = match last_finalized_height.cmp(&prev_l1_height) {
             Ordering::Less => {
--- a/crates/sequencer/src/sequencer.rs
+++ b/crates/sequencer/src/sequencer.rs
@@ -18,6 +18,7 @@ use reth_primitives::{FromRecoveredPooledTransaction, IntoRecoveredTransaction};
 use reth_provider::BlockReaderIdExt;
 use reth_transaction_pool::{BestTransactionsAttributes, EthPooledTransaction, PoolTransaction};
 use shared_backup_db::{CommitmentStatus, PostgresConnector, SharedBackupDbConfig};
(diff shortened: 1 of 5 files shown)


crates/sequencer/src/sequencer.rs
```diff
@@ -464,12 +474,22 @@ where
...
-        let l1_fee_rate = self
+        let fee_rate_range = self.get_l1_fee_rate_range()?;
+
+        let mut l1_fee_rate = self
             .da_service
             .get_fee_rate()
             .await
             .map_err(|e| anyhow!(e))?;
 
+        l1_fee_rate = if l1_fee_rate < *fee_rate_range.start() {
```
kpp, line 485: Why not clamp? =)

crates/sequencer/src/sequencer.rs
```diff
@@ -464,12 +474,22 @@ where
             last_finalized_height
         );
 
-        let l1_fee_rate = self
+        let fee_rate_range = self.get_l1_fee_rate_range()?;
+
+        let mut l1_fee_rate = self
```
kpp, line 479: rust allows "redefining" a variable. Remove this let and create a new l1_fee_rate below

crates/soft-confirmation-rule-enforcer/src/query.rs
```diff
@@ -52,4 +54,30 @@ impl<C: Context, Da: DaSpec> SoftConfirmationRuleEnforcer<C, Da> {
...
+
+    /// function to get min and max for next L1 fee rate
+    pub fn get_next_min_max_l1_fee_rate(
+        &self,
+        working_set: &mut WorkingSet<C>,
+    ) -> RpcResult<RangeInclusive<u128>> {
+        let last_l1_fee_rate = self.last_l1_fee_rate.get(working_set).unwrap_or(0);
+
+        if last_l1_fee_rate == 0 {
+            // on the first soft confirmtaion, we don't have a last fee rate
```
kpp, line 66:
```suggestion
            // on the first soft confirmation, we don't have a last fee rate
```

crates/sequencer/src/sequencer.rs
eyusufatik: bc I couldn't find the proper clamp 😓

crates/sequencer/src/sequencer.rs
```diff
@@ -464,12 +474,22 @@ where
...
             .await
             .map_err(|e| anyhow!(e))?;
 
+        l1_fee_rate = if l1_fee_rate < *fee_rate_range.start() {
+            *fee_rate_range.start()
+        } else if l1_fee_rate > *fee_rate_range.end() {
+            *fee_rate_range.end()
+        } else {
+            l1_fee_rate
+        };
```
kpp, line 491:
```suggestion
        let fee_rate_range = self.get_l1_fee_rate_range()?;

        let l1_fee_rate = self
            .da_service
            .get_fee_rate()
            .await
            .map_err(|e| anyhow!(e))?;

        let l1_fee_rate = if l1_fee_rate < *fee_rate_range.start() {
            *fee_rate_range.start()
        } else if l1_fee_rate > *fee_rate_range.end() {
            *fee_rate_range.end()
        } else {
            l1_fee_rate
        };
```

kpp (approved):

Issue #562, opened 2024-05-14
EVM: L1 fee for tx resul Halt is 0


eyusufatik:
For success and revert we calculate correct diff size and l1 fee, but for txs that halted (gas finished without contract execution ending) l1 fee is calculated to be 0, which is not 0 a tx on a valid block can halt and the state of the account would change. (Pay full gas, no contract state changes but account nonce is incremented by 1)

PR #571, opened 2024-05-15
Pay l1 fee for a halted tx

kpp:
# Description

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/562

## Testing
`make test`


Makefile
```diff
@@ -23,7 +23,7 @@ test-legacy: ## Runs test suite with output from tests printed
 	@cargo test -- --nocapture -Zunstable-options --report-time
 
 test: $(EF_TESTS_DIR) ## Runs test suite using next test
-	@cargo nextest run --workspace --all-features --no-fail-fast
+	@cargo nextest run --workspace --all-features --no-fail-fast $(filter-out $@,$(MAKECMDGOALS))
```
eyusufatik, line 26: what is this?

crates/evm/src/evm/handler.rs
```diff
@@ -356,19 +356,16 @@ impl<SPEC: Spec, EXT: CitreaExternalExt, DB: Database> CitreaHandler<SPEC, EXT,
         let l1_fee_rate = context.external.l1_fee_rate();
         let l1_fee = U256::from(diff_size) * U256::from(l1_fee_rate);
         context.external.set_tx_info(TxInfo { diff_size, l1_fee });
-        if result.interpreter_result().is_ok() {
```
eyusufatik, line 359: in post execution output, we can't have txs that failed right?

assume we run a tx with wrong nonce, that doesnt come into the post execution output, right?

kpp: Right, it fails far before execution, it's basically free for us

eyusufatik (approved):

PR #577, opened 2024-05-16
fix for bitcoin da get block by hash

eyusufatik:
endinanness problem


kpp (approved):

crates/bitcoin-da/src/service.rs
```diff
@@ -546,9 +546,9 @@ impl DaService for BitcoinService {
     async fn get_block_by_hash(&self, hash: [u8; 32]) -> Result<Self::FilteredBlock, Self::Error> {
```
kpp, line 546: It would be better to also accept BlockHash as an arg

eyusufatik: needs change on BlockHashTrait. too much work for now.

PR #609, opened 2024-05-20
Add `web3.py` tests

okkothejawa:
# Description
Achieves the same purpose as #536 but for `web3.py`.


kpp (dismissed):

bin/citrea/tests/evm/web3.py_tests/test.py
```diff
@@ -0,0 +1,242 @@
...
+        # Check if 1 WETH is minted
+        self.assertEqual(contract.functions.balanceOf("0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266").call(), self.web3.to_wei(1, 'ether'))
+
+    def test_call_errors_correctly(self):
+        try:
+            selector = self.web3.keccak(text='ERRONEUS_FUNC()')[:4]
+            self.web3.eth.call({'value': 0, 'to': '0x3100000000000000000000000000000000000001', 'data': selector})
+            self.fail("Expected call to fail, but it succeeded.")
+        except Exception as e:
+            self.assertEqual(str(e), "('execution reverted', 'no data')")
```
eyusufatik, line 180: should we be comparing e.to_string()?

is there another error type?

we should be testing if our error can be converted to the expected type in the library

eyusufatik (approved):

Issue #509, opened 2024-05-06
Enhance logging


otaliptus:
## Description
We need to log a couple more stuff to trace how things pan out better, along with reconsidering the return part.

eyusufatik:
Changing this issue for general refactoring of our logs.

Let's do more & better logs, the tracing library has cool features, let's use them.

Span of this log refactoring includes the following paths and never includes tests:

- [ ] bin/citrea
- [ ] crates/bitcoin-da
- [ ] crates/citrea-stf
- [ ] crates/ethereum-rpc
- [ ] crates/evm
- [ ] crates/sequencer (very important)
- [ ] crates/sequencer-client (don't think much to log here)
- [ ] crates/shared-backup-db
- [ ] crates/soft-confirmation-rule-enforcer
- [ ] crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db
- [ ] crates/sovereign-sdk/full-node/sov-ledger-rpc
- [ ] crates/sovereign-sdk/full-node/sov-stf-runner
- [ ] crates/sovereign-sdk/module-system/module-implementations/sov-accounts
- [ ] crates/sovereign-sdk/module-system/sov-modules-rollup-blueprint
- [ ] crates/sovereign-sdk/module-system/sov-modules-stf-blueprint

PR #615, opened 2024-05-22
Add logs tier 2 and 3

kpp:
# Description
Adds more instrument attribute for better log fmt.

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/509

### DIFF
--- a/crates/sovereign-sdk/module-system/sov-modules-api/src/lib.rs
+++ b/crates/sovereign-sdk/module-system/sov-modules-api/src/lib.rs
@@ -13,6 +13,156 @@ mod reexport_macros;
 #[cfg(feature = "macros")]
 pub use reexport_macros::*;
 
+#[macro_export]
+macro_rules! native_debug {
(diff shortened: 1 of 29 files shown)


crates/sovereign-sdk/module-system/sov-modules-api/src/lib.rs
```diff
@@ -13,6 +13,156 @@ mod reexport_macros;
 #[cfg(feature = "macros")]
 pub use reexport_macros::*;
 
+#[macro_export]
+macro_rules! native_debug {
+    (?$($k:ident).+) => (
```
eyusufatik, line 18: have you considered putting these under sov-modules-macros?

kpp: I did but macro crate can only export procedural macros, not declarative macros

eyusufatik (approved):

PR #619, opened 2024-05-23
Retry all da-service requests

### LINKED ISSUE #618: StateTransitionRunner retry on da service errors
Scenario:

Full node queries a block height from the da service and the bitcoin node is down at that moment. Runner throws an error and the node shuts down.

Instead of this the runner should retry calls like get_block_at. (look out for other da_service calls, cant remember on the top of my head atm)

rakanalh:
# Description

Retries requests made to the DA service in exponential backoff fashion.

## Linked Issues
- Fixes #618

### DIFF
--- a/crates/sovereign-sdk/full-node/sov-stf-runner/src/prover_service/parallel/mod.rs
+++ b/crates/sovereign-sdk/full-node/sov-stf-runner/src/prover_service/parallel/mod.rs
@@ -1,9 +1,7 @@
 mod prover;
 use std::sync::Arc;
-use std::time::Duration;
 
 use async_trait::async_trait;
-use backoff::ExponentialBackoffBuilder;
 use borsh::BorshSerialize as _;
 use prover::Prover;
 use serde::de::DeserializeOwned;
@@ -161,10 +159,6 @@ where
         da_service: &Self::DaService,
     ) -> Result<(<Da as DaService>::TransactionId, Proof), anyhow::Error> {
         loop {
-            let exponential_backoff = ExponentialBackoffBuilder::new()
-                .with_initial_interval(Duration::from_secs(1))
-                .with_max_elapsed_time(Some(Duration::from_secs(5 * 60)))
-                .build();
             let status = self
                 .prover_state
                 .get_prover_status_for_da_submission(block_header_hash.clone())?;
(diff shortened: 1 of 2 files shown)


crates/sovereign-sdk/full-node/sov-stf-runner/src/prover_service/parallel/mod.rs
```diff
@@ -167,12 +173,16 @@ where
...
-                    let tx_id = da_service
-                        .send_transaction(
-                            da_data.try_to_vec().expect("Should serialize").as_slice(),
-                        )
-                        .await
-                        .map_err(|e| anyhow::anyhow!(e))?;
+                    let tx_id = backoff::future::retry(exponential_backoff, || async {
+                        da_service
+                            .send_transaction(
+                                da_data.try_to_vec().expect("Should serialize").as_slice(),
```
eyusufatik, line 179: @kpp given that we retry sending txs at bitcoin-da level, does it make sense to retry here?

rakanalh: We don't seem to be retrying sending transactions.

The `send_transaction` functions sends to the bitcoin-da receiver loop to `send_transaction_with_fee` which would happen only once and would not be retried.

1.
https://github.com/chainwayxyz/citrea/blob/2e7fa5afc35cdccf9938fe51af6b8261c78654c7/crates/bitcoin-da/src/service.rs#L502-L513

2.
https://github.com/chainwayxyz/citrea/blob/2e7fa5afc35cdccf9938fe51af6b8261c78654c7/crates/bitcoin-da/src/service.rs#L502-L513

3
https://github.com/chainwayxyz/citrea/blob/2e7fa5afc35cdccf9938fe51af6b8261c78654c7/crates/bitcoin-da/src/service.rs#L106-L138

eyusufatik: see in 3, if there is an error from any rpc calls, the loop continues to try

rakanalh: oh i see your point, we have
```
while receive {
   loop {}
}

rakanalh: It doesn't make sense to retry outside since the outside async task would retry sending through the channel in this case.

rakanalh: Removed

eyusufatik (approved):

PR #661, opened 2024-06-03
Esad/fix mempool maintenance

eyusufatik:
# Description
Some improvements to our mempool usage, not sure about account updates actually.

### DIFF
--- a/crates/sequencer/src/mempool.rs
+++ b/crates/sequencer/src/mempool.rs
@@ -7,9 +7,9 @@ use reth_tasks::TokioTaskExecutor;
 use reth_transaction_pool::blobstore::NoopBlobStore;
 use reth_transaction_pool::error::PoolError;
 use reth_transaction_pool::{
-    BestTransactions, BestTransactionsAttributes, CoinbaseTipOrdering, EthPooledTransaction,
-    EthTransactionValidator, Pool, PoolConfig, PoolResult, SubPoolLimit, TransactionPool,
-    TransactionValidationTaskExecutor, ValidPoolTransaction,
+    BestTransactions, BestTransactionsAttributes, ChangedAccount, CoinbaseTipOrdering,
+    EthPooledTransaction, EthTransactionValidator, Pool, PoolConfig, PoolResult, SubPoolLimit,
+    TransactionPool, TransactionPoolExt, TransactionValidationTaskExecutor, ValidPoolTransaction,
 };
 
 use crate::config::SequencerMempoolConfig;
@@ -115,6 +115,10 @@ impl<C: sov_modules_api::Context> CitreaMempool<C> {
         tx_hashes: Vec<TxHash>,
     ) -> Vec<Arc<ValidPoolTransaction<Transaction<C>>>> {
         self.0.remove_transactions(tx_hashes)
+    }
+
+    pub(crate) fn update_accounts(&self, account_updates: Vec<ChangedAccount>) {
+        self.0.update_accounts(account_updates);
     }
 
     pub(crate) fn best_transactions_with_attributes(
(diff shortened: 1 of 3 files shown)


crates/sequencer/src/sequencer.rs
```diff
@@ -962,4 +968,31 @@ where
             .get_next_min_max_l1_fee_rate(&mut working_set)
             .map_err(|e| anyhow::anyhow!("Error reading min max l1 fee rate: {}", e))
     }
+
+    fn get_account_updates(&self) -> Result<Vec<ChangedAccount>, anyhow::Error> {
+        let head = self.db_provider.last_block()?.expect("Head must exist");
```
rakanalh, line 973:
```suggestion
        let head = self.db_provider.last_block()?.ok_or_else(anyhow!("Could not retrieve last block"))?;
```

crates/sequencer/src/sequencer.rs
```diff
@@ -962,4 +968,31 @@ where
...
+                .basic_account(address)?
+                .expect("Account must exist");
+            updates.push(ChangedAccount {
+                address,
+                nonce: account.nonce,
+                balance: account.balance,
+            });
+        }
+
+        Ok(vec![])
```
rakanalh, line 996:
```suggestion
        Ok(updates)
```

rakanalh (changes requested):

crates/sequencer/src/mempool.rs
```diff
@@ -81,16 +81,14 @@ impl<C: sov_modules_api::Context> CitreaMempool<C> {
...
-                client,
-                Arc::new(chain_spec),
-                blob_store,
-                TokioTaskExecutor::default(),
-            ),
-            blob_store,
-            pool_config,
-        )))
+        let validator = TransactionValidationTaskExecutor::eth_builder(Arc::new(chain_spec))
+            .no_cancun()
```
rakanalh, line 85: I am curious was to how this affected us previously

crates/sequencer/src/sequencer.rs
```diff
@@ -962,4 +968,31 @@ where
             .get_next_min_max_l1_fee_rate(&mut working_set)
             .map_err(|e| anyhow::anyhow!("Error reading min max l1 fee rate: {}", e))
     }
+
+    fn get_account_updates(&self) -> Result<Vec<ChangedAccount>, anyhow::Error> {
+        let head = self.db_provider.last_block()?.expect("Head must exist");
+
+        let addresses: HashSet<Address> = match head.transactions {
+            reth_rpc_types::BlockTransactions::Full(ref txs) => {
+                txs.iter().map(|tx| tx.from).collect()
```
kpp, line 977: We already have a list of senders in `batch_receipt.tx_receipts`?

crates/sequencer/src/sequencer.rs
```diff
@@ -962,4 +968,31 @@ where
             .get_next_min_max_l1_fee_rate(&mut working_set)
             .map_err(|e| anyhow::anyhow!("Error reading min max l1 fee rate: {}", e))
     }
+
+    fn get_account_updates(&self) -> Result<Vec<ChangedAccount>, anyhow::Error> {
+        let head = self.db_provider.last_block()?.expect("Head must exist");
+
+        let addresses: HashSet<Address> = match head.transactions {
```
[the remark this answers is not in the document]
eyusufatik: no, native only since it's the sequencer

crates/sequencer/src/sequencer.rs
eyusufatik: those are for sovereign transactions

crates/sequencer/src/mempool.rs
eyusufatik: I guess the mempool would expect blob transactions but since we passed noop blob store never return them.

kpp (approved):

rakanalh (approved):

PR #674, opened 2024-06-06
exclude system transactions from tip cap calc.

eyusufatik:
# Description

Fixes priority fee calculation by ignoring system transactions, because it was bumping up suggested fee cap.

### DIFF
--- a/crates/ethereum-rpc/src/gas_price/gas_oracle.rs
+++ b/crates/ethereum-rpc/src/gas_price/gas_oracle.rs
@@ -8,7 +8,7 @@ use std::sync::Arc;
 use citrea_evm::Evm;
 use reth_primitives::basefee::calc_next_block_base_fee;
 use reth_primitives::constants::GWEI_TO_WEI;
-use reth_primitives::{BlockNumberOrTag, B256, U256};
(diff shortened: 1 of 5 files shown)


rakanalh (approved):
LGTM

crates/ethereum-rpc/src/gas_price/gas_oracle.rs
```diff
@@ -363,6 +363,7 @@ impl<C: sov_modules_api::Context> GasPriceOracle<C> {
                 // check if coinbase
                 let sender = tx.from;
                 sender != block.header.miner
+                    && sender != address!("deaddeaddeaddeaddeaddeaddeaddeaddeaddead")
```
rakanalh, line 366: Suggestion here:
- Make this address a constant value that we can import and use rather than hard code.

rakanalh: https://github.com/chainwayxyz/citrea/issues/675

kpp, line 366: There is a special constant for it

PR #677, opened 2024-06-07
refactor(sequencer.rs): use if let instead of match

hkey0:
# Description
Since we are not dealing with the `None` case, using `if let` expression instead of `match` expression will make the code more readable.


kpp (approved):

ercecan (approved):

eyusufatik:
Open source contribution is always appreciated around here!

If you want to contribute more please check our issues.

PR #708, opened 2024-06-11
dprint format TOML files (Excluding sovereign-sdk ones)

rakanalh:
# Description
Trying to get this merged this time but excluding files from sovereign-sdk so that we have better and consistent TOML file format.

To see how sovereign-sdk is excluded, please look into [dprint.json](https://github.com/chainwayxyz/citrea/blob/758467ee731a227d93281cd11ccd129d587d262c/dprint.json)

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -92,6 +92,8 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2
+      - name: Check TOML
+        uses: dprint/check@v2.2
       - name: Run lint
         run: |
           if ! SKIP_GUEST_BUILD=1 make lint ; then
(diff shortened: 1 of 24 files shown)


.github/workflows/checks.yml
```diff
@@ -92,6 +92,8 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2
+      - name: Check TOML
+        uses: dprint/check@v2.2
```
kpp, line 96: I am fine with dprint fmt for now, but a new check can be annoying.

kpp (approved):

.github/workflows/checks.yml
rakanalh: I have it installed & my editor would automatically execute it after i save files. You will not feel it. Or i think we can set it up pre-commit. wdyt?

eyusufatik: lets go the precommit way

rakanalh: Added

eyusufatik (approved):

PR #720, opened 2024-06-12
Move fullnode to it's own crate

### LINKED ISSUE #685: Separate fullnode into it's own crate
As the title says

rakanalh:
# TODO
- [x] Move fullnode into it's own crate
- [x] Use `CitreaRollupBlueprint` to spawn `CitreaSequencer` and cleanup `sov-stf-runner`

## Linked Issues
- Fixes #685


crates/sequencer/src/lib.rs
```diff
@@ -7,5 +7,54 @@ mod rpc;
...
+
 pub use config::{SequencerConfig, SequencerMempoolConfig};
 pub use sequencer::CitreaSequencer;
+use sov_modules_rollup_blueprint::RollupBlueprint;
+use sov_modules_stf_blueprint::StfBlueprint;
+use tokio::sync::oneshot;
+use tracing::{instrument, Instrument};
+
+/// Sequencer stf runner
+pub struct Sequencer<S: RollupBlueprint> {
```
ercecan, line 20: Out of curiosity:

Since we are now moving forward with `CitreaRollupBlueprint`, wouldn't it be better to use that extended trait instead of `RollupBlueprint`?

rakanalh: We could. However, it wouldn't make so much of a difference unless we're overriding methods from the old trait.

eyusufatik (approved):

Issue #651, opened 2024-06-03
Pass Bridge::deposit data via sol types


kpp:
Now we pass it via a big blob:

1. https://github.com/chainwayxyz/citrea/blob/91632a4c1372cea81f5218c570f49e1af5205f75/crates/evm/src/tests/sys_tx_tests.rs#L392
2. https://github.com/chainwayxyz/citrea/blob/91632a4c1372cea81f5218c570f49e1af5205f75/crates/evm/src/evm/system_contracts/src/Bridge.sol#L17

However there is a way for us to make it transparent: https://docs.rs/alloy-sol-macro/latest/alloy_sol_macro/macro.sol.html

Issue #645, opened 2024-05-31
Switch to alloy-primitives from crates.io


kpp:
https://github.com/chainwayxyz/citrea/pull/636 introduced a patch for alloy-primitives from github. We need to remove this patch after a new version of alloy-primitives.

Issue #468, opened 2024-04-29
Add check_sys_hashes for Deposit contract


kpp:
Add public Bridge methods to crates/evm/src/evm/system_contracts/test/check_sys_hashes.py:

- [ ] initialize
- [ ] deposit

Linked issue: https://github.com/chainwayxyz/citrea/pull/445

How it was implemented for BitcoinLightClient: https://github.com/chainwayxyz/citrea/pull/447

Issue #329, opened 2024-04-01
Migrate ethers-rs to alloy-rs 


kpp:
Ethers is being deprecated: https://github.com/gakonst/ethers-rs/issues/2667

PR #726, opened 2024-06-12
Remove ethers -> alloy

kpp:
# Description

Use alloy instead of ethers.

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/105
- Fixes https://github.com/chainwayxyz/citrea/issues/329
- Fixes https://github.com/chainwayxyz/citrea/issues/651
- Fixes https://github.com/chainwayxyz/citrea/issues/468
- Fixes https://github.com/chainwayxyz/citrea/issues/645

## Testing
`make test`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -126,13 +126,6 @@ risc0-build = "0.21"
 bonsai-sdk = "0.7.0"
 
 # EVM dependencies
-# ethereum-types = "0.14.1"
-ethers = "2.0"
-ethers-core = { version = "2.0", default-features = false }
-# ethers-contract = "2.0"
-# ethers-providers = { version = "2.0", default-features = false }
-ethers-signers = { version = "2.0", default-features = false }
-# ethers-middleware = { version = "2.0", default-features = false }
 
 revm-inspectors = { git = "https://github.com/paradigmxyz/evm-inspectors", rev = "7168ac5", default-features = false }
 reth-primitives = { git = "https://github.com/paradigmxyz/reth", tag = "v0.2.0-beta.7", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -126,13 +126,13 @@ risc0-build = "0.21"
 bonsai-sdk = "0.7.0"
 
 # EVM dependencies
-ethereum-types = "0.14.1"
+# ethereum-types = "0.14.1"
 ethers = "2.0"
 ethers-core = { version = "2.0", default-features = false }
-ethers-contract = "2.0"
-ethers-providers = { version = "2.0", default-features = false }
+# ethers-contract = "2.0"
+# ethers-providers = { version = "2.0", default-features = false }
 ethers-signers = { version = "2.0", default-features = false }
-ethers-middleware = { version = "2.0", default-features = false }
+# ethers-middleware = { version = "2.0", default-features = false }
 
 revm-inspectors = { git = "https://github.com/paradigmxyz/evm-inspectors", rev = "7168ac5", default-features = false }
 reth-primitives = { git = "https://github.com/paradigmxyz/reth", tag = "v0.2.0-beta.7", default-features = false }
@@ -151,11 +151,11 @@ reth-stages = { git = "https://github.com/paradigmxyz/reth", tag = "v0.2.0-beta.
 revm = { version = "8", features = ["serde"], default-features = false }
 # forcing cargo for this version or else chooses 3.1.1 and there is some dependency conflicts
 revm-primitives = { version = "3.1", default-features = false }
-# forcing cargo for this version or else chooses 0.3.1 and there is some dependency conflicts
-alloy-trie = { version = "=0.3.0", default-features = false }
+alloy-trie = { version = "0.3", default-features = false }
 alloy-rlp = { version = "0.3", default-features = false }
-alloy-primitives = { version = "0.7.4", default-features = false }
-alloy-sol-types = { version = "0.7.4", default-features = false }
+alloy-primitives = { version = "0.7.6", default-features = false }
+alloy-sol-types = { version = "0.7.6", default-features = false, features = ["json"] }
+alloy = { git = "https://github.com/alloy-rs/alloy", rev = "77c1240", default-features = false }
 
 ed25519-dalek = { version = "2", default-features = false, features = ["serde"] }
 secp256k1 = { version = "0.28.2", default-features = false, features = ["global-context", "recovery"] }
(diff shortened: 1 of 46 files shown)


crates/evm/src/hooks.rs
```diff
@@ -62,38 +64,16 @@ where
...
-            ));
+            system_events.push(SystemEvent::BridgeInitialize);
         }
 
         soft_confirmation_info
             .deposit_data
             .iter()
             .for_each(|deposit_data| {
-                system_events.push(SystemEvent::BridgeDeposit(deposit_data.clone()));
+                let params = DepositParams::abi_decode(deposit_data, false)
```
kpp, line 74: I didn't go further - I still store raw deposit params in `soft_confirmation_info.deposit_data`, which is done explicitly for backward compat for now.

crates/sequencer/src/rpc.rs
```diff
@@ -117,28 +119,21 @@ pub(crate) fn create_rpc_module<C: sov_modules_api::Context>(
...
-            let evm = Evm::<C>::default();
-            let mut working_set = WorkingSet::<C>::new(ctx.storage.clone());
-
-            let dep_tx = ctx
-                .deposit_mempool
-                .lock()
-                .await
-                .make_deposit_tx_from_data(deposit.clone().into());
-
-            let tx_res = evm.get_call(dep_tx, None, None, None, &mut working_set);
```
kpp, line 129: Since there is a nice way to parse and validate deposit params - no need to make a virtual evm call to check it.

Cargo.toml
```diff
@@ -151,11 +144,11 @@ reth-stages = { git = "https://github.com/paradigmxyz/reth", tag = "v0.2.0-beta.
 revm = { version = "8", features = ["serde"], default-features = false }
 # forcing cargo for this version or else chooses 3.1.1 and there is some dependency conflicts
 revm-primitives = { version = "3.1", default-features = false }
-# forcing cargo for this version or else chooses 0.3.1 and there is some dependency conflicts
-alloy-trie = { version = "=0.3.0", default-features = false }
+alloy-trie = { version = "0.3", default-features = false }
```
rakanalh, line 147: Why are we not using 0.7.6 similar to primities and sol-types crates?

bin/citrea/provers/risc0/guest-bitcoin/Cargo.toml
```diff
@@ -8,6 +8,7 @@ resolver = "2"
 
 [dependencies]
 # forcing cargo for this version
+alloy-primitives = { version = "0.7.6", default-features = false }
```
rakanalh, line 11: Can we not use `workspace = true, default-features = false` here?

zerosnacks:
Hi all,

Happy to announce we've released Alloy `0.1` today: https://www.paradigm.xyz/2024/06/alloy-release

crates/sequencer/src/rpc.rs
eyusufatik: the evm call was not to check params, it was a way for us to verify the deposit. we must have it here. pls add it back

crates/sequencer/src/deposit_data_mempool.rs
```diff
@@ -18,15 +14,6 @@ impl DepositDataMempool {
         }
     }
 
-    pub fn make_deposit_tx_from_data(&mut self, deposit_tx_data: Vec<u8>) -> TransactionRequest {
```
eyusufatik, line 21: again, we need this back. this is not doing abi decode, however, it's building a deposit transaction from the system signer with the given data from the request to be passed to eth_call

crates/evm/src/evm/system_contracts/mod.rs
```diff
@@ -58,20 +72,22 @@ impl Bridge {
...
     }
 
     /// Return data to deposit
-    pub fn deposit(data: Vec<u8>) -> Bytes {
-        let mut func_selector: Vec<u8> = vec![0xdd, 0x95, 0xc7, 0xc6]; // deposit((bytes4,bytes2,bytes,bytes,bytes,bytes4,bytes,uint256,uint256)) dd95c7c6
-
-        func_selector.extend_from_slice(data.as_slice());
-
-        Bytes::from(func_selector)
+    pub fn deposit(params: BridgeContract::DepositParams) -> Bytes {
```
eyusufatik, line 88: all these changes on this file from a readability perspective is 1000x better.

however, how do you think the performance is effected?

don't forget this part runs in zk context.

crates/evm/src/hooks.rs
eyusufatik: should stay as is imo

eyusufatik, line 74: i think I'm against the idea of serializing into `DepositParams` here. What's the benefit?

we serialize here, then it's going to be deserialized back to bytes right before this goes into EVM.

crates/evm/src/evm/system_contracts/mod.rs
eyusufatik, line 1: why not do this for BitcoinLightClient?

crates/evm/src/tests/tx_tests.rs
```diff
@@ -19,39 +22,47 @@ fn tx_rlp_encoding_test() {
...
+
+    let decoded = TransactionSigned::decode(&mut bytes.as_ref()).unwrap();
+
+    let decoded_signed = decoded.try_ecrecovered().unwrap();
+    let decoded_signer = decoded_signed.signer();
+    assert_eq!(decoded_signer, wallet.address());
 
-    assert_eq!(tx, decoded_tx);
+    let decoded_envelope = TxEnvelope::decode(&mut bytes.as_ref()).unwrap();
+    assert_eq!(envelope, decoded_envelope);
```
eyusufatik, line 65: are we going to wait for your fix on alloy for this?

eyusufatik (changes requested):

bin/citrea/provers/risc0/guest-bitcoin/Cargo.toml
eyusufatik: guests are not part of the workspace afaik

crates/evm/src/tests/tx_tests.rs
kpp: No, no need, I found a workaround

crates/evm/src/evm/system_contracts/mod.rs
```diff
@@ -10,45 +19,49 @@ impl BitcoinLightClient {
...
-        println!("block_number: {:?}", block_number);
-        func_selector.extend_from_slice(&block_number.to_be_bytes::<32>());
-
-        Bytes::from(func_selector)
+        BitcoinLightClientContract::getBlockHashCall {
+            _blockNumber: U256::from(block_number),
+        }
+        .abi_encode()
+        .into()
     }
```
eyusufatik, line 45: lets revert first 2 to the previous version





get block hash can be test only, no?

eyusufatik (changes requested):

eyusufatik (approved):

PR #773, opened 2024-06-19
Fullnode: Separate L1 / L2 sync

### LINKED ISSUE #741: Fullnode: Split L1 / L2 sync
Similar to #738, we want to split the sync and be able to also split soft confirmations sync from soft confirmations processing to improve the flow and simplify the code.

rakanalh:
# Description

## Linked Issues
- Fixes #741


bin/citrea/tests/e2e/mod.rs
```diff
@@ -1090,21 +1090,6 @@ async fn test_soft_confirmations_status_one_l1() -> Result<(), anyhow::Error> {
...
     wait_for_l1_block(&da_service, 3, None).await;
 
-    // now retrieve confirmation status from the sequencer and full node and check if they are the same
-    for i in 1..=6 {
-        let status_node = full_node_test_client
-            .ledger_get_soft_confirmation_status(i)
-            .await
-            .unwrap();
-
-        assert_eq!(SoftConfirmationStatus::Trusted, status_node.unwrap());
```
rakanalh, line 1100: For the record, the sync on full node was happening faster in a way which gets those blocks to a `Finalized` state rather than `Trusted`. Therefore, i opted for removing these checks as we're going to check for finalized state at the end of this test.

crates/sovereign-sdk/full-node/sov-stf-runner/src/config.rs
```diff
@@ -96,6 +96,8 @@ pub struct FullNodeConfig<DaServiceConfig> {
     pub da: DaServiceConfig,
     /// Important pubkeys
     pub public_keys: RollupPublicKeys,
+    /// Number of blocks to request during sync
+    pub sync_blocks_count: Option<u64>,
```
eyusufatik, line 100: we can make this u64 and default to the other number by a serde macro.

cleaner approach imo. what do you think?

crates/fullnode/src/runner.rs
```diff
@@ -186,467 +195,625 @@ where
...
+                Some(l1_block) = l1_rx.recv() => {
+                    // Set the l1 height of the l1 hash
+                    self.ledger_db
+                        .set_l1_height_of_l1_hash(l1_block.header().hash().into(), l1_block.header().height())
+                        .unwrap();
+
+                    // Merkle root hash - L1 start height - L1 end height
+                    // TODO: How to confirm this is what we submit - use?
+                    // TODO: Add support for multiple commitments in a single block
+
```
eyusufatik, line 578: these todos are removable.

crates/fullnode/src/runner.rs
```diff
@@ -186,467 +195,625 @@ where
...
+                },
+                Some((l2_height, l2_block)) = l2_rx.recv() => {
+                    let l1_block = get_da_block_at_height(&self.da_service, l2_block.da_slot_height).await?;
+                    if let Err(e) = self.process_l2_block(l2_height, l2_block, l1_block).await {
+                        error!("Could not process L2 block: {}", e);
                     }
-                }
-
-                cur_l1_block = Some(filtered_block);
+                },
```
eyusufatik, line 663: two things here:

1. assume there is 100 L2 blocks on an L1 block, we fetch the L1 block 100 times, this used to be just once. (on l1 height changed)
2. L1 syncer is already fetching the L1 blocks, can't we have a shared L1 block fetcher that fetches a given height/hash just once and serves both tasks?

crates/fullnode/src/runner.rs
```diff
@@ -186,467 +195,625 @@ where
...
+            .ledger_db
+            .get_l2_range_by_l1_height(SlotNumber(start_l1_height))?
+        {
+            Some((start_l2_height, _)) => start_l2_height,
+            None => {
+                return Err(anyhow!(
+                        "Sequencer commitment verification: L1 L2 connection does not exist. L1 height = {}. Skipping commitment.",
+                        start_l1_height
+                ).into());
+            }
```
eyusufatik, line 389: this can also happen due to a missing L2 block, can retry later if this happens

crates/fullnode/src/runner.rs
```diff
@@ -186,467 +195,625 @@ where
...
+            .ledger_db
+            .get_l2_range_by_l1_height(SlotNumber(end_l1_height))?
+        {
+            Some((_, end_l2_height)) => BatchNumber(end_l2_height.0 + 1),
+            None => {
+                return Err(anyhow!(
+                    "Sequencer commitment verification: L1 L2 connection does not exist. L1 height = {}. Skipping commitment.",
+                    end_l1_height
+                ).into());
+            }
```
eyusufatik, line 402: same as previous comment

bin/citrea/tests/e2e/mod.rs
eyusufatik: we can check for trusted state before a sequencer commitment is sent maybe

eyusufatik (changes requested):

crates/fullnode/src/runner.rs
rakanalh: I have added those cases to `SyncError::MissingL2` so that they can be retried.

bin/citrea/tests/e2e/mod.rs
rakanalh: Restored

crates/fullnode/src/runner.rs
rakanalh: Handled

rakanalh: Added LRU cache

eyusufatik (approved):

PR #778, opened 2024-06-21
Implement Utility http server

### LINKED ISSUE #721: Health endpoint
should be based on main

ercecan:
# Description
Implements Utility http server for health endpoint, and more endpoints to come in the future

## Linked Issues
- Closes #721

### DIFF
--- a/bin/citrea/tests/test_helpers/mod.rs
+++ b/bin/citrea/tests/test_helpers/mod.rs
@@ -19,7 +19,7 @@ use tempfile::TempDir;
 use tokio::sync::oneshot;
 use tokio::time::sleep;
 use tracing::{debug, warn};
-use utility_server::config::UtilityServerConfig;
+use utility_server::UtilityServerConfig;
 
 use crate::test_client::TestClient;
 
@@ -34,6 +34,7 @@ pub enum NodeMode {
 #[allow(clippy::too_many_arguments)]
 pub async fn start_rollup(
     rpc_reporting_channel: oneshot::Sender<SocketAddr>,
+    utility_server_reporting_channel: oneshot::Sender<SocketAddr>,
     rt_genesis_paths: GenesisPaths,
     rollup_prover_config: Option<ProverConfig>,
     node_mode: NodeMode,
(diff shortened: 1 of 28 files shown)


eyusufatik (changes requested):
since the utility server only responds with ok when we start using this for regular healthchecks, if rpc server goes down we won't be able to detect it anymore.

bin/citrea/tests/test_helpers/mod.rs
```diff
@@ -33,6 +34,7 @@ pub enum NodeMode {
 #[allow(clippy::too_many_arguments)]
 pub async fn start_rollup(
     rpc_reporting_channel: oneshot::Sender<SocketAddr>,
+    utility_server_reporting_channel: Option<oneshot::Sender<SocketAddr>>,
```
rakanalh, line 37: IMO, RPC & Utility server channels should be passed as last params and not first. Contextually `start_rollup` called by passing some secondary values does not provide the best clarify on what this does.

crates/utility-server/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "utility-server"
```
rakanalh, line 2: I don't agree with this being called `utility-server`. Should be called `http-server` (if we're going to use some generic name) or `health-check-server` if we are being specific. `utility` does not say anything about what this does.

rakanalh (changes requested):

crates/utility-server/Cargo.toml
ercecan: We discussed that exact same thing with @eyusufatik :D The name for him `http-server` was too generic but I also believe it is a good name since it is a different protocol and the `rpc-server` name is based on another protocol `jsonrpc` which makes sense.

What are your final thoughts on naming @eyusufatik ?

I would also like your @rakanalh @rakanalh @kpp  opininon on health-checking logic for prover nodes, current implementation checks if there is a new soft confirmation after 3 seconds, but since that logic won't work with prover we need either new method for prover or a new health-check logic entirely.

eyusufatik:
i think we should somehow make this a part of our RPC server. not a seperate server.

eyusufatik:
closed in favor of #834

PR #779, opened 2024-06-23
Reorganize repo

rakanalh:
# Description
@eyusufatik thoughts on this?

- Added:
  - `data` from (postgres) to `resources/dbs`
- Moved:
  - `bin/rollup-constants` to `crates/primitives` (This would enable us to group common functionality & constants)
  - `hive` to `resources`
  - `test-data` to `resources/test-data`
 - `configs/{X}/genesis-files` to `resources/genesis/{X}`
  - `configs` to `resources/configs`
- Removed:
  - test-da-dbs


rakanalh:
I'll fix CI if the ideas are appealing.

eyusufatik (commented):
everything looks good but I don't think rosources/genesis is appropriate naming for test-data for this reason:

under configs/ we have genesis-files/ directories for *real* environments we launch citrea on. and then with this change, on ../../ we have a directory called genesis/ which works for tests. doesn't make sense

we can move genesis-files/ under configs/ to resources/genesis/ and test-data to resources/test-data maybe. what do you think?


also can you elaborate what will be in primitives/ please?

right now and in the future the rollup-constants/ crate will be used for things we want to bake in to the zk geust. right now it's rollup name and da tx id bytes length. but on mainnet this can include sequencer public key etc.

rakanalh:
> everything looks good but I don't think rosources/genesis is appropriate naming for test-data for this reason:
> 
> under configs/ we have genesis-files/ directories for _real_ environments we launch citrea on. and then with this change, on ../../ we have a directory called genesis/ which works for tests. doesn't make sense
> 
> we can move genesis-files/ under configs/ to resources/genesis/ and test-data to resources/test-data maybe. what do you think?

Yea this sounds good to me.

> also can you elaborate what will be in primitives/ please?

Primitives in substrate / reth among other repos is a crate which usually contains constants and generic functions which can be shared across the whole repo. Kind of like a `common` crate but the functionality contained there is usually "simpler" and more primitive.

> right now and in the future the rollup-constants/ crate will be used for things we want to bake in to the zk geust. right now it's rollup name and da tx id bytes length. but on mainnet this can include sequencer public key etc.

Sure, compilation flags such as `#[cfg(feature = "native")]` can be added in that crate to define values which are only used in the runtime.

eyusufatik:
in that case primitives/ is fine

eyusufatik (approved):

PR #781, opened 2024-06-23
Add cors layer to rpc

### LINKED ISSUE #722: Add support for OPTIONS requests
basically allow cors on RPC side.

Right now we handle this on the infra level, but it would be better if we let the RPC server handle it.

PR should be based on main

ercecan:
# Description
Adds cors layer to rpc

## Linked Issues
- Fixes #722


eyusufatik (changes requested):
Make an rpc middleware crate and use it in both sequencer and full nodes

In the future we can add health endpoints etc through this middleware I guess

rakanalh:
> Make an rpc middleware crate and use it in both sequencer and full nodes
> 
> In the future we can add health endpoints etc through this middleware I guess

Maybe rather than adding a crate only for RPC middleware, we can have a `common` crate which contains common functionality. 
I am working on adding this in one of my PRs.

ercecan:
> Maybe rather than adding a crate only for RPC middleware, we can have a `common` crate which contains common functionality. I am working on adding this in one of my PRs.

I believe this sounds good, but since we need cors on main, for now we can duplicate middlewares and then move the middleware logic to `common` when we merge nightly

kpp (approved):
I don't think this small change deserves moving to a `common` crate. If it works it works. LGTM

eyusufatik (approved):
let's merge this and think about moving to  common etc. later on

PR #784, opened 2024-06-24
Use the same action for toolchain

rakanalh:
# Description
Also remove redundant rustup install commands

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -75,13 +75,9 @@ jobs:
           version: "23.2"
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Toolchain
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@nightly
         with:
-          toolchain: stable
-          override: true
           components: rustfmt, clippy
-      - name: Install Rust
-        run: rustup show && rustup install nightly && rustup component add rustfmt --toolchain nightly-x86_64-unknown-linux-gnu # Nightly is needed for our configuration of cargo fmt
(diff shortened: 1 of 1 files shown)


.github/workflows/checks.yml
```diff
@@ -75,13 +75,9 @@ jobs:
           version: "23.2"
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Toolchain
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@nightly
```
kpp, line 78: We need both stable and nightly

.github/workflows/checks.yml
```diff
@@ -75,13 +75,9 @@ jobs:
...
         with:
-          toolchain: stable
-          override: true
           components: rustfmt, clippy
-      - name: Install Rust
-        run: rustup show && rustup install nightly && rustup component add rustfmt --toolchain nightly-x86_64-unknown-linux-gnu # Nightly is needed for our configuration of cargo fmt
```
rakanalh, line 84: @kpp i am basing my command on this.

.github/workflows/checks.yml
rakanalh: For lints? that's not actually the case. See my other comment

kpp:
Does make test also requires only nightly?

rakanalh:
> Does make test also require only nightly?

I have made it explicit now

kpp (approved):
LGMT but I think there is no need to add toolchain: stable for a tagged @stable toolchain =)

eyusufatik (approved):

PR #804, opened 2024-06-25
Fix missing alloy feature for tests

kpp:
# Description
Reported by Esad

## Testing

`cargo test --package citrea --test all_tests --all-features -- e2e::full_node_verify_proof_and_store --exact --show-output`


eyusufatik:
how did this work in CI before ????

kpp:
> how did this work in CI before ????

Because we have `alloy = { workspace = true, features = ["consensus"` in evm/Cargo.toml. Which is enough when we build the whole workspace.

Issue #810, opened 2024-06-26
Allow RocksDB max open file limit to be set through config


yaziciahmet:
We should allow user to set the RocksDB max open file limit through config. It should be kept in mind that there is a possibility that this provided number might be higher than what OS allows. In that case a panic with a proper message seems to be appropriate.

eyusufatik:
closed by #825

PR #811, opened 2024-06-26
Fetch RocksDB max file limit from OS

### LINKED ISSUE #731: RocksDB hits open file descriptor limit
## Summary

On OSX default file descriptor limit is very low, so it's easy to hit that limit.

## Steps to reproduce

1. Sync your node on MacOS, around 500kth block you'll hit the limit (I guess so. this only happened on my machine so far)

## Actual Result
```sh
2024-06-13T07:22:48.093113Z  INFO sov_modules_stf_blueprint: tx #0 hash: 0xa5d2bd846f27861845b0609ce2d8d1cc776b07b807f83747b1ab83b9c7a793c9 result Successful
2024-06-13T07:22:48.103357Z  INFO save_change_set_l2{l2_block_height=474637}: sov_prover_storage_manager: return=()
2024-06-13T07:22:48.103915Z  INFO sov_stf_runner::runner: New State Root after soft confirmation #474637 is: RootHash("6a5f56b8adbfd101fbc8cafee2a88fdd2f98937777b84a73bd9caec50243e368")
2024-06-13T07:22:48.104032Z ERROR finalize_l2{l2_block_height=474637}:finalize_by_l2_height{l2_block_height=474637}: sov_prover_storage_manager: error=IO error: While open a file for appending: full-node-db/state/001016.sst: Too many open files
2024-06-13T07:22:48.104060Z ERROR finalize_l2{l2_block_height=474637}: sov_prover_storage_manager: error=IO error: While open a file for appending: full-node-db/state/001016.sst: Too many open files
2024-06-13T07:22:48.104087Z ERROR sov_stf_runner::runner: error=IO error: While open a file for appending: full-node-db/state/001016.sst: Too many open files
2024-06-13T07:22:48.105350Z ERROR sov_modules_rollup_blueprint: error=IO error: While open a file for appending: full-node-db/state/001016.sst: Too many open files
2024-06-13T07:22:48.105369Z ERROR citrea: Error: IO error: While open a file for appending: full-node-db/state/001016.sst: Too many open files

```
## Additional Information
We might want to look into if we can configure rocks db to use N file descriptors.

yaziciahmet:
# Description

RocksDB had a default open file limit of 5000, but in some scenarios this number is higher than what OS allows, which caused the full node to crash. With this PR, open file limit for the process is read from the OS and 80% is set as the limit to RocksDB.

## Linked Issues
- Fixes #731 
- Related to #810 

## Testing
I lowered my own OS file limit temporarily and reproduced the #731. I did the same test after this fix, and there were no more problems.


crates/sovereign-sdk/full-node/db/sov-db/src/rocks_db_config.rs
```diff
@@ -20,9 +23,25 @@ pub struct RocksdbConfig {
 
 impl Default for RocksdbConfig {
     fn default() -> Self {
+        let mut rlim: rlimit = rlimit {
+            rlim_cur: 0,
+            rlim_max: 0,
+        };
+        let result = unsafe { getrlimit(RLIMIT_NOFILE, &mut rlim) };
```
ercecan, line 30: Is there a way to get resource limit without unsafe code?

configs/mock/rollup_config.toml
```diff
@@ -5,7 +5,7 @@ prover_da_pub_key = ""
 
 [da]
 sender_address = "0000000000000000000000000000000000000000000000000000000000000000"
-db_path = "da-db"
+db_path = "test-da-dbs"
```
rakanalh, line 8: Please revert changes in those config files.

`test-da-dbs` is removed in a different PR and this value is set to `resources/dbs/da-db`

ercecan (commented):
lgtm left just left one question

rakanalh (changes requested):

configs/mock/rollup_config.toml
yaziciahmet: Done

crates/sovereign-sdk/full-node/db/sov-db/src/rocks_db_config.rs
yaziciahmet: Found another crate called `rlimit`. It still uses unsafe within itself, but it provides a wrapper around unsafe methods, hence we can use those methods without the use of unsafe blocks. See https://github.com/chainwayxyz/citrea/pull/811/commits/e50bf1d21da62172d34d208fed1441a0d48e3b0e

eyusufatik (approved):

rakanalh (approved):

PR #822, opened 2024-06-28
Esad/fix deadlock block cache

eyusufatik:


eyusufatik:
should have opened to main :)

Issue #824, opened 2024-06-28
Reduce mutexes to 1 in BlockCache


eyusufatik:
See https://github.com/chainwayxyz/citrea/pull/823#discussion_r1658451747

PR #826, opened 2024-06-28
Backport lru deadlock fix

kpp:
# Description

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/824

### DIFF
--- a/crates/ethereum-rpc/src/gas_price/cache.rs
+++ b/crates/ethereum-rpc/src/gas_price/cache.rs
@@ -8,8 +8,10 @@ use sov_modules_api::WorkingSet;
 
 /// Cache for gas oracle
 pub struct BlockCache<C: sov_modules_api::Context> {
-    // Assuming number_to_hash and cache are always in sync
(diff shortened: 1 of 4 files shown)


eyusufatik:
are we sure can remove the mutex inside blockcache

kpp:
Yes

rakanalh:
> are we sure can remove the mutex inside blockcache

I have the same concern.

What is the justification of removing the mutex? I wanted to delay my review just to check why it was there to begin with and what potential race conditions the mutex is avoiding.

kpp:
> What is the justification of removing the mutex?

What is the justification of having the mutex?

rakanalh (approved):
Apparently, the block cache populates itself through the `provider` internally.

Given that the cache is `LruMap`, this means that race conditions to insert blocks would be basically inserting the same piece of data. Therefore, i see @kpp's point in removing the Mutex. 

Based on @kpp's explanation, the mutex was used to get mutable access to the internal LRU maps rather than borrowing self as mutable. So fixing that makes sense as well.

LGTM.

Issue #737, opened 2024-06-13
Prover: Save witness generated data in Ledger DB


rakanalh:
The prover currently calculates witness data on the fly and uses this data to generate proof.
However, in case of prover failure, we are not going to be able to regenerate the witness data again and therefore we should be able to restore the witness data from DB and generate prove after failure.

eyusufatik:
For every L2 block the prover is running on its state, it should save the generated witness in its DB.

When it encounters a sequencer commitment on DA it can use these witnesses to generate a proof.

This will enable #738

PR #830, opened 2024-06-28
Store L2=>Witness in ledgerdb

kpp:
# Description

6740398174d0ee011b671ecf3df8ac14052d4b94 removes mutex from witness
c46cb814c33b0e4df4740fe90a36a430d8f1db13 actually implements the PR

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/737

## Testing
`make test`

### DIFF
--- a/crates/prover/src/runner.rs
+++ b/crates/prover/src/runner.rs
@@ -469,6 +469,8 @@ where
                         &mut signed_soft_confirmation,
                     );
 
+                    self.ledger_db
+                        .set_l2_witness(l2_height, &slot_result.witness)?;
                     state_transition_witnesses_to_push.push(slot_result.witness);
 
                     for receipt in slot_result.batch_receipts {
--- a/crates/prover/src/runner.rs
+++ b/crates/prover/src/runner.rs
@@ -332,7 +332,7 @@ where
 
             let mut soft_confirmations: VecDeque<Vec<SignedSoftConfirmationBatch>> =
                 VecDeque::new();
-            let mut state_transition_witnesses: VecDeque<Vec<Stf::Witness>> = VecDeque::new();
+            let mut commitments_l2_heights = vec![];
             let mut da_block_headers_of_soft_confirmations: VecDeque<
                 Vec<<<Da as DaService>::Spec as DaSpec>::BlockHeader>,
             > = VecDeque::new();
@@ -340,8 +340,8 @@ where
             let mut traversed_l1_tuples = vec![];
 
             for sequencer_commitment in sequencer_commitments.clone().into_iter() {
+                let mut commitment_l2_heights = vec![];
                 let mut sof_soft_confirmations_to_push = vec![];
-                let mut state_transition_witnesses_to_push = vec![];
                 let mut da_block_headers_to_push: Vec<
                     <<Da as DaService>::Spec as DaSpec>::BlockHeader,
                 > = vec![];
@@ -434,6 +434,7 @@ where
                     let mut signed_soft_confirmation: SignedSoftConfirmationBatch =
                         soft_batch.clone().into();
 
+                    commitment_l2_heights.push(l2_height);
                     sof_soft_confirmations_to_push.push(signed_soft_confirmation.clone());
 
                     // The filtered block of soft batch, which is the block at the da_slot_height of soft batch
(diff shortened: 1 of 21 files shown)


crates/prover/src/runner.rs
```diff
@@ -469,6 +469,8 @@ where
                         &mut signed_soft_confirmation,
                     );
 
+                    self.ledger_db
+                        .set_l2_witness(l2_height, &slot_result.witness)?;
                     state_transition_witnesses_to_push.push(slot_result.witness);
```
eyusufatik, line 474: you can get rid of line 474 and everything related to that 

because we can pull witness from db

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
```diff
@@ -564,6 +564,23 @@ impl LedgerDB {
...
 
+    /// Set the last scanned slot by the prover
+    /// Called by the prover.
+    #[instrument(level = "trace", skip_all, err, ret)]
+    pub fn set_l2_witness<Witness: Serialize>(
+        &self,
+        l2_height: u64,
+        witness: &Witness,
+    ) -> anyhow::Result<()> {
+        let buf = bincode::serialize(witness)?;
```
eyusufatik, line 575: `SequencerCommitment`did not need a serialization, are we sure we need this? also I thought we said we'd use borsh, why bincode?

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
```diff
@@ -249,6 +255,11 @@ define_table_with_default_codec!(
     (L2RangeByL1Height) SlotNumber => L2HeightRange
 );
 
+define_table_with_default_codec!(
+    /// The primary source of witness by L2 height
+    (L2Witness) BatchNumber => Vec<u8>
```
eyusufatik, line 260: why two different Witness tables

crates/sovereign-sdk/module-system/sov-state/src/witness.rs
eyusufatik, line 1: why are we getting rid of Mutex's here?

eyusufatik (commented):
I guess this PR is still a draft so just leaving as comment

crates/sovereign-sdk/module-system/sov-state/src/witness.rs
kpp: Because no need to keep mutex. I may do it in another PR

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
kpp: Oops

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
kpp: > SequencerCommitment did not need a serialization, are we sure we need this?

Because `SequencerCommitment` is a concrete type. Inside prover we don't have it, we have only some generic associated type.

Bincode instead of borsh because borsh requires BorshSerialize instead of Serialize. The Serialize trait is implemented for most of types in the project.

eyusufatik (approved):

crates/prover/src/runner.rs
```diff
@@ -521,15 +523,28 @@ where
                     l2_height += 1;
                 }
 
+                commitments_l2_heights.push(commitment_l2_heights);
```
rakanalh, line 526: We can do 
```suggestion
                commitments_l2_heights.extend(commitment_l2_heights);
```

crates/prover/src/runner.rs
```diff
@@ -521,15 +523,28 @@ where
...
             }
 
             info!("Sending for proving");
 
             let hash = da_block_header_of_commitments.hash();
 
+            let mut state_transition_witnesses = VecDeque::new();
+            for l2_heights in commitments_l2_heights {
+                let mut witnesses = vec![];
+                for l2_height in l2_heights {
```
rakanalh, line 538: Based on my previous comment, we can remove the second loop and just have one.

rakanalh (changes requested):

rakanalh (approved):
Approving to address my comments later

PR #834, opened 2024-07-01
Add health check API

### LINKED ISSUE #721: Health endpoint

yaziciahmet:
# Description
Health check API is added to both sequencer and full node. `ProxyGetRequestLayer` from jsonrpsee crate is used to proxy `GET /health` requests to the `health_check` rpc method, which just returns `Ok` value.

## Linked Issues
- Fixes #721 

## Testing
I ran a sequencer and a fullnode, and sent a GET request to `/health`.


da-db/mock_da.db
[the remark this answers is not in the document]
yaziciahmet: mb :) deleted https://github.com/chainwayxyz/citrea/pull/834/commits/40a8361e3e6961e4fc674de558e31e56827113a8

crates/sequencer/src/sequencer.rs
```diff
@@ -918,14 +922,50 @@ where
...
         let rpc_context = self.create_rpc_context().await;
-        let rpc = create_rpc_module(rpc_context)?;
-        rpc_methods.merge(rpc)?;
-        Ok(rpc_methods)
+        let mut rpc = create_rpc_module(rpc_context)?;
+
+        // TODO: create register_healthcheck_rpc helper function
+        let ledger_db = self.ledger_db.clone();
+        rpc.register_method("health_check", move |_, _| {
+            let next_soft_batch_num = ledger_db.get_next_items_numbers().soft_batch_number;
```
eyusufatik, line 935: using this method might be problematic. can't you use `ledger.getHeadSoftBatch()`?

crates/sequencer/src/sequencer.rs
```diff
@@ -918,14 +922,50 @@ where
...
+                .map_err(|err| {
+                    ErrorObjectOwned::owned(
+                        INTERNAL_ERROR_CODE,
+                        INTERNAL_ERROR_MSG,
+                        Some(format!("Failed to get soft batch range: {}", err)),
+                    )
+                })?;
+
+            let block_time_s = soft_batches[1].timestamp - soft_batches[0].timestamp;
+            thread::sleep(Duration::from_millis(block_time_s * 1500));
```
eyusufatik, line 953: do async sleep with tokio, you can do `rpc.register_async_method()`

eyusufatik (changes requested):
healthcheck is fine, left small nits

eyusufatik:
also we should have a test like this:

in bin/citrea/tests/e2e

create a new test

set up a sequencer and full node, set the sequencer to test_mode = false so that it produces blocks on its own.

send health check request to the full node and check that it's 200.

abort the sequencer task and send health check request to the full node and check that it's 400/500 whatever

crates/sequencer/src/sequencer.rs
```diff
@@ -931,36 +930,57 @@ where
...
-
-            let soft_batches = ledger_db
-                .get_soft_batch_range(
-                    &(BatchNumber(next_soft_batch_num - 2)..BatchNumber(next_soft_batch_num)),
-                )
-                .map_err(|err| {
+        rpc.register_async_method("health_check", move |_, _| {
+            let ledger_db = ledger_db.clone();
+            async move {
+                let head_batch = ledger_db.get_head_soft_batch().map_err(|err| {
```
eyusufatik, line 936: why not make the whole closure async move?

yaziciahmet: It caused all move and lifetime issues. I tried Arc and everything but only cloning `ledger_db` before every invocation of the async part of the code solved the issue.

eyusufatik: @kpp @rakanalh any better way to do this?

if not, @yaziciahmet make sure clone on ledger db is a simple operation

yaziciahmet: we can also use the RpcContext to share the ledger_db. The reason I didn't go that way initially is, full node is not using RpcContext but sequencer is using it, hence it was easier to clone the ledger_db and use it. I will try to make RpcContext shared in common crate and use it in both places.

eyusufatik: that makes more sense, however, the sequencer has mempool in its context and the prover does not. what about that?

yaziciahmet: Might be a bit dirty, but we can have a trait in common called `LedgerDBAccessor`, and the actual context implementations can be different for Sequencer and FullNode. And these contexts would implement this trait simply returning the LedgerDB.

eyusufatik: maybe.

your call :)

yaziciahmet: I believe it is unnecessary to deal with contexts right now as cloning LedgerDB seems to be really cheap due to its' `db` and `next_item_numbers` fields being encapsulated with an Arc, and as far as I researched, it is also a cheap operation to clone a broadcast channel. Hence I will keep the implementation as it is.

bin/citrea/tests/e2e/mod.rs
```diff
@@ -3164,3 +3164,78 @@ async fn test_full_node_sync_status() {
...
+    let full_node_url = format!("http://localhost:{}/health", full_node_addr.port());
+    let resp = reqwest::get(&full_node_url).await.unwrap();
+    assert_eq!(resp.status().as_u16(), 200);
+
+    wait_for_l2_block(&full_node_test_client, 4, None).await;
+
+    seq_task.abort();
+
+    let resp = reqwest::get(&full_node_url).await.unwrap();
+    assert_eq!(resp.status().as_u16(), 500);
```
eyusufatik, line 3238: instead of using reqwest directly here, can you put health request to test_client.rs ?

eyusufatik (changes requested):

bin/citrea/tests/e2e/mod.rs
yaziciahmet: Done https://github.com/chainwayxyz/citrea/pull/834/commits/fdb5e83a496122a74aead8061c2489ab7270fbd0

eyusufatik (approved):

PR #836, opened 2024-07-01
better coinbase diff calc

eyusufatik:
# Description
fixes #266 's mistake

### DIFF
--- a/crates/evm/src/tests/queries/estimate_gas_tests.rs
+++ b/crates/evm/src/tests/queries/estimate_gas_tests.rs
@@ -93,7 +93,7 @@ fn test_tx_request_fields_gas() {
     );
     assert_eq!(
         contract_diff_size.unwrap(),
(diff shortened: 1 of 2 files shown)


crates/evm/src/tests/queries/estimate_gas_tests.rs
```diff
@@ -93,7 +93,7 @@ fn test_tx_request_fields_gas() {
     );
     assert_eq!(
         contract_diff_size.unwrap(),
-        serde_json::from_value::<EstimatedDiffSize>(json![{"gas":"0x6601","diffSize":"0xa8"}])
+        serde_json::from_value::<EstimatedDiffSize>(json![{"gas":"0x6601","diffSize":"0xdc"}])
```
eyusufatik, line 96: here for example, call was made with non eip1550 tx request and the previous diff size calculation failed to add coinbase (addr. and balance change, 20 + 32)

ercecan (approved):

PR #839, opened 2024-07-02
Use tree based structures in EVM handler

### LINKED ISSUE #835: Use map instead of hash_set in diff size calc
map is better in zk context

eyusufatik:
# Description
**Hash based structure**
user cycles: 21452123
total cycles: 30408704
segments: 29

**Tree based structure**
user cycles: 21240308
total cycles: 30408704
segments: 29

200k less cycles on test `e2e::test_prover_sync_with_commitments`. didn't test but on a test with more contract storage changes we'd have a better discount :) 

PS: see [When you need a map, use BTreeMap instead of HashMap](https://dev.risczero.com/api/zkvm/optimization#tldr-and-quick-wins)
## Linked ### Issues
Fixes #835 (title was a bit wrong)

### DIFF
--- a/crates/evm/src/evm/handler.rs
+++ b/crates/evm/src/evm/handler.rs
@@ -1,5 +1,5 @@
 use std::cmp::Ordering;
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeMap, BTreeSet};
 use std::mem::size_of;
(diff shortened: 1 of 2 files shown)


kpp (approved):

rakanalh (approved):
LGTM

PR #840, opened 2024-07-02
impl. make pr

eyusufatik:
# Description
I'm tired of running 

```sh
make lint

make test
```


rakanalh (approved):
You could have done `make lint && make test` but introducing this makes sense as well.

eyusufatik:
> You could have done `make lint && make test` but introducing this makes sense as well.

movin' my fingers less :D

PR #853, opened 2024-07-04
Fix intersecting l2 commitment blocks

### LINKED ISSUE #842: Commitments are created on intersecting L2 ranges
## Summary

When running #831, specifically `sequencer_crash_and_replace_full_node`, i noticed that there are 2 sequencer commitments created:

```
2024-07-02T12:24:12.731711Z DEBUG Sequencer: citrea_sequencer::sequencer: Sequencer: new L1 block, checking if commitment should be submitted
2024-07-02T12:24:12.732119Z DEBUG Sequencer:get_commitment_info: citrea_sequencer::commitment_controller: Last commitment L1 height: None
2024-07-02T12:24:12.732279Z DEBUG Sequencer:get_commitment_info: citrea_sequencer::commitment_controller: L2 range to submit: (BatchNumber(1), BatchNumber(4))
2024-07-02T12:24:12.732309Z DEBUG Sequencer:get_commitment_info: citrea_sequencer::commitment_controller: L1 height range: (1, 1)
```

And then the second one:
```
2024-07-02T12:24:18.521313Z DEBUG Sequencer: citrea_sequencer::sequencer: Sequencer: new L1 block, checking if commitment should be submitted
2024-07-02T12:24:18.521518Z DEBUG Sequencer:get_commitment_info: citrea_sequencer::commitment_controller: Last commitment L1 height: None
2024-07-02T12:24:18.521682Z DEBUG Sequencer:get_commitment_info: citrea_sequencer::commitment_controller: L2 range to submit: (BatchNumber(1), BatchNumber(9))
```

The second one happens after restarting the sequencer which somewhat tells that the start L2 block at which the commitment should be made is not initialized properly.

Investigate & resolve

## Steps to reproduce
1. run `sequencer_crash_and_replace_full_node` with `DEBUG` logs and check logs

yaziciahmet:
# Description

The issue was caused by the fact that when the sequencer is restarted, the sync of ledger db from postgres was problematic. This PR fixes the `compare_commitments_from_db` method.

## Linked Issues
- Fixes #842 

## Testing
Existing test `sequencer_crash_and_replace_full_node` works now.

### DIFF
--- a/crates/sequencer/src/sequencer.rs
+++ b/crates/sequencer/src/sequencer.rs
@@ -1016,22 +1016,32 @@ where
         &self,
         pg_connector: PostgresConnector,
     ) -> Result<(), anyhow::Error> {
+        let db_commitment = match pg_connector.get_last_commitment().await? {
+            Some(comm) => comm,
+            None => return Ok(()),
+        };
         let ledger_commitment_l2_height = self
             .ledger_db
             .get_last_sequencer_commitment_l2_height()?
-            .ok_or(anyhow!("No commitment exists"))?;
-
-        let db_commitment = pg_connector.get_last_commitment().await?;
-        // check if last commitment in db matches sequencer's last commitment
-        if let Some(db_commitment) = db_commitment {
-            // this means that the last commitment in the db is not the same as the sequencer's last commitment
-            if db_commitment.l2_start_height > ledger_commitment_l2_height.0 {
-                self.ledger_db
-                    .set_last_sequencer_commitment_l2_height(BatchNumber(
-                        db_commitment.l2_end_height,
-                    ))?
-            }
+            .unwrap_or_default();
+        if ledger_commitment_l2_height.0 >= db_commitment.l2_end_height {
+            return Ok(());
         }
+
+        self.ledger_db
+            .set_last_sequencer_commitment_l2_height(BatchNumber(db_commitment.l2_end_height))?;
+
+        let batches = self
+            .ledger_db
+            .get_soft_batch_range(
+                &(BatchNumber(db_commitment.l2_start_height)
+                    ..BatchNumber(db_commitment.l2_start_height + 1)),
+            )
+            .unwrap();
+        let start_batch = &batches[0];
+        self.ledger_db
+            .set_last_sequencer_commitment_l1_height(SlotNumber(start_batch.da_slot_height))?;
+
         Ok(())
     }
 
--- a/crates/sequencer/src/sequencer.rs
+++ b/crates/sequencer/src/sequencer.rs
@@ -1036,8 +1036,7 @@ where
             .get_soft_batch_range(
                 &(BatchNumber(db_commitment.l2_start_height)
                     ..BatchNumber(db_commitment.l2_start_height + 1)),
-            )
-            .unwrap();
+            )?;
         let start_batch = &batches[0];
         self.ledger_db
             .set_last_sequencer_commitment_l1_height(SlotNumber(start_batch.da_slot_height))?;
--- a/crates/sequencer/src/sequencer.rs
+++ b/crates/sequencer/src/sequencer.rs
@@ -1032,12 +1032,12 @@ where
             .set_last_sequencer_commitment_l2_height(BatchNumber(db_commitment.l2_end_height))?;
 
         let batches = self.ledger_db.get_soft_batch_range(
-            &(BatchNumber(db_commitment.l2_start_height)
-                ..BatchNumber(db_commitment.l2_start_height + 1)),
+            &(BatchNumber(db_commitment.l2_end_height)
+                ..BatchNumber(db_commitment.l2_end_height + 1)),
         )?;
-        let start_batch = &batches[0];
+        let l2_end_batch = &batches[0];
         self.ledger_db
-            .set_last_sequencer_commitment_l1_height(SlotNumber(start_batch.da_slot_height))?;
+            .set_last_sequencer_commitment_l1_height(SlotNumber(l2_end_batch.da_slot_height))?;
 
         Ok(())
     }
--- a/crates/sequencer/src/sequencer.rs
+++ b/crates/sequencer/src/sequencer.rs
@@ -1031,12 +1031,10 @@ where
         self.ledger_db
             .set_last_sequencer_commitment_l2_height(BatchNumber(db_commitment.l2_end_height))?;
 
-        let batches = self
-            .ledger_db
-            .get_soft_batch_range(
-                &(BatchNumber(db_commitment.l2_start_height)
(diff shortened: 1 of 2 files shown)


crates/sequencer/src/sequencer.rs
```diff
@@ -1016,22 +1016,29 @@ where
...
+        self.ledger_db
+            .set_last_sequencer_commitment_l2_height(BatchNumber(db_commitment.l2_end_height))?;
+
+        let batches = self.ledger_db.get_soft_batch_range(
+            &(BatchNumber(db_commitment.l2_start_height)
+                ..BatchNumber(db_commitment.l2_start_height + 1)),
+        )?;
+        let start_batch = &batches[0];
+        self.ledger_db
+            .set_last_sequencer_commitment_l1_height(SlotNumber(start_batch.da_slot_height))?;
```
rakanalh, line 1040: I believe you would want to take the last batch instead of the first one and set the da_slot_height to that.

crates/sequencer/src/sequencer.rs
```diff
@@ -1016,22 +1016,29 @@ where
...
         }
+
+        self.ledger_db
+            .set_last_sequencer_commitment_l2_height(BatchNumber(db_commitment.l2_end_height))?;
+
+        let batches = self.ledger_db.get_soft_batch_range(
+            &(BatchNumber(db_commitment.l2_start_height)
+                ..BatchNumber(db_commitment.l2_start_height + 1)),
+        )?;
+        let start_batch = &batches[0];
```
rakanalh, line 1038:
```suggestion
        let last_soft_batch = &batches.last();
```

rakanalh (changes requested):
Since `compare_commitments_from_db` isn't really just comparing anymore, i suggest calling it, `adjust_commitments_from_db` or moving the set logic outside of the `compare_` method.

crates/sequencer/src/sequencer.rs
```diff
@@ -1016,22 +1016,29 @@ where
         &self,
         pg_connector: PostgresConnector,
     ) -> Result<(), anyhow::Error> {
+        let db_commitment = match pg_connector.get_last_commitment().await? {
+            Some(comm) => comm,
+            None => return Ok(()),
```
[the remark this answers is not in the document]
rakanalh: Re-thinking my comment, i believe it's okay to return Ok here just to indicate they we're not making any adjustments here. Maybe place a comment around that return saying that we do nothing if postgres has no information about the last commitment.

rakanalh:
Question: We re-initialize the sequencer ledger db on startup from postgres... i believe the proper bug fix would be for the sequencer ledger DB to not be out of sync to begin with. Why does postgres have information ledger db doesn't have?

crates/sequencer/src/sequencer.rs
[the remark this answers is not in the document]
yaziciahmet: I know and I talked that with Esad as well and he suggested for us to sync-back the postgres in a different issue. Although I have no opposition to either ideas, my question would be, even in that is syncing necessary? Do we ever need all commitments and not just the last one from postgres outside of tests? Because if we only technically need the last one, the sequencer will eventually will publish a commitment and it will be the latest.

crates/sequencer/src/sequencer.rs
yaziciahmet: If you check the range statement above, we are only getting 1 batch. Hence, it should be safe to get the value right out.

crates/sequencer/src/sequencer.rs
yaziciahmet: I was confused thinking that each commitment maps 1-1 with an L1 block, that is why I didn't care and just used start. If you check the range statement in the above line, we are just getting 1 batch, which is the L2 start batch. I believe the fix should be just getting L2 end instead of start. See https://github.com/chainwayxyz/citrea/pull/853/commits/9b16ae4d480d2931a18b4ae2b3c1089511b76aa8 for a fix.

rakanalh:
> @rakanalh what would you say to naming it `sync_commitments_from_db`?

Yea that works i guess.

crates/sequencer/src/sequencer.rs
```diff
@@ -1032,12 +1032,12 @@ where
             .set_last_sequencer_commitment_l2_height(BatchNumber(db_commitment.l2_end_height))?;
 
         let batches = self.ledger_db.get_soft_batch_range(
-            &(BatchNumber(db_commitment.l2_start_height)
-                ..BatchNumber(db_commitment.l2_start_height + 1)),
+            &(BatchNumber(db_commitment.l2_end_height)
+                ..BatchNumber(db_commitment.l2_end_height + 1)),
         )?;
-        let start_batch = &batches[0];
+        let l2_end_batch = &batches[0];
```
rakanalh, line 1038: You're still using the first one here.

yaziciahmet: `batches` variable only contains one batch. The range statement I used to query the batches is:
```
&(BatchNumber(db_commitment.l2_end_height)
                ..BatchNumber(db_commitment.l2_end_height + 1))
```

So it is actually only getting one batch, which is the l2 end batch.

yaziciahmet:
> Question: We re-initialize the sequencer ledger db on startup from postgres... i believe the proper bug fix would be for the sequencer ledger DB to not be out of sync to begin with. Why does postgres have information ledger db doesn't have?

If you restart a sequencer, there is no problem. The problem is when full node is upgraded to be a sequencer. Full node does not track commitment related stuff in its LedgerDB. That is why this type of sync is necessary.

rakanalh (approved):
Makes sense. Approving.

Please wait for @eyusufatik to review as well before merging.

crates/sequencer/src/sequencer.rs
```diff
@@ -1016,22 +1016,29 @@ where
...
         }
+
+        self.ledger_db
+            .set_last_sequencer_commitment_l2_height(BatchNumber(db_commitment.l2_end_height))?;
+
+        let batches = self.ledger_db.get_soft_batch_range(
+            &(BatchNumber(db_commitment.l2_end_height)
+                ..BatchNumber(db_commitment.l2_end_height + 1)),
+        )?;
+        let l2_end_batch = &batches[0];
```
eyusufatik, line 1038: can just use `get_soft_batch_by_number`

eyusufatik (changes requested):

crates/sequencer/src/sequencer.rs
yaziciahmet: Done https://github.com/chainwayxyz/citrea/pull/853/commits/41df0da498e6b49dab39bf4dcda949b4c849543a

eyusufatik (approved):

PR #854, opened 2024-07-04
Commit when diff size > 300KB

### LINKED ISSUE #739: Sequencer: Keep state diff < 400KB
The sequencer should watch it's state diff and invoke publishing a commitment as soon as it sees that it's about to reach this limit.

rakanalh:
# Description

## Linked Issues
- Fixes #739

### DIFF
--- a/crates/sequencer/src/commitment_controller.rs
+++ b/crates/sequencer/src/commitment_controller.rs
@@ -26,6 +26,7 @@ pub fn get_commitment_info(
     ledger_db: &LedgerDB,
     min_soft_confirmations_per_commitment: u64,
     prev_l1_height: u64,
+    state_diff_threshold_reached: bool,
 ) -> anyhow::Result<Option<CommitmentInfo>> {
     // first get when the last merkle root of soft confirmations was submitted
     let last_commitment_l1_height = ledger_db
@@ -98,8 +99,9 @@ pub fn get_commitment_info(
     debug!("L2 range to submit: {:?}", l2_range_to_submit);
     debug!("L1 height range: {:?}", l1_height_range);
 
-    if (l2_range_to_submit.1 .0 + 1)
-        < min_soft_confirmations_per_commitment + l2_range_to_submit.0 .0
+    if !state_diff_threshold_reached
+        && (l2_range_to_submit.1 .0 + 1)
+            < min_soft_confirmations_per_commitment + l2_range_to_submit.0 .0
     {
         return Ok(None);
     }
--- a/crates/sequencer/src/sequencer.rs
+++ b/crates/sequencer/src/sequencer.rs
@@ -56,7 +56,7 @@ use crate::mempool::CitreaMempool;
 use crate::rpc::{create_rpc_module, RpcContext};
 use crate::utils::recover_raw_transaction;
 
-const COMMITMENT_THRESHOLD: u64 = 300 * 1024;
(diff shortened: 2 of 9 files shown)


crates/sequencer/src/sequencer.rs
```diff
@@ -575,6 +601,10 @@ where
         let inscription_queue = self.da_service.get_send_transaction_queue();
         let min_soft_confirmations_per_commitment =
             self.config.min_soft_confirmations_per_commitment;
+
+        // Clear state diff
+        self.ledger_db.set_state_diff(vec![])?;
```
eyusufatik, line 606: maybe this should be done after commitment is sent

eyusufatik: so that if something happens while sending the diff, we don't lose what's on db

eyusufatik (commented):
look good to me overall, only need to implement what we talked in DM.

crates/sequencer/src/commitment_controller.rs
```diff
@@ -26,6 +26,7 @@ pub fn get_commitment_info(
     ledger_db: &LedgerDB,
     min_soft_confirmations_per_commitment: u64,
     prev_l1_height: u64,
+    state_diff_threshold_reached: bool,
```
rakanalh, line 29: Note to reviewers: This is more like a `force` flag to skip check for `min_soft_confirmations_per_commitment`

crates/sequencer/src/sequencer.rs
```diff
@@ -541,6 +558,15 @@ where
 
                 self.mempool.update_accounts(account_updates);
 
+                // If we exceed the threshold, we should notify the commitment
+                // worker to initiate a commitment.
+                #[allow(clippy::collapsible_if)]
+                if serialized_state_diff.len() as u64 > STATEDIFF_SIZE_COMMITMENT_THRESHOLD {
```
kpp, line 564: I think the logic should be:

```
if last_state_diff + new_state_diff > 400k {
   commit last_state_diff
   current_diff = new_state_diff
} else {
   current_diff = last_state_diff + new_state_diff
}
```

rakanalh: That's done somewhere else.

eyusufatik: where?

rakanalh: https://github.com/chainwayxyz/citrea/pull/854/files#diff-2039c1f53f29d88ff08ce7de70335902799a60dec1fcf87ba0f54ec87a292745R490

eyusufatik (approved):

crates/sequencer/src/sequencer.rs
eyusufatik: Doesn't Roman's control flow make more sense?

rakanalh: No because:
```
commit last_state_diff
current_diff = new_state_diff
```

The commit part happens asynchronously. Which means that we override `current_diff` in this version after we send the notification over the commitment channel. When the commitment channel receives the message, we will have overridden current diff with the very small `new_state_diff`

kpp: I understand it goes async, however we may end up with > 400kb because of how it is implemented. No?

rakanalh: Changed logic to how Roman suggested. I was misunderstanding some details about how this suggestion would work.

eyusufatik (approved):

PR #857, opened 2024-07-05
Implement separate l1 l2 sync for prover

### LINKED ISSUE #738: Prover: Separate L1 / L2 sync
Currently, there's a huge loop in the prover where L1 / L2 syncs are done sequentially and one depends on the other.

We would like to separate those so that the prover syncs L1 / L2 separately and starts generating proofs when commitments are found on L1 based on the L2 blocks previously synced.

ercecan:
# Description
Separates l1 l2 sync for prover

## Linked Issues
- Fixes #738


crates/sovereign-sdk/full-node/sov-ledger-rpc/Cargo.toml
```diff
@@ -26,6 +26,7 @@ futures = { version = "0.3", optional = true }
 sov-modules-api = { path = "../../module-system/sov-modules-api", features = [
     "native",
 ], optional = true }
+tracing = { workspace = true }
```
eyusufatik, line 29: seems like unused, remove

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
```diff
@@ -57,6 +57,7 @@ pub const LEDGER_TABLES: &[&str] = &[
     SoftBatchByHash::table_name(),
     L2RangeByL1Height::table_name(),
     L2Witness::table_name(),
+    L2StateRoot::table_name(),
```
eyusufatik, line 60: let's try not to introduce another table for this :D

bin/citrea/provers/risc0/guest-bitcoin/Cargo.lock
eyusufatik, line 1: how come bitcoin guest changed but mock da guest didn't?

crates/prover/src/runner.rs
```diff
@@ -202,85 +299,136 @@ where
             .get_prover_last_scanned_l1_height()
             .unwrap_or_else(|_| panic!("Failed to get last scanned l1 height from the ledger db"));
 
-        let mut l1_height = match last_scanned_l1_height {
-            Some(height) => height.0 + 1,
+        let start_l1_height = match last_scanned_l1_height {
+            Some(height) => height.0,
```
eyusufatik, line 303: should be +1, no?

crates/prover/src/runner.rs
```diff
@@ -202,85 +299,136 @@ where
...
+            let l1_height = l1_block.header().height();
 
-            // map the height to the hash
+            // Set the l1 height of the l1 hash
             self.ledger_db
-                .set_l1_height_of_l1_hash(filtered_block.header().hash().into(), l1_height)
+                .set_l1_height_of_l1_hash(
+                    l1_block.header().hash().into(),
+                    l1_block.header().height(),
+                )
```
eyusufatik, line 386: talked about this, just pop front is fine, no need to find index and remove

crates/prover/src/runner.rs
```diff
@@ -290,212 +438,146 @@ where
...
-
-                l2_height += 1;
+                da_block_headers_of_soft_confirmations.push_back(da_block_headers_to_push);
+                for l2_height in sequencer_commitment.l2_start_block_number
+                    ..=sequencer_commitment.l2_end_block_number
+                {
+                    let witness = self
+                        .ledger_db
+                        .get_l2_witness::<Stf::Witness>(l2_height)
+                        // Handle error
```
eyusufatik, line 553: handle these unwraps

bin/citrea/tests/e2e/mod.rs
```diff
@@ -1220,7 +1221,7 @@ async fn test_soft_confirmations_status_two_l1() -> Result<(), anyhow::Error> {
 
 #[tokio::test(flavor = "multi_thread")]
 async fn test_prover_sync_with_commitments() -> Result<(), anyhow::Error> {
-    // citrea::initialize_logging(tracing::Level::INFO);
+    citrea::initialize_logging(tracing::Level::INFO);
```
eyusufatik, line 1224: comment this out

bin/citrea/tests/e2e/mod.rs
```diff
@@ -1367,7 +1365,7 @@ async fn test_prover_sync_with_commitments() -> Result<(), anyhow::Error> {
 
 #[tokio::test(flavor = "multi_thread")]
 async fn test_reopen_prover() -> Result<(), anyhow::Error> {
-    // citrea::initialize_logging(tracing::Level::INFO);
+    citrea::initialize_logging(tracing::Level::INFO);
```
eyusufatik, line 1368: comment out

eyusufatik (changes requested):

crates/prover/src/runner.rs
ercecan: no need, I increment the for loop starting range by one, otherwise things get complicated and I need to pass ledger db inside l1_sync, something I can do but unnecessary overhead
This works fine btw no extra proving whatsoever

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
ercecan: https://github.com/chainwayxyz/citrea/issues/880

bin/citrea/provers/risc0/guest-bitcoin/Cargo.lock
ercecan: mockda guest don't use primitives, these changes are from primitives deps

eyusufatik (approved):

crates/sovereign-sdk/full-node/db/sov-db/src/schema/types.rs
```diff
@@ -187,6 +188,28 @@ pub struct StoredSoftBatch {
...
+        SignedSoftConfirmationBatch::new(
+            value.hash,
+            value.da_slot_height,
+            value.da_slot_hash,
+            value.da_slot_txs_commitment,
+            value.pre_state_root,
+            value.l1_fee_rate,
+            value
+                .txs
+                .iter()
```
kpp, line 202: Does it work with into_iter + without clone()?

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
```diff
@@ -258,6 +259,11 @@ define_table_with_default_codec!(
     (L2Witness) BatchNumber => Vec<u8>
 );
 
+define_table_with_default_codec!(
+    /// The primary source of witness by L2 height
+    (L2StateRoot) BatchNumber => Vec<u8>
```
kpp, line 264: Can witness be of an arbitrary size?

crates/fullnode/src/runner.rs
```diff
@@ -517,89 +496,7 @@ where
...
-                                    },
-                                    SyncError::Error(e) => {
-                                        error!("Could not process sequencer commitments: {}... skipping", e);
-                                        pending_sequencer_commitments.remove(index);
-                                    }
-                                }
-                            }
-                        }
-                    }
+                   self.run_inner(&mut pending_sequencer_commitments,&mut pending_zk_proofs, l1_block).await;
```
rakanalh, line 499:
```suggestion
                   self.process_l1_block(&mut pending_sequencer_commitments,&mut pending_zk_proofs, l1_block).await;
```
run in this case indicates this is a worker which this is not.

The L2 block below calls `process_l2_block`

crates/prover/src/runner.rs
```diff
@@ -202,85 +299,139 @@ where
...
+        let l1_handle = tokio::spawn(async move {
+            l1_sync(start_l1_height, da_service, l1_tx, l1_block_cache).await;
+        });
+        tokio::pin!(l1_handle);
+
+        let (l2_tx, mut l2_rx) = mpsc::channel(1);
+        let start_l2_height = self.start_l2_height;
+        let sequencer_client = self.sequencer_client.clone();
+        let sync_blocks_count = self.sync_blocks_count;
+        let l2_handle = tokio::spawn(async move {
```
rakanalh, line 335: This and the L1 handler are not needed. Just get the future from `l1_sync` and `l2_sync` without awaiting them and use those in the loop/select.

crates/prover/src/runner.rs
```diff
@@ -202,85 +299,139 @@ where
...
+            VecDeque::<Da::FilteredBlock>::new();
+        let pending_l1 = &mut pending_l1_blocks;
+        loop {
+            select! {
+                _ = &mut l1_handle => {panic!("l1 sync handle exited unexpectedly");},
+                _ = &mut l2_handle => {panic!("l2 sync handle exited unexpectedly");},
+                Some(l1_block) = l1_rx.recv() => {
+                    pending_l1.push_back(l1_block);
+                 },
+                _ = sleep(Duration::from_secs(1)) => {
```
rakanalh, line 351: This would instantiate a sleep future every iteration of the loop. Please use `tokio::time::interval` by instantiating it outside of the select. Like what the sequencer does.

crates/prover/src/runner.rs
```diff
@@ -202,85 +299,139 @@ where
...
+                    self.run_inner(
```
rakanalh, line 352: What does this do? Maybe use a more expressive name for this?

crates/prover/src/runner.rs
```diff
@@ -631,9 +741,143 @@ where
...
+            };
+
+        if soft_batches.is_empty() {
+            debug!(
+                "Soft Batch: no batch at starting height {}, retrying...",
+                l2_height
+            );
+
+            // We wait for 2 seconds and then return a Permanent error so that we exit the retry.
+            // This should not backoff exponentially
```
rakanalh, line 867: This comment doesn't make sense here because we do not return a Permanent error

crates/prover/src/runner.rs
```diff
@@ -650,3 +894,32 @@ async fn get_initial_slot_height<Da: DaSpec>(client: &SequencerClient) -> u64 {
         }
     }
 }
+
+async fn get_da_block_at_height<Da: DaService>(
```
rakanalh, line 898: I believe that this function is also declared in `fullnode/runner.rs`.

What i wanted to do with this is move it to `crates/primitives` where shared functionality between sequencer / fullnode / prover would be placed.

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
```diff
@@ -258,6 +259,11 @@ define_table_with_default_codec!(
     (L2Witness) BatchNumber => Vec<u8>
 );
 
+define_table_with_default_codec!(
+    /// The primary source of witness by L2 height
```
kpp, line 263: Is it witness?

eyusufatik: good catch :)

crates/prover/src/runner.rs
```diff
@@ -202,85 +299,139 @@ where
...
-            let filtered_block = retry_backoff(exponential_backoff.clone(), || async {
-                da_service
-                    .get_block_at(l1_height)
-                    .await
-                    .map_err(backoff::Error::transient)
-            })
-            .await?;
-
-            // map the height to the hash
+    async fn run_inner(
```
rakanalh, line 370: I would have hoped that this would be broken down into different components:
1. X would be executed only when an L2 is received
2. Y would be executed only when an L1 is received
3. Z would wait for both L1/L2

This is basically the meat of the prover,,, IMO we did split L1/L2 sync in this PR but we didn't split this.

rakanalh (changes requested):

crates/prover/src/runner.rs
eyusufatik: erce would explain better but from my understanding, 

if an arm of the select! macro starts execution, await in those arms does not yield execution to other async tasks.

so the arm has to finish execution before execution is yielded to another task. which was not really intended behaviour.

@ercecan

crates/primitives/src/cache.rs
```diff
@@ -0,0 +1,33 @@
+use std::num::NonZeroUsize;
+
+use lru::LruCache;
+use sov_rollup_interface::services::da::DaService;
+
+pub struct L1BlockCache<Da>
```
kpp, line 6: I am against that struct in `primitives` crate because it's not used anywhere but in fullnode-runner AND it's used only by_number, not by_hash.

crates/prover/src/runner.rs
rakanalh: No that's actually not true.

A `Future` has both `Ready` and `Pending` state and a `poll` method. `select` calls `poll` on each of those futures and if the future returns `Ready` then it executes the body until the first `.await` which yields control over to the runtime. If the `Future` returns `Pending` it moves on to the arm of the select and the same thing happens.

crates/primitives/src/cache.rs
rakanalh: Prover is also using the L1 block hash.

But i agree that by_hash is not needed anymore. Which makes #878 irrelevant.

crates/prover/src/runner.rs
kpp: No, that's true. We discovered it with Erce, once `select!` selected a branch it stops polling futures.

ercecan: Yes, `select!` macro will not yield until the branch finishes the execution.
From Tokio Documentation:
> The select! macro runs all branches concurrently on the same task. Because all branches of the select! macro are executed on the same task, they will never run simultaneously. The select! macro multiplexes asynchronous operations on a single task.

https://tokio.rs/tokio/tutorial/select

crates/prover/src/runner.rs
ercecan: could you elaborate please?

crates/prover/src/runner.rs
ercecan: Not sure why we need interval here, if what we need is the inner function to run every second (except when it detects commitments and tries to prove them ofc) then this works fine.

In essence is this any different than the receiver? It awaits that branch for a second just like receiver does until it receives a message, am I missing something here?

crates/prover/src/runner.rs
rakanalh: This method:
1. Processes pending L1 blocks
2. Processes sequencer commitments
3. Applies soft confirmations

- It can be broken down into multiple methods
- Applying soft confirmation can happen when we receive an L2
- Processing commitments can happen when we receive an L1
So calling the sub methods can be called in their more appropriate places.

crates/prover/src/runner.rs
rakanalh: You're instantiating a sleep instance on every call to select which could potentially lead to misbehavior. It is always recommended to instantiate the future outside the select and `.await` it inside the select. That way you guarantee it will not misbehave. 

However, when thinking about it some more, here what you're doing is that you're always waiting for 1 second if the `select` randomly selects this match arm to execute first. Which might not be necessary. Using the interval is much more suited to say that i am doing to process pending L1 blocks every 1 second.

crates/prover/src/runner.rs
```diff
@@ -202,85 +299,139 @@ where
...
+        let l1_block_cache = self.l1_block_cache.clone();
+        let mut pending_l1_blocks: VecDeque<<Da as DaService>::FilteredBlock> =
+            VecDeque::<Da::FilteredBlock>::new();
+        let pending_l1 = &mut pending_l1_blocks;
+        loop {
+            select! {
+                _ = &mut l1_handle => {panic!("l1 sync handle exited unexpectedly");},
+                _ = &mut l2_handle => {panic!("l2 sync handle exited unexpectedly");},
+                Some(l1_block) = l1_rx.recv() => {
+                    pending_l1.push_back(l1_block);
```
rakanalh, line 349: @eyusufatik @kpp do you guys think this is a potential race condition?

1 worker is pushing to the list and another is processing until this list is empty.

However, every 1 second we are spawning a new task to process the list which means that we might have multiple tasks processing the same items in the list.

crates/prover/src/runner.rs
rakanalh: Thanks for clarifying... i didn't know this about select but this is weird IMO

crates/prover/src/runner.rs
ercecan: `l1_sync` worker cannot push before the list is processed i.e. before the `process_l1_block_and_generate_proof` function returns

We are not spawning a new task, `select!` macro itself is one task

crates/prover/src/runner.rs
ercecan: Okay, just pushed a commit, hope I got it right

eyusufatik (approved):

rakanalh (approved):

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
ercecan: This will be handled with https://github.com/chainwayxyz/citrea/issues/880

PR #860, opened 2024-07-06
Use verifier in citrea-stf

### LINKED ISSUE #858: Move STFVerifier to its own crate

rakanalh:
# Description
This is just copy pasting the `StateTransitionVerifier` into a module inside `Citrea-STF` which should be sufficient for us to customize rather than updating the sov-sdk version.

Thoughts @eyusufatik?

## Linked Issues
- Fixes #858

### DIFF
--- a/crates/citrea-stf/src/lib.rs
+++ b/crates/citrea-stf/src/lib.rs
@@ -5,9 +5,10 @@
 pub mod genesis_config;
 mod hooks_impl;
 pub mod runtime;
+mod verifier;
 use sov_modules_stf_blueprint::StfBlueprint;
 use sov_rollup_interface::da::DaVerifier;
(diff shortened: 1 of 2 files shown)


eyusufatik:
Make sense

can you also update our guest code (bin/citrea/provers)

rakanalh:
> Makes sense
> 
> can you also update our guest code (bin/citrea/provers)

That's the beauty of it,,, these don't need any update because they import `StfVerifier`
https://github.com/chainwayxyz/citrea/blob/478e47055c621b5349e7a243a2794dbc61bb1506/bin/citrea/provers/risc0/guest-mock/src/bin/mock_da.rs#L39

from citrea-stf

eyusufatik (approved):

PR #863, opened 2024-07-08
Commitment rules

### LINKED ISSUE #856: Consolidate commitment logic
Based on a conversation with @eyusufatik and as a follow up to #739 

> right now in this pr it’s either over 300kb or (when we are finished with an L1 block and there is more commitments than min_soft_confirmations_per_commitment )
> we’ll need to remove any logic, ledger and postgres stuff related to L1 on this “when to commit” logic then we are fine
> over 300kb or non-committed is over min_soft_confirmations_per_commitment no more crazy L1 stuff

rakanalh:
# Description

Simplifies commitment rules by removing L1 dependency and ruling to commit based on minimum soft confirmations per commitment and/or commitment threshold

## Linked Issues
- Fixes #856 

## TODOs:
- [x] Removing dependency on L1 related stuff in our commit/don't commit logic and controller.
- [x] Removing last committed l1 height table, and code where it's used.
- [x] This might affect some tests because with the new logic the sequencer might commit before the test expects :). But should be fairly easy to fix.
- [x] Simplify tests, we would not need the "publish L2, publish L1, publish L2" logic anymore.

The last TODO is partially done since the prover L1 / L2 sync was not separated so we still need to publish L1 / L2 blocks for the prover to prove commitments.

### DIFF
--- a/crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
+++ b/crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
@@ -13,10 +13,9 @@ use tracing::instrument;
 use crate::rocks_db_config::gen_rocksdb_options;
 use crate::schema::tables::{
     BatchByHash, BatchByNumber, CommitmentsByNumber, EventByKey, EventByNumber, L2RangeByL1Height,
-    L2Witness, LastSequencerCommitmentSent, LastSequencerCommitmentSentL2, LastStateDiff,
-    ProofBySlotNumber, ProverLastScannedSlot, SlotByHash, SlotByNumber, SoftBatchByHash,
-    SoftBatchByNumber, SoftConfirmationStatus, TxByHash, TxByNumber, VerifiedProofsBySlotNumber,
-    LEDGER_TABLES,
+    L2Witness, LastSequencerCommitmentSentL2, LastStateDiff, ProofBySlotNumber,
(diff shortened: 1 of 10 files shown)


crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
```diff
@@ -470,22 +469,6 @@ impl LedgerDB {
         Ok(())
     }
 
-    /// Used by the sequencer to record that it has committed to soft confirmations on a given L2 height
-    #[instrument(level = "trace", skip(self), err, ret)]
```
eyusufatik, line 474: we can also remove the table and everything related to it

and rename `LastSequencerCommitmentSentL2`to just `LastSequencerCommitmentSent`

rakanalh:
> didn't really read tests, I assume they are fine if they are passing though some tests might be obsolote now, I'll be creating an issue or making a pr about them soon

We need to cleanup a bit further after #857

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
rakanalh: Done.

eyusufatik (approved):

yaziciahmet (approved):
LGTM

PR #864, opened 2024-07-09
Adopt transparent upgradeability pattern

### LINKED ISSUE #833: E2E bridge contract upgradability tests
test bridge and light client contracts' upgrade features on e2e or integration tests.

okkothejawa:
# Description
- Adopts transparent upgradeability pattern instead of UUPS pattern in system contracts, as `UUPSUpgradeable` contract which needs to be inherited in all implementation contracts have an `immutable` variable which makes our pre-deployment scheme inconvenient. Instead, we adopt transparent proxies like other rollups such as Optimism and Scroll which doesn't have this issue.
- Introduces two scripts, `GenesisGenerator.s.sol` and `GenesisToEvmJson.py`. The former is a Foundry script that sets the genesis state directly through Solidity testing environment, and this approach has significant advantages over the old method of manually setting genesis files as it is less error-prone and more intuitive to set. The latter takes the generated genesis data and appends it with extra information and formats it to be compatible with our existing  `evm.json` format.
- Introduces a new make command, `make genesis` to trigger the scripts above. This command should be run after all changes to the system contracts and any genesis change.
- Adds upgradeability test with a scenario that upgrades the `BitcoinLightClient` with an obviously false implementation that returns `0xdeaddeaddeaddeaddeaddeaddeaddeaddeaddeaddeaddeaddeaddeaddeaddead` for any queried block hash. As this is not possible for any actual block, successful confirmation of this shows that the behavior is updated. 

## Linked Issues
- Fixes #833 

## Testing
Existing tests were adapted.

## Docs
-


bin/citrea/tests/evm/mod.rs
```diff
@@ -188,46 +187,45 @@ async fn test_genesis_contract_call() -> Result<(), Box<dyn std::error::Error>>
...
-        .await
-        .unwrap();
-    assert_eq!(storage_value, U256::from(4660));
-
     let storage_value = seq_test_client
         .eth_get_storage_at(
             contract_address,
-            U256::from_str("0x6661e9d6d8b923d5bbaab1b96e1dd51ff6ea2a93520fdc9eb75d059238b8c5e9")
-                .unwrap(),
+            U256::from_be_slice(&hex!(
```
eyusufatik, line 215: can you revert this back to `U256::from_str`

crates/evm/src/evm/system_contracts/Makefile
eyusufatik, line 1: imo don't introduce a new makefile, 

just use the genesis command in the original make file

crates/evm/src/evm/system_contracts/foundry.toml
```diff
@@ -2,5 +2,7 @@
 src = "src"
 out = "out"
 libs = ["lib"]
+bytecode_hash = "none"
```
eyusufatik, line 5: what are these? can you add comment in the toml?

crates/evm/src/evm/system_contracts/script/GenesisGenerator.s.sol
eyusufatik, line 1: is this for test_data only?

i see `dealDevAddress()`

eyusufatik (changes requested):

bin/citrea/tests/evm/ethers_js/test.js
```diff
@@ -238,7 +238,7 @@ describe("RpcTests", function() {
...
...on="sendTransaction", data=null, reason=null, invocation=null, revert=null, transaction={ "data": "", "from": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266", "to": "0x3100000000000000000000000000000000000002" }, receipt={ "_type": "TransactionReceipt", "blobGasPrice": null, "blobGasUsed": null, "blockHash": "0x[0-9a-fA-F]+", "blockNumber": \d+, "contractAddress": null, "cumulativeGasUsed": "\d+", "from": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266", "gasPrice": "1000000", "gasUsed": "\d+", "hash": "0x[0-9a-fA-F]+", "index": 0, "logs": \[  ], "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000", "root": null, "status": 0, "to": "0x3100000000000000000000000000000000000002" }, code=CALL_EXCEPTION, version=6\.12\.1\)/);
```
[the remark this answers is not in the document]
eyusufatik: turns out it's a regex

crates/evm/src/evm/system_contracts/Makefile
eyusufatik: needs too much effort to make work.

crates/evm/src/evm/system_contracts/script/GenesisGenerator.s.sol
eyusufatik: decided on a seperate production mode or script fort this generator scripts. issue coming.

eyusufatik (approved):

eyusufatik:
`evm::test_genesis_contract_call` fails on CI 

I believe it's not flaky.

eyusufatik (changes requested):
see my above comment

eyusufatik (approved):

Issue #537, opened 2024-05-10
Bitcoin DA: get last tx after restart


kpp:
Rn we use utxo as inputs for a new tx. We should be able to build a chain of utxos to get the last tx and attach the first sequencer tx on DA to it.

eyusufatik:
TODOs:

- [ ] Get all pending txs of wallet from RPC.
- [ ] Find the last tx of the tx chain(s). Use that as the last utxo to use.
- [ ] Create a function called `get_pending_txs` for DA adapters. This should return the parsed info inside the txs. (This is to be used for re-submitting pending commitments)

PR #883, opened 2024-07-12
Async commitment submission

### LINKED ISSUE #819: DA send blocks block production
## Summary

I've noticed block production halts when there is a send tx on bitcoin da (keep in mind these are cpu heavy, pow-like transaction creation). This didn't use to happen.

## Steps to reproduce

1. Run sequencer on bitcoin regtest
2. Publish block on bitcoin regtest

## Expected Result
Bitcoin send happens on a seperate thread.

## Actual Result
It did not I suppose

yaziciahmet:
# Description

Made waiting for commitment result asynchronous, so the L2 block production can continue.

## Linked Issues
- Fixes #819 

## Testing
I tested by reproducing #819 and observing that commitments are resolved successfully without blocking L2 block production. I also observed when restarted when there are pending commitments, it blocks the start and waits until the commitment is finalized.

### DIFF
--- a/crates/sequencer/src/commitment_controller.rs
+++ b/crates/sequencer/src/commitment_controller.rs
@@ -23,6 +23,7 @@ pub fn get_commitment_info(
     min_soft_confirmations_per_commitment: u64,
     state_diff_threshold_reached: bool,
 ) -> anyhow::Result<Option<CommitmentInfo>> {
+    // TODO: take pending commitments into account as well
     let Some((head_soft_batch_number, _)) = ledger_db.get_head_soft_batch()? else {
         // No soft batches have been created yet.
         return Ok(None);
--- a/crates/sequencer/src/commitment_controller.rs
+++ b/crates/sequencer/src/commitment_controller.rs
@@ -22,12 +22,15 @@ pub fn get_commitment_info(
     ledger_db: &LedgerDB,
     min_soft_confirmations_per_commitment: u64,
     state_diff_threshold_reached: bool,
+    last_committed_l2_height: Option<BatchNumber>,
 ) -> anyhow::Result<Option<CommitmentInfo>> {
-    // Based on heights stored in ledger_db, decided which L2 blocks
-    // to commit.
-    let last_committed_l2_height = ledger_db
-        .get_last_sequencer_commitment_l2_height()?
-        .unwrap_or(BatchNumber(0));
+    // If last commited l2 height is not passed, query it from ledger db
+    // to calculate l2 block range to commit
+    let last_committed_l2_height = if let Some(l2_height) = last_committed_l2_height {
+        l2_height
+    } else {
+        ledger_db.get_last_sequencer_commitment_l2_height()?.unwrap_or(BatchNumber(0))
+    };
 
     let Some((head_soft_batch_number, _)) = ledger_db.get_head_soft_batch()? else {
         // No soft batches have been created yet.
--- a/crates/sequencer/src/commitment_controller.rs
+++ b/crates/sequencer/src/commitment_controller.rs
@@ -35,13 +35,9 @@ pub fn get_commitment_info(
         .unwrap_or(BatchNumber(0));
     let last_pending_l2_height = ledger_db
         .get_pending_commitments_l2_range()?
-        .map(|comms| {
-            comms
-                .iter()
-                .map(|(_, end)| *end)
-                .max()
-                .unwrap_or(BatchNumber(0))
-        })
+        .iter()
+        .map(|(_, end)| *end)
+        .max()
         .unwrap_or(BatchNumber(0));
     let last_committed_l2_height = cmp::max(last_finalized_l2_height, last_pending_l2_height);
 
--- a/crates/sequencer/src/commitment_controller.rs
+++ b/crates/sequencer/src/commitment_controller.rs
@@ -23,17 +23,23 @@ pub fn get_commitment_info(
     min_soft_confirmations_per_commitment: u64,
     state_diff_threshold_reached: bool,
 ) -> anyhow::Result<Option<CommitmentInfo>> {
-    // TODO: take pending commitments into account as well
     let Some((head_soft_batch_number, _)) = ledger_db.get_head_soft_batch()? else {
         // No soft batches have been created yet.
         return Ok(None);
     };
 
-    // If last commited l2 height is not passed, query it from ledger db
+    // If last committed l2 height is not passed, query it from ledger db
     // to calculate l2 block range to commit
-    let last_committed_l2_height = ledger_db
-        .get_last_sequencer_commitment_l2_height()?
-        .unwrap_or(BatchNumber(0));
+    let last_committed_l2_height = match ledger_db.get_pending_commitments_l2_range()? {
+        Some(pending_commitments) if !pending_commitments.is_empty() => *pending_commitments
+            .iter()
+            .map(|(_, end)| end)
+            .max()
+            .unwrap(),
+        _ => ledger_db
+            .get_last_sequencer_commitment_l2_height()?
+            .unwrap_or(BatchNumber(0)),
+    };
 
     // If the last commitment made is on par with the head
     // soft batch, we have already committed the latest block.
--- a/crates/sequencer/src/commitment_controller.rs
+++ b/crates/sequencer/src/commitment_controller.rs
@@ -31,8 +31,8 @@ pub fn get_commitment_info(
     // If last commited l2 height is not passed, query it from ledger db
     // to calculate l2 block range to commit
     let last_committed_l2_height = ledger_db
-            .get_last_sequencer_commitment_l2_height()?
-            .unwrap_or(BatchNumber(0));
+        .get_last_sequencer_commitment_l2_height()?
+        .unwrap_or(BatchNumber(0));
 
     // If the last commitment made is on par with the head
     // soft batch, we have already committed the latest block.
--- a/crates/sequencer/src/commitment_controller.rs
+++ b/crates/sequencer/src/commitment_controller.rs
@@ -1,3 +1,4 @@
+use std::cmp;
 use std::ops::RangeInclusive;
 
 use anyhow::anyhow;
@@ -28,18 +29,21 @@ pub fn get_commitment_info(
         return Ok(None);
     };
 
-    // First check if there are any pending commitments and use that,
-    // if there are none, use last finalized commitment height
-    let last_committed_l2_height = match ledger_db.get_pending_commitments_l2_range()? {
-        Some(pending_commitments) if !pending_commitments.is_empty() => *pending_commitments
-            .iter()
-            .map(|(_, end)| end)
-            .max()
-            .unwrap(),
-        _ => ledger_db
-            .get_last_sequencer_commitment_l2_height()?
-            .unwrap_or(BatchNumber(0)),
-    };
+    // Get latest finalized and pending commitments and find the max height
+    let last_finalized_l2_height = ledger_db
+        .get_last_sequencer_commitment_l2_height()?
+        .unwrap_or(BatchNumber(0));
+    let last_pending_l2_height = ledger_db
+        .get_pending_commitments_l2_range()?
+        .map(|comms| {
+            comms
+                .iter()
+                .map(|(_, end)| *end)
+                .max()
+                .unwrap_or(BatchNumber(0))
+        })
+        .unwrap_or(BatchNumber(0));
+    let last_committed_l2_height = cmp::max(last_finalized_l2_height, last_pending_l2_height);
 
     // If the last commitment made is on par with the head
     // soft batch, we have already committed the latest block.
--- a/crates/sequencer/src/commitment_controller.rs
+++ b/crates/sequencer/src/commitment_controller.rs
@@ -28,8 +28,8 @@ pub fn get_commitment_info(
         return Ok(None);
     };
 
-    // If last committed l2 height is not passed, query it from ledger db
-    // to calculate l2 block range to commit
+    // First check if there are any pending commitments and use that,
+    // if there are none, use last finalized commitment height
     let last_committed_l2_height = match ledger_db.get_pending_commitments_l2_range()? {
         Some(pending_commitments) if !pending_commitments.is_empty() => *pending_commitments
             .iter()
--- a/crates/sequencer/src/commitment_controller.rs
+++ b/crates/sequencer/src/commitment_controller.rs
@@ -29,7 +29,9 @@ pub fn get_commitment_info(
     let last_committed_l2_height = if let Some(l2_height) = last_committed_l2_height {
         l2_height
     } else {
-        ledger_db.get_last_sequencer_commitment_l2_height()?.unwrap_or(BatchNumber(0))
+        ledger_db
+            .get_last_sequencer_commitment_l2_height()?
+            .unwrap_or(BatchNumber(0))
     };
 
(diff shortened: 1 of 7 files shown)


crates/sequencer/src/sequencer.rs
```diff
@@ -629,53 +630,77 @@ where
...
+            info!("Sent commitment to DA queue");
+
+            // TODO: this causes state diff to be lost on restart/error, fix that
             self.ledger_db.set_state_diff(vec![])?;
             self.last_state_diff = vec![];
 
-            info!("New commitment. L2 range: #{}-{}", l2_start, l2_end,);
+            // Store this only in memory to indicate there is an in-progress
+            // commitment process happenning
+            self.last_committed_l2_height = Some(*commitment_info.l2_height_range.end());
```
rakanalh, line 646: This does not indicate that there is an in-progress commitment but rather sets the value of the "already committed". That's how we interpret this.

What happens if this value was set but submitting the commitment failed inside the async block below? The sequencer would fall out of sync when looking at the value in this var vs what is stored inside ledger_db.

Assuming it goes through, there's no need to introduce `self.last_committed_l2_height` member variable since `get_commitment_info` will use the ID that we've just written to ledger db.

In short, i do believe we can make this much better.

rakanalh (changes requested):

crates/sequencer/src/commitment_controller.rs
```diff
@@ -22,12 +22,17 @@ pub fn get_commitment_info(
...
-        .unwrap_or(BatchNumber(0));
+    // If last commited l2 height is not passed, query it from ledger db
+    // to calculate l2 block range to commit
+    let last_committed_l2_height = if let Some(l2_height) = last_committed_l2_height {
+        l2_height
+    } else {
+        ledger_db
+            .get_last_sequencer_commitment_l2_height()?
+            .unwrap_or(BatchNumber(0))
+    };
```
eyusufatik, line 35: everything else looks good to me I don't understand why we started passing this BatchNumber tho.

crates/sequencer/src/sequencer.rs
yaziciahmet: The thing is, we need something that indicates that there is an asynchronous commit happening right now and check based on that. Otherwise it is gonna commit every time a new block is produced once the commitment conditions are satisfied.

But we also don't want it to block the further commitments incoming if the previous one didn't finish yet, hence we can't just use a simple boolean to indicate that. We can rename this to `last_commit_try_l2_height`, and reset this back to its old value if the previous one failed, but that is still problematic. Because if second commitment arrived and pushed to queue, right after that first commitment failed, and second commitment succeded. This scenario will cause a gap for the l2 height range of the first commitment.

Maybe we can requeue the commitment if it failed with the hopes that everything is gonna turn out fine?

Just putting my though flow out there. LMK what you think.

crates/sequencer/src/commitment_controller.rs
yaziciahmet: Please check above discussion with Rakan

crates/sequencer/src/sequencer.rs
yaziciahmet: Or for now, maybe the best course of action is to block the commitment submission if there is one going on right now, even though its not optimal.

eyusufatik: > The thing is, we need something that indicates that there is an asynchronous commit happening right now and check based on that. Otherwise it is gonna commit every time a new block is produced once the commitment conditions are satisfied.

that's fine?

yaziciahmet: wait how is that fine? lets say the min soft confirmation limit is 100, and we reached the block 100 and got ourselves a 1-100 commitment. when 101 comes, it will submit another commitment for 1-101, 102 will submit 1-102 etc.

eyusufatik: ah, you mean that.

yeah, that's actually not fine.

what about another ledger db method that returns a vec of pending "commitment ranges" so that we can check that on commitment controller.

then in the next prs we can build on that, for restarts etc.

rakanalh: I see your point.

We need a way to lock specific commitments to prevent running multiple commitments txs for the same range.

How about you add a HashSet for in-progress commitments where the entry is the l2 end range and just check if that one has been submitted already then you skip. If it wasn't, add and then spawn.
Once the spawned task is done, remove it from the list. This should serve as some kind of a locking mechanism.

yaziciahmet: > I see your point.
> 
> We need a way to lock specific commitments to prevent running multiple commitments txs for the same range.
> 
> How about you add a HashSet for in-progress commitments where the entry is the l2 end range and just check if that one has been submitted already then you skip. If it wasn't, add and then spawn. Once the spawned task is done, remove it from the list. This should serve as some kind of a locking mechanism.

We definitely need to store the range as you said. But if we only store it in memory, it is going to be problematic to handle error cases, as it could leave gaps in between commits. So I believe it is best to store it in a db as Esad suggested above like "in_progress_commitment_ranges". And maybe now, or at least in the future, that would allow us to compare that table with the finalized commitments and figure out what to do from there on.

crates/sequencer/src/commitment_controller.rs
```diff
@@ -23,17 +23,24 @@ pub fn get_commitment_info(
...
+        Some(pending_commitments) if !pending_commitments.is_empty() => *pending_commitments
+            .iter()
+            .map(|(_, end)| end)
+            .max()
+            .unwrap(),
+        _ => ledger_db
+            .get_last_sequencer_commitment_l2_height()?
+            .unwrap_or(BatchNumber(0)),
+    };
+
```
eyusufatik, line 43: imo, getting both and using the bigger one is more error prone

crates/sovereign-sdk/module-system/sov-modules-rollup-blueprint/src/lib.rs
```diff
@@ -38,14 +38,14 @@ pub trait RollupBlueprint: Sized + Send + Sync {
...
-    type NativeContext: Context;
+    type NativeContext: Context + Send;
 
     /// Manager for the native storage lifecycle.
     type StorageManager: HierarchicalStorageManager<
         Self::DaSpec,
         NativeStorage = <Self::NativeContext as Spec>::Storage,
         NativeChangeSet = <Self::NativeContext as Spec>::Storage,
-    >;
+    > + Send;
```
eyusufatik, line 48: i don't know what thse + Sends would imply. 


@kpp @rakanalh is this ok?

crates/sequencer/src/sequencer.rs
```diff
@@ -629,54 +631,194 @@ where
...
                 .send(request)
                 .map_err(|_| anyhow!("Bitcoin service already stopped!"))?;
-            let tx_id = rx
-                .await
-                .map_err(|_| anyhow!("DA service is dead!"))?
-                .map_err(|_| anyhow!("Send transaction cannot fail"))?;
 
+            info!("Sent commitment to DA queue");
+
+            // TODO: this causes state diff to be lost on restart/error, fix that
```
eyusufatik, line 641: remove todo

crates/sequencer/src/sequencer.rs
```diff
@@ -629,54 +631,194 @@ where
...
+            self.ledger_db.set_state_diff(vec![])?;
+            self.last_state_diff = vec![];
+
+            // Add commitment pending commitments
+            let mut pending_commitments = self.pending_commitments_l2_range.lock().await;
+            pending_commitments.push((*l2_range_to_submit.start(), *l2_range_to_submit.end()));
```
eyusufatik, line 647: maybe push to pending first then delete the state diff stuff 🤔

crates/sequencer/src/sequencer.rs
```diff
@@ -629,54 +631,194 @@ where
...
+                    );
+                }
+            });
+        }
+        Ok(())
+    }
 
-            info!("New commitment. L2 range: #{}-{}", l2_start, l2_end,);
+    #[instrument(level = "trace", skip(self), err, ret)]
+    pub async fn resubmit_pending_commitments(&mut self) -> anyhow::Result<()> {
```
eyusufatik, line 720: for simplicity, this function does not need to spawn a task

since it's called on restart, i think we can burden it taking a long time to start producing blocks. 

this is up for discussion tho.

eyusufatik (changes requested):
a few minor nits. looks good

crates/sovereign-sdk/module-system/sov-modules-rollup-blueprint/src/lib.rs
yaziciahmet: TBH, I did this change completely unaware of what the hell is going on. It didn't allow me to call `seq.resubmit_pending_commitments().await` in `create_new_sequencer` method until I did this. I believe it has smh. to due with this method mutating the `seq` asynchronously at the point the instance is initialized, but no idea in the end :).

crates/sequencer/src/sequencer.rs
yaziciahmet: I would disagree with this suggestion. Because I can foresee that being really annoying.
1. developer restarts the app while there was a pending commitment to test a change
2. pending commitment takes like 5-10 minutes and blocks everything
3. developer angry

crates/sequencer/src/sequencer.rs
yaziciahmet: done

crates/sequencer/src/sequencer.rs
yaziciahmet: done

crates/sequencer/src/sequencer.rs
rakanalh: We don't care much about the scenario you're proposing. The sequencer should not cater for the develop / test cycle compared to catering for consistency. We should wait for the pending commitments to finish before starting the sequencer normal operations. This could also prevent potential race conditions.

crates/sequencer/src/sequencer.rs
```diff
@@ -591,94 +597,151 @@ where
...
+    }
+
+    #[instrument(level = "trace", skip(self), err, ret)]
+    pub async fn resubmit_pending_commitments(&mut self) -> anyhow::Result<()> {
+        let pending_commitments = self.pending_commitments_l2_range.lock().await.clone();
+        for (l2_start, l2_end) in pending_commitments {
+            let commitment_info = commitment_controller::CommitmentInfo {
+                l2_height_range: l2_start..=l2_end,
+            };
+            self.submit_commitment(commitment_info, true).await?;
```
rakanalh, line 624: IMO, this is what i would call blind re-submission. 

Scenario:
- Sequencer sees a commitment is due
- Submits commitment
- Sequencer is stopped while commitment TX is in DA layer
- TX in DA layer goes through while sequencer is stopped
- Sequencer is started
- Sequencer resubmits an already successful commitment tx (pays fee as well)

Last step is redundant and costly.

I would be against merging this approach completely.

The alternative is coming up with a way to let the sequencer figure out which pending blocks were included on DA vs the ones that were not and resubmitting those.

rakanalh (changes requested):
Somehow the scope of this PR is getting bigger as we discuss more issues.

Do we need to discuss this offline? i don't really like the way we implement async commitments, feels rushed.

crates/sequencer/src/sequencer.rs
yaziciahmet: Is there a way to query transactions by the address from DA, and the details of it, so at the restart we can ask it and configure sequencer accordingly? I understand your point and I too agree that this doesn't feel nice. But I think this problem can be generalized as atomicity problem across different components of the system. I don't know of a nice solution to this problem.

rakanalh: One option is to wait for L1 block sync of the sequencer which pretty much does contain information about commitments. Another approach would be to query the Da with the TX hashes we have but that's an extra step.

eyusufatik: I left the comment on #537 for this possible scenario actually :D

crates/sequencer/src/sequencer.rs
eyusufatik: > The sequencer should not cater for the develop / test cycle compared to catering for consistency.

I agree with this tbh.

But again @yaziciahmet has a good point on how it doesn't make a difference if we do this asynchronously.

however, I believe we should act defensively on these kind of issues.

since we don't expect this feature to be used frequently (well, hopefully never) it's a good idea to implement the simplest way possible.

let's go with the sync way.

eyusufatik (approved):

crates/sequencer/src/sequencer.rs
```diff
@@ -591,94 +597,153 @@ where
...
+
+        let ledger_db = self.ledger_db.clone();
+        let db_config = self.config.db_config.clone();
+        let pending_commitments_l2_range = self.pending_commitments_l2_range.clone();
+        let handle_da_response = async move {
+            let result: anyhow::Result<()> = async move {
+                let l2_start = *commitment_info.l2_height_range.start();
+                let l2_end = *commitment_info.l2_height_range.end();
+
+                let tx_id = rx
```
rakanalh, line 676: This is a oneshot channel which is going to be closed when a single signal is received.... cannot be reused.

crates/sequencer/src/sequencer.rs
```diff
@@ -591,94 +597,153 @@ where
...
+            self.ledger_db
+                .set_pending_commitments_l2_range(&pending_commitments)?;
+
+            // Clear state diff
             self.ledger_db.set_state_diff(vec![])?;
             self.last_state_diff = vec![];
 
-            info!("New commitment. L2 range: #{}-{}", l2_start, l2_end,);
+            // Handle DA response asynchronously
+            tokio::spawn(handle_da_response);
```
rakanalh, line 741: Here we're removing the current commitment from pending. However, `handle_da_response` here would be called again and we would remove the same commitment from the list again which has already been removed. I also don't see how calling it again would process other commitments?

rakanalh (dismissed):
How about we write a test for this?

crates/sequencer/src/sequencer.rs
yaziciahmet: AFAIK, `handle_da_response`, which is a Future, will not execute until await is called on it. Hence we are actually not calling it twice, but we are either handing the future to tokio to handle it in a separate green thread (case of regular submission), or we are blocking further execution by polling the Future within the current thread.

I believe this should also answer the previous conversation about oneshot channels.

eyusufatik (approved):

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
```diff
@@ -536,6 +537,40 @@ impl LedgerDB {
...
     }
 
+    /// Gets all pending commitments' l2 ranges.
+    /// Returns start-end L2 heights.
+    #[instrument(level = "trace", skip(self), err)]
+    pub fn get_pending_commitments_l2_range(&self) -> anyhow::Result<Vec<L2HeightRange>> {
+        let mut iter = self.db.iter::<PendingSequencerCommitmentL2Range>()?;
+        iter.seek_to_first();
+
+        let mut l2_ranges = vec![];
```
eyusufatik, line 547: collect() ?

yaziciahmet: done ser

PR #885, opened 2024-07-12
Improve prover implementation

rakanalh:
# Description

Follow up to #857

### DIFF
--- a/crates/prover/src/runner.rs
+++ b/crates/prover/src/runner.rs
@@ -407,28 +407,13 @@ where
                 )
                 .unwrap();
 
-            let mut sequencer_commitments = Vec::<SequencerCommitment>::new();
-
             let mut da_data = self.da_service.extract_relevant_blobs(l1_block);
             // if we don't do this, the zk circuit can't read the sequencer commitments
             da_data.iter_mut().for_each(|blob| {
                 blob.full_data();
             });
-            da_data.iter_mut().for_each(|tx| {
-                let data = DaData::try_from_slice(tx.full_data());
-                // Check for commitment
-                if tx.sender().as_ref() == self.sequencer_da_pub_key.as_slice() {
-                    if let Ok(DaData::SequencerCommitment(seq_com)) = data {
-                        sequencer_commitments.push(seq_com);
-                    } else {
-                        tracing::warn!(
-                            "Found broken DA data in block 0x{}: {:?}",
-                            hex::encode(l1_block.hash()),
-                            data
-                        );
-                    }
-                }
-            });
+            let mut sequencer_commitments: Vec<SequencerCommitment> =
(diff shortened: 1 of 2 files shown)


crates/prover/src/runner.rs
```diff
@@ -205,94 +202,6 @@ where
         });
     }
 
-    async fn process_l2_block(
```
rakanalh, line 208: This was just moved below `run` closer to `process_l1_block`

crates/prover/src/runner.rs
```diff
@@ -411,9 +408,27 @@ where
...
             let mut sequencer_commitments = Vec::<SequencerCommitment>::new();
-            let mut zk_proofs = Vec::<Proof>::new();
 
-            self.extract_relevant_l1_data(l1_block, &mut sequencer_commitments, &mut zk_proofs);
+            let mut da_data = self.da_service.extract_relevant_blobs(l1_block);
+            // if we don't do this, the zk circuit can't read the sequencer commitments
+            da_data.iter_mut().for_each(|blob| {
+                blob.full_data();
+            });
+            da_data.iter_mut().for_each(|tx| {
```
rakanalh, line 417: There is no need to extract `zk_proofs` in prover as it does not process them. 

This code has been simplfied

crates/prover/src/runner.rs
```diff
@@ -572,23 +578,21 @@ where
...
-                        {
-                            Ok(block) => block,
-                            Err(_) => {
-                                return Err(anyhow!(
-                                    "Error while fetching DA block at height: {}",
-                                    soft_batch.da_slot_height
-                                ));
-                            }
-                        };
+                    let filtered_block = match get_da_block_at_height(
```
rakanalh, line 581: This uses the `l1_block_cache` rather than fetching block through da directly.

ercecan (approved):

crates/prover/src/runner.rs
```diff
@@ -411,9 +408,27 @@ where
...
+                if tx.sender().as_ref() == self.sequencer_da_pub_key.as_slice() {
+                    if let Ok(DaData::SequencerCommitment(seq_com)) = data {
+                        sequencer_commitments.push(seq_com);
+                    } else {
+                        tracing::warn!(
+                            "Found broken DA data in block 0x{}: {:?}",
+                            hex::encode(l1_block.hash()),
+                            data
+                        );
+                    }
```
eyusufatik, line 429: imo it would look better if we had a function `extract_sequencer_commitments`

eyusufatik (changes requested):

crates/prover/src/runner.rs
rakanalh: Done

eyusufatik (approved):

PR #891, opened 2024-07-16
Async commitment submission improvements

rakanalh:
# Description

Improvements on the async commitment submission #883

### DIFF
--- a/crates/sequencer/src/sequencer.rs
+++ b/crates/sequencer/src/sequencer.rs
@@ -829,6 +829,11 @@ where
                         (last_finalized_block, l1_fee_rate) = l1_data;
                         last_finalized_height = last_finalized_block.header().height();
 
+                        // An L1 block could contain a commitment that we can clear out of the pending list
+                        if let Err(e) = self.process_l1_block(last_finalized_block.clone()).await {
+                            error!("Could not process L1 block data: {}", e);
+                        }
+
                         if last_finalized_block.header().height() > last_used_l1_height {
                             let skipped_blocks = last_finalized_height - last_used_l1_height - 1;
                             if skipped_blocks > 0 {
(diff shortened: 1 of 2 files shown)


crates/sequencer/src/sequencer.rs
```diff
@@ -954,6 +969,68 @@ where
...
+                        sequencer_commitments.push(seq_com);
+                    } else {
+                        tracing::warn!(
+                            "Found broken DA data in block 0x{}: {:?}",
+                            hex::encode(l1_block.hash()),
+                            data
+                        );
+                    }
+                }
+            });
```
eyusufatik, line 995: I don't think we should scan L1 like  this in the sequencer.

we want to remove the submission from the pending list as soon as the tx lands in the L1 mempool so that if the sequencer restarts again it wont resubmit the commitment. 


again, after #537 this will be handled much better

rakanalh: We are receiving L1 events anyway... we might as well use this to:
1. Confirm a commitment has been included in a new DA block
2. Remove the L2 ranges of those commitments from pending
3. Update PG to set the status of a commitment as mined.

The current way of removing a range from pending is blind (There's no confirmation the tx has been mined, could potentially stay in the mempool for a long time if not dropped).

eyusufatik: > There's no confirmation the tx has been mined, could potentially stay in the mempool for a long time if not dropped

we will handle that with RBF or CPFP on the bitcoin da adapter level.

> so that if the sequencer restarts again it wont resubmit the commitment.

the problem is we are not gonna be able to replace a commitment that has been sent, we will be double sending the commitment

eyusufatik:
closing since concerns of this pr will be handled later on. + don't want to introduce another thing for the sequencer to do (scan L1 and mark db)

PR #893, opened 2024-07-16
Get last tx after restart & chain txs on that utxo

### LINKED ISSUE #537: Bitcoin DA: get last tx after restart

ercecan:
# Description
Gets last tx after restart & chain txs on that utxo

## Linked Issues
- Fixes #537

### DIFF
--- a/crates/bitcoin-da/src/service.rs
+++ b/crates/bitcoin-da/src/service.rs
@@ -437,43 +437,13 @@ impl DaService for BitcoinService {
         &self,
         block: &Self::FilteredBlock,
     ) -> Vec<<Self::Spec as sov_rollup_interface::da::DaSpec>::BlobTransaction> {
-        let mut txs = Vec::new();
-
         info!(
             "Extracting relevant txs from block {:?}",
             block.header.block_hash()
         );
 
-        // iterate over all transactions in the block
-        for tx in block.txdata.iter() {
-            if !tx
-                .txid()
-                .to_byte_array()
-                .as_slice()
-                .starts_with(self.reveal_tx_id_prefix.as_slice())
-            {
-                continue;
-            }
-
-            // check if the inscription in script is relevant to the rollup
-            let parsed_inscription = parse_transaction(tx, &self.rollup_name);
-
-            if let Ok(inscription) = parsed_inscription {
-                if inscription.get_sig_verified_hash().is_some() {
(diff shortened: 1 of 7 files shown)


crates/bitcoin-da/src/service.rs
```diff
@@ -545,6 +589,34 @@ impl DaService for BitcoinService {
...
+                if inscription.get_sig_verified_hash().is_some() {
+                    // Decompress the blob
+                    let decompressed_blob = decompress_blob(&inscription.body);
+
+                    let relevant_tx = BlobWithSender::new(
+                        decompressed_blob,
+                        inscription.public_key,
+                        sha256d::Hash::hash(&inscription.body).to_byte_array(),
+                    );
+
```
eyusufatik, line 613: we do this in another function as well. make a shared function among these two.

eyusufatik (changes requested):

eyusufatik (approved):

crates/bitcoin-da/src/service.rs
```diff
@@ -98,7 +99,21 @@ impl BitcoinService {
             tokio::runtime::Handle::current().block_on(async move {
                 // TODO https://github.com/chainwayxyz/citrea/issues/537
                 // TODO find last tx by utxo chain
-                let mut prev_tx = None;
+                let mut prev_tx = match this.get_pending_transactions().await {
+                    Ok(pending_txs) => {
+                        if !pending_txs.is_empty() {
+                            let tx = pending_txs.first().unwrap().clone();
```
yaziciahmet, line 105: is pending_txs guaranteed to be ordered?

kpp, line 105: The logic is unclear to me. Shouldnt we sort by height desc?

crates/bitcoin-da/src/spec/transaction.rs
```diff
@@ -24,6 +24,10 @@ impl TransactionWrapper {
         };
         Self { tx }
     }
+
+    pub fn inner(&self) -> &BitcoinTransaction {
```
kpp, line 28: No need, there is Deref trait

kpp (approved):

PR #896, opened 2024-07-17
Fix commitment threshold test

rakanalh:
# Description
- Sequencer was not clearing state diffs properly which is why diff threshold test was failing.
- Some tests are failing because we've changed `wait_for_l2_block` from `get_eth_number` to `get_head_soft_batch_height` which blocks and might cause some delays being written to the filesystem.

### DIFF
--- a/crates/sequencer/src/sequencer.rs
+++ b/crates/sequencer/src/sequencer.rs
@@ -625,10 +625,14 @@ where
     async fn submit_commitment(
         &mut self,
         commitment_info: commitment_controller::CommitmentInfo,
-        resubmit: bool,
+        wait_for_da_response: bool,
     ) -> anyhow::Result<()> {
         let l2_start = *commitment_info.l2_height_range.start();
         let l2_end = *commitment_info.l2_height_range.end();
+
+        // Clear state diff early
+        self.ledger_db.set_state_diff(vec![])?;
+        self.last_state_diff = vec![];
 
         // calculate exclusive range end
         let range_end = BatchNumber(l2_end.0 + 1); // cannnot add u64 to BatchNumber directly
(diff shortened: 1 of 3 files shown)


crates/sequencer/src/sequencer.rs
```diff
@@ -625,11 +625,15 @@ where
...
-        resubmit: bool,
+        wait_for_da_response: bool,
     ) -> anyhow::Result<()> {
         let l2_start = *commitment_info.l2_height_range.start();
         let l2_end = *commitment_info.l2_height_range.end();
 
+        // Clear state diff early
+        self.ledger_db.set_state_diff(vec![])?;
+        self.last_state_diff = vec![];
+
```
yaziciahmet, line 636: I am not sure I understand why this fixes stuff. Can you explain this to me plz?

Because from this location to its previous location which is inside the if-else branch, everything is synchronous, and the only mutative thing happening is sending to transaction queue. Only thing I can think of is that for MockDA finding reveal nonce and submitting tx is instantenous, but I guess I am not properly informed about what is exactly happenning and which states are changing on the DA side.

eyusufatik:
> Some tests are failing because we've changed wait_for_l2_block from get_eth_number to get_head_soft_batch_height which blocks and might cause some delays being written to the filesystem.

very weird. I thought I fixed that race condition (don't know if we can call it that)

crates/sequencer/src/sequencer.rs
rakanalh: There are multiple reasons:
1. In the previous version, clearing the state diff would happen when `resubmit` (as it was called) and if we pass `rebusmit` as `false` it would not clear the state diff.
2. There are writes to storage which internally call `block_in_place` that might cause some small delays and therefore the whole function would take more time to finish
3. Meanwhile L2 blocks get published leading to the current L2 block and the previous detecting that a state diff threshold is reached and therefore submitting two consecutive commitments which is unintended behavior.

rakanalh: Also, even if we submit commitments asynchronously, we should also clear the state diff because we start calculating the size starting from the last commitment up to the current point.

yaziciahmet: I don't really understand the 1st point since so far I haven't got into storage side of things, but regarding 2nd point, since this gets executed within a branch of select statement, isn't that blocking? Hence there will be no L2 block creation while this piece of code is being executed.

eyusufatik (approved):

PR #898, opened 2024-07-17
Erce/last verified proof

### LINKED ISSUE #691: Implement ledger_getLastVerifiedProof
I suggest we add last verified proof rpc endpoint

ercecan:
# Description
Implements get last verified proof

## Linked Issues
- Fixes #691


crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/rpc.rs
```diff
@@ -428,6 +428,14 @@ impl LedgerRpcProvider for LedgerDB {
         }
     }
 
+    fn get_last_verified_proof(&self) -> Result<Option<VerifiedProofResponse>, anyhow::Error> {
```
eyusufatik, line 431: what if there wasn't a proof on the last L1 block?

this should return the *last* proof. maybe along with the L1 height it was found on

eyusufatik: maybe changing L1 heigh => verified proof table encoding to be seek encoding is enough to pull the last proof in that table.

eyusufatik (changes requested):

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/rpc.rs
ercecan: Done, now it uses seek to last to get the latest proof also returns l1 height with it.

eyusufatik (approved):

PR #899, opened 2024-07-17
Check pending da commitments before resubmit

### LINKED ISSUE #889: Implement proper resubmission of pending sequencer commitments
after #883 and #537 is closed, we'll need to do better resubmissions of sequencer commitments. For a commitment in sequencer's pending commitment table, there is 2 possible ways why it wasn't removed from the table:

1. While trying to find a reveal tx with correct nonce, the sequencer shutdown and never actually submitted the tx to the bitcoin network.
2. The bitcoin da adapter broadcasted the tx however before the sequencer could remove the commitment from pending table, it shutdown.

What we can do is; in the resubmit sequencer commitment function, we can call `get_pending_transactions` which when parsed, would give us the ranges of sequencer commitments that are in the mempool. Compare it to the list from ledger db and *resubmit* the ones that are in the table but not in the bitcoin mempool.

One possible scenario with this approach:

1. sequencer shutdowns with 2. (first list)
2. before the sequencer could reboot, the tx was included in a bitcoin block.
3. sequencer calls get pending txs, filters out the table accordingly.
4. resubmits the commitment commitment in 2.  (this list)

to solve this problem, we'll need to implement sequencer commitment tracking on the sequencer (whatever the full node does + for reboot scenario scan non-finalized bitcoin blocks.)

yaziciahmet:
# Description

Before resubmission of pending commitments, we now check whether this already exists in DA mempool to avoid paying for a fee again.

## Linked Issues-
- Fixes #889

### DIFF
--- a/crates/sequencer/src/sequencer.rs
+++ b/crates/sequencer/src/sequencer.rs
@@ -615,7 +615,7 @@ where
         let pending_db_commitments = self.ledger_db.get_pending_commitments_l2_range()?;
         debug!("Pending db commitments: {:?}", pending_db_commitments);
         let pending_da_commitments = self.get_pending_da_commitments().await;
-        debug!("Pending da commitments: {:?}", pending_db_commitments);
+        debug!("Pending da commitments: {:?}", pending_da_commitments);
         // TODO: also take mined DA blocks into account
         for (l2_start, l2_end) in pending_db_commitments {
             if pending_da_commitments.iter().any(|commitment| {
--- a/crates/sequencer/src/sequencer.rs
+++ b/crates/sequencer/src/sequencer.rs
@@ -616,11 +616,21 @@ where
         debug!("Pending db commitments: {:?}", pending_db_commitments);
         let pending_da_commitments = self.get_pending_da_commitments().await;
         debug!("Pending da commitments: {:?}", pending_db_commitments);
+        // TODO: also take mined DA blocks into account
         for (l2_start, l2_end) in pending_db_commitments {
             if pending_da_commitments.iter().any(|commitment| {
                 commitment.l2_start_block_number == l2_start.0
                     && commitment.l2_end_block_number == l2_end.0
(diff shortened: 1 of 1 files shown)


crates/sequencer/src/sequencer.rs
```diff
@@ -611,13 +612,35 @@ where
...
+        let pending_da_commitments = self.get_pending_da_commitments().await;
+        debug!("Pending da commitments: {:?}", pending_da_commitments);
+        // TODO: also take mined DA blocks into account
+        for (l2_start, l2_end) in pending_db_commitments {
+            if pending_da_commitments.iter().any(|commitment| {
+                commitment.l2_start_block_number == l2_start.0
+                    && commitment.l2_end_block_number == l2_end.0
+            }) {
+                // Update last sequencer commitment l2 height
+                match self.ledger_db.get_last_sequencer_commitment_l2_height()? {
```
eyusufatik, line 626: I'd get last sequencer commitment l2 height before the for loop then compare to that variable and set if needed

```
last = get_last_commitment()

for commitment in pending():
  if !commitment.in_mempool():
      resubmit()
  else:
      delete commitment
   
  if commitment >= last:
      set_last(commitment)
      last = commitment
         

```

eyusufatik (changes requested):

crates/sequencer/src/sequencer.rs
yaziciahmet: makes 1000% sense. done

yaziciahmet: now i think, makes 10% sense. because because in the else branch submit_commitment could also potentially update last_commitment_l2_height. we could also track and retrieve updates after submit_commitment finishes, but imo its better as it is right now, especially considering that it will almost always go to else { submit_commitmen } branch

eyusufatik (approved):

PR #903, opened 2024-07-19
CI Check genesis files

### LINKED ISSUE #828: CI for bytecode matching in genesis files against Foundry compilation outputs
The process of pre-deploying contracts is now taking the deployed bytecodes from Foundry compilation output, and copy pasting them into genesis files, but this process is error-prone and as in code reviews these bytecodes are not checked against the proper contract its also opaque and introduces some risks. CI that compares these would be beneficial. Having a script that autogenerate these genesis files can also be considered and would significantly improve the developer experience.

ercecan:
# Description
Checks if make genesis was executed correctly

## Linked Issues
- Fixes #828

### DIFF
--- a/.github/scripts/check_genesis.sh
+++ b/.github/scripts/check_genesis.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+# Paths
+paths=( "./resources/genesis/bitcoin-regtest/evm.json" 
+        "./resources/genesis/mock/evm.json" 
+        "./resources/genesis/mock-dockerized/evm.json" 
+        "./resources/test-data/demo-tests/bitcoin-regtest/evm.json" 
+        "./resources/test-data/demo-tests/mock/evm.json" 
+        "./resources/test-data/integration-tests/evm.json" 
+        "./resources/test-data/integration-tests-low-block-gas-limit/evm.json" 
+        "./resources/test-data/integration-tests-low-max-l2-blocks-per-l1/evm.json")
+
+# Function to clean up copied directories
+cleanup() {
+    for path in "${paths[@]}"; 
+    do
+        rm -rf "${path}_tmp"
+    done
+}
+cleanup
+
+sleep 1
+
+# Copy directories
+for path in "${paths[@]}"; 
+do
+    cp -r "$path" "${path}_tmp"
+done
+
+sleep 5
+
+# Run make command
+make genesis
+
+# Compare directories
+for path in "${paths[@]}"; 
+do
+    diff -r "$path" "${path}_tmp"
+    if [ $? -ne 0 ]; then
+        echo "Error: Differences found between $path and ${path}_tmp"
+        cleanup
+        exit 1
+    fi
+done
+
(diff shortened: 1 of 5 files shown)


okkothejawa (approved):

eyusufatik (approved):

.github/scripts/check_genesis.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+
+# Paths
+paths=( "./resources/genesis/bitcoin-regtest/evm.json"
```
kpp, line 4:
```
>>> import glob
>>> lst = glob.glob('**/evm.json', recursive=True)
>>> print(lst)
['resources/genesis/mock/evm.json', 'resources/genesis/bitcoin-regtest/evm.json', 'resources/genesis/mock-dockerized/evm.json', 'resources/genesis/devnet/evm.json', 'resources/test-data/integration-tests-low-block-gas-limit/evm.json', 'resources/test-data/integration-tests/evm.json', 'resources/test-data/demo-tests/mock/evm.json', 'resources/test-data/demo-tests/bitcoin-regtest/evm.json', 'resources/test-data/integration-tests-low-max-l2-blocks-per-l1/evm.json', 'crates/sovereign-sdk/examples/test-data/genesis/integration-tests-low-limiting-number/evm.json', 'crates/sovereign-sdk/examples/test-data/genesis/integration-tests/evm.json', 'crates/sovereign-sdk/examples/test-data/genesis/demo-tests/mock/evm.json', 'crates/sovereign-sdk/examples/test-data/genesis/demo-tests/celestia/evm.json', 'crates/sovereign-sdk/examples/test-data/genesis/demo-tests/bitcoin/evm.json']
```

.github/scripts/check_genesis.sh
```diff
@@ -0,0 +1,52 @@
...
+
+sleep 1
+
+# Run make command
+make genesis
+
+# Compare directories
+for path in "${paths[@]}"; 
+do
+    diff -r "$path" "${path}_tmp"
```
kpp, line 38: We don't have to do that thing with tmp, we need to `git diff --exit-code` and check if exit code = 1, then there are some changes.

kpp (changes requested):
`git diff --exit-code` is actually better because we can diff even files not listed in paths

.github/scripts/check_genesis.sh
okkothejawa: I'm not sure if we should do this, as for example `resources/genesis/devnet/evm.json` folder is an exception as it is not affected by automated genesis construction. In this case, I think it would work as `make genesis` wouldn't attempt to change that file either, so I'm not sure.

okkothejawa: wdyt @eyusufatik

eyusufatik: yes. production genesis files e.g. devnet or mainnet

should be only produced by our trigger, not auto generated

kpp (approved):

PR #909, opened 2024-07-19
Implement debug_traceChain

### LINKED ISSUE #704: Implement trace_filter

yaziciahmet:
# Description

Implements debug_traceChain subscription rpc.

## Linked Issues
- Fixes #704

### DIFF
--- a/crates/ethereum-rpc/src/lib.rs
+++ b/crates/ethereum-rpc/src/lib.rs
@@ -11,7 +11,7 @@ pub use gas_price::fee_history::FeeHistoryCacheConfig;
 use gas_price::gas_oracle::GasPriceOracle;
 pub use gas_price::gas_oracle::GasPriceOracleConfig;
 use jsonrpsee::types::ErrorObjectOwned;
-use jsonrpsee::RpcModule;
+use jsonrpsee::{RpcModule, SubscriptionMessage};
 use reth_primitives::{keccak256, BlockNumberOrTag, Bytes, B256, U256};
 use reth_rpc::eth::error::EthApiError;
 use reth_rpc_types::trace::geth::{
@@ -489,10 +489,10 @@ fn register_rpc_methods<C: sov_modules_api::Context, Da: DaService>(
 
     rpc.register_async_method::<Result<Vec<GethTrace>, ErrorObjectOwned>, _, _>(
         "debug_traceBlockByHash",
-        |parmaeters, ethereum| async move {
+        |parameters, ethereum| async move {
             info!("eth module: debug_traceBlockByHash");
 
-            let mut params = parmaeters.sequence();
+            let mut params = parameters.sequence();
 
             let block_hash: B256 = params.next()?;
             let evm = Evm::<C>::default();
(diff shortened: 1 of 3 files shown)


crates/ethereum-rpc/src/lib.rs
```diff
@@ -638,52 +565,135 @@ fn register_rpc_methods<C: sov_modules_api::Context, Da: DaService>(
...
+                Ok(v) => v,
+                Err(err) => {
+                    pending.reject(err).await;
+                    return Ok(());
+                }
+            };
+            match topic.as_str() {
+                "traceChain" => {
+                    let start_block: BlockNumberOrTag = match params.next() {
+                        Ok(v) => v,
```
eyusufatik, line 597: move this match arm to some function like `handle_trace_chain`

eyusufatik: so that when we have another arm we respond to, code is more readable

crates/ethereum-rpc/src/lib.rs
```diff
@@ -638,52 +565,135 @@ fn register_rpc_methods<C: sov_modules_api::Context, Da: DaService>(
...
-            )?;
+            Ok(traces[0].clone())
+        },
+    )?;
 
-            Ok(traces.into_iter().next().unwrap())
+    rpc.register_subscription(
+        "debug_subscribe",
+        "debug_subscription",
+        "debug_unsubscribe",
```
rakanalh, line 583: How is `unsubscribe` handled here?

rakanalh (approved):
LGTM

crates/ethereum-rpc/src/lib.rs
yaziciahmet: Once the client sends a rpc request with method `debug_unsubscribe` and the subscription id as the `params`, jsonrpsee will automatically unsubscribe the client.

For this endpoint, even though this is a subscription endpoint due to potentially really long operation, it is finite because start and end blocks are specified, hence it makes sense for this subscription or connection to be closed by the server once the last block is sent. But it appears that jsonrpsee does not provide closing functionality on the server side.

It makes sense for server to not close for real-time endpoints, but for this case I believe it should be done. But such functionality simply doesn't exist for some reason.

crates/ethereum-rpc/src/lib.rs
yaziciahmet: Done.

yaziciahmet:
> single nit. merge after fix. also, have you tested this query with something like ethers.js?

I wasn't able to. ethers.js doesn't support this out of the box. It supports custom rpc requests with custom method and params, and it supports custom subscription topics for `eth_subscribe`, but it doesn't support `debug_subscribe`. When I sent rpc request with simply `provider.send("debug_subscribe", [<args>])`, it just returns the subscription id as that is the first response returned, but it doesn't know that it should wait, hence it ignores the rest.

I tested with wscat using the following command:
```
wscat -c ws://localhost:12345 -w 10 -x '{"jsonrpc": "2.0", "id": 1, "method": "debug_subscribe", "params": ["traceChain", "0x1", "latest"]}'
```

eyusufatik:
>  wasn't able to. ethers.js doesn't support this out of the box. It supports custom rpc requests with custom method and params, and it supports custom subscription topics for eth_subscribe, but it doesn't support debug_subscribe. When I sent rpc request with simply provider.send("debug_subscribe", [<args>]), it just returns the subscription id as that is the first response returned, but it doesn't know that it should wait, hence it ignores the rest.

I am suspecting more and more if this is going to replace trace filter 😅 

anyways let's call it the more the merrier

Issue #871, opened 2024-07-10
Bitcoin DA supports "versions"


eyusufatik:
Before we start supporting chunk da transactions, we need a few features and optimizations regarding the way we do DA transactions.
TODOs:

- [ ] Get rid of “tags” in transactions, enforce a certain order.
- [ ] Add 1 byte type descriptor, so that we can have different type of DA transactions. (this can both be utilized for future upgrades) -- by the way let's discuss this. we can have the upgradeability on the DaData struct.
- [ ] The new order for DA transaction fields should be rollup name, da tx type, public key, signature (remind you these are different than bitcoin signatures and public keys), nonce, chunks of 520 bytes of body.

Should be merged before #743

kpp:
In order to distinguish different transactions, we need to define a header, which will affect transaction body.

New bitcoin transaction header:

```
PushBytes(TagType[8 bytes, 6 for rollup name, 2 for transaction type])
```

1. Legacy reveal tx (when blob is less than 400kb):

Type: 0

Body:

```
PushBytes(XOnlyPublicKey)
OP_CHECKSIG
OP_FALSE
OP_IF
PushBytes(signature)
PushBytes(sequencer_public_key)
PushBytes(1; - up to 520-bytes chunk of blob)
...
PushBytes(N; - up to 520-bytes chunk of blob)
OP_ENDIF
PushBytes(8 random bytes)
```

2. Chunk-aggregated reveal tx(when blob is greater than 400kb):

Type: 1

Body:

```
PushBytes(XOnlyPublicKey)
OP_CHECKSIG
OP_FALSE
OP_IF
PushBytes(signature)
PushBytes(sequencer_public_key)
PushBytes(len of chunks [2 bytes in little endian])
PushBytes(1; tx_id [64 bytes])
...
PushBytes(N; tx_id [64 bytes])
OP_ENDIF
```

2. Chunk-aggregated reveal tx(a part of Type-1):

Type: 2

Body:

```
PushBytes(XOnlyPublicKey)
OP_CHECKSIG
OP_FALSE
PushBytes(1; - up to 520-bytes chunk of blob)
...
PushBytes(N; - up to 520-bytes chunk of blob)
```

PR #912, opened 2024-07-19
Bitcoin DA versions

kpp:
# Description

TODOs:

- [x] Get rid of “tags” in transactions, enforce a certain order.
- [x] Add 2 byte type descriptor, so that we can have different type of DA transactions. 
- [x] The new order for DA transaction fields are: header(da tx type, rollup name), signature, public key, chunks of 520 bytes of body, nonce.
- [ ] Fix tests

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/871

### DIFF
--- a/crates/bitcoin-da/src/helpers/builders.rs
+++ b/crates/bitcoin-da/src/helpers/builders.rs
@@ -5,7 +5,7 @@ use std::io::{BufWriter, Write};
 
 use anyhow::anyhow;
 use bitcoin::absolute::LockTime;
-use bitcoin::blockdata::opcodes::all::{OP_CHECKSIG, OP_ENDIF, OP_IF};
+use bitcoin::blockdata::opcodes::all::{OP_CHECKSIG, OP_DROP, OP_ENDIF, OP_IF};
 use bitcoin::blockdata::opcodes::OP_FALSE;
 use bitcoin::blockdata::script;
 use bitcoin::hashes::{sha256d, Hash};
@@ -22,7 +22,7 @@ use bitcoin::{
 };
 use tracing::{instrument, trace, warn};
 
-use crate::helpers::{BODY_TAG, PUBLICKEY_TAG, RANDOM_TAG, ROLLUP_NAME_TAG, SIGNATURE_TAG};
+use super::{TransactionHeader, TransactionType};
 use crate::spec::utxo::UTXO;
 use crate::REVEAL_OUTPUT_AMOUNT;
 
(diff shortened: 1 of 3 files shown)


crates/bitcoin-da/src/helpers/mod.rs
```diff
@@ -1,13 +1,51 @@
...
 pub mod builders;
 pub mod compression;
 pub mod parsers;
 #[cfg(test)]
 pub mod test_utils;
+
+/// Header - first 8 bytes of any rollup transaction
+struct TransactionHeader<'a> {
+    pub(crate) rollup_name: &'a [u8],
+    pub(crate) typ: TransactionType,
```
eyusufatik, line 13: type*

crates/bitcoin-da/src/helpers/mod.rs
```diff
@@ -1,13 +1,51 @@
...
+        };
+        Some(Self { rollup_name, typ })
+    }
+}
+
+/// Type represents
+#[repr(u16)]
+enum TransactionType {
+    /// This type of transaction includes full body (< 400kb)
+    Inscribed = 0,
```
eyusufatik, line 49: can we call this "regular" or maybge smth else? 

Open to suggestions

Inscribed is just too vague :D

crates/bitcoin-da/src/helpers/builders.rs
```diff
@@ -336,34 +336,76 @@ pub fn create_inscription_transactions(
...
+// TODO: parametrize hardness
+// so tests are easier
+// Creates the inscription transactions (commit and reveal) Type 0
+#[allow(clippy::too_many_arguments)]
+#[instrument(level = "trace", skip_all, err)]
+pub fn create_inscription_type_0(
+    rollup_name: &[u8],
+    body: Vec<u8>,
+    signature: Vec<u8>,
+    sequencer_public_key: Vec<u8>,
```
eyusufatik, line 366: can you call this signer_public_key?

bc prover also uses this method. since we are refactoring here, we can rename

crates/bitcoin-da/src/helpers/builders.rs
```diff
@@ -336,34 +336,76 @@ pub fn create_inscription_transactions(
...
+            .push_slice(PushBytesBuf::try_from(chunk.to_vec()).expect("Cannot push body chunk"));
+    }
+    // push end if
+    reveal_script_builder = reveal_script_builder.push_opcode(OP_ENDIF);
+
+    // This envelope is not finished yet. The random number will be added later
 
     // Start loop to find a 'nonce' i.e. random number that makes the reveal tx hash starting with zeros given length
-    let mut nonce: i64 = 0;
+    let mut nonce: i64 = 16;
```
eyusufatik, line 408: add comment explaining why this starts from 16.

also on the verifier side (parser) do we enforce nonce being bigger than 16?

crates/bitcoin-da/src/helpers/parsers.rs
```diff
@@ -71,111 +84,109 @@ fn get_script(tx: &Transaction) -> Result<&Script, ParserError> {
...
+
+fn read_opcode(
+    instructions: &mut dyn Iterator<Item = Result<Instruction<'_>, ParserError>>,
+) -> Result<Opcode, ParserError> {
+    let instr = read_instr(instructions)?;
+    let Op(op) = instr else {
+        return Err(ParserError::UnexpectedOpcode);
+    };
+    Ok(op)
+}
```
eyusufatik, line 136: do we benefit from these read_* functions being inline?

crates/bitcoin-da/src/helpers/parsers.rs
```diff
@@ -71,111 +84,109 @@ fn get_script(tx: &Transaction) -> Result<&Script, ParserError> {
...
+    if !op_false.is_empty() {
+        return Err(ParserError::UnexpectedOpcode);
+    }
+
+    if OP_IF != read_opcode(instructions)? {
+        return Err(ParserError::UnexpectedOpcode);
+    }
+
+    let signature = read_push_bytes(instructions)?;
+    let public_key = read_push_bytes(instructions)?;
```
eyusufatik, line 158: how are you reading public key 2 times?

crates/bitcoin-da/src/helpers/parsers.rs
```diff
@@ -71,111 +84,109 @@ fn get_script(tx: &Transaction) -> Result<&Script, ParserError> {
...
+    if OP_DROP != read_opcode(instructions)? {
+        return Err(ParserError::UnexpectedOpcode);
+    }
+    // END of transaction
+    if instructions.next().is_some() {
+        return Err(ParserError::UnexpectedOpcode);
+    }
+
+    let body_size: usize = chunks.iter().map(|c| c.len()).sum();
+    let mut body = Vec::with_capacity(body_size);
```
eyusufatik, line 182: are we really gonna benefit from creating a vec with capacity while we are iterating over chunks and summing lengths?

calculating length vs increasing the capacity of a vector.

wdyt? which has less cycles

eyusufatik (changes requested):
great pr

crates/bitcoin-da/src/helpers/mod.rs
kpp: Type is a reserved word. I can still use it but I have to write `r#type` instead.

crates/bitcoin-da/src/helpers/builders.rs
kpp: No, we don't enforce it on the parser side because we don't care as long as prefix is fine.

crates/bitcoin-da/src/helpers/mod.rs
eyusufatik: that's fine, leave is as it is

crates/bitcoin-da/src/helpers/parsers.rs
kpp: This is for readability only. But I believe rustc should inline them.

crates/bitcoin-da/src/helpers/parsers.rs
kpp: The first one is xonlypublickey, the second one is sequencer/signer public key

crates/bitcoin-da/src/helpers/parsers.rs
kpp: increasing the capacity of a vector on the fly requires reallocation which is not free. Storing refs of chunks, summing their length is almost a no op for risc0

crates/bitcoin-da/src/helpers/parsers.rs
eyusufatik: we are parsing the second script in the witness stack. it does not start with an xonly public key

crates/bitcoin-da/src/helpers/mod.rs
kpp: Inlined as an opposite to Chunked?

crates/bitcoin-da/src/helpers/parsers.rs
eyusufatik: my bad, got confused

crates/bitcoin-da/src/helpers/mod.rs
eyusufatik: what about "unified" or "complete"

crates/bitcoin-da/src/helpers/builders.rs
```diff
@@ -336,34 +336,76 @@ pub fn create_inscription_transactions(
...
 
+    let header = TransactionHeader {
+        rollup_name,
+        typ: TransactionType::Inscribed,
+    };
+
     // start creating inscription content
-    let reveal_script_builder = script::Builder::new()
+    let mut reveal_script_builder = script::Builder::new()
+        .push_slice(PushBytesBuf::try_from(header.to_bytes()).expect("Cannot push header"))
```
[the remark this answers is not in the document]
kpp: It's not and error, it must always succeed.

crates/bitcoin-da/src/helpers/mod.rs
```diff
@@ -1,13 +1,51 @@
...
+        }
+        let mut type_bytes = [0; 2];
+        type_bytes.copy_from_slice(type_slice);
+        let typ = match u16::from_le_bytes(type_bytes) {
+            0 => TransactionType::Inscribed,
+            n => TransactionType::Unknown(NonZeroU16::new(n).expect("Is not zero")),
+        };
+        Some(Self { rollup_name, typ })
+    }
+}
```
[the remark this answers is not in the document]
kpp: It may fail IFF len is < 2, so no need to use those context and Result. Match checks for 0, so NonZeroU16::new won't fail.

kpp:
Closing in favor of https://github.com/chainwayxyz/citrea/issues/871

PR #913, opened 2024-07-22
Add change upgradeability admin test

okkothejawa:
# Description
Adds a test scenario that transfers the ownership of the `ProxyAdmin` to another address and attempts a upgrade a with the new owner.

## Linked Issues
- Partially fixes 869


eyusufatik (approved):
looks ok but why don't you call ProxyAdmin::owner()?

is there no such function?

Issue #117, opened 2024-01-16
Websocket RPC


eyusufatik:
Ethereum clients support RPC through websockets. However, it's not the case that it's a stream kind of websocket server/client relationships we usually see on CEX APIs or many other websocket use cases. Rather, it's just sending requests through the websocket. 

Not sure if Sovereign SDK has plans to implement this, also we don't have to.

If it turns out to be an easy thing we should, if not we can postpone this until we have everything else ready.

PR #916, opened 2024-07-22
Realtime Subscription RPCs

### LINKED ISSUE #118: Subscription based Ethereum RPCs
Ethereum has a few RPC calls that let's users to subscribe to topics. e.g. [eth_newFilter](https://ethereum.org/en/developers/docs/apis/json-rpc#eth_newfilter). These can be useful for full node runners, like API providers and indexers.

The way these RPC's works is user sends something to subscribe to and the node returns an ID for this topic. 

When the user calls the corresponding "get new events" RPC with the ID, the node returns new events matching the topic given in the first RPC call, since the last time the events were queried.

We can leverage the accessory state for this, since this is a node unique operation.

The RPCs we must implement are:

- [ ] `eth_newFilter` -> `eth_getFilterChanges`
- [ ]  `eth_newBlockFilter` -> `eth_getFilterChanges`
- [ ]  `eth_uninstallFilter`

yaziciahmet:
# Description

Initialized realtime RPCs. Currently only contains `newHeads` and `logs` subscriptions. I also introduced new config parameters: `disable_subscriptions` and `max_subscription_connections`.

Node implementations (sequencer, fullnode, prover) now receives a sender broadcast channel to notify about new L2 soft confirmations. RPC layer has its own wrapper around a receiver to this channel to actually use this notifications to query Ethereum blocks and send them through the Websocket connections.

In the future, when different types of notifications proves to be needed, it would make sense to replace this broadcast channel with ChainEvents struct which would handle different types of events happening on the chain.

## Linked Issues
- Fixes #118 
- Related to #117


yaziciahmet:
Q: Does this handle reorg out of the box? If not, what to do?

yaziciahmet:
Q: Other than Ethereum block fields, this additionally returns l1FeeRate and l1Hash. Are we fine with that?

bin/citrea/src/rollup/mod.rs
```diff
@@ -37,15 +38,21 @@ pub trait CitreaRollupBlueprint: RollupBlueprint {
         // Maybe whole "prev_root" can be initialized inside runner
         // Getting block here, so prover_service doesn't have to be `Send`
 
-        let ledger_db = self.create_ledger_db(&rollup_config);
+        let (soft_confirmation_tx, _) = broadcast::channel(10);
```
rakanalh, line 41: The channel size of 10 i believe should be either configurable or a much larger number

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
```diff
@@ -353,6 +359,9 @@ impl LedgerDB {
 
         self.db.write_schemas(schema_batch)?;
 
+        // Error can be ignored becaue it only errors when there are no receivers
+        let _ = self.soft_confirmation_tx.send(batch_to_store.l2_height);
```
rakanalh, line 363: Why do this here and not where we call `commit_soft_batch` (sequencer)? Separation of concerns here is violated because we're broadcasting in the storage layer which is not part of it's job. Also this added some extra parameters to initializing the storage layer which are not part of storage requirements.

rakanalh (changes requested):

bin/citrea/src/rollup/mod.rs
yaziciahmet: IMHO it doesn't make sense to allow user to configure this. We can definitely increase this number, but I am not sure why you think it should much larger? This is only the size of the message buffer, meaning that the requirement for buffer size doesn't increase with the number of consumers. So one block comes to this buffer, is cloned and sent to each consumer, and dropped from the buffer.

rakanalh: It is not the size of the message buffer. 
https://docs.rs/tokio/latest/tokio/sync/broadcast/fn.channel.html

A channel's capacity is how many messages you can push into the channel before `send` starts blocking.
> The channel is created with a capacity of 32. If messages are sent faster than they are received, the channel will store them. Once the 32 messages are stored in the channel, calling send(...).await will go to sleep until a message has been removed by the receiver.

yaziciahmet: I don't think that is the case. We have only 1 sender, and we will push to this channel newly created blocks 1 by 1. And internally this broadcast channel will distribute that message to each receiver. It wouldn't make sense for it to create the same message for 100 subscriber, and push 100 of the same message. So only send rate matters, consumer doesn't matter (or it matters as much as it prolongs the blocking `tx.send()` call).

I tested it just in case with the following piece of code:
```rs
async fn test_sub() -> anyhow::Result<()> {
    let (tx, _) = broadcast::channel(1);

    let msg_count = 3;

    // Run 10 consumers
    for i in 0..10 {
        let mut rx = tx.subscribe();
        tokio::spawn(async move {
            for _ in 0..msg_count {
                let value = rx.recv().await.unwrap();
                println!("Consumer {} received value {}", i, value);
            }
        });
    }

    // Run the producer
    for i in 0..msg_count {
        tx.send(i).unwrap();
        sleep(Duration::from_millis(10)).await;
    }

    Ok(())
}
```
and the output is:
```
Consumer 4 received value 0
Consumer 1 received value 0
Consumer 6 received value 0
Consumer 5 received value 0
Consumer 0 received value 0
Consumer 3 received value 0
Consumer 2 received value 0
Consumer 7 received value 0
Consumer 8 received value 0
Consumer 9 received value 0
Consumer 4 received value 1
Consumer 6 received value 1
Consumer 0 received value 1
Consumer 1 received value 1
Consumer 8 received value 1
Consumer 9 received value 1
Consumer 3 received value 1
Consumer 7 received value 1
Consumer 5 received value 1
Consumer 2 received value 1
Consumer 4 received value 2
Consumer 0 received value 2
Consumer 8 received value 2
Consumer 9 received value 2
Consumer 3 received value 2
Consumer 7 received value 2
Consumer 2 received value 2
Consumer 5 received value 2
Consumer 6 received value 2
Consumer 1 received value 2
```

I also tested with 100 consumers, all of them still received all messages.

rakanalh: I stand corrected. Thank you for putting the effort to show this.

eyusufatik:
> Q: Does this handle reorg out of the box? If not, what to do?

I don't think so. But we don't reorg as of now. don't need to think about it

eyusufatik:
> Q: Other than Ethereum block fields, this additionally returns l1FeeRate and l1Hash. Are we fine with that?
yes we should

eyusufatik (changes requested):
good implementation. pls also think about @rakanalh 's comment on where the send on channel should happen. 

I think we should have one more feature on this PR. 

I don't want our rpc nodes to handle subscriptions. let's put a new boolean config on rollup config for RPC. you can call it `allow_subscriptions` or smth like that.

also there should be a max_subscriptions on jsonrpsee config, you can pass that through rollup config as well, check `max_connections` deafult can be something like 100 idk

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
rakanalh: Yup, correct

bin/citrea/src/rollup/mod.rs
```diff
@@ -37,15 +38,21 @@ pub trait CitreaRollupBlueprint: RollupBlueprint {
...
+        let ledger_db = self.create_ledger_db(&rollup_config, soft_confirmation_tx.clone());
         let genesis_config = self.create_genesis_config(runtime_genesis_paths, &rollup_config)?;
 
         let mut storage_manager = self.create_storage_manager(&rollup_config)?;
         let prover_storage = storage_manager.create_finalized_storage()?;
 
         // TODO(https://github.com/Sovereign-Labs/sovereign-sdk/issues/1218)
-        let rpc_methods =
-            self.create_rpc_methods(&prover_storage, &ledger_db, &da_service, None)?;
+        let rpc_methods = self.create_rpc_methods(
```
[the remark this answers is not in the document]
yaziciahmet: honestly I don't know. I still don't know much about prover, hence I implemented this for all types of nodes. If prover won't be used as an rpc endpoint, there is no reason for it to subscribe to blocks at all. I will change it.

yaziciahmet:
After all the comments and discussions, here is how I see this PR going:
1. I add `disable_subscriptions` and `max_subscription_count` config parameters, which defaults to `false` and `100` respectively.
2. I create a `EthereumSubscriptionManager` type. Fullnode and Sequencer have an instance of this manager, and broadcasts l2 height when a new soft confirmation is committed using the manager. RPC methods also have access to this manager, and are able to subscribe to different types of events. 
3. `EthereumSubscriptionManager`, based on the subscription type, queries the evm storage and broadcasts it to all the subscribers. This way, each subscriber won't have query the same evm block for instance again and again. We will be able to query it once, and copy it in memory, and send it to all subscribers.

With this approach, we don't need cache whatsoever since we just query once and send to everyone.

crates/prover/src/runner.rs
eyusufatik, line 1: let's not have subscriptions on prover

yaziciahmet:
IMO there are 2 places where performance can be improved:
1. In logs subscription, for each filter in the hash map, we call `evm.eth_get_logs(filter)`. Even though each logs has to be filtered repeatedly for each one of the filters, calling the `eth_get_logs` high level method causes the block, transactions, and receipts to be queried again and again for each filter. With the cost of potential duplication, and the dirtiness of exposing some internal methods from evm, we can query all the receipts and even all the logs just once at the beginning, and just apply filters onto them.
2. Implementation of Hash trait for Filter. Currently, it is serialized into json and the bytes are hashed. But json is definitely not the optimal data type for serialization and in terms of byte-size. We can use Borsh maybe, or to avoid serialization completely, we can use unsafe Rust and get the memory layout of the struct and use that as the hash value.

crates/evm/src/rpc_helpers/filter.rs
```diff
@@ -555,6 +555,13 @@ impl<'de> Deserialize<'de> for Filter {
     }
 }
 
+impl std::hash::Hash for Filter {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
```
eyusufatik, line 559: I guess this is used for constructing a hashmap over Filter. are we sure we have to impl. this trait? how did reth handle log subscriptions for example?

crates/prover/src/runner.rs
yaziciahmet: IDK who is meant to run prover, but if it's meant to be run by potentially anyone, could there be a scenario where someone wants to run a prover on the chain, but also wants to use that as their RPC source so that they don't have to run another fullnode just for that?

crates/evm/src/rpc_helpers/filter.rs
yaziciahmet: Check the following code they are using:
```
/// Returns a stream that yields all logs that match the given filter.
    fn log_stream(&self, filter: FilteredParams) -> impl Stream<Item = Log> {
        BroadcastStream::new(self.chain_events.subscribe_to_canonical_state())
            .map(move |canon_state| {
                canon_state.expect("new block subscription never ends").block_receipts()
            })
            .flat_map(futures::stream::iter)
            .flat_map(move |(block_receipts, removed)| {
                let all_logs = logs_utils::matching_block_logs_with_tx_hashes(
                    &filter,
                    block_receipts.block,
                    block_receipts.tx_receipts.iter().map(|(tx, receipt)| (*tx, receipt)),
                    removed,
                );
                futures::stream::iter(all_logs)
            })
    }
```
As far as I understood, they have some sort of internal subscription mechanism to changes to the chain such as new block or reorgs. They first subscribe to this, and for each subscription, they are querying all receipts, and applying the filter on top of them. So the only difference (other than implementation details) is that we store subscriptions by filter, which allows us to group the same filters into the same operation. The reason I went with this is, most of the time subscriptions will be for many different types of filters. But high percentage of the subscriptions will be the same for some popular DEX like Uniswap. So I though we can get away with applying the same filter once.

Being said that, I also understand that when assumptions are forced on the behavior implemented, it doesn't yield the best result. Hence I am open to changing this to a simple vector of subscriptions and iterating over it.

yaziciahmet:
> are we sure if the logs subscription uses the same filter with `eth_getLogs`? I saw it has a field called `pub block_option: FilterBlockOption` does having this make sense on the subscription? should we handle it?

They also accept block number and filter by that. Even though it doesn't make much sense, I think it is the general behavior of eth nodes, including reth implementation.

crates/evm/src/rpc_helpers/filter.rs
yaziciahmet: One thing we can do is, once the soft confirmation signal arrives, SubscriptionManager can query all the logs in the block and send them all to each subscriber. And then subscribers can apply their filters onto these logs themselves. This way we don't have to have hash map, nor Hash trait on the filter.

yaziciahmet:
I have 3 questions left for this PR:
1. Should we spawn senders in a separate thread to not block receiving of blocks? Because currently if for some reason notifying subscribers on the SubscriptionManager takes some time, it prevents receiving incoming blocks.
2. Should we count debug_traceChain as subscription? It is not a realtime subscription, but it is handled through subscription. Should we also disable it through the enable_subscription flag?
3. Should we just disable subscriptions in prover? Check my comment above about this: https://github.com/chainwayxyz/citrea/pull/916#discussion_r1690203512
4. Do we implement config parameters disable_subscriptions and max_subscription_count in this PR?

yaziciahmet:
1. I don't think its necessary as it is doing very light work, if it takes longer than producing a block we have a completely irrelevant problem we have to attend to. And also these broadcast channels are buffered hence there is a lower chance these messages will be lost.
2. I think we should count debug_traceChain as subscription, and disable it if specified in config. Simply because it is still a websocket connection, and jsonrpsee does not provide a way to close the subscription from the server side (a bit weird imo). Hence, if client fails to handle the subscription closing, there could be empty websocket connections lying around. Even though websocket connections have a default timeout, imo best to consider this as subscription.
3. I think we should have the capability of subscriptions on prover, but if we or whoever is gonna run it wants to disable subscriptions, they can do it through rollup config with disable_subscriptions parameter.
4. I already did.

crates/ethereum-rpc/src/subscription.rs
```diff
@@ -0,0 +1,140 @@
...
+        tokio::spawn(async move {
+            let evm = Evm::<C>::default();
+            loop {
+                let Ok(height) = soft_confirmation_rx.recv().await else {
+                    return;
+                };
+
+                let mut working_set = None;
+
+                if new_heads_tx.receiver_count() != 0 {
```
eyusufatik, line 36: pr has been going on for too long so let's create a new issue for separate this senders for different things

eyusufatik:
btw keep the subscriptions on prover. we run the prover for the short-mid term future. we just won't enable subscriptions.

 doesn't matter that much.

crates/ethereum-rpc/src/subscription.rs
yaziciahmet: wdym? new_heads and logs subscriptions are already separated.

crates/ethereum-rpc/src/subscription.rs
```diff
@@ -0,0 +1,140 @@
...
+        // Spawn the task that will listen for new soft confirmation heights
+        // and send the corresponding ethereum block to subscribers
+        tokio::spawn(async move {
+            let evm = Evm::<C>::default();
+            loop {
+                let Ok(height) = soft_confirmation_rx.recv().await else {
+                    return;
+                };
+
+                let mut working_set = None;
```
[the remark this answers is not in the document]
yaziciahmet: Changed it as I learned that cloning storage is cheap operation.

crates/ethereum-rpc/src/subscription.rs
```diff
@@ -0,0 +1,140 @@
...
+            let Ok(logs) = rx.recv().await else {
+                // Connection closed
+                return;
+            };
+
+            for log in logs {
+                if log_matches_filter(
+                    &log.clone().try_into().unwrap(),
+                    &filter,
+                    &filter.topics,
```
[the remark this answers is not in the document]
yaziciahmet: It was how log_matches_filter was written before. Changed it to only pass filter.

eyusufatik (approved):

bin/citrea/src/rollup/mod.rs
```diff
@@ -43,9 +44,21 @@ pub trait CitreaRollupBlueprint: RollupBlueprint {
         let mut storage_manager = self.create_storage_manager(&rollup_config)?;
         let prover_storage = storage_manager.create_finalized_storage()?;
 
+        let (soft_confirmation_tx, soft_confirmation_rx) = broadcast::channel(10);
+        // If subscriptions disabled, pass None
+        let soft_confirmation_rx = if !rollup_config.rpc.disable_subscriptions {
```
rakanalh, line 49: I would use positive or optimistic language instead of the opposite. `enable_subscriptions` where the default is true. That way we don't have to read `if not disable` but rather, `if enable`.

rakanalh (approved):
Looks good with small nits

bin/citrea/src/rollup/mod.rs
yaziciahmet: The reason I went with disable is simply because I believe it is more common that the default value of a boolean configuration parameter is false, and it made more sense for subscriptions to be enabled by default. Hence I used `disable_subscriptions`. Wdyt @rakanalh  @eyusufatik ?

bin/citrea/tests/test_client/mod.rs
```diff
@@ -691,6 +693,50 @@ impl TestClient {
...
 
+    pub(crate) async fn subscribe_new_heads(&self) -> mpsc::Receiver<RichBlock> {
+        let (tx, rx) = mpsc::channel();
+        let mut subscription = self
+            .ws_client
+            .subscribe("eth_subscribe", rpc_params!["newHeads"], "eth_unsubscribe")
+            .await
+            .unwrap();
+
+        tokio::spawn(async move {
```
rakanalh, line 704: We have to find a way to decrease the number of `tokio::spawn` usages in our codebase and replace it somehow with workers we know are running. Point being, it's harder to debug randomly spawned tasks all over the codebase.

bin/citrea/src/rollup/mod.rs
yaziciahmet: done

PR #917, opened 2024-07-23
LedgerDB namespacing

### LINKED ISSUE #911: LedgerDb namespacing for sequencer, node, prover

jfldde:
# Description

- Split up LedgerDB operations by usage
- Introduces `SharedLedgerOps` for any method shared between Prover/Sequencer/Node.
- Adds `NodeLedgerOps`, `ProverLedgerOps` and `SequencerLedgerOps` to constraint the associated `CitreaFullNode`, `CitreaProver` and `CitreaSequencer`

## Linked Issues
- Fixes #911


rakanalh (commented):
I just gonna leave a comment here without "requesting changes" since i don't agree with the way things have been done here but i don't want to block the PR either.

The reason i have is:
We have Node,Prover,SequencerTrait... etc which are all implemented for `LedgerDB` which doesn't make that much of a difference from the previous version which `LedgerDB` had by containing all the methods. Limiting the access to methods using traits in prover for example by defining `ledger_db: DB` is okay.

The way i would have done it is as follows:
One trait, 3 structs
```
/// Common ledger ops, leaving out the word `common`/`shared` from the name.
pub trait Ledger {
  ... common stuff
}

pub struct ProverLedger {
}
impl ProverLedger {
}
impl LedgerOps for ProverLedger {
}
```
Same for sequencer and fullnode. I would also try to namespace tables
```
sequencer_xyz
prover_abc`
```
so that prover tables are only accessed within `ProverLedger`.

eyusufatik (approved):

PR #918, opened 2024-07-23
Use key hash for lookup of JmtValues

### LINKED ISSUE #712: Research if sov state db key to key-hash mapping is performant
There is a key to key hash mapping in the statedb, research if it actually makes sense and remove if we are actually better without it

rakanalh:
# Description

## Linked Issues
- Fixes #712


rakanalh:
Closing

Issue #235, opened 2024-03-12
Create rollup wide hardfork spec


eyusufatik:
If anything goes wrong in production, we'd like to have a system that will allow us to fix our mistakes.

Ideally this fork system has access to state and accessory state directly (both state db and ledger db) and underlying modules can query this system like 

```rust
hard_fork_manager.is_fork1_enabled()
```

See revm and reth's hardfork managers.

More details from @eyusufatik (I guess someone added these lines besides me xd):
Create something that is accessible from many different parts of our software. All of these will probably need to know if a hardfork is activated / when it’s gonna be activated:
1. citrea evm
2. soft confirmation rule enforcer
3. STF blueprint
4. any RPC
5. anything related to storage
6. ledger db tables

reasoning:
1 and 2 are simple, change behavior of something when we need to. Set an activation block height in the future and everybody is gonna be upgraded to the version by then.
3. We might want to change the behavior of something in the sdk level.
4. obvious after all these points
5-6. when we want to create new tables etc. we’ll need to do changes without breaking anything. but these can be handled outside the “hardfork system” tho. these can be database migrations i guess.

TODOs:
- [ ] Vector of hardfork activations that map to onetime running functions, if software is run with a new hardfork immideately calls that function.
- [ ] Systemwide access to this vector is important.

rakanalh:
# Regarding RPC:

I suggest we use versioning to make sure that our RPC calls are compatible with specific parts of the codebase.

Example:
`/v1/ledger_getHeadSoftBatch` can be different than `/v2/ledger_getHeadSoftBatch` is they use different tables to fetch data from.

Both can be maintained at the same time to support old & new clients.

That should be pretty easy to implement.

rakanalh:
# Fork Manager

A fork manager is a component which decides when to activate a specific fork by calling a set of one-off functions that would set specific states in order for the new functionality to start running.

Assuming we define forks similar to how revem implements it:
```rust
pub enum SpecId {
   Spec1,
   Spec2,
   Spec3
}
```

At genesis, the components we have start their state at the first spec. 

But in order for us to activate a fork (spec), we need to manage how each component would migrate it's internals to work with that new speci.

```rust
/// Defines the interface of a migrations manager
pub trait ForkMigrationManager {
  /// Invoked before a spec is activated.
  fn pre_spec_activation(&self, spec_id: SpecId) -> anyhow::Result<()>;
  /// Invoked after a spec is activated.
[...]

PR #922, opened 2024-07-25
Fork Manager

rakanalh:
# Description

## Linked Issues
- Related to #235


crates/primitives/src/error.rs
```diff
@@ -1,8 +1,6 @@
-use sov_db::schema::types::BatchNumber;
-
 #[derive(Debug)]
 pub enum SyncError {
-    MissingL2(&'static str, BatchNumber, BatchNumber),
+    MissingL2(&'static str, u64, u64),
```
kpp, line 3: Can we have named fields?

crates/primitives/src/fork/migration.rs
```diff
@@ -0,0 +1,23 @@
...
+impl ForkMigration for NoOpMigration {
+    fn pre_spec_activation(&self, _spec_id: SpecId) -> anyhow::Result<()> {
+        // Do nothing
+        Ok(())
+    }
+
+    fn post_spec_activation(&self, _spec_id: SpecId) -> anyhow::Result<()> {
+        // Do nothing
+        Ok(())
+    }
```
eyusufatik, line 22: i gave this some thought, what about only having a single on activation fork?

when spec activate height is reached, function is called before everything else.

eyusufatik (changes requested):

eyusufatik (approved):

Issue #743, opened 2024-06-13
Bitcoin DA: Transaction setup to support chunks


rakanalh:
TBD

PR #925, opened 2024-07-25
Chunked DA txs

kpp:
# Description

- [x] Fix https://github.com/chainwayxyz/citrea/issues/871
- [x] Fix https://github.com/chainwayxyz/citrea/issues/743
- [x] Fix https://github.com/chainwayxyz/citrea/issues/972
- [x] Fix https://github.com/chainwayxyz/citrea/issues/971
- [x] Fix https://github.com/chainwayxyz/citrea/issues/993

This PR implements a new DA tx format. Here is the format: https://github.com/chainwayxyz/citrea/issues/871#issuecomment-2233488255

Also instead of txids we use wtxids for prefixes. So we don't try to find nonces for txid, we find nonces so that wtxid matches a specific prefix. It was done for https://github.com/chainwayxyz/citrea/pull/991 better inclusion proof (O(log N) instead of O(N)).

In order to speed up proofs when a DA block has both zkproof and seq commitment I implemented a trick to parse only DA tx that are related to our proofs. Before we had 1 prefix, but now there are two of them (https://github.com/chainwayxyz/citrea/issues/972). Before the prefix was 00, now they are: 
- light_client: 22
- batch_proof: 11

Where light_client - zkproofs (diffs), batch_proof - currently only a tx with sequencer commitment, but will be extended with forced transactions sometime in the future.


rakanalh (commented):
Genuine question: Is there an E2E test that uses bitcoin-da?

kpp:
> Genuine question: Is there an E2E test that uses bitcoin-da?

Not really. Esad dumped a bitcoin block with txs (see mock_txs.txt) and all tests passed (except for CI but I will fix it). Tests are: parse seq commitment, verify inclusion etc.

eyusufatik:
> Genuine question: Is there an E2E test that uses bitcoin-da?

waitin on @jfldde for e2e bitcoin framework 🙄

crates/bitcoin-da/src/helpers/builders.rs
```diff
@@ -457,38 +957,235 @@ pub fn create_inscription_transactions(
...
+                    tx: reveal_tx,
+                },
+            });
+        }
+
+        nonce += 1;
+    }
+}
 
+// TODO: parametrize hardness
```
eyusufatik, line 1019: todo is irrelevant now

crates/bitcoin-da/src/helpers/builders.rs
```diff
@@ -457,38 +957,235 @@ pub fn create_inscription_transactions(
...
+    commit_fee_rate: f64,
+    reveal_fee_rate: f64,
+    network: Network,
+    reveal_tx_prefix: &[u8],
+) -> Result<BatchProvingTxs, anyhow::Error> {
+    debug_assert!(
+        body.len() < 520,
+        "The body of a serialized sequencer commitment exceeds 520 bytes"
+    );
+    // Create commit key
```
eyusufatik, line 1041: i believe this creates the internal private key to use in the reveal tx

crates/bitcoin-da/src/helpers/builders.rs
```diff
@@ -457,38 +957,235 @@ pub fn create_inscription_transactions(
...
+            if nonce > 65536 {
+                warn!("Too many iterations finding nonce");
+            }
+        }
+        let utxos = utxos.clone();
+        let recipient = recipient.clone();
+        // ownerships are moved to the loop
+        let mut reveal_script_builder = reveal_script_builder.clone();
+
+        // push first random number and body tag
```
eyusufatik, line 1084: fix comment -- no body tag now

crates/bitcoin-da/src/helpers/builders.rs
```diff
@@ -457,38 +957,235 @@ pub fn create_inscription_transactions(
...
+        reveal_script_builder = reveal_script_builder
+            .push_slice(nonce.to_le_bytes())
+            .push_opcode(OP_DROP);
```
eyusufatik, line 1087: just now realized that we can actually have the nonce inside if / endif for sequencer commitments since we know the body is a single push so we don't need to do OP_DROP.

if this is correct please open an issue we can merge like this np

crates/bitcoin-da/src/helpers/builders.rs
```diff
@@ -457,38 +957,235 @@ pub fn create_inscription_transactions(
...
+        // build commit tx
+        // we don't need leftover_utxos because they will be requested from bitcoind next call
+        let (unsigned_commit_tx, _leftover_utxos) = build_commit_transaction(
+            prev_utxo.clone(),
+            utxos,
+            commit_tx_address.clone(),
+            recipient.clone(),
+            commit_value,
+            commit_fee_rate,
+        )?;
```
eyusufatik, line 1142: since we decided to go with wtxid for prefixes, now we have to choose utxos every time we try a nonce.

this can be improved actually. 

pls open an issue for this

crates/bitcoin-da/src/helpers/builders.rs
```diff
@@ -87,7 +94,7 @@ fn choose_utxos(
         sum += req_amount;
     }
     if sum >= amount {
-        return Ok((chosen_utxos, sum));
+        return Ok((chosen_utxos, sum, utxos.to_vec()));
```
eyusufatik, line 97: there is a good chance the `required_utxo` is in utxos. pls remove it from utxos.

crates/bitcoin-da/src/helpers/builders.rs
```diff
@@ -124,20 +129,26 @@ fn choose_utxos(
         if sum < amount {
             return Err(anyhow!("not enough UTXOs"));
         }
-
-        Ok((chosen_utxos, sum))
     }
+
+    let input_set: HashSet<_> = utxos.iter().collect();
+    let chosed_set: HashSet<_> = chosen_utxos.iter().collect();
```
eyusufatik, line 135: chosen_set*

crates/bitcoin-da/src/verifier.rs
```diff
@@ -79,7 +80,8 @@ impl DaVerifier for BitcoinVerifier {
     fn new(params: <Self::Spec as DaSpec>::ChainParams) -> Self {
         Self {
             rollup_name: params.rollup_name,
-            reveal_tx_id_prefix: params.reveal_tx_id_prefix,
+            // TODO
```
kpp, line 83: Remove TODO because we don't verify light client txs

crates/bitcoin-da/src/helpers/builders.rs
```diff
@@ -378,17 +518,377 @@ pub fn create_inscription_transactions(
...
+
+        // build commit tx
+        let (unsigned_commit_tx, leftover_utxos) = build_commit_transaction(
+            prev_utxo.clone(),
+            utxos,
+            commit_tx_address.clone(),
+            recipient.clone(),
+            commit_value,
+            commit_fee_rate,
+        )?;
```
eyusufatik, line 746: for commit of chunks we don't really need to chain utxos.

it'd be a good idea to use different utxos as inputs for chunks

and THEN have a chunk aggregator that uses the prev_utxo for it's input.

eyusufatik: see the image I'll send you on slack

crates/bitcoin-da/src/verifier.rs
```diff
@@ -94,68 +96,63 @@ impl DaVerifier for BitcoinVerifier {
...
+                            let blob_content = blob_content.accumulator();
+
+                            // assert tx content is not modified
+                            if blob_content != seq_comm.body {
+                                return Err(ValidationError::BlobContentWasModified);
+                            }
+                        }
                     }
                 }
             }
```
kpp, line 153: TODO what if a tx wasn't parsed?

crates/bitcoin-da/src/verifier.rs
```diff
@@ -181,76 +178,59 @@ impl DaVerifier for BitcoinVerifier {
...
+        // the one with highest output index is assumed to be the commitment.
+        // That  is why the iterator is reversed.
+        let commitment_idx = coinbase_tx.output.iter().rev().position(|output| {
+            output
+                .script_pubkey
+                .as_bytes()
+                .starts_with(WITNESS_COMMITMENT_PREFIX)
+        });
+        match commitment_idx {
+            // If commitmet does not exist
```
kpp, line 192:
```suggestion
            // If commitment does not exist
```

crates/bitcoin-da/src/verifier.rs
```diff
@@ -181,76 +178,59 @@ impl DaVerifier for BitcoinVerifier {
...
+            None => {
+                // Relevant txs should be empty if there is no wtiness data because data is inscribed in the witness
```
kpp, line 194:
```suggestion
                // Relevant txs should be empty if there is no witness data because data is inscribed in the witness
```

crates/bitcoin-da/src/helpers/builders.rs
eyusufatik: we can open an issue and do it later though :)

crates/bitcoin-da/src/service.rs
```diff
@@ -312,52 +354,135 @@ impl BitcoinService {
...
+                        reveal_chunks,
+                        commit,
+                        reveal,
+                    } => {
+                        for (commit, reveal) in commit_chunks.into_iter().zip(reveal_chunks) {
+                            // sign inscribe transactions
+                            let signed_raw_commit_tx = client
+                                .sign_raw_transaction_with_wallet(&commit, None, None)
+                                .await?;
+
```
eyusufatik, line 410: can you also write the txs in a file just like we do on seq. comm.

crates/bitcoin-da/src/verifier.rs
```diff
@@ -181,76 +178,59 @@ impl DaVerifier for BitcoinVerifier {
...
+                .as_raw_hash()
+                .to_byte_array(),
+            0,
+            inclusion_proof.coinbase_merkle_proof,
+        );
 
-        let root_from_inclusion = root_from_inclusion.to_raw_hash().to_byte_array();
-        // Check that the tx root in the block header matches the tx root in the inclusion proof.
-        if root_from_inclusion != tx_root {
+        if block_header.merkle_root() != claimed_root {
```
kpp, line 233:
```suggestion
        // Check that the tx root in the block header matches the tx root in the inclusion proof.
        if block_header.merkle_root() != claimed_root {
```

crates/bitcoin-da/src/service.rs
```diff
@@ -789,14 +1132,18 @@ mod tests {
 
         let txs = da_service.extract_relevant_blobs(&block);
 
-        assert_eq!(txs, relevant_txs)
+        assert_eq!(txs, relevant_txs);
+
+        // kill tokio runtime
+        drop(da_service);
```
eyusufatik, line 1138: remove this

crates/bitcoin-da/src/service.rs
```diff
@@ -146,13 +145,27 @@ impl BitcoinService {
...
                         {
                             Ok(tx) => {
                                 let tx_id = TxidWrapper(tx.id);
                                 info!(%tx.id, "Sent tx to BitcoinDA");
-                                prev_tx = Some(tx);
+                                prev_utxo = Some(UTXO {
+                                    tx_id: tx.id,
+                                    vout: 0,
+                                    script_pubkey: tx.tx.output[0].script_pubkey.to_hex_string(),
+                                    address: None,
```
kpp, line 162: @ercecan you set address to Some("ANY"). Can I use None here as well?

crates/bitcoin-da/src/service.rs
```diff
@@ -299,7 +333,15 @@ impl BitcoinService {
         let da_private_key = self.da_private_key.expect("No private key set");
 
         // Compress the blob
```
kpp, line 335:
```suggestion
```

crates/bitcoin-da/src/helpers/parsers.rs
```diff
@@ -45,22 +86,94 @@ impl ParsedInscription {
...
 
 pub fn parse_transaction(
     tx: &Transaction,
     rollup_name: &str,
-) -> Result<ParsedInscription, ParserError> {
+) -> Result<ParsedLightClientTransaction, ParserError> {
+    let script = get_script(tx)?;
+    let instructions = script.instructions().peekable();
+    // Map all Instructions errors into ParserError::ScriptError
+    let mut instructions = instructions.map(|r| r.map_err(ParserError::from));
```
eyusufatik, line 162: name it parse_light_client_transaction -- or something else if we come up with another naming (see other comments)

crates/bitcoin-da/src/helpers/parsers.rs
```diff
@@ -71,119 +184,315 @@ fn get_script(tx: &Transaction) -> Result<&Script, ParserError> {
...
+    };
+
+    // Parse transaction body of Type0
+    pub(super) fn parse_type_0_body(
+        instructions: &mut dyn Iterator<Item = Result<Instruction<'_>, ParserError>>,
+    ) -> Result<ParsedComplete, ParserError> {
+        let op_false = read_push_bytes(instructions)?;
+        if !op_false.is_empty() {
+            return Err(ParserError::UnexpectedOpcode);
+        }
```
eyusufatik, line 298: you don't verify if it's really OP_FALSE

crates/bitcoin-da/src/helpers/parsers.rs
```diff
@@ -200,338 +509,219 @@ pub fn parse_hex_transaction(
...
 
-        let result =
-            parse_relevant_inscriptions(&mut reveal_script.instructions().peekable(), "sov-btc");
-
-        assert!(result.is_err(), "Failed to error on no random tag.");
-        assert_eq!(result.unwrap_err(), ParserError::EnvelopeHasIncorrectFormat);
-    }
-
-    #[test]
-    fn non_parseable_tx() {
```
eyusufatik, line 398: this one too

crates/bitcoin-da/src/service.rs
kpp, line 1138: Does it work?

eyusufatik (changes requested):

crates/bitcoin-da/src/helpers/builders.rs
```diff
@@ -124,20 +129,26 @@ fn choose_utxos(
...
             return Err(anyhow!("not enough UTXOs"));
         }
-
-        Ok((chosen_utxos, sum))
     }
+
+    let input_set: HashSet<_> = utxos.iter().collect();
+    let chosed_set: HashSet<_> = chosen_utxos.iter().collect();
+    let leftovers_set = input_set.difference(&chosed_set);
+    let leftovers: Vec<_> = leftovers_set.copied().cloned().collect();
```
kpp, line 137:
```suggestion
    let chosen_set: HashSet<_> = chosen_utxos.iter().collect();
    let leftovers_set = input_set.difference(&chosen_set);
    let leftovers: Vec<_> = leftovers_set.copied().cloned().collect();
```

crates/bitcoin-da/src/service.rs
eyusufatik: does nothing, just delete

crates/bitcoin-da/src/helpers/builders.rs
kpp, line 1084:
```suggestion
        // push nonce
```

crates/bitcoin-da/src/helpers/builders.rs
kpp: It is true but parsing in my case looks nice and neat. In case of nonce in the body we need to take pushbuf except for the last one and the last one will be nonce

crates/bitcoin-da/src/helpers/parsers.rs
kpp: OP_FALSE is pushbytes 0 actually. I will add comments to exclude confusion

crates/bitcoin-da/src/helpers/parsers.rs
kpp: What's with this one?

crates/bitcoin-da/src/service.rs
kpp: I do with write_inscription_txs above

crates/bitcoin-da/src/helpers/builders.rs
eyusufatik: I mean just update the comment

crates/bitcoin-da/src/helpers/builders.rs
eyusufatik: hmm, resolve this conversation but pls first open issue with low priority, we probably won't implement it right away. but maybe one day.

crates/bitcoin-da/src/helpers/builders.rs
kpp: We already do that in:

```
    if let Some(req_utxo) = &prev_utxo {
        // if we don't do this, then we might end up using the required utxo twice
        // which would yield an invalid transaction
        // however using a different txo from the same tx is fine.
        utxos.retain(|utxo| !(utxo.vout == req_utxo.vout && utxo.tx_id == req_utxo.tx_id));
    }
```

crates/bitcoin-da/src/helpers/parsers.rs
eyusufatik: i guess some of my comments are missing...

non_parseable_tx + leave_out_tags

we still need these tests.

crates/bitcoin-da/src/helpers/builders.rs
kpp: https://github.com/chainwayxyz/citrea/issues/1022

crates/bitcoin-da/src/helpers/builders.rs
kpp: https://github.com/chainwayxyz/citrea/issues/1023

crates/bitcoin-da/src/helpers/builders.rs
kpp: https://github.com/chainwayxyz/citrea/issues/1024

crates/bitcoin-da/src/helpers/parsers.rs
kpp: non_parseable_tx - for this one we need to create a tx with a correct kind/rollup name but with a mess inside the script. is it useful?

leave_out_tags - since our tx format changed for a better parsing there is no need for a state machine to match tags as before. Signature, pub key, nonce - everything has its own place in the script instead of [tag] -> [data].

crates/bitcoin-da/src/service.rs
ercecan: I believe it is okay yes

eyusufatik (approved):

PR #926, opened 2024-07-26
WitnessUtils library improvement

### LINKED ISSUE #384: Tests for WitnessUtils.lib

erdkocak:
# Description
- Fixed a bug in `determineWitnessLengthAt` function.
- Added tests for `WitnessUtils` library.
- Created helper random generating Python scripts for random witness generation.

## Linked Issues
- Fixes #384

## Testing
- Added unit and fuzz tests.


crates/evm/src/evm/system_contracts/test/WitnessUtils.t.sol
```diff
@@ -0,0 +1,162 @@
...
+        inputs[1] = "test/randomWitnessGenerator.py";
+        
+        bytes memory result = vm.ffi(inputs);
+        return result;
+    }
+
+    function getRandomStackItem() public returns (bytes memory) {
+        string[] memory inputs = new string[](2);
+        inputs[0] = "python3";
+        inputs[1] = "test/randomStackItemGenerator.py";
```
kpp, line 36: Is it really a proper way of gen random hex strings?

erdkocak: I could not find a native random library in foundry to use for complex statements like this (fuzzing was not enough, it is generally deterministic and gives the same randomness in each iteration). This approach was also used by a lot of people, that's why I choose to do like this.

okkothejawa (approved):

eyusufatik (approved):

PR #927, opened 2024-07-26
update deny.toml allowed licences

eyusufatik:
# Description
comments out unused allow fields in the deny.toml

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -3,22 +3,22 @@
 
 allow = [
   "Apache-2.0",
-  "Apache-2.0 WITH LLVM-exception",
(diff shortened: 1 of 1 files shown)


kpp (approved):
Is it that much required? I would imaging our ci breaks in the future bc new deps would require new licenses. LGTM

eyusufatik:
> Is it that much required? I would imaging our ci breaks in the future bc new deps would require new licenses. LGTM

we can add those license when it's needed. 

I needed to see which licenses we currently used, and for this reason in the future, it's a good idea to have the minimum viable set.

PR #928, opened 2024-07-26
Sequencer scan L1 and resubmit mined commitments

### LINKED ISSUE #897: Sequencer scan L1 on reboot
in #889 I mentioned a possible scenario, 

1. sequencer sends commitment.
2. sequencer shuts down and reboots
3. right before it does the reboot sequence, one of the pending commitments gets mined.
4. it resends the same "pending commitment" because it was in his table but not in its bitcoin pending txs list.

To circumvent this issue, the sequencer should scan the L1 blocks for sequencer commitments. (we can discuss if this should be a reboot only scanning or the sequencer should always do it and record data etc.)

Ideating on the reboot only scanning:

- Sequencer shutsdown
- Sequencer restarts
- Get pending da transactions (we do this before scanning L1 so that if another block gets mined while we are doing the L1 scan we don't have the same problem we are resolving)
- Checks the height of last finalized sequencer commitment (this requires scanning all the time I guess ;) )
- from this height to the bitcoin tip -> search for sequencer commitments. remove from pending commitments table & pending da txs variable.
- then do the "resubmits"

yaziciahmet:
# Description

We already had resubmission logic for commitments in local database. We also cross-checked whether if any pending marked commitments actually made it to mempool without sequencer being able to update its local database.

This PR adds also checks the potentially mined commitments. If a sequencer:
1. submitted a commitment
2. valid tx nonce is found and commitment submitted to DA
3. restarted right before it was able to update its local pending commitments database
4. during the off time of the restart, the commitment is mined and included in a block
this causes sequencer to double commit the same l2 heights, which is a gas expensive operation.

Now we check the last committed L1 height to the head L1 block to see if there was any commitments included and update their pending status if so. 

## Linked Issues
- Fixes #897

### DIFF
--- a/crates/sequencer/src/sequencer.rs
+++ b/crates/sequencer/src/sequencer.rs
@@ -634,7 +634,7 @@ where
         let last_commitment_l1_height = self
             .ledger_db
             .get_last_sequencer_commitment_l1_height()?
-            .unwrap_or(SlotNumber(0));
+            .unwrap_or(SlotNumber(1));
         let mined_commitments = self
             .get_mined_commitments_from(last_commitment_l1_height)
             .await?;
--- a/crates/sequencer/src/sequencer.rs
+++ b/crates/sequencer/src/sequencer.rs
@@ -622,12 +622,14 @@ where
 
     #[instrument(level = "trace", skip(self), err, ret)]
     pub async fn resubmit_pending_commitments(&mut self) -> anyhow::Result<()> {
+        info!("Resubmitting pending commitments");
+
         let pending_db_commitments = self.ledger_db.get_pending_commitments_l2_range()?;
         debug!("Pending db commitments: {:?}", pending_db_commitments);
 
         let pending_mempool_commitments = self.get_pending_mempool_commitments().await;
         debug!(
-            "Pending mempool commitments: {:?}",
+            "Commitments that are already in DA mempool: {:?}",
             pending_mempool_commitments
         );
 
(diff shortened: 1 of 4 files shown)


crates/sequencer/src/sequencer.rs
```diff
@@ -622,11 +622,29 @@ where
...
+            .get_last_sequencer_commitment_l1_height()?
+            .unwrap_or(SlotNumber(1));
+        let mined_commitments = self
+            .get_mined_commitments_from(last_commitment_l1_height)
+            .await?;
+        debug!("Unsaved mined commitments: {:?}", mined_commitments);
+
+        let mut pending_commitments_to_remove = vec![];
+        pending_commitments_to_remove.extend(pending_mempool_commitments);
+        pending_commitments_to_remove.extend(mined_commitments);
```
eyusufatik, line 643: why are you adding mined_commitments  here? shouldn't they be handled seperately?

yaziciahmet: what separate functionality is needed for mined commitments? we normally save them to pg but we decided to not deal with that since we plan to remove it anyways.

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
```diff
@@ -578,6 +578,22 @@ impl LedgerDB {
...
+            Some(l2_height) => {
+                let soft_confirmation = self
+                    .get_soft_batch_by_number(&l2_height)?
+                    .expect("Expected soft confirmation to exist");
+                Ok(Some(SlotNumber(soft_confirmation.da_slot_height)))
+            }
+            None => Ok(None),
+        }
+    }
+
```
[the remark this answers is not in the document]
yaziciahmet: renamed to `get_l1_height_of_last_commitment`

crates/sequencer/src/sequencer.rs
eyusufatik: misunderstood

eyusufatik (approved):

PR #930, opened 2024-07-29
Added optimized byte comparison function to Bridge contract

erdkocak:
# Optimized the byte comparison function in the bridge contract

## Testing
-Changed some of the old tests that expect revert in the case of a length mismatch between the inputs to expecting function to return false. Also added a test for the edge cases.


crates/evm/src/evm/system_contracts/src/Bridge.sol
```diff
@@ -166,22 +166,38 @@ contract Bridge is Ownable2StepUpgradeable {
...
         require(a.length == b.length, "Lengths do not match");
 
-        // Cannot use keccak as its costly in ZK environment
-        uint length = a.length;
-        for (uint i = 0; i < length; i++) {
-            if (a[i] != b[i]) {
-                result = false;
-                return result;
+        uint256 fullChunks = a.length / 32;
+        for (uint256 i = 0; i < fullChunks; i++) {
```
yaziciahmet, line 177: no need for `i = 0` as that is already the default value, and assigning costs extra.

crates/evm/src/evm/system_contracts/src/Bridge.sol
```diff
@@ -166,22 +166,38 @@ contract Bridge is Ownable2StepUpgradeable {
...
+                return false;
             }
         }
-        result = true;
+
+        // Check remaining bytes (if any)
+        uint remainingBytes = a.length % 32;
+        if (remainingBytes > 0) {
+            uint start = fullChunks * 32;
+            for (uint i = 0; i < remainingBytes; i++) {
```
yaziciahmet, line 193: same i = 0 as above

yaziciahmet (commented):
I don't know what is our motto when developing contracts, but if we want extreme optimizations, following could be of use

crates/evm/src/evm/system_contracts/src/Bridge.sol
```diff
@@ -166,22 +166,38 @@ contract Bridge is Ownable2StepUpgradeable {
...
-        for (uint i = 0; i < length; i++) {
-            if (a[i] != b[i]) {
-                result = false;
-                return result;
+        uint256 fullChunks = a.length / 32;
+        for (uint256 i = 0; i < fullChunks; i++) {
+            bytes32 chunkA;
+            bytes32 chunkB;
+            assembly {
+                chunkA := mload(add(a, add(32, mul(i, 32))))
```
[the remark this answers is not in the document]
yaziciahmet: and to even further optimize, offset can be calculated outside of for loop once, and 32 could be added on every loop instead of calculating the multiplication every time. basically instead of having i variable in the loop, you can start the offset from 32, and increment 32 in the for loop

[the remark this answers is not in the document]
erdkocak: As this requires an additional memory load and memory store, its actually more expensive when comapred to one add and one mul. We just implemented both to test practically and it seems that this version is just a little more expensive.

crates/evm/src/evm/system_contracts/src/Bridge.sol
erdkocak: just also tried this, did not change the cost at all. I guess the new solidity compiler automatically optimizes this.

erdkocak:
> I don't know what is our motto when developing contracts, but if we want extreme optimizations, following could be of use

For this function, we were using a simple check for each byte before. This optimization reduces the gas cost almost %50 in this function for the inputs that we generally use in our smart contracts. That's why we worked on this one.

crates/evm/src/evm/system_contracts/src/Bridge.sol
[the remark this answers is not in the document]
erdkocak: That was my guess for the reason why it was more costly. I just checked one more and even if the gas costs are super close this one (using offset) is slightly more expensive. I can share my testing code with you if you want and maybe you can try it out.

kpp (approved):

kpp:
This PR can be linked to https://github.com/chainwayxyz/citrea/pull/930#pullrequestreview-2204910834 if we want further opts.

yaziciahmet (approved):

crates/evm/src/evm/system_contracts/src/Bridge.sol
```diff
@@ -171,29 +171,29 @@ contract Bridge is Ownable2StepUpgradeable {
...
+        uint256 len = a.length;
+        if (len != b.length) {
+            return false;
+        }
 
-        uint256 fullChunks = a.length / 32;
-        for (uint256 i = 0; i < fullChunks; i++) {
-            bytes32 chunkA;
-            bytes32 chunkB;
+        uint256 offset = 32;
```
kpp, line 179: In case a,b len = 32, cmp will be 1 per byte?

erdkocak: just made offset < len to offset <= len. I think that solves this

kpp (changes requested):
While in general this looks good, can you add 3 corner cases to cmp str len 31, 32 and 33?

erdkocak:
> While in general this looks good, can you add 3 corner cases to cmp str len 31, 32 and 33?

just added them, let me know if you think this is not enough

kpp (approved):

eyusufatik (approved):

PR #931, opened 2024-07-29
chore: Bitcoin RPC cleanup

jfldde:
# Description

- Consistent use of `anyhow::Result`
- Use type inference 
- Stricter type usage

# Notes:

Have we considered using a library such as like https://github.com/rust-bitcoin/rust-bitcoincore-rpc/ instead of re-implementing it ourselves ? It comes with a few footguns we will need to be mindful of

### DIFF
--- a/crates/bitcoin-da/src/rpc.rs
+++ b/crates/bitcoin-da/src/rpc.rs
@@ -1,6 +1,7 @@
 use core::fmt::{Debug, Display};
 use core::str::FromStr;
 
+use anyhow::Result;
 use bitcoin::block::{Header, Version};
 use bitcoin::hash_types::{TxMerkleNode, WitnessMerkleNode};
 use bitcoin::hashes::Hash;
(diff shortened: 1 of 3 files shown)


kpp (changes requested):
This makes little sense if we may use https://github.com/Sovereign-Labs/sov-bitcoin-adapter/blob/nightly/src/service/rpc/mod.rs .

I am going to replace anyhow to a proper RpcError, because we need to se-send tx (reveal) in case the previous (commit) hasn't been processed yet.

eyusufatik:
> This makes little sense if we may use https://github.com/Sovereign-Labs/sov-bitcoin-adapter/blob/nightly/src/service/rpc/mod.rs . I am going to replace anyhow to a proper RpcError, because we need to se-send tx (reveal) in case the previous (commit) hasn't been processed yet.

i also didn't understand. @kpp can you please elobrate?

eyusufatik (approved):

eyusufatik:
@kpp explained to me on slack what he meant. we'll do that on another pr

kpp (approved):

PR #933, opened 2024-07-31
Use Borsh ser/deser in TransactionWrapper and BitcoinHeaderWrapper

jfldde:
# Description

- Implement `BorshSerialize` and `BorshDeserializaiton` on `TransactionWrapper` and `BitcoinHeaderWrapper` by manually ser/deser every inner fields
- Remove usage of bitcoin consensus encoding


eyusufatik (commented):
what gains do we get from not using consensus encode? which is very similar to borsh encode when you think about it btw.

crates/bitcoin-da/src/spec/header.rs
```diff
@@ -71,7 +70,7 @@ impl HeaderWrapper {
 }
 
 /// BitcoinHeaderWrapper is a wrapper around BitcoinHeaderWrapper to implement borsh serde
-#[derive(Clone, PartialEq, Eq, Debug, Hash, Deserialize, Serialize)]
+#[derive(Clone, PartialEq, Eq, Debug, Hash, Serialize, Deserialize)]
 #[repr(transparent)]
 #[serde(transparent)]
 pub struct BitcoinHeaderWrapper {
```
yaziciahmet, line 76: If it's just a wrapper, imo it's better to have it as `pub struct BitcoinHeaderWrapper(BitcoinHeader)`

crates/bitcoin-da/src/spec/header.rs
```diff
@@ -71,7 +70,7 @@ impl HeaderWrapper {
 }
 
 /// BitcoinHeaderWrapper is a wrapper around BitcoinHeaderWrapper to implement borsh serde
-#[derive(Clone, PartialEq, Eq, Debug, Hash, Deserialize, Serialize)]
+#[derive(Clone, PartialEq, Eq, Debug, Hash, Serialize, Deserialize)]
```
yaziciahmet, line 73: Do we still need serde derives?

crates/bitcoin-da/src/spec/transaction.rs
```diff
@@ -11,12 +12,12 @@ use serde::{Deserialize, Serialize};
 #[repr(transparent)]
 #[serde(transparent)]
 pub struct TransactionWrapper {
```
yaziciahmet, line 14: Same wrapper suggestion as above

yaziciahmet (changes requested):

eyusufatik:
can you do benchmarks on test_all_flow? if you don't have the right commands let me know over slack. also @yaziciahmet recently asked me those commands as well.

yaziciahmet (approved):

kpp:
Actually it would be better to test borsh in isolation - only to de-serialize inputs and exit, because now it takes only a small portion of our execution (less then 1% of total cycles). @yaziciahmet did it in https://github.com/chainwayxyz/citrea/pull/943 for comparison.

eyusufatik:
> Actually it would be better to test borsh in isolation - only to de-serialize inputs and exit, because now it takes only a small portion of our execution (less then 1% of total cycles). @yaziciahmet did it in https://github.com/chainwayxyz/citrea/pull/943 for comparison.

yes let's do it

eyusufatik:
> can you do benchmarks on test_all_flow? if you don't have the right commands let me know over slack. also @yaziciahmet recently asked me those commands as well.

forgot these were tests on MockDA. we should wait for merge until @jfldde 's bitcoin e2e testing stuff is ready and then benchmark.

yaziciahmet:
With the merge of #973 , new bitcoin 0.32.2 version has changed the consensus_encode mechanism, and it is no longer usable with the Borsh ser/de. We tried couple of different methods to read/write from and to Borsh to consensus encodable, but it failed. I tested this PR and it seems to solve this issue. It might be required now that we merge this PR in @eyusufatik @kpp

PR #941, opened 2024-08-01
[WIP] risc0 read optimization

yaziciahmet:
# Description

Run the test with:
```
cargo test --package citrea --test all_tests -- e2e::proving::full_node_verify_proof_and_store --exact --show-output
```

Play around with `min_soft_confirmation_per_commitment` value in the test to test with larger proof sizes.

## Linked Issues
- Fixes # (issue, if applicable)
- Related to # (issue) 

## Testing
Describe how these changes were tested. If you've added new features, have you added unit tests?

## Docs
Describe where this code is documented. If it changes a documented interface, have the docs been updated?

PR #943, opened 2024-08-02
RISC0 guest read optimization

### LINKED ISSUE #876: Research: Guest code borsh - avoid preallocations if possible
Follow up of https://github.com/chainwayxyz/citrea/pull/851. What we do in guest code:

1. read len(u64) in LE
2. alloc `Vec<u32>` with the given len
3. read len u32 into that vec
4. cast `Vec<u32>` to &[u8]
5. deserialize from slice

This can be sub-optimal because we alloc a vec in step 2. Borsh is capable of de-serializing from a Reader, avoiding preallocs if possible. Instead a tricky `Read` can be implemented that reads only aligned to 4 data (that's the most optimal way for risc0 - https://dev.risczero.com/api/zkvm/optimization#unaligned-data-access-is-significantly-more-expensive) to bypass the vec. 

Further reading: https://github.com/risc0/risc0/pull/1303

yaziciahmet:
# Description

When sending data from host to guest, we are length prefixing the data at the host, and allocating the whole data at once when reading back at the guest.

Instead of this, implementing a reader proved to be a more efficient solution in terms of cycles. You can see the #876 about the details of the cycle improvement.

To see the code to test the cycles in isolation (where the only operation is ser-de), see #941 

## Linked Issues
- Fixes #876

### DIFF
--- a/crates/risc0-bonsai/src/host.rs
+++ b/crates/risc0-bonsai/src/host.rs
@@ -331,22 +331,15 @@ impl<'a> Risc0BonsaiHost<'a> {
         // Prepare input data and upload it.
         let client = self.client.as_ref().unwrap();
 
-        let mut input_data = vec![];
-        let mut buf = borsh::to_vec(&item).unwrap();
-        // append [0..] alignment
+        let mut buf = borsh::to_vec(&item).expect("Risc0 hint serialization is infallible");
+        // append [0..] alignment to cast &[u8] to &[u32]
         let rem = buf.len() % 4;
         if rem > 0 {
             buf.extend(vec![0; 4 - rem]);
         }
-        let buf_u32: &[u32] = bytemuck::cast_slice(&buf);
-        // write len(u64) in LE
-        let len = buf_u32.len() as u64;
-        input_data.extend(len.to_le_bytes());
-        // write buf
-        input_data.extend(buf);
 
         // handle error
-        let input_id = client.upload_input(input_data);
+        let input_id = client.upload_input(buf);
         tracing::info!("Uploaded input with id: {}", input_id);
         self.last_input_id = Some(input_id);
     }
@@ -365,11 +358,6 @@ impl<'a> ZkvmHost for Risc0BonsaiHost<'a> {
             buf.extend(vec![0; 4 - rem]);
         }
         let buf: &[u32] = bytemuck::cast_slice(&buf);
-        // write len(u64) in LE
-        let len = buf.len() as u64;
-        let len_buf = &len.to_le_bytes()[..];
-        let len_buf: &[u32] = bytemuck::cast_slice(len_buf);
-        self.env.extend_from_slice(len_buf);
         // write buf
         self.env.extend_from_slice(buf);
 
--- a/crates/risc0-bonsai/src/host.rs
+++ b/crates/risc0-bonsai/src/host.rs
@@ -372,12 +372,12 @@ impl<'a> ZkvmHost for Risc0BonsaiHost<'a> {
 
             let session = executor.run()?;
             // don't delete useful while benchmarking
-            println!(
-                "user cycles: {}\ntotal cycles: {}\nsegments: {}",
-                session.user_cycles,
-                session.total_cycles,
-                session.segments.len()
-            );
+            // println!(
+            //     "user cycles: {}\ntotal cycles: {}\nsegments: {}",
+            //     session.user_cycles,
+            //     session.total_cycles,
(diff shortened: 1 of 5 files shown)


crates/risc0-bonsai/src/host.rs
```diff
@@ -331,22 +331,15 @@ impl<'a> Risc0BonsaiHost<'a> {
         // Prepare input data and upload it.
         let client = self.client.as_ref().unwrap();
 
-        let mut input_data = vec![];
-        let mut buf = borsh::to_vec(&item).unwrap();
-        // append [0..] alignment
+        let mut buf = borsh::to_vec(&item).expect("Risc0 hint serialization is infallible");
+        // append [0..] alignment to cast &[u8] to &[u32]
         let rem = buf.len() % 4;
         if rem > 0 {
```
kpp, line 337: I am not sure if we need this anymore

kpp:
For consistency: there is a zkvm impl for tests, I think it should be changed too

crates/risc0-bonsai/src/host.rs
yaziciahmet: IMO it should be kept like this. Even if it doesn't benefit us anymore, if we were to ever send more data for any reason, it could be good to have the pre-knowledge that every message is aligned properly.

yaziciahmet:
Those tests feels useless isn't it? Talking about `crates/sovereign-sdk/adapters/risc0/tests/native.rs`. It uses an unused host implementation, with a mock guest. The whole test is mockery and not testing anything. We can delete this test and add a new test to Risc0BonsaiHost by using a mock guest. How does that sound @kpp ?

kpp:
Until you replace the old test you need to keep it consistent. Also that new test should go in a separate pr

crates/risc0-bonsai/src/host.rs
yaziciahmet: But... next reads will always fail as there will be dummy 0s. Hence with the reader we either find and read those alignment zeroes after every read, which doesn't seem good to me tbh, or we actually remove alignment. I will still test, but probably will go for 2nd.

crates/risc0-bonsai/src/host.rs
```diff
@@ -358,23 +343,12 @@ impl<'a> ZkvmHost for Risc0BonsaiHost<'a> {
...
-        let buf: &[u32] = bytemuck::cast_slice(&buf);
-        // write len(u64) in LE
-        let len = buf.len() as u64;
-        let len_buf = &len.to_le_bytes()[..];
-        let len_buf: &[u32] = bytemuck::cast_slice(len_buf);
-        self.env.extend_from_slice(len_buf);
+        let buf = borsh::to_vec(&item).expect("Risc0 hint serialization is infallible");
         // write buf
-        self.env.extend_from_slice(buf);
+        self.env.extend_from_slice(&buf);
```
kpp, line 348: Extend, not extend_from_slice

crates/sovereign-sdk/adapters/risc0/src/host.rs
```diff
@@ -71,20 +71,9 @@ impl<'a> ZkvmHost for Risc0Host<'a> {
...
-        let buf: &[u32] = bytemuck::cast_slice(&buf);
-        // write len(u64) in LE
-        let len = buf.len() as u64;
-        let len_buf = &len.to_le_bytes()[..];
-        let len_buf: &[u32] = bytemuck::cast_slice(len_buf);
-        self.env.extend_from_slice(len_buf);
+        let buf = borsh::to_vec(&item).expect("Risc0 hint serialization is infallible");
         // write buf
-        self.env.extend_from_slice(buf);
+        self.env.extend_from_slice(&buf);
```
kpp, line 76: extend

kpp (approved):

crates/sovereign-sdk/adapters/risc0/src/host.rs
yaziciahmet: yes to this one

crates/risc0-bonsai/src/host.rs
yaziciahmet: no to this one, because we would need to do a clone since buf is used after this statement. clone bad, clone very very very bad.

crates/risc0-bonsai/src/host.rs
yaziciahmet: went with the 2nd option, no more alignments on host.

kpp (approved):

Issue #945, opened 2024-08-02
Meta: Optimize guest code cycles


yaziciahmet:
Go through the guest code to find potential optimizations, and implement them.

**Achievable:**
Much lower cycle counts.

- [x] https://github.com/chainwayxyz/citrea/issues/751
- [x] https://github.com/chainwayxyz/citrea/issues/876
- [x] https://github.com/chainwayxyz/citrea/pull/968
- [x] https://github.com/chainwayxyz/citrea/pull/933
- [x] https://github.com/chainwayxyz/citrea/pull/959
- [x] https://github.com/chainwayxyz/citrea/issues/952

rakanalh:
@eyusufatik Every single sub-issue here was done. Shall we close this one?

PR #959, opened 2024-08-07
RISC0 guest optimization: Compiler parameter optimizations & Lower memory copies

yaziciahmet:
# Description

Introduced optimizations to RISC0 guest code. Here are the results:

**Testing runs with 10 blocks within a proof**

**nightly**
data size: 296637
user cycles: 52044801
total cycles: 82313216
segments: 79

**+ compiler optimizations**
data size: 297619
user cycles: 49478904
total cycles: 78643200
segments: 75

**+ lower mem copies in BitcoinVerifier**
data size: 298711
user cycles: 49470336
total cycles: 78118912
segments: 75

**+ BTreeSet instead of HashSet in BitcoinVerifier**
data size: 297440
user cycles: 49278291
total cycles: 77856768
segments: 75

*NOTE:*
- BTreeSet is suggested to be used by RISC0 docs. Lower mem copies and BTreeSet didn't provide big optimizations. But I believe that is because these tests were run against empty blocks with no transaction. Hence there was nothing to very little to copy anyways. Despite that it can be seen that they still provided very little improvement.
- Also this PR fixes a bug as well. If you check the following piece of code
```rs
let mut iter = inclusion_proof.wtxids.iter();
completeness_proof
            .iter()
            .all(|tx| iter.any(|&y| y == tx.wtxid().to_byte_array()));
```
you can see that the output of the `completeness_proof_iter().all(...)` is unused, so even if it returns false, the execution continues. The fix is to zip the txid and wtxids in inclusion proof, and searching for that exact tx in completeness proof.

## Linked Issues
- Related to #945 

## Testing
- Run BitcoinDA in regtest mode
- Run sequencer and prover
- Wait until a commitment is created and prover proves it

### DIFF
--- a/bin/citrea/provers/risc0/guest-bitcoin/Cargo.toml
+++ b/bin/citrea/provers/risc0/guest-bitcoin/Cargo.toml
@@ -33,19 +33,9 @@ lto = true
 opt-level = 3
 codegen-units = 1
 
-[profile.dev.build-override]
-debug = 0
-opt-level = 3
-codegen-units = 1
-
 [profile.release]
 debug = 0
 lto = true
-opt-level = 3
-codegen-units = 1
-
-[profile.release.build-override]
-debug = 0
 opt-level = 3
 codegen-units = 1
 
--- a/bin/citrea/provers/risc0/guest-bitcoin/Cargo.toml
+++ b/bin/citrea/provers/risc0/guest-bitcoin/Cargo.toml
@@ -28,17 +28,26 @@ secp256k1_v028 = { package = "secp256k1", version = "0.28", git = "https://githu
 k256 = { package = "k256", version = "0.13.3", git = "https://github.com/risc0/RustCrypto-elliptic-curves", tag = "k256/v0.13.3-risczero.0" }
 
 [profile.dev]
+debug = 0
+lto = true
 opt-level = 3
+codegen-units = 1
 
 [profile.dev.build-override]
+debug = 0
 opt-level = 3
+codegen-units = 1
 
 [profile.release]
-debug = 1
+debug = 0
 lto = true
+opt-level = 3
+codegen-units = 1
 
 [profile.release.build-override]
+debug = 0
 opt-level = 3
+codegen-units = 1
 
 [features]
 bench = [
--- a/bin/citrea/provers/risc0/guest-bitcoin/Cargo.toml
+++ b/bin/citrea/provers/risc0/guest-bitcoin/Cargo.toml
@@ -27,12 +27,6 @@ crypto-bigint = { git = "https://github.com/risc0/RustCrypto-crypto-bigint", tag
 secp256k1_v028 = { package = "secp256k1", version = "0.28", git = "https://github.com/Sovereign-Labs/rust-secp256k1.git", branch = "risc0-compatible-0-28-2" }
 k256 = { package = "k256", version = "0.13.3", git = "https://github.com/risc0/RustCrypto-elliptic-curves", tag = "k256/v0.13.3-risczero.0" }
 
-[profile.dev]
-debug = 0
-lto = true
-opt-level = 3
-codegen-units = 1
-
 [profile.release]
 debug = 0
 lto = true
--- a/crates/bitcoin-da/src/verifier.rs
+++ b/crates/bitcoin-da/src/verifier.rs
@@ -89,20 +89,92 @@ impl DaVerifier for BitcoinVerifier {
         inclusion_proof: <Self::Spec as DaSpec>::InclusionMultiProof,
         completeness_proof: <Self::Spec as DaSpec>::CompletenessProof,
     ) -> Result<<Self::Spec as DaSpec>::ValidityCondition, Self::Error> {
-        let validity_condition = ChainValidityCondition {
-            prev_hash: block_header.prev_hash().to_byte_array(),
-            block_hash: block_header.block_hash().to_byte_array(),
-        };
+        // create hash set of blobs
+        let mut blobs_iter = blobs.iter();
 
-        // check that wtxid's of transactions in completeness proof are included in the InclusionMultiProof
-        // and are in the same order as in the completeness proof
-        let mut iter = inclusion_proof.wtxids.iter();
-        let is_wtxs_found_in_block = completeness_proof.iter().all(|tx| {
+        let mut inclusion_iter = inclusion_proof
+            .txids
+            .iter()
+            .zip(inclusion_proof.wtxids.iter());
+
+        let prefix = self.reveal_tx_id_prefix.as_slice();
+        // Check starting bytes tx that parsed correctly is in blobs
+        let mut completeness_tx_hashes = BTreeSet::new();
+
+        for (index_completeness, tx) in completeness_proof.iter().enumerate() {
+            let txid = tx.txid().to_byte_array();
+            // make sure it starts with the correct prefix
+            if !txid.starts_with(prefix) {
+                return Err(ValidationError::NonRelevantTxInProof);
+            }
+
             let wtxid = tx.wtxid();
-            iter.any(|y| y == wtxid.as_byte_array())
-        });
-        if !is_wtxs_found_in_block {
-            return Err(ValidationError::RelevantTxNotFoundInBlock);
+            let wtxid = wtxid.as_byte_array();
+
+            // make sure completeness txs are ordered same in inclusion proof
+            // this logic always start seaching from the last found index
+            // ordering should be preserved naturally
+            let is_found_in_block =
+                inclusion_iter.any(|(txid_inc, wtxid_inc)| *txid_inc == txid && wtxid_inc == wtxid);
+            if !is_found_in_block {
+                return Err(ValidationError::RelevantTxNotFoundInBlock);
+            }
+
+            // it must be parsed correctly
+            if let Ok(parsed_tx) = parse_transaction(tx, &self.rollup_name) {
+                if let Some(blob_hash) = parsed_tx.get_sig_verified_hash() {
+                    let blob = blobs_iter.next();
+
+                    if blob.is_none() {
+                        return Err(ValidationError::ValidBlobNotFoundInBlobs);
+                    }
+
+                    let blob = blob.unwrap();
+                    if blob.hash != blob_hash {
+                        return Err(ValidationError::BlobWasTamperedWith);
+                    }
+
+                    if parsed_tx.public_key != blob.sender.0 {
+                        return Err(ValidationError::IncorrectSenderInBlob);
+                    }
+
+                    // decompress the blob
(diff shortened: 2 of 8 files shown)


bin/citrea/provers/risc0/guest-bitcoin/Cargo.toml
```diff
@@ -28,17 +28,26 @@ secp256k1_v028 = { package = "secp256k1", version = "0.28", git = "https://githu
 k256 = { package = "k256", version = "0.13.3", git = "https://github.com/risc0/RustCrypto-elliptic-curves", tag = "k256/v0.13.3-risczero.0" }
 
 [profile.dev]
+debug = 0
+lto = true
 opt-level = 3
+codegen-units = 1
 
 [profile.dev.build-override]
```
kpp, line 36: Only for build scripts, proc macros, and their dependencies. Does not do anything for the binary itself.

bin/citrea/provers/risc0/guest-bitcoin/Cargo.toml
```diff
@@ -28,17 +28,26 @@ secp256k1_v028 = { package = "secp256k1", version = "0.28", git = "https://githu
 k256 = { package = "k256", version = "0.13.3", git = "https://github.com/risc0/RustCrypto-elliptic-curves", tag = "k256/v0.13.3-risczero.0" }
 
 [profile.dev]
```
kpp, line 30: We never build risc0 binary in dev mode, that's how risc0 build scripts work

bin/citrea/provers/risc0/guest-bitcoin/Cargo.toml
```diff
@@ -28,17 +28,26 @@ secp256k1_v028 = { package = "secp256k1", version = "0.28", git = "https://githu
...
+codegen-units = 1
 
 [profile.release]
-debug = 1
+debug = 0
 lto = true
+opt-level = 3
+codegen-units = 1
 
 [profile.release.build-override]
```
kpp, line 47: same

crates/bitcoin-da/src/verifier.rs
```diff
@@ -136,14 +139,13 @@ impl DaVerifier for BitcoinVerifier {
                     let wtxids = inclusion_proof
                         .wtxids
                         .iter()
-                        .copied()
-                        .map(Txid::from_byte_array);
+                        .map(|wtxid| Txid::from_slice(wtxid).unwrap());
```
kpp, line 142:
```
            fn from_slice(sl: &[u8]) -> Result<Hash<$($gen),*>, FromSliceError> {
                if sl.len() != $bits / 8 {
                    Err(FromSliceError{expected: Self::LEN, got: sl.len()})
                } else {
                    let mut ret = [0; $bits / 8];
                    ret.copy_from_slice(sl);
                    Ok(Self::internal_new(ret))
                }
            }
                        fn from_byte_array(bytes: Self::Bytes) -> Self {
                Self::internal_new(bytes)
            }
```

So you added an extra .unwrap

crates/bitcoin-da/src/verifier.rs
```diff
@@ -252,10 +254,9 @@ impl DaVerifier for BitcoinVerifier {
         let tx_hashes = inclusion_proof
             .txids
             .iter()
-            .map(|tx| Txid::from_slice(tx).unwrap())
-            .collect::<Vec<_>>();
+            .map(|txid| Txid::from_slice(txid).unwrap());
```
kpp, line 257: from_bytes_array

crates/citrea-stf/src/verifier.rs
```diff
@@ -37,7 +37,7 @@ where
         zkvm: Zk,
         pre_state: Stf::PreState,
     ) -> Result<(), Da::Error> {
-        println!("Running sequencer commitments in DA slot");
+        // println!("Running sequencer commitments in DA slot");
```
kpp, line 40: That's what Esad asked not to remove. Please add a commend above "Don't remove"

crates/citrea-stf/src/verifier.rs
```diff
@@ -46,7 +46,7 @@ where
             data.completeness_proof,
         )?;
 
-        println!("going into apply_soft_confirmations_from_sequencer_commitments");
+        // println!("going into apply_soft_confirmations_from_sequencer_commitments");
```
kpp, line 49: Same

crates/citrea-stf/src/verifier.rs
```diff
@@ -63,7 +63,7 @@ where
                 data.soft_confirmations,
             );
 
-        println!("out of apply_soft_confirmations_from_sequencer_commitments");
+        // println!("out of apply_soft_confirmations_from_sequencer_commitments");
```
kpp, line 66: Same

crates/bitcoin-da/src/verifier.rs
yaziciahmet: But if we don't use from_slice, we have to copy the array as it was before. Even if we do not use `copied()` and use it as `from_byte_array(*wtxid)`, but this will again copy the data.

yaziciahmet: fixed

crates/bitcoin-da/src/verifier.rs
yaziciahmet: fixed

crates/citrea-stf/src/verifier.rs
yaziciahmet: fixed

crates/bitcoin-da/src/verifier.rs
```diff
@@ -96,10 +96,13 @@ impl DaVerifier for BitcoinVerifier {
...
-        let mut iter = inclusion_proof.wtxids.iter();
-        completeness_proof
-            .iter()
-            .all(|tx| iter.any(|&y| y == tx.wtxid().to_byte_array()));
+        completeness_proof.iter().all(|tx| {
+            let wtxid = tx.wtxid();
+            inclusion_proof
+                .wtxids
+                .iter()
+                .any(|y| y == wtxid.as_byte_array())
```
kpp, line 104: That's a good fix. This can be simplified with inclusion_proof.wtxids.contains(...).

And it shows that this part is O(N*M).

yaziciahmet: Will look into using BTreeMap/BTreeSet's instead of looping over arrays in a separate PR, but it's a bit more tricky to test that. As we will need transactions in the blocks, it will probably require writing some sort of script/test.

But I suspect and it says in docs that HashSet's are inefficient compared to their efficiency it brings in normal architectures. But it's still worth to test imo.

yaziciahmet: and now I actually suspect its a fix at all. because if you read the comment, it should also validate the order of items, in which makes it sensible to actually define iterator outside of loop to validate that the next element actually comes after.

but what I don't understand is, the output of `all` method is not used anywhere, there is no panic no errors. so even if it is false, we are not doing anything. anyone know why @eyusufatik @kpp @rakanalh @ercecan ?

kpp: Not using all is a bug

yaziciahmet: I don't see it boss. in outer iter, we expect everything to be true by using all. but in inner iter, we only need to find one equal value, so any.

crates/bitcoin-da/src/verifier.rs
```diff
@@ -252,10 +255,9 @@ impl DaVerifier for BitcoinVerifier {
         let tx_hashes = inclusion_proof
             .txids
             .iter()
-            .map(|tx| Txid::from_slice(tx).unwrap())
-            .collect::<Vec<_>>();
+            .map(|txid| Txid::from_byte_array(*txid));
```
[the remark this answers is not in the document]
yaziciahmet: nice catch

crates/risc0-bonsai/src/host.rs
```diff
@@ -368,12 +370,12 @@ impl<'a> ZkvmHost for Risc0BonsaiHost<'a> {
...
-            //     "user cycles: {}\ntotal cycles: {}\nsegments: {}",
-            //     session.user_cycles,
-            //     session.total_cycles,
-            //     session.segments.len()
-            // );
+            println!(
+                "user cycles: {}\ntotal cycles: {}\nsegments: {}",
+                session.user_cycles,
+                session.total_cycles,
+                session.segments.len()
```
eyusufatik, line 377: don't forget to comment these out

crates/risc0-bonsai/src/host.rs
```diff
@@ -412,6 +414,15 @@ impl<'a> ZkvmHost for Risc0BonsaiHost<'a> {
                         .expect("API error, missing receipt on completed session");
 
                     tracing::info!("Receipt URL: {}", receipt_url);
+                    if let Some(stats) = res.stats {
```
eyusufatik, line 417: this existed??????? 😫 

great find thanks

eyusufatik (approved):
good catches. tho some of the compiler options might not really be relevant.

kpp (approved):

crates/bitcoin-da/src/verifier.rs
```diff
@@ -251,11 +254,10 @@ impl DaVerifier for BitcoinVerifier {
...
         let tx_hashes = inclusion_proof
             .txids
-            .iter()
-            .map(|tx| Txid::from_slice(tx).unwrap())
-            .collect::<Vec<_>>();
+            .into_iter()
+            .map(Txid::from_byte_array);
 
-        if let Some(root_from_inclusion) = merkle_tree::calculate_root(tx_hashes.into_iter()) {
+        if let Some(root_from_inclusion) = merkle_tree::calculate_root(tx_hashes) {
```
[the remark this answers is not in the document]
yaziciahmet: done ser

eyusufatik (approved):

crates/bitcoin-da/src/verifier.rs
```diff
@@ -249,26 +177,85 @@ impl DaVerifier for BitcoinVerifier {
...
+        if !completeness_proof.is_empty() {
+            let coinbase_tx = &inclusion_proof.coinbase_tx;
+            // If there are more than one scriptPubKey matching the pattern,
+            // the one with highest output index is assumed to be the commitment.
+            // That  is why the iterator is reversed.
+            let commitment_idx = coinbase_tx.output.iter().rev().position(|output| {
+                output
+                    .script_pubkey
+                    .as_bytes()
+                    .starts_with(&[0x6a, 0x24, 0xaa, 0x21, 0xa9, 0xed])
```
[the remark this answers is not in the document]
yaziciahmet: double done ser

kpp (approved):

PR #961, opened 2024-08-08
Add vout index for prev tx

### LINKED ISSUE #958: BitDA: Witness program was passed an empty witness
## Summary

Seems like some tx output cannot be used to chain txs with `prev_tx`.

## Steps to reproduce

1. send btc to someone but don't produce block
2. try to put some data on DA:

```
    #[tokio::test]
    async fn send_transaction() {
        let da_service = get_service().await;

        let size = 2000;

        // create random bytes with size of size variable
        let blob = (0..size).map(|_| rand::random::<u8>()).collect::<Vec<u8>>();

        let txid = da_service.send_transaction(&blob).await;

        if txid.is_ok() {
            println!("Transaction sent successfully");
        } else {
            panic!("Failed to send transaction: {:?}", txid.err());
        }
    }
```

## Expected Result

Ok

## Actual Result

Error `Witness program was passed an empty witness`.

ercecan:
# Description
Fixes used output issue in prev tx

## Linked Issues
- Fixes #958

### DIFF
--- a/crates/bitcoin-da/src/spec/utxo.rs
+++ b/crates/bitcoin-da/src/spec/utxo.rs
@@ -13,40 +13,6 @@ pub struct UTXO {
     pub solvable: bool,
 }
 
-// Temporary struct to deserialize UTXO from JSON
-#[derive(Deserialize)]
-struct RawUTXO {
-    txid: String,
-    vout: u32,
-    address: String,
-    #[serde(rename = "scriptPubKey")]
-    script_pub_key: String,
-    amount: f64,
-    confirmations: u64,
-    spendable: bool,
-    solvable: bool,
-}
-
-// Deserialize UTXO from JSON
-impl<'de> serde::Deserialize<'de> for UTXO {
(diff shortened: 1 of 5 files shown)


eyusufatik (commented):
I think we can simplify a lot of things by just rewriting the "prev_tx" logic with a "prev_utxo" logic. I'll implement that soon

crates/bitcoin-da/src/spec/utxo.rs
```diff
@@ -39,7 +39,7 @@ impl<'de> serde::Deserialize<'de> for UTXO {
             vout: raw_utxo.vout,
             address: raw_utxo.address,
             script_pubkey: raw_utxo.script_pub_key,
-            amount: (raw_utxo.amount * 100_000_000.0) as u64, // satoshis to bitcoin
```
kpp, line 42: We actually get ListUnspentEntry here, not RawUTXO, right?

kpp (approved):

PR #962, opened 2024-08-08
Remove postgres

### LINKED ISSUE #960: get rid of postgres

rakanalh:
# Description

## Linked Issues
- Fixes #960 

## TODO
- [x] Replace `wait_for_postgres_commitment`
- [x] Replace `wait_for_postgres_proofs`
- [x] Save mempool transactions to ledger DB and restore


yaziciahmet (approved):
Feels so nice... LGTM

eyusufatik:
One thing we should implement is to saving the mempool transactions to ledger db.

Before removal of postgres we'd write them to postgres so that if we need to reboot the sequencer the memmpool transactions are not lost.

We can now put them in ledger db (in full node launched as sequencer this would mean we'd lose the mempool txs, tho that's fine I guess.)

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
```diff
@@ -466,6 +466,24 @@ impl SharedLedgerOps for LedgerDB {
...
+    }
+
+    /// Used by the sequencer to record that it has committed to soft confirmations on a given L2 height
+    #[instrument(level = "trace", skip(self), err, ret)]
+    fn set_last_commitment_l2_height(&self, l2_height: BatchNumber) -> Result<(), anyhow::Error> {
+        let mut schema_batch = SchemaBatch::new();
+
+        schema_batch.put::<LastSequencerCommitmentSent>(&(), &l2_height)?;
+        self.db.write_schemas(schema_batch)?;
+
```
eyusufatik, line 484: we can add to the function's documentation that the for  a sequencer  the "last" commitment is set when it is sent but for a full node the last commitment is set when a commitment is read from a finalized DA layer block.

eyusufatik (approved):

PR #980, opened 2024-08-13
Use bonsai sdk blocking client

eyusufatik:
# Description
Removes the bonsai sdk client wrapper queue system we have because we couldn't go async in that part.

Feels so nice 😇

# testing
I haven't tested this code even a bit @kpp  please double review the macro :)

### DIFF
--- a/crates/risc0-bonsai/src/host.rs
+++ b/crates/risc0-bonsai/src/host.rs
@@ -1,10 +1,8 @@
 //! This module implements the [`ZkvmHost`] trait for the RISC0 VM.
-
-use std::sync::mpsc::{self, Sender};
-use std::sync::Arc;
 use std::time::Duration;
 
 use anyhow::anyhow;
+use bonsai_sdk::blocking::Client;
 use borsh::{BorshDeserialize, BorshSerialize};
 use risc0_zkvm::sha::Digest;
 use risc0_zkvm::{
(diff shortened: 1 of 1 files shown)


crates/risc0-bonsai/src/host.rs
```diff
@@ -13,278 +11,40 @@ use risc0_zkvm::{
...
-                            // TODO: think about whether we should have a case where we use Bonsai with only execute mode
-                            let res = client.create_session(img_id, input_id, assumptions, false);
-                            let res = unwrap_bonsai_response!(res, 'client, 'queue);
-                            let _ = notify.send(res);
+use tracing::{error, info, warn};
+
+macro_rules! unwrap_bonsai_response {
+    ($response:expr) => (
+        loop {
+            match $response {
```
kpp, line 19:
```suggestion
    ($bonsai_call:expr) => (
        loop {
            let response = $bonsai_call;
            match response {
```

crates/risc0-bonsai/src/host.rs
```diff
@@ -13,278 +11,41 @@ use risc0_zkvm::{
...
-                            // TODO: think about whether we should have a case where we use Bonsai with only execute mode
-                            let res = client.create_session(img_id, input_id, assumptions, false);
-                            let res = unwrap_bonsai_response!(res, 'client, 'queue);
-                            let _ = notify.send(res);
+use tracing::{error, info, warn};
+
+macro_rules! unwrap_bonsai_response {
+    ($bonsai_call:expr) => (
+        loop {
+            let response = $bonsai_call;
```
rakanalh, line 19: Instead of `loop`ing forever until a good response is received... why don't we use backoff crate to retry some number of times before failing?

rakanalh (changes requested):

crates/risc0-bonsai/src/host.rs
eyusufatik: let's say we use backoff (I agree that's best practice), what happens when we fail due to some connection problems etc.? what should the prover do? I thought it will just retry proving --- which is nearly the same as just retrying indefinitely

rakanalh: Right... but backoff retry:
1. Simplifies / potentially removes the need for this macro.
2. Retries in a better way by increasing the time between retries so that we have higher potential of succeeding at the next retry than when retrying every 5 seconds or something like that.

rakanalh (approved):
LGTM

Issue #744, opened 2024-06-13
Chunks for proofs


rakanalh:
TBD

eyusufatik:
After #814 is impelmented, we will be able to create different types of DA transactions.

One of these types can be "chunk" transactions which would be type 1.

You simply make a type 0 transaction with a hash commitment (of data from all chunks) + the number of chunks this tx would require.

The type 1 transaction will have some extra fields right before pushing the body:
1. 32 bytes txid of the type 0 comitment transactions txid. (assumes witness is always in the first input of the reveal transaction) 
2. single byte showing the order of the "chunk".

This way on light client proofs we can show that the prover committed to n chunks to be written to bitcoin and we can give each chunk as input to the circuit to be used inside.

eyusufatik:
@ekrembal suggestion sounds even bettter, again this also requires #814 to be implemented.

we use type 0 transactions to write "chunks" to DA.

we define a new transaction type 1, which pushes n 32 bytes (bitcoin tx ids) indicating the order of the chunks.

the light client proof is inputted with all the transactions the type 1 transaction points to, the body of the "chunks" are simply concatenated and used inside the proof.

no need for commitments etc. and this is easier to use in light cilent proofs. I also favor this option now.

ekrembal:
I think it can also be improved with with the ability to type 1 transaction to add data,

For example if there is a need to write a 700kb data, instead of doing 3 transactions, we should be able to do it in 2 transactions

kpp:
Fixed in https://github.com/chainwayxyz/citrea/pull/925

PR #983, opened 2024-08-13
Prover commitments grouping 

### LINKED ISSUE #965: Prover pre-calculates size of sequencer commitments
We want to avoid sending data over 400kb's to the bitcoin network.

On our rollup this can only happen due to state diffs outputted by the batch proof.

A batch proof basically runs all the soft confirmations commited inside a (or multiple) sequencer commitments found in a finalized DA block.

Our batch proofs have the option to prove a range of sequencer commitments inside a DA block. (think of running only the first one among the 2 found in a DA block.)

A prover can pre-calculate the size of a state diff and decide which commitment(s) to prove. 

Since L2 sync and L1 scanning happens in parallel, there is a few ways a prover can know the state diff size of a given L2 range (which is what the commitments commit to):

1. Save state diffs of individual L2 blocks in its db, merge when needed.
2. Prover already saves Witness (trace of get sets to the state db), find a way to convert these to state diffs so we don't save (kind of) the same data twice.
3. There is an option to run the zk guest in "simulation" mode, which does not run the ZkVm at all, just invokes the same function ZkVm would call in native mode. We can call this function and simply get the state diff from its output.

When the prover has the ability to get the state diff size of certain ranges of L2 blocks, it can utilize a simple algorithm to select sequencer commitments to prove:

```
let seq_comms be array of sequencer commitments in a DA block

size = MAX_U64;
remaining = []

while calc_diff_size(seq_comms) > MAX_DATA_SIZE:
    remaining.push(seq_comms.pop())

prove(seq_comms)

# repeat above for remaining
```

Note: The prover usually uses only L1 height to record the last "proving" it made. With these changes, it will also need to know (record to DB) the index of the last sequencer commitment it proved.

rakanalh:
# Description
Split commitments into groups where the cumulative diff size does not exceed 400KB.

## Linked Issues
- Fixes #965 

## TODO
- [ ] Figure out how to write a test where DA block has multiple sequencer commitments
  - Currently, a DA block is created once every single sequencer commitment.


crates/sovereign-sdk/rollup-interface/src/state_machine/da.rs
```diff
@@ -42,7 +42,7 @@ pub trait DaSpec:
     type BlockHeader: BlockHeaderTrait<Hash = Self::SlotHash> + Send + Sync;
 
     /// The transaction type used by the DA layer.
-    type BlobTransaction: BlobReaderTrait<Address = Self::Address> + Send + Sync;
+    type BlobTransaction: BlobReaderTrait<Address = Self::Address> + Send + Sync + Clone;
```
kpp, line 45: Why?

rakanalh: Because `da_data` needs to be used in multiple iterations of the loop in the `process_l1_block` inside the prover. This value is used to among others to create `StateTransitionData` for each group of sequencer commitments.

eyusufatik:
> Currently, a DA block is created once every single sequencer commitment.

this is gonna be a big struggle with mockda 

will be much easier to do once bitcoin e2e framework is merged @jfldde 👀 

please open a seperate issue for this @rakanalh

crates/prover/src/runner.rs
```diff
@@ -786,6 +800,55 @@ where
...
+                    .get_l2_state_diff(BatchNumber(l2_height))?
+                    .ok_or(anyhow!(
+                        "Could not find state diff for L2 range {}-{}",
+                        sequencer_commitment.l2_start_block_number,
+                        sequencer_commitment.l2_end_block_number
+                    ))?;
+                cumulative_state_diff = merge_state_diffs(cumulative_state_diff, state_diff);
+            }
+
+            let serialized_state_diff = bincode::serialize(&cumulative_state_diff)?;
```
eyusufatik, line 828: we use borsh for this inside apply_sequencer_commitments_from_da I believe. so we should use that as well. pls double check

crates/prover/src/runner.rs
```diff
@@ -786,6 +800,55 @@ where
...
+
+            // The threshold is reached when we hit 95% of the limit to prevent exceeding it.
+            let state_diff_threshold_reached = serialized_state_diff.len() as u64
+                > (MAX_STATEDIFF_SIZE_PROOF_THRESHOLD
+                    .saturating_mul(95)
+                    .saturating_div(100));
+
```
eyusufatik, line 835: instead checking this way, we should check if the addition of nth commitment makes the merged commitment go over the limit

in that case we group until n-1.

in the case that a single commitment by itself goes over the limit, we have to accept that and have a single commitment group. (and that's solved by having [chainwayxyz/citrea#744](https://github.com/chainwayxyz/citrea/issues/744))

crates/prover/src/runner.rs
```diff
@@ -432,13 +436,15 @@ where
             da_data.iter_mut().for_each(|blob| {
                 blob.full_data();
             });
-            let sequencer_commitments: Vec<SequencerCommitment> =
+            let mut sequencer_commitments: Vec<SequencerCommitment> =
                 self.extract_sequencer_commitments(l1_block.header().hash().into(), &mut da_data);
+            // Make sure all sequencer commitments are stored in ascending order.
+            sequencer_commitments.sort_by_key(|commitment| commitment.l2_start_block_number);
```
eyusufatik, line 442: good catch, however if commitments in a group are not in same order also in da, it will break the proving. I'll make an issue about it 👍🏻

eyusufatik (approved):

crates/prover/src/runner.rs
rakanalh: bincode in this case is only used to serialize and then calculate the diff size. This is also used in the sequencer the same way. Do you want to switch to borsh?

eyusufatik: yes should have been borsh, probably we fdorgot to update those parts

eyusufatik (approved):

PR #987, opened 2024-08-14
Sort commitments in zk

### LINKED ISSUE #985: Sort commitments by height in zk
If the commitments are out of order, we can sort them inside `apply_sequencer_commitments_on_da_slot`. A good way to mitigate a potential problem for our rollup -- tho we try to prevent this already by chaining UTXOs.

yaziciahmet:
# Description
Sorting commitments in zk env just in case.

## Linked Issues
- Fixes #985

### DIFF
--- a/crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
+++ b/crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
@@ -523,8 +523,7 @@ where
         }
 
         // Sort commitments just in case
-        sequencer_commitments
-            .sort_unstable_by(|c1, c2| c1.l2_end_block_number.cmp(&c2.l2_end_block_number));
+        sequencer_commitments.sort_unstable();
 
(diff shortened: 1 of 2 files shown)


crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
```diff
@@ -522,6 +522,10 @@ where
             }
         }
 
+        // Sort commitments just in case
+        sequencer_commitments
+            .sort_unstable_by(|c1, c2| c1.l2_end_block_number.cmp(&c2.l2_end_block_number));
```
rakanalh, line 527: In #983 i am sorting by key
```
sequencer_commitments.sort_by_key(|commitment| commitment.l2_start_block_number);
```

I do think we should use the same sort for consistency

rakanalh (changes requested):

eyusufatik (approved):
aggree with rakan.

rakanalh (approved):

PR #989, opened 2024-08-14
Prevent processing pre-proven commitments

### LINKED ISSUE #977: Skip proving already proved sequencer commitments
## Summary

When we publish sequencer commitment on DA there is an ec signature of body(signed with sequencer secret key, it is not a bitcoin sig) + body itself. One can’t use a signature for another body (the sig check will fail in zk) but nothing stops Mallory from replaying the whole body of sequencer commitment forcing us to prove an already proved range.

## Steps to reproduce

1. sequencer pushes a sequencer commitment on DA
2. produce block
3. sequencer pushes a sequencer commitment on DA
4. mallory pushes a copy of sequencer commitment on DA
5. produce block
6. prove

rakanalh:
# Description

## Linked Issues
- Fixes #977

### DIFF
--- a/crates/prover/src/runner.rs
+++ b/crates/prover/src/runner.rs
@@ -472,6 +472,18 @@ where
                 break;
             }
 
+            // Make the decision to prove all sequencer commitment groups
+            let should_prove: bool = {
+                let mut rng = rand::thread_rng();
+                // if proof_sampling_number is 0, then we always prove and submit
+                // otherwise we submit and prove with a probability of 1/proof_sampling_number
+                if prover_config.proof_sampling_number == 0 {
+                    true
+                } else {
+                    rng.gen_range(0..prover_config.proof_sampling_number) == 0
+                }
+            };
+
             let sequencer_commitments_groups =
                 self.break_sequencer_commitments_into_groups(sequencer_commitments)?;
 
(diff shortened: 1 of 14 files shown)


crates/prover/src/runner.rs
```diff
@@ -492,14 +491,25 @@ where
...
 
+            // if proof_sampling_number is 0, then we always prove and submit
+            // otherwise we submit and prove with a probability of 1/proof_sampling_number
+            let should_prove = prover_config.proof_sampling_number == 0
+                || rand::thread_rng().gen_range(0..prover_config.proof_sampling_number) == 0;
+
+            let mut sequencer_commitments =
+                self.filter_out_proven_commitments(&sequencer_commitments)?;
+            // Make sure all sequencer commitments are stored in ascending order.
+            sequencer_commitments.sort_unstable();
```
kpp, line 502: But we still can have duplicates of unproved commitments.

rakanalh: The line you're addressing is irrelevant & you don't provide details on why you think that is the case. Can you elaborate please?

kpp: No idea where to put my comment. So I left it there.

1. sequencer pushes a sequencer commitment on DA
2. mallory scans mempool and pushes a copy of sequencer commitment on DA

rakanalh: Ok so AFAIU 2 transactions with the same sequencer commitment into the DA layer. What happens is that one or both of these transactions is going to be included into the DA layer blocks. Here we have 2 cases:

Case 1: Commitment txs are included in different DA blocks:
In this case, the fullnode & prover will read the first block,extract commitments and mark L2 range as finalized. The second block containing the replayed commitment would be filtered out since the L2 range is already finalized.

We agree this is handled?

Case 2: Commitment txs are included in the same DA block

What we can do in this case is to filter out commitments which have the same L2 range that we've just extracted from the current DA block. what do you think?

kpp: 1) yes
2) yes. I believe we don't care which one seq commitment to prove as long as they are valid

rakanalh: Handled that scenario

eyusufatik (approved):
It'll work fine, what about just passing indices of the sequencer commitments to be skipped? does this have any benefits over that?

rakanalh:
> It'll work fine, what about just passing indices of the sequencer commitments to be skipped? does this have any benefits over that?

I am passing the preproven ranges because the STF structure was also accepting ranges so i followed the same approach.

PR #997, opened 2024-08-15
Check SKIP_GUEST_BUILD value

jfldde:
# Description

- Current behaviour does not check for the inside Ok(value). It will skip guest build for `SKIP_GUEST_BUILD=false` and `SKIP_GUEST_BUILD=0`.


rakanalh (approved):
LGTM

PR #1010, opened 2024-08-18
Erce/minimum basefee

### LINKED ISSUE #998: Set a minimum value for base fee
Currently we are using eip-1559 for fee calculations, however when there are a lot of empty blocks in a row the base fee drops down rapidly and it is hard to get it back up. That's why we can set a minimum value to our base fee like [arbitrum](https://docs.arbitrum.io/how-arbitrum-works/gas-fees#gas-price-floor) did.

ercecan:
# Description
Sets a minimum value to basefee

## Linked Issues
- Fixes #998


crates/sovereign-sdk/fuzz/Cargo.lock
kpp, line 1: Why is it so big?

crates/primitives/src/basefee.rs
```diff
@@ -0,0 +1,16 @@
...
+
+use crate::MIN_BASE_FEE_PER_GAS;
+
+pub fn calculate_next_block_base_fee(
+    gas_used: u128,
+    gas_limit: u128,
+    base_fee: Option<u64>,
+    base_fee_params: BaseFeeParams,
+) -> Option<u64> {
+    Some(std::cmp::max(
```
eyusufatik, line 12: why option?

+ shouldn't this return u128?

bin/citrea/tests/evm/fee.rs
```diff
@@ -0,0 +1,60 @@
...
+    assert!(block.header.base_fee_per_gas.unwrap() >= 10000000);
+    for _ in 0..100 {
+        test_client.spam_publish_batch_request().await.unwrap();
+    }
+
+    let block = test_client
+        .eth_get_block_by_number(Some(BlockNumberOrTag::Latest))
+        .await;
+    // Base fee should at most be 0.01 gwei
+    assert!(block.header.base_fee_per_gas.unwrap() >= 10000000);
```
eyusufatik, line 57: make this assert base_fee == 10000000

eyusufatik (changes requested):

crates/sovereign-sdk/fuzz/Cargo.lock
ercecan: added reth primitives to primitives, and it is big

crates/primitives/src/basefee.rs
ercecan: The old way we calc basefee is usually something like this:
```rust
parent_block
.header
.next_block_base_fee(cfg.base_fee_params)
```
the function  next_block_base_fee returns `Option<u64>` that's why I tried to make it sth like that

also  pretty much everywhere basefee is u64

eyusufatik: ok, I'll respect that

eyusufatik (approved):

PR #1011, opened 2024-08-19
Enable verifying proofs based on the fork

### LINKED ISSUE #967: Native fork spec for method ids
A full node needs to know what verifying key to use while verifying a batch proof.

Most of the changes to the non-native feature gated code changes the verifying key.

An easy to edit and use way of doing something like this makes sense.

We can look into using the fork manager for this. (but the manager of this should be native-gated).

TODOs:
- [ ] Add a new output to the batch proof, the latest fork spec used.
- [ ] Add a get verifying key method to the fork manager based on the active spec. (careful should be native only)
- [ ] Use this method in the full nodes to get verifying key to verify zk proofs.

yaziciahmet:
# Description

This PR adds a mapping from the fork spec id to risc0 code commitment. This will enable full-nodes to verify the proofs retrospectively.

Also, this PR clears up the ForkManager implementation a bit.

There exists couple of concerns and potential future issues:
1. Our current fork implementation does not account for different environments (devnet, testnet, mainnet, local etc.). It is out of the scope of this PR, but we should keep this in mind.
2. Currently, since we have only one fork, Genesis, its corresponding method id is just read from the build output of the guest code. But we need it hardcoded in order to be able to not lose the previous fork. During development phase it is prone to change, so it seems ok to keep it like this for now. But before testnet, we need to have it as hardcoded, and also have some way to verify that these hardcoded values actually match.

## Linked Issues
- Fixes #967


crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
```diff
@@ -50,10 +53,11 @@ pub trait ZkvmHost: Zkvm + Clone {
     /// with some mild performance overhead and is not as easy to debug as [`simulate_with_hints`](ZkvmHost::simulate_with_hints).
     fn run(&mut self, with_proof: bool) -> Result<Proof, anyhow::Error>;
 
-    /// Extracts public input form the proof.
+    /// Extracts public input and receipt from the proof.
+    #[allow(clippy::type_complexity)]
     fn extract_output<Da: DaSpec, Root: BorshDeserialize>(
         proof: &Proof,
-    ) -> Result<StateTransition<Da, Root>, Self::Error>;
+    ) -> Result<(StateTransition<Da, Root>, Option<Receipt>), Self::Error>;
```
eyusufatik, line 60: I don't really like binding *Risc0* receipts here tbh. what if we use another zkvm?

we should keep these kind of trait generic

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
```diff
@@ -63,7 +67,11 @@ pub trait ZkvmHost: Zkvm + Clone {
 /// Must support recursive proofs.
 pub trait Zkvm: Send + Sync {
     /// A commitment to the zkVM program which is being proven
-    type CodeCommitment: Clone + Debug + Serialize + DeserializeOwned;
+    type CodeCommitment: Clone
+        + Debug
+        + Serialize
+        + DeserializeOwned
+        + Into<risc0_zkvm::sha::Digest>;
```
eyusufatik, line 74: again here also, let's use something like Into<[u8; 32]>

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
```diff
@@ -144,6 +152,8 @@ pub struct StateTransition<Da: DaSpec, Root> {
     /// to be checked outside of the zkVM circuit. This typically corresponds to
     /// some claim about the DA layer history, such as (X) is a valid block on the DA layer
     pub validity_condition: Da::ValidityCondition,
+    /// The final spec id after state transition is completed.
+    pub final_spec_id: SpecId,
```
eyusufatik, line 156: let's call this last_active_spec_id

crates/sovereign-sdk/adapters/risc0/src/host.rs
```diff
@@ -92,18 +92,24 @@ impl<'a> ZkvmHost for Risc0Host<'a> {
...
             }
             Proof::Full(data) => {
                 let receipt: Receipt = bincode::deserialize(data)?;
-                receipt.journal
+                (receipt.journal.clone(), Some(receipt))
             }
         };
-        Ok(BorshDeserialize::deserialize(&mut journal.bytes.as_ref())?)
+        Ok((BorshDeserialize::try_from_slice(&journal.bytes)?, receipt))
     }
```
eyusufatik, line 113: this method just keeps getting uglier.

you want to add another method to the trait where it returns the receipt itself, not just the output?

eyusufatik (changes requested):

crates/primitives/src/fork/manager.rs
```diff
@@ -1,83 +1,66 @@
...
-        self.active_spec
+    pub fn active_fork(&self) -> &Fork {
+        &self.forks[self.active_fork_idx]
     }
 
-    fn register_block(&mut self, height: u64) -> anyhow::Result<()> {
-        if let Some((new_spec, activation_block_height)) = self.specs.front() {
-            if height == *activation_block_height {
-                #[cfg(feature = "native")]
-                info!("Activating fork {:?} at height: {}", *new_spec, height);
```
rakanalh, line 57: This would be very useful IMO so let's restore it.

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
yaziciahmet: Then we should be either accepting serializing the proof twice to not expose the Receipt, or should be exposing It by creating a `VerifiableReceipt` trait, and implementing the trait on the risc0::Receipt type.

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
yaziciahmet: If we don't expose the receipt, this won't be needed, else, will update this.

crates/sovereign-sdk/adapters/risc0/src/host.rs
yaziciahmet: no longer relevant with the current version

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
yaziciahmet: done

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
yaziciahmet: no longer relevant with the current version

crates/primitives/src/fork/manager.rs
yaziciahmet: agreed, restored.

rakanalh (approved):
LGTM

crates/fullnode/src/runner.rs
```diff
@@ -230,47 +231,37 @@ where
...
                 if !self.accept_public_input_as_proven {
                     return Err(anyhow!(
-                        "Found public input in da block number: {:?}, Skipping to next proof..",
+                        "Found public input in da block number: {}, Skipping to next proof..",
                         l1_block.header().height(),
                     )
                     .into());
                 }
-                // public input is accepted only in tests, so ok to expect
-                Vm::extract_output(&proof).expect("Proof should be deserializable")
```
eyusufatik, line 271: if you remove this, tests are going to be broken, no?

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
```diff
@@ -50,7 +52,8 @@ pub trait ZkvmHost: Zkvm + Clone {
     /// with some mild performance overhead and is not as easy to debug as [`simulate_with_hints`](ZkvmHost::simulate_with_hints).
     fn run(&mut self, with_proof: bool) -> Result<Proof, anyhow::Error>;
 
-    /// Extracts public input form the proof.
+    /// Extracts public input and receipt from the proof.
+    #[allow(clippy::type_complexity)]
```
eyusufatik, line 56: no need for this now that you reverted the changes I assume?

eyusufatik (approved):
LGTM
left a few nits

crates/fullnode/src/runner.rs
eyusufatik: ah i see, we were unnecessarily checking if we could deserialize the "receipt" from execute mode.

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
yaziciahmet: nice catch, removed

crates/risc0-bonsai/src/host.rs
```diff
@@ -405,11 +405,16 @@ impl<'host> Zkvm for Risc0BonsaiHost<'host> {
...
-        _code_commitment: &Self::CodeCommitment,
-    ) -> Result<&'a [u8], Self::Error> {
-        unimplemented!();
+    fn verify(
+        serialized_proof: &[u8],
+        code_commitment: &Self::CodeCommitment,
+    ) -> Result<Vec<u8>, Self::Error> {
+        let receipt: Receipt = bincode::deserialize(serialized_proof)?;
+
+        #[allow(clippy::clone_on_copy)]
```
kpp, line 414: What's that?

yaziciahmet: CodeCommitment type is copyable (inner is [u32;8]), but we are calling clone on it. Hence it is unnecessary and can be just dereferenced with *. It was like that previously hence didn't give much attention just copy-pasted.

PR #1012, opened 2024-08-19
Pass index of preproven commitments instead of L2 range

rakanalh:
# Description

## Linked Issues
- Related to #977

### DIFF
--- a/crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
+++ b/crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
@@ -531,7 +531,7 @@ where
         // commitment index is not at the beginning of the list of preproven indicies.
         let mut filtered = vec![];
         for (index, sequencer_commitment) in sequencer_commitments.into_iter().enumerate() {
-            if let Some(exclude_index) = preproven_commitment_indicies.get(0) {
+            if let Some(exclude_index) = preproven_commitment_indicies.first() {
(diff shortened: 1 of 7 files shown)


crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
```diff
@@ -526,27 +523,16 @@ where
...
-        }
-
-        let mut sequencer_commitments = filtered;
         // Sort commitments just in case
         sequencer_commitments.sort_unstable();
 
+        let sequencer_commitments: Vec<_> = sequencer_commitments
+            .into_iter()
+            .enumerate()
+            .filter(|(index, _)| !preproven_commitment_indicies.contains(index))
```
eyusufatik, line 532: I guess we can use Iter::any instead of contains so that we don't rescan the preproven indices from the start. might be wrong on the function name but I'm %100 sure there is something like that. 

we'd just need to make sure the list is ever increasing.

eyusufatik (approved):

crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
eyusufatik: well couldn't really find what I was looking for, maybe it was an iter trick for some other condition..

kpp:
Don't we need exactly one range instead of vec<range> how it used to be or even vec<idxs>?

PR #1031, opened 2024-08-21
Fix prover commitment serialization

yaziciahmet:
# Description

This pr fixes the DaData ser/de. BitcoinService serialized the inner values of DaData, while other components tried to deserialize it using `DaData::try_from_slice`.

### DIFF
--- a/crates/bitcoin-da/src/service.rs
+++ b/crates/bitcoin-da/src/service.rs
@@ -308,15 +308,7 @@ impl BitcoinService {
         let rollup_name = self.rollup_name.clone();
         let da_private_key = self.da_private_key.expect("No private key set");
 
-        let blob = match da_data {
-            DaData::ZKProof(proof) => {
-                let blob = borsh::to_vec(&proof).expect("Should serialize");
-                compress_blob(&blob)
-            }
-            DaData::SequencerCommitment(commitment) => {
-                borsh::to_vec(&commitment).expect("Should serialize")
-            }
-        };
+        let blob = borsh::to_vec(da_data).expect("DaData serialize must not fail");
 
         // get all available utxos
         let utxos = self.get_utxos().await?;
@@ -331,6 +323,7 @@ impl BitcoinService {
 
         match da_data {
             DaData::ZKProof(_) => {
+                let blob = compress_blob(&blob);
                 // create inscribe transactions
                 let inscription_txs = create_zkproof_transactions(
                     &rollup_name,
--- a/crates/prover/src/prover_service/parallel/mod.rs
+++ b/crates/prover/src/prover_service/parallel/mod.rs
@@ -215,7 +215,7 @@ where
         &self,
         da_service: &Arc<Self::DaService>,
     ) -> Result<Vec<(<Da as DaService>::TransactionId, Proof)>, anyhow::Error> {
-        tracing::info!("Checking if ongoing bonsai session exists");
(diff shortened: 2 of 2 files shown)


crates/bitcoin-da/src/service.rs
```diff
@@ -308,15 +308,10 @@ impl BitcoinService {
...
-                let blob = borsh::to_vec(&proof).expect("Should serialize");
-                compress_blob(&blob)
-            }
-            DaData::SequencerCommitment(commitment) => {
-                borsh::to_vec(&commitment).expect("Should serialize")
-            }
-        };
+        let mut blob = borsh::to_vec(da_data).expect("DaData must serialize");
+        if let DaData::ZKProof(_) = da_data {
+            blob = compress_blob(&blob);
```
kpp, line 313: We may do that on line 329

crates/bitcoin-da/src/service.rs
```diff
@@ -308,15 +308,10 @@ impl BitcoinService {
...
-        let blob = match da_data {
-            DaData::ZKProof(proof) => {
-                let blob = borsh::to_vec(&proof).expect("Should serialize");
-                compress_blob(&blob)
-            }
-            DaData::SequencerCommitment(commitment) => {
-                borsh::to_vec(&commitment).expect("Should serialize")
-            }
-        };
+        let mut blob = borsh::to_vec(da_data).expect("DaData must serialize");
```
kpp, line 311: Thats a fix, it should work. However the actual fix should be where we call DaData::from_slice, because it’s not DaData anymore. It’s a SequencerCommitment | ForcedTransaction

yaziciahmet: To separate ZKProof and SequencerCommitment, checking whether the tx sender is sequencer or prover is sufficient, and we can deserialize directly into the respective struct rather than DaData. But I do not know whether we will be able to discriminate the ForcedTransaction in a similar way.

Even if we can distinguish that one somehow as well, I am not sure about the benefit we have from that. IMO having some sort of variant discriminator in the blob might be useful.

kpp: That's totally my bad.

Yes, I agree there should be a discriminator. But it shouldn't be of DaData.

This is how it works for DA now:
1. We send DaData tx
2. It is split in either TransactionKindLightClient(Complete, Chunks) or TransactionKindBatchProof(SequencerCommitment, "ForcedTransaction").
3. extract_sequencer_commitments scans for TransactionKindBatchProof only, filtering out zkproofs.
4. So we are left with SequencerCommitment and "ForcedTransaction". So I believe we should add TransactionKindBatchProof as a discriminator.

kpp: Or we should wrap it into enum BatchProof { SequencerCommitment, ForcedTransaction }

eyusufatik (approved):

eyusufatik:
I'd expect the bitcoin da verifier tests to fail?

Do you know if bitcoin da parsers should also change?

yaziciahmet:
@eyusufatik This is specific to method `send_transaction_with_fee_rate`, which is not used anywhere other than the DA queue worker, so there are no tests for this.

kpp (approved):
Okay as a tmp fix to make things work. However a proper fix needs to be implemented. Will do tomorrow.

Issue #999, opened 2024-08-16
Seperate AccountInfo struct


eyusufatik:
balance, code_hash, and nonce are all grouped in a single struct.

so if only one of them changes, we have to write the whole struct again in a state diff.

We can have separate state values for these in the DbAccount.

I am not exactly sure about this because this would require storage read from more keys resulting in more cycle counts but less data write to da.

ercecan:
In order to see which is better I ran the benchmark by using `test_all_flow` in e2e tests:
### Nightly Results
In nightly db account is sth like this:
```rust
pub(crate) struct DbAccount {
    pub(crate) info: AccountInfo,
    pub(crate) storage: StateMap<U256, U256, BcsCodec>,
    pub(crate) keys: StateVec<U256, BcsCodec>,
}
```
Anytime an update is made to `nonce`, `balance` or `codehash` which are normal values in `AccountInfo`, the whole of `DbAccount` is written to state diff

```
---- e2e::test_all_flow stdout ----


user cycles: 21790935
total cycles: 32505856
segments: 31


State diff size: 6809


user cycles: 16386437
total cycles: 24641536
segments: 24
State diff size: 4813

Elapsed: 63.94s
```
### Updated DbAccount branch results
updated DbAccount  looks like this:
```rust
pub(crate) struct DbAccount {
    pub(crate) balance: StateValue<U256, BcsCodec>,
    pub(crate) code_hash: StateValue<B256, BcsCodec>,
    pub(crate) nonce: StateValue<u64, BcsCodec>,
    pub(crate) storage: StateMap<U256, U256, BcsCodec>,
    pub(crate) keys: StateVec<U256, BcsCodec>,
}
```
With this change, changes on nonce, balance and codehash are all atomic and do not effect state diffs, but this increases the number of reads and writes quite a bit.

```
---- e2e::test_all_flow stdout ----


user cycles: 25136290
total cycles: 36700160
segments: 35 

 
State diff size: 6579

user cycles: 18790439
total cycles: 27262976
segments: 26 

State diff size: 3512

Elapsed: 64.00s
```

PR #1034, opened 2024-08-22
Use state values for account info

ercecan:
# Description
Uses state values for account info in dbaccount to lower state diff (but increases cycle counts)

## Linked Issues
- Related to #999


eyusufatik:
too small gains over size for the amount of cycles we get

PR #1039, opened 2024-08-22
Adjust diff size

### LINKED ISSUE #1035: Adjust diff size calculation for l1 fee compensation
Normally when an accounts balance changes we should only write the 20 byte address as key and 32 byte balance value to the state diff, however in our current implementation when we update any single one of the following fields like balance, nonce or code hash, even though only one of them changes, we write all of them to the state diff [#999].

We need to compensate this in L1 fees by adjusting our diff size calculation.

For example right now we only add the balance if balance changes of an account , but we should also be adding the size of code_hash and nonce.
So when a `balance` changes add extra `size_of(code_hash) + size_of(nonce)`
when a `nonce` changes add extra `size_of(balance) + size_of(code_hash)`
when a `code_hash` changes add extra `size_of(balance) + size_of(nonce)`

ercecan:
# Description
Adjusts diff sizes on balance, nonce and code changes
This pr is introduced because due to our architecture whenever an accounts balance, nonce or code changes the entire struct DbAccount is written to the storage with it's key and then they are added to the state diff.
DbAccount in total is 256 bytes, the key has 24 bytes of prefix and 1 byte of size descriptor and 20 bytes of address. I removed 20 bytes from address as we default add it for every operation. 
So in this pr we add 256 + 25 = 281 more bytes for every nonce, balance or code changed.

## Linked Issues
- Fixes #1035

### DIFF
--- a/crates/evm/src/evm/handler.rs
+++ b/crates/evm/src/evm/handler.rs
@@ -23,6 +23,12 @@ use tracing::instrument;
 use crate::system_events::SYSTEM_SIGNER;
 use crate::{BASE_FEE_VAULT, L1_FEE_VAULT};
 
+const DB_ACCOUNT_SIZE: usize = 256;
+
+// Normally db account key is: 24 bytes of prefix + 1 byte for size of remaining data + 20 bytes of address = 45 bytes
+// But we already add address size to diff size, so we don't need to add it here
+const DB_ACCOUNT_KEY_SIZE: usize = 25;
+
 #[derive(Copy, Clone, Default, Debug)]
 pub struct TxInfo {
     pub l1_diff_size: u64,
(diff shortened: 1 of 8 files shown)


crates/evm/src/tests/call_tests.rs
```diff
@@ -215,7 +221,7 @@ fn call_test() {
                 },
                 gas_used: 132943,
                 log_index_start: 0,
-                l1_diff_size: 565,
+                l1_diff_size: 1023,
```
eyusufatik, line 224: why only one of the transactions size changed?

crates/evm/src/evm/handler.rs
```diff
@@ -449,6 +449,10 @@ fn calc_diff_size<EXT, DB: Database>(
 
     let slot_size = 2 * size_of::<U256>(); // key + value;
     let mut diff_size = 0usize;
+    let db_account_size = 256;
+    // Normally db account key is: 24 bytes of prefix + 1 byte for size of remaining data + 20 bytes of address = 45 bytes
+    // But we already add address size to diff size, so we don't need to add it here
```
eyusufatik, line 454: these two should be constants

eyusufatik (changes requested):

crates/evm/src/tests/call_tests.rs
ercecan: Fixed

eyusufatik (approved):

PR #1032, opened 2024-08-21
Enable blocking RPC methods

rakanalh:
# Description

## Linked Issues
- Fixes #1018

Issue #1018, opened 2024-08-19
Resolve long running RPC calls halting node sync


rakanalh:
It was noted that calls such as:
- `eth_call`
- `eth_estimateGas`

among others cause the node's sync to halt / stall. 

Investigate & resolve

PR #1042, opened 2024-08-25
Convert async RPC into blocking RPC

rakanalh:
# Description

Tokio's default blocking threads is 512 threads. These threads are created on demand when `spawn_blocking` is called.

This PR converts some RPC calls into blocking methods because the only async part of them was using `tokio::sync::Mutex` which is now replaced with `parking_lot::Mutex` since there's nothing else async about it.

This would merge into #1032 

## Linked Issues
- Related to #1018

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1994,6 +1994,7 @@ dependencies = [
  "hex",
  "hyper 1.4.1",
  "jsonrpsee",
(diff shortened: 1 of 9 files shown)


kpp (approved):
LGTM. We may try to use RwLock in the future (it has a cool upgrade read->write method)

PR #1044, opened 2024-08-26
da remove rollup name and use single sha256

### LINKED ISSUE #1004: use sha256 instead of sha256d while signing for da

### LINKED ISSUE #1041: DA: remove rollup name from header
We don't need it anymore.

eyusufatik:
# Description
removes rollup name from the da transactions beause we namespace with wtxid prefixes

## Linked Issues
closes #1041 
closes #1004 

## Testing
existing tests are updated


crates/bitcoin-da/src/helpers/builders.rs
```diff
@@ -519,7 +507,8 @@ pub fn create_inscription_type_0(
         // push nonce
         reveal_script_builder = reveal_script_builder
             .push_slice(nonce.to_le_bytes())
-            .push_opcode(OP_DROP);
+            // drop the second item, bc there is a big change it's 0 (tx kind) and nonce is >= 16
```
rakanalh, line 510:
```suggestion
            // drop the second item, bc there is a big chance it's 0 (tx kind) and nonce is >= 16
```
I think this is what was meant. Please change in all other places as well.

rakanalh (approved):
LGTM

crates/bitcoin-da/src/helpers/builders.rs
eyusufatik: good catch

kpp (approved):

Issue #1015, opened 2024-08-19
Improving fees


eyusufatik:
## Proving fee
Currently we charge:
- Base fee
- Priority fee - if the user wants to pay it ofc
- L1 fee

We neglected another important fee though, proving fee.

On our latest profiling map, we can see that most of the proving time is spent on JMT verification and update.

For this reason, for now, it's a good enough estimation to charge the proving fee based on the diff size. 

```
proving_fee = l1_fee * k
```

Implemented, it would be something like this:

```
l1_fee = (l1_fee_overhead  + l1_diff_size) *  l1_fee_rate
proving_fee = l1_diff_size * k

decrease_caller_balance(&from, l1_fee + proving_fee)
```

TODOs:

- [ ] Decide on a constant for the proving fee.
- [ ] Implement proving fee, it should be pretty much the same as L1 fee in terms of deduction etc.
- [ ] Update rpc's with proving fee. (add new fields)
- [ ] For things named like transactions failing on L1 fee, rename those things to Rollup fees (open for discussion), since they now fail on a total of l1 fees and proving fees. (think about removing some txs from the mempool bc they pass base + priority fee balance check but fail after execution on L1 fees.)

## L1 fee overhead
For every data write operation we do, there is data we have to write to bitcoin besides the state diff itself.

We need to charge this for the average number of transactions we want to put on a batch proof.

```
overhead_per_tx = total_overhead / avg_no_txs_in_batch_proof
```

Total overhead consists of non-push-data parts of the bitcoin transaction (so all of commit + some parts of the reveal transaction).

TODOs:
- [ ] Calculate the size of the non-push-data parts of the bitcoin transaction, this can be a rough estimation (5-6 bytes difference is fine + we can never be sure of how many inputs the txs are gonna have so there is no point of finding an exact number)
- [ ] Figure out how many txs are going to be processed on a 10 min basis.
- [ ] Similar to above but keep track of the cumulative state diff size and find avg. number of txs whenever we reach 400kb. Let's talk if there is a really big difference between the two numbers.
- [ ] Update L1 fee calculation with the findings.

### Notes
Overall, we're going to see increase in the fees paid by transactions, however, this decision is really helpful for making the rollup sustainable and ready for making it decentralized.

eyusufatik:
For testnet, I'm reducing this issue to the overhead part only.

@exeokan has the necessary statistics. 

This can be implemented along with #966

orkunkilic:
[...]

PR #1054, opened 2024-08-29
Implement l1 fee overhead

ercecan:
# Description
Adds L1 fee overhead

## Linked Issues
- Related to #1015


kpp (approved):

kpp:
What I dont understand is why we need this in pur code instead of slightly increasing l1fee rate

ercecan:
> What I dont understand is why we need this in our code instead of slightly increasing l1fee rate

I believe it is to be as precise as possible

rakanalh (approved):

eyusufatik (approved):

PR #1067, opened 2024-09-02
Get soft confirmation signature as ref

jfldde:
# Description

- Prevent cloning the signature

### DIFF
--- a/crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
+++ b/crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
@@ -181,6 +181,11 @@ impl SignedSoftConfirmation {
         self.signature.clone()
     }
 
+    /// Signature of the sequencer as ref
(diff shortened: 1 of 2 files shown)


crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
```diff
@@ -181,6 +181,11 @@ impl SignedSoftConfirmation {
         self.signature.clone()
     }
 
+    /// Signature of the sequencer as ref
+    pub fn signature_as_ref(&self) -> &[u8] {
```
kpp, line 185:
```suggestion
    pub fn signature(&self) -> &[u8] {
```

kpp (approved):

crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
eyusufatik: i think there is also a method named `signature` right above

PR #1068, opened 2024-09-02
change types of fee rates and fix the rounding to 2 problem

eyusufatik:
# Description
without the new round to decimals, we would get 1.00000...02 and that would ceil to 2 sat per vbyte

now it rounds correctly to 8 decimal points + I've changed the return types of fee rates to u64 which should be enough for our calcs


ercecan (approved):

crates/bitcoin-da/src/lib.rs
```diff
@@ -12,3 +12,8 @@ const REVEAL_OUTPUT_AMOUNT: u64 = 546;
 
 #[cfg(feature = "native")]
 const MAX_TXBODY_SIZE: usize = 397000;
+
+fn round(x: f64, decimals: u32) -> f64 {
```
kpp, line 16: https://docs.rs/rust_decimal/latest/rust_decimal/struct.Decimal.html#method.rescale

kpp (changes requested):
I doubt the current fn round works as expected for all floats

kpp (approved):

eyusufatik:
Refactored for better readability and fixed a few simple mistakes in our tx builders.

We should only use taproot addresses in our wallets from now on.

PR #1077, opened 2024-09-02
Update get_relevant_blobs_from_txs

### LINKED ISSUE #1040: DA: get_relevant_blobs_from_txs must accept &[Tx] not Vec<Tx>
To avoid unnecessary copy up to 4mb of data in zk context.

exeokan:
# Description
Modifies get_relevant_blobs_from_txs to accept &[&Transaction] and avoids copying in extract_relevant_blobs

## Linked Issues
- Fixes #1040

### DIFF
--- a/crates/bitcoin-da/src/service.rs
+++ b/crates/bitcoin-da/src/service.rs
@@ -583,9 +583,8 @@ impl DaService for BitcoinService {
             "Extracting relevant txs from block {:?}",
             block.header.block_hash()
         );
-
-        let txs = block.txdata.iter().map(|tx| tx.inner().clone()).collect();
-        get_relevant_blobs_from_txs(txs, &self.reveal_batch_prover_prefix)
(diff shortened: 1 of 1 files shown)


crates/bitcoin-da/src/service.rs
```diff
@@ -583,9 +583,8 @@ impl DaService for BitcoinService {
             "Extracting relevant txs from block {:?}",
             block.header.block_hash()
         );
-
-        let txs = block.txdata.iter().map(|tx| tx.inner().clone()).collect();
-        get_relevant_blobs_from_txs(txs, &self.reveal_batch_prover_prefix)
+        let txs: Vec<&Transaction> = block.txdata.iter().map(|tx| tx.inner()).collect(); // get a list of references to the transactions
```
kpp, line 586: The trick is to get rid of .collect() if possible. And it is possible

eyusufatik: 😁
don't tell him how tho :p

kpp:
I should say I made a mistake. `get_relevant_blobs_from_txs` is not called in a zk context so there is no need to optimize this function.

PR #1098, opened 2024-09-04
Move fork module to rollup interface

rakanalh:
# Description
Citrea primitives crate should not be imported by anything sov-sdk related. This would later lead to cyclic dependencies.


eyusufatik (approved):

bin/citrea/provers/risc0/guest-bitcoin/Cargo.lock
```diff
@@ -2919,6 +2919,7 @@ dependencies = [
  "serde",
  "sha2",
  "thiserror",
+ "tracing",
```
kpp, line 2922: Wow

rakanalh: The dependency is optional and would only be used in native mode

eyusufatik: then it shouldn't have ended up in the bitcoin prover cargo.lock

crates/sovereign-sdk/rollup-interface/Cargo.toml
```diff
@@ -43,14 +44,15 @@ fuzzing = ["proptest", "proptest-derive", "sha2", "std"]
...
+  "borsh/default",
+  "borsh/bytes",
+  "bytes",
+  "digest/default",
+  "hex/default",
+  "proptest?/default",
+  "serde/default",
+  "sha2?/default",
+  "thiserror",
+  "tracing",
```
eyusufatik, line 57: only put in native and it'll be solved

eyusufatik: I mean tracing

rakanalh: i did

rakanalh: Ahh i see what you mean... it's in std so it should also be included with native

eyusufatik: no its in std which is in native. I believe we should put it explicitly in native, like tokio

PR #1100, opened 2024-09-04
Fullnode: Skip preproven commitments

### LINKED ISSUE #1075: Full nodes should not mark `preproven_commitments`
TODOs:

- [ ] In the zk circuit output the field `preproven_commitments` from the input
- [ ] In the full nodes, don't mark these indices as proven

rakanalh:
# Description

~~Merge AFTER #1098~~

## Linked Issues
- Fixes #1075


crates/fullnode/src/runner.rs
eyusufatik, line 1: you should update `StoredStateTransition` and related structs so that it also holds preproven_commitments

crates/prover/src/runner.rs
```diff
@@ -474,7 +474,7 @@ where
             sequencer_commitments.sort_unstable();
 
             let (sequencer_commitments, preproven_commitments) =
-                self.filter_out_proven_commitments(&sequencer_commitments)?;
+                filter_out_finalized_commitments(&self.ledger_db, &sequencer_commitments)?;
```
eyusufatik, line 477: this should have stayed filter out proven

eyusufatik: and on line 831 the prover should mark the ranges as proven, since it knows it already proved them.

eyusufatik: a mistake we did on not proving preproven_commitments pr :D

crates/prover/src/runner.rs
eyusufatik, line 1: line 604 through line 608 needs fixing, we break down into groups, but we don't set the correct ranges for the sequencer commitments

crates/fullnode/src/runner.rs
```diff
@@ -570,6 +574,16 @@ where
...
 
+            // Make sure all sequencer commitments are stored in ascending order.
+            sequencer_commitments.sort_unstable();
+
+            let Ok((sequencer_commitments, _)) =
+                filter_out_finalized_commitments(&self.ledger_db, &sequencer_commitments)
+            else {
+                warn!("Could not filter out finalized commitments");
+                return;
+            };
```
eyusufatik, line 585: we don't need this at all

crates/fullnode/src/runner.rs
```diff
@@ -300,6 +301,9 @@ where
             }
         };
 
+        let (proven_commitments, _) =
+            filter_out_proven_commitments(&self.ledger_db, &proven_commitments)?;
```
eyusufatik, line 305: and this should be filtered by index from zk output `preprover_commitments` of sorted list of sequencer commitments on the da slot

eyusufatik (changes requested):

crates/fullnode/src/runner.rs
```diff
@@ -290,39 +289,73 @@ where
...
+        let mut filtered_commitments = vec![];
+        for (index, sequencer_commitment) in commitments_on_da_slot.into_iter().enumerate() {
+            if let Some(exclude_index) = excluded_commitment_indices.first() {
+                if index == *exclude_index {
+                    excluded_commitment_indices.remove(0);
+                    continue;
+                }
+            }
+            filtered_commitments.push(sequencer_commitment);
+        }
```
rakanalh, line 321:
```
    commitments_on_da_slot
        .into_iter()
        .enumerate()
        .filter(|(index, _)| !excluded_commitment_indices.contains(index))
        .map(|(_, commitment)| commitment.clone())
        .collect()
```

crates/fullnode/src/runner.rs
rakanalh: Added to `StoredStateTransition` but nothing else.

bin/citrea/tests/bitcoin_e2e/tests/prover_test.rs
```diff
@@ -98,7 +106,163 @@ impl TestCase for BasicProverTest {
...
+        let fee_sat_per_vbyte = bitcoin_da_service.get_fee_rate().await.unwrap();
+        bitcoin_da_service
+            .send_transaction_with_fee_rate(
+                None,
+                DaData::SequencerCommitment(commitments.first().unwrap().clone()),
+                fee_sat_per_vbyte,
+            )
+            .await
+            .unwrap();
+
```
eyusufatik, line 237: right here, also trigger a sequencer commitment
and wait for mempool len 4

so that the prover will generate a proof for that block leaving out preproven commmitment (the one you sent)

eyusufatik (approved):

PR #1106, opened 2024-09-05
Log Fee rate with debug instead of info

kpp:
# Description


ercecan (approved):

eyusufatik:
the test this pr is breaking, I've seen it for the first time please fix @kpp

PR #1110, opened 2024-09-05
non state module fields

eyusufatik:
# Description
shared memory between hooks and module calls 

no need for pending_transactions and block_env

TODOs:
- [x] make block_env #[memory]

### DIFF
--- a/crates/evm/src/call.rs
+++ b/crates/evm/src/call.rs
@@ -147,8 +147,7 @@ impl<C: sov_modules_api::Context> Evm<C> {
         let mut cumulative_gas_used = 0;
         let mut log_index_start = 0;
 
-        if !self.pending_transactions.is_empty() {
-            let tx = self.pending_transactions.last().unwrap();
+        if let Some(tx) = self.pending_transactions.last() {
             cumulative_gas_used = tx.receipt.receipt.cumulative_gas_used;
             log_index_start = tx.receipt.log_index_start + tx.receipt.receipt.logs.len() as u64;
         }
--- a/crates/evm/src/call.rs
+++ b/crates/evm/src/call.rs
@@ -31,7 +31,7 @@ pub struct CallMessage {
 impl<C: sov_modules_api::Context> Evm<C> {
     /// Executes system events for the current block and push tx to pending_transactions.
     pub(crate) fn execute_system_events(
-        &self,
+        &mut self,
         system_events: Vec<SystemEvent>,
         l1_fee_rate: u128,
         working_set: &mut WorkingSet<C>,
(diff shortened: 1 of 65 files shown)


eyusufatik:
I suggest only reviewing [0d04e0a](https://github.com/chainwayxyz/citrea/pull/1110/commits/0d04e0a1be960e49b432185645378286504a2a78)

crates/evm/src/call.rs
```diff
@@ -80,7 +80,9 @@ impl<C: sov_modules_api::Context> Evm<C> {
 
         let mut cumulative_gas_used = 0;
         let mut log_index_start = 0;
-        if let Some(tx) = self.pending_transactions.last(working_set) {
+
+        if !self.pending_transactions.is_empty() {
```
kpp, line 84: If they are not empty we are doing smth wrong

crates/evm/src/call.rs
```diff
@@ -149,7 +150,9 @@ impl<C: sov_modules_api::Context> Evm<C> {
         let block_number = block_env.number;
         let mut cumulative_gas_used = 0;
         let mut log_index_start = 0;
-        if let Some(tx) = self.pending_transactions.last(working_set) {
+
+        if !self.pending_transactions.is_empty() {
```
kpp, line 154: if let Some(tx) = self.pending_transactions.last()

crates/evm/src/call.rs
```diff
@@ -149,7 +150,9 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+            let tx = self.pending_transactions.last().unwrap();
```
kpp, line 155:
```suggestion
        if let Some(tx) = self.pending_transactions.last() {
```

crates/evm/src/call.rs
eyusufatik: remember when you were implementing sys txs and I asked you why you did some stuff in a certain way you said maybe we'll run sys txs at the end of the blocks in the future.

:)

eyusufatik:
[b59d67d](https://github.com/chainwayxyz/citrea/pull/1110/commits/b59d67d07db28a448c99e4798a7183f348720718)
[da3bd19](https://github.com/chainwayxyz/citrea/pull/1110/commits/da3bd19315c565b2cd3dc2245bc8c5b5c83e6f39)

these two are also worty of review

crates/evm/src/call.rs
kpp: I was wrong =)

eyusufatik: I can make another pr that makes sure execute sys txs is called at the beginning. it will also save us some cycles

kpp: Yes

PR #1113, opened 2024-09-05
Esad/fix tests

eyusufatik:
# Description
we make big copies while trying to find nonce for proof inscribe txs so bumped proof wait time to 10 mins

### DIFF
--- a/bin/citrea/tests/all_tests.rs
+++ b/bin/citrea/tests/all_tests.rs
@@ -12,5 +12,5 @@ mod test_helpers;
 
 const DEFAULT_MIN_SOFT_CONFIRMATIONS_PER_COMMITMENT: u64 = 1000;
 const DEFAULT_DEPOSIT_MEMPOOL_FETCH_LIMIT: usize = 10;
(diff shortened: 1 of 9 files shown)


bin/citrea/tests/all_tests.rs
```diff
@@ -12,5 +12,5 @@ mod test_helpers;
 
 const DEFAULT_MIN_SOFT_CONFIRMATIONS_PER_COMMITMENT: u64 = 1000;
 const DEFAULT_DEPOSIT_MEMPOOL_FETCH_LIMIT: usize = 10;
-const DEFAULT_PROOF_WAIT_DURATION: u64 = 300; // 5 minutes
+const DEFAULT_PROOF_WAIT_DURATION: u64 = 600; // 10 minutes
```
rakanalh, line 15: Mostly, this looks like the fix... am i correct?

eyusufatik: yeah. we'll see on ci if this is enough

eyusufatik: for mock da 5 min was more than enough

PR #1115, opened 2024-09-05
Add a `testmempoolaccept` check before sending raw txs

### LINKED ISSUE #1083: Use `testmempoolaccept` before sending txs
https://developer.bitcoin.org/reference/rpc/testmempoolaccept.html

use the above rpc before sending any txs for the set, so that we know all will be accepted.

jfldde:
# Description

- Add a `testmempoolaccept` check before sending raw txs
- Make sure all TXs will be accepted by mempool before sending any of them

## Linked Issues
- Fixes #1083 

## Testing
- WIP, needs to find the right assertion where we make sure that either both TXs go on, or none of them


bin/citrea/tests/bitcoin_e2e/tests/mempool_accept.rs
```diff
@@ -0,0 +1,56 @@
...
+
+struct MempoolAcceptTest;
+
+#[async_trait]
+impl TestCase for MempoolAcceptTest {
+    fn bitcoin_config() -> BitcoinConfig {
+        BitcoinConfig {
+            extra_args: vec![
+                // Restrictive mempool policy
+                "-limitancestorcount=0",
```
eyusufatik, line 19: a very smart way to do this :)

bin/citrea/tests/bitcoin_e2e/tests/mempool_accept.rs
```diff
@@ -0,0 +1,56 @@
...
+        }
+        sequencer
+            .wait_for_l2_height(min_soft_conf_per_commitment, None)
+            .await;
+
+        da.generate(FINALITY_DEPTH + 1, None).await?;
+
+        // TODO find the right assertions here
+        // Should be either 2 or 0
+        // Before this PR and the addition of testmempoolaccept, first tx would go in and second would be rejected due to mempool policy set above
```
eyusufatik, line 44: my suggestion is 


first set up the mempool with normal configs

see 2 txs of seqeuncer commitment

generate that block

shutdown bitcoin node

reboot it with thise mempool policy args above

trigger a commitment again, wait 3 mins or something and ceckmempoolsize is 0

that way test knows the sequencer can send commitments, and can check it can't produce a seq com for 3 mins -- due to `testmempoolaccept` (not guaranteed because of our nonce thing, but good enough for this feature)

eyusufatik (approved):

PR #1125, opened 2024-09-07
Fix rzup

rakanalh:
# Description


kpp:
Do we need cpp? Why?

rakanalh:
cpp removed.

TBH i don't feel like rzup is usable. Yesterday the same command used was working and today it's not. 

Seems like the script download is inconsistent.
[install-scripts.tar.gz](https://github.com/user-attachments/files/16921618/install-scripts.tar.gz)

PR #1127, opened 2024-09-08
Optimize rocksdb

### LINKED ISSUE #1043: Investigate node crashes caused by OOM
Logs have been provided privately.

rakanalh:
# Description

Based on profiles generated for this code, `write_schemas` was the biggest allocator of memory in our sequencer node.

without

As apparent from the above SVG, malloc calls caused by `write_schemas` is reported to consume >90% of memory during the lifecycle of the sequencer.

With the optimizations in this PR, we gain around 10% improvement on the memory usage of the sequencer. Options set are explained in the code and linked to rocksdb documentation.

Result after:

with

## Profiling

The above SVGs were generated as follows:

1. Using https://github.com/chainwayxyz/citrea/pull/1087
2. Run bitcoin regtest
3. Run sequencer
4. `bitcoin-cli -regtest -rpcconnect=127.0.0.1 -rpcport=18443 -rpcuser=citrea -rpcpassword=citrea -generate 10`
5. Deploy [Multicall3](https://github.com/mds1/multicall/blob/main/src/Multicall3.sol#L195) contract: `forge create --rpc-url localhost:12345 --private-key MY_PRIVATE_KEY src/Multicall3.sol:Multicall3`
6. Generate 500 accounts & fund them 
```sh
cast wallet new -n 500 -j accounts.json wallets
```

```sh
items=$(cat wallets/all.json | jq -c -r '.[].address')
for item in ${items[@]}; do
    echo $item
    cast send --rpc-url http://localhost:12345 --private-key MY_PRIVATE_KEY --value 0.1ether $item
done
```

7. Call `aggregate` in `Multicall3`: `forge script --rpc-url http://localhost:12345 --private-key MY_PRIVATE_KEY --broadcast script/Multicall3.s.sol:Multicall3Script`
```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity 0.8.12;

import {Script, console} from "forge-std/Script.sol";
import {Multicall3} from "../src/Multicall3.sol";

contract Multicall3Script is Script {
    Multicall3 public multicall;

    address[] addresses = [
       .. 500 accounts
    ];

    function setUp() public {}

    function run() public {
        vm.startBroadcast();

        address multicallAddress = 0x5FbDB2315678afecb367f032d93F642f64180aa3;
        multicall = Multicall3(multicallAddress);

        Multicall3.Call[] memory calls = new Multicall3.Call[](500);
        for (uint256 i = 0; i < 500; i++) {
            address current_address = addresses[i];
            bytes memory balancePayload = abi.encodeWithSignature("getEthBalance(address)", current_address);
            calls[i] = Multicall3.Call({
                target: multicallAddress,
                callData: balancePayload
            });

        }

        vm.stopBroadcast();
    }
}

```
8. Download profile from sequencer: `curl localhost:3000/debug/pprof/heap > heap.pb.gz`
9. Generate SVG: `go tool pprof -svg heap.pb.gz > heap.svg`

## Linked Issues
- Fixes #1043


crates/sovereign-sdk/full-node/db/sov-db/src/rocks_db_config.rs
```diff
@@ -44,10 +44,33 @@ impl<'a> RocksdbConfig<'a> {
     /// Build [`rocksdb::Options`] from [`RocksdbConfig`]
     pub fn as_rocksdb_options(&self, readonly: bool) -> Options {
         let mut db_opts = Options::default();
+
+        let mut block_based_options = BlockBasedOptions::default();
+        /*
+         * The following settings are recommended in:
+         * https://github.com/facebook/rocksdb/wiki/memory-usage-in-rocksdb
+         */
+        // Since our database is write heavy, disable cache to decrease memory footprint
```
eyusufatik, line 53: are you sure about this?

when we get more usage (not like on our tests where we produce mostly empty blocks) reads will also be very common, especially when we start to get a lot of transactions

rakanalh: Maybe i mis-worded the comment. But in general, disabling cache would remove the possibility of OOM errors.

Instead of doing that, i have set an LRU cache to a larger value (100 MB) rather than the default (8 MB). However, this would not remove the possibility of OOM errors.

eyusufatik (approved):

PR #1128, opened 2024-09-08
Fullnode: bypass wallet check for bitcoin DA

### LINKED ISSUE #1089: Full node doesn't need a bitcoin wallet
on startup, for full nodes, we require the node to have a wallet loaded.

full nodes doesn't need a wallet on their bitcoin nodes. only the sequencer and the prover does.

rakanalh:
# Description

## Linked Issues
- Fixes #1089

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -82,10 +82,13 @@ jobs:
           components: rustfmt, clippy
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2
-      - name: Install risc0-zkvm toolchain # Use the risc0 cargo extension to install the risc0 std library for the current toolchain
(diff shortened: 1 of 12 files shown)


eyusufatik:
what about instead of adding a param to the rollup config for this, putting a new arg to bitcoin da new function?

sequencer and prover always needs the wallet but the full node never needs it. so true on sequencer and prover false on full node?

rakanalh:
> what about instead of adding a param to the rollup config for this, putting a new arg to bitcoin da new function?
> 
> sequencer and prover always needs the wallet but the full node never needs it. so true on sequencer and prover false on full node?

You're right... this was better. Done.

eyusufatik (approved):

PR #1138, opened 2024-09-09
E2E Add support for env

### LINKED ISSUE #1133: E2E Output to log files without colour

### LINKED ISSUE #1134: E2E Support different env per test case and/or binary

jfldde:
# Description

- Add support for env per test case and/or per binary
- Add a `base_env` method where env variables that should always be on must be added (In this PR case, `NO_COLOR=1` to fix #1133)
- Add a `test` env field which will be added to every binary's env.
- Add a env field per binary that will be appeneded to these two previous ones.
- Full env used to run the binary will be -> [base_env + test_env + binary's env]
- Config cleanup

- Usage :

```
    fn test_env() -> TestCaseEnv {
        TestCaseEnv {
            sequencer: vec![("RUST_LOG", "trace")],
            prover: vec![("JSON_LOGS", "true")],
            ..Default::default()
        }
    }
```

## Linked Issues
- Fixes #1134
- Fixes #1133

### DIFF
--- a/bin/citrea/tests/bitcoin_e2e/tests/sequencer_test.rs
+++ b/bin/citrea/tests/bitcoin_e2e/tests/sequencer_test.rs
@@ -2,7 +2,6 @@ use anyhow::bail;
 use async_trait::async_trait;
 use bitcoincore_rpc::RpcApi;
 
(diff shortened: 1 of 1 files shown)


rakanalh (approved):
LGTM

PR #1141, opened 2024-09-09
check if soft confirmation heights are sequential in zk

### LINKED ISSUE #1136: Check L2 height in zk
we check soft confirmation orders (prev hash holds for soft conf n-1 and n), da slot height and hashes but not the height. We should 

Fix is very simple

eyusufatik:
# Description

## Linked Issues
closes #1136 
## Testing

## Docs

### DIFF
--- a/crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
+++ b/crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
@@ -748,6 +748,12 @@ where
                     da_block_header = da_block_headers_iter.next().unwrap();
                 }
 
+                assert_eq!(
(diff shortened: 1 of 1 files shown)


crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
```diff
@@ -748,6 +748,12 @@ where
                     da_block_header = da_block_headers_iter.next().unwrap();
                 }
 
+                assert_eq!(
+                    soft_confirmation.l2_height(),
```
eyusufatik, line 752: seqeuncer commitment start and end are also checked because of the way I added these. see lines above and below

rakanalh (approved):
LGTM

PR #1146, opened 2024-09-10
Add make clean-all target

jfldde:
# Description

- Adds make clean-all target
- Adds clean-txs target


Makefile
```diff
@@ -23,6 +23,11 @@ clean-node: ## Cleans local dbs needed for sequencer and nodes
 	rm -rf resources/dbs/full-node-db
 	rm -rf resources/dbs/citrea-bitcoin-regtest-data
```
rakanalh, line 24: Maybe introduce `clean-docker` and execute this line... because this one requires `sudo`. Don't include it as a dependency for `clean-all` though

rakanalh (approved):

PR #1153, opened 2024-09-10
Add cycle-diff script and perf CI

jfldde:
# Description

- Add `cycle-diff.sh` used to catch any change in cycles between working branch and nightly
- Add a CI job as `workflow_dispatch` to run this script on any PR
- Currently uses `basic_prover` test but can be configured to use a more significant test with the `TEST_NAME` env variable 
- Add new `resources/scripts` directory
- Averages the run over NUM_RUNS

Usage :
```
./cycle-diff.sh generate
./cycle-diff.sh check
```

### DIFF
--- a/resources/scripts/cycle-diff.sh
+++ b/resources/scripts/cycle-diff.sh
@@ -153,7 +153,7 @@ check_regression() {
 
 check_performance_regression() {
     local failed=false
-    if ! check_regression "Execution Time"; then failed=true; fi
+    # if ! check_regression "Execution Time"; then failed=true; fi
     if ! check_regression "Number of Segments"; then failed=true; fi
     if ! check_regression "Total Cycles"; then failed=true; fi
     if ! check_regression "User Cycles"; then failed=true; fi


resources/scripts/cycle-diff.sh
```diff
@@ -0,0 +1,166 @@
...
+
+# Used to silence custom risc0 build.rs output
+export RISC0_GUEST_LOGFILE=$(mktemp)
+
+run_test_and_extract() {
+    local command="cargo test $TEST_NAME -- --nocapture"
+    local output_file="output.log"
+
+    # Aggressively silence build output
+    script -q /dev/null make build > /dev/null 2>&1
```
kpp, line 52: `make build` will build with profile `dev`, then `cargo test` will rebuild it with profile `test`.

resources/scripts/cycle-diff.sh
```diff
@@ -0,0 +1,166 @@
...
+BASE_BRANCH=${BASE_BRANCH:-"nightly"}
+TEST_NAME=${TEST_NAME:-"bitcoin_e2e::tests::prover_test::basic_prover_test"}
+TARGET_PCT=${TARGET_PCT:-3}
+COMPARISON_FILE=${COMPARISON_FILE:-"comparison_results.log"}
+
+# Used to silence custom risc0 build.rs output
+export RISC0_GUEST_LOGFILE=$(mktemp)
+
+run_test_and_extract() {
+    local command="cargo test $TEST_NAME -- --nocapture"
```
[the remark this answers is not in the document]
kpp: Oh, right. You expect it to run an E2E test, so `-p citrea` is actually would be better.

resources/scripts/cycle-diff.sh
rakanalh: Why not introduce `make release` which builds with `--release` since we want to optimize as much as possible for the cycle counts?

kpp: guest code is built with a separate build.rs script which builds it in `--release` anyway.

kpp (approved):
Very good. Thanks!

PR #1160, opened 2024-09-11
full node fix l1 start height

eyusufatik:
# Description
full node would start scanning from first l1 always

### DIFF
--- a/crates/fullnode/src/runner.rs
+++ b/crates/fullnode/src/runner.rs
@@ -52,7 +52,6 @@ where
     DB: NodeLedgerOps,
 {
     start_l2_height: u64,
-    start_l1_height: u64,
     da_service: Arc<Da>,
     stf: Stf,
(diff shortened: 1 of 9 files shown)


crates/fullnode/src/runner.rs
```diff
@@ -802,3 +820,16 @@ async fn sync_l2<Da>(
         }
     }
 }
+
+async fn get_initial_slot_height<Da: DaSpec>(client: &SequencerClient) -> u64 {
+    loop {
+        match client.get_soft_confirmation::<Da>(1).await {
+            Ok(Some(batch)) => return batch.da_slot_height,
```
ercecan, line 827:
```suggestion
            Ok(Some(soft_confirmation)) => return soft_confirmation.da_slot_height,
```

ercecan (approved):

PR #1171, opened 2024-09-12
Enable prover simulation

### LINKED ISSUE #1170: Re-enable prover simulation mode

rakanalh:
# Description

## Linked Issues
- Fixes #1170

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1865,6 +1865,7 @@ dependencies = [
  "hex",
  "jsonrpsee",
  "num_cpus",
+ "parking_lot",
  "rand 0.8.5",
  "rayon",
  "rs_merkle",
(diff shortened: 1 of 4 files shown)


rakanalh:
> Couldn't we keep prover_state behind a mutex instead ? I'm not seeing any reads. I also see a nested `prover_state` naming. Could we rename `prover_state: Prover<StateRoot, Witness, Da>,` to `prover: Prover<StateRoot, Witness, Da>,` and rename inner Arc> to state directly. A bit out of scope for this PR though, just opening the discussion

For `prover_state` i didn't change anything apart from using `parking_lot` which doesn't poison.
But you're probably right. Might be better off using a Mutex for that.

eyusufatik:
@jfldde agree with jeremy. not in this pr but we should seriously reconsider rewriting the whole prover service tbh.

eyusufatik (commented):
LGTM. again merging after checking bitcoin da ids

PR #1172, opened 2024-09-12
use block id for some evm queries

eyusufatik:
# Description
didnt test yet.

we also need to put this in nightly somehow


crates/evm/src/tests/queries/evm_call_tests.rs
```diff
@@ -96,7 +97,7 @@ fn test_state_change() {
             value: Some(U256::from(123134235)),
             ..Default::default()
         },
-        Some(BlockNumberOrTag::Latest),
+        Some(BlockId::Number(BlockNumberOrTag::Latest)),
```
rakanalh, line 100: This would be `None` in this case, no?

crates/evm/src/query.rs
```diff
@@ -615,37 +513,67 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+                            working_set,
+                        )
+                        .unwrap()
+                        .expect("Block must be set"),
+                )
+            }
+            None => self.block_env.get(working_set).unwrap_or_else(|| {
+                BlockEnv::from(
+                    &self
+                        .get_sealed_block_by_number(Some(BlockNumberOrTag::Earliest), working_set)
```
rakanalh, line 572:
```suggestion
                        .get_sealed_block_by_number(Some(BlockNumberOrTag::Latest), working_set)
```
Did you this?

rakanalh (dismissed):

ercecan (approved):

crates/evm/src/query.rs
ercecan: If block Id is None we should return latest but in this case since there is no block is produced we return the genesis block env, the same logic in lines 527-535
(If the working set exists it is already in the latest state so we don't need to do anything else)

crates/evm/src/query.rs
```diff
@@ -615,37 +513,67 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+
+                BlockEnv::from(
+                    &self
+                        .get_sealed_block_by_number(
+                            Some(BlockNumberOrTag::Number(block_number)),
+                            working_set,
+                        )
+                        .unwrap()
+                        .expect("Block must be set"),
+                )
```
eyusufatik, line 567: I should consider getting the block env before setting state version

crates/evm/src/tests/queries/evm_call_tests.rs
eyusufatik: Yeah None is interpreted the same as latest in most queries

Issue #15, opened 2023-11-30
eth_sendRawTransaction doesn't check the TX


eyusufatik:
We should check the incoming transactions for correct nonce and signature at the RPC level and return an error accordingly.

Issue #12, opened 2023-11-28
Refactor test cases


eyusufatik:
For instance, `sov-evm/tests/call_tests.rs` has many different test cases that start exactly the same way. 

Below is nearly in every test in `call_tests.rs`
```rust
let signer_balance: u64 = 10000000000000000;

let dev_signer: TestSigner = TestSigner::new_random();

let contract_addr: Address = Address::from_slice(
    hex::decode("819c5497b157177315e1204f52e588b393771719")
        .unwrap()
        .as_slice(),
);
let config = EvmConfig {
    data: vec![AccountData {
        address: dev_signer.address(),
        balance: U256::from(signer_balance),
        code_hash: KECCAK_EMPTY,
        code: Bytes::default(),
        nonce: 0,
    }],
    spec: vec![(0, SpecId::SHANGHAI)].into_iter().collect(),
    ..Default::default()
};

let (evm, mut working_set) = get_evm(&config);

......
```

Let's detect cases like these and merge them in a function called `create_test_date()` or something like that.

PR #1180, opened 2024-09-13
Don't insert duplicate account codes

kpp:
# Description

Check if account.code(hash) exists before inserts

### DIFF
--- a/crates/evm/src/evm/db_commit.rs
+++ b/crates/evm/src/evm/db_commit.rs
@@ -51,8 +51,8 @@ impl<'a, C: sov_modules_api::Context> DatabaseCommit for EvmDb<'a, C> {
             let account_info = account.info;
 
             if let Some(ref code) = account_info.code {
-                if !code.is_empty() && !self.code.exists(&account_info.code_hash, self.working_set)
-                {
+                if !code.is_empty() {
+                    // TODO: would be good to have a contains_key method on the StateMap that would be optimized, so we can check the hash before storing the code
                     self.code
                         .set(&account_info.code_hash, code, self.working_set);
                 }
--- a/crates/evm/src/evm/db_commit.rs
+++ b/crates/evm/src/evm/db_commit.rs
@@ -51,8 +51,8 @@ impl<'a, C: sov_modules_api::Context> DatabaseCommit for EvmDb<'a, C> {
             let account_info = account.info;
 
             if let Some(ref code) = account_info.code {
-                if !code.is_empty() {
-                    // TODO: would be good to have a contains_key method on the StateMap that would be optimized, so we can check the hash before storing the code
+                if !code.is_empty() && !self.code.exists(&account_info.code_hash, self.working_set)
+                {
                     self.code
                         .set(&account_info.code_hash, code, self.working_set);
                 }
--- a/crates/evm/src/evm/db_commit.rs
+++ b/crates/evm/src/evm/db_commit.rs
@@ -52,9 +52,14 @@ impl<'a, C: sov_modules_api::Context> DatabaseCommit for EvmDb<'a, C> {
 
             if let Some(ref code) = account_info.code {
                 if !code.is_empty() {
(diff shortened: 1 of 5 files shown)


crates/evm/src/evm/handler.rs
```diff
@@ -557,9 +558,13 @@ fn calc_diff_size<EXT, DB: Database>(
             let account = &state[addr];
 
             if let Some(code) = account.info.code.as_ref() {
-                // if code is eoa code
-                diff_size += CODE_KEY_SIZE;
-                diff_size += code.len();
+                // Don't charge for account code if it is already in DB.
+                let db_code = db.code_by_hash(account.info.code_hash)?;
```
eyusufatik, line 562: db is revm's internal cache. we have to check if !self.code.exists(code_hash, working_set)

kpp: We don't have an access to self.code here

eyusufatik: wait sorry, I got confused with journaled state. this is correct

crates/sovereign-sdk/module-system/sov-modules-core/src/storage/scratchpad.rs
```diff
@@ -27,6 +27,9 @@ pub trait StateReaderAndWriter {
     /// Deletes a storage value.
     fn delete(&mut self, key: &StorageKey);
 
+    /// Checks a storage key for existance
```
eyusufatik, line 30:
```suggestion
    /// Checks a storage key for existence
```

crates/sovereign-sdk/module-system/sov-modules-core/src/storage/cache.rs
```diff
@@ -419,6 +419,29 @@ impl StorageInternalCache {
         self.tx_cache.add_write(cache_key, None);
     }
 
+    /// Check a key for existance.
```
eyusufatik, line 422:
```suggestion
    /// Check a key for existence.
```

eyusufatik (approved):

crates/evm/src/evm/handler.rs
kpp, line 562: Unfortunately there is no way to optimize this, we have to deser the code anyway. So there is no use in those .exists/ contains_key methods at all...

eyusufatik: look, selfdestruct is very very rarely used. so it's not a big deal

crates/evm/src/evm/db_commit.rs
```diff
@@ -51,8 +51,8 @@ impl<'a, C: sov_modules_api::Context> DatabaseCommit for EvmDb<'a, C> {
             let account_info = account.info;
 
             if let Some(ref code) = account_info.code {
-                if !code.is_empty() && !self.code.exists(&account_info.code_hash, self.working_set)
-                {
+                if !code.is_empty() {
+                    // TODO: would be good to have a contains_key method on the StateMap that would be optimized, so we can check the hash before storing the code
                     self.code
                         .set(&account_info.code_hash, code, self.working_set);
```
eyusufatik: I wonder if we set the code state map to the same code again in a batch proof, do we include that in the state diff?

kpp: No, it's gonna be only 1, because we keep only the last (k, v) pair.

eyusufatik: no but imagine there is two different proofs 

proof 1: sequencer commitmenti in L1 #12
proof 2: sequencer commitment in L1 #15 

in proof one I deploy contract x 

state diff will have the code (as it should be)

in proof two I also deploy contract x
the value in the state tree does not change but will it be in the state diff????

eyusufatik: it's ok. sorry for the misunderstanding. we have the if check now

PR #1192, opened 2024-09-17
Add l1 status to citrea_syncStatus

### LINKED ISSUE #1186: Sync status should also show l1 sync status
Currently we have an rpc call `citrea_syncStatus` which shows l2 syncing status of full nodes to sequencer.

It would be also nice to see the status of l1 sync.

- [ ] In the rpc find a way to access which l1 block the node last synced
- [ ] Update the `CitreaStatus` in a way that includes l1 sync status
- [ ] Update rpc endpoint 
- [ ] Update tests using the rpc endpoint and test l1 sync feature

exeokan:
# Description

Update the citrea_syncStatus rpc to also show l1 sync status.

## Linked Issues
- Fixes #1186

### DIFF
--- a/crates/ethereum-rpc/src/lib.rs
+++ b/crates/ethereum-rpc/src/lib.rs
@@ -630,29 +630,26 @@ fn register_rpc_methods<C: sov_modules_api::Context, Da: DaService>(
                 };
 
                 // get l2 synced block number
-                let evm = Evm::<C>::default();
-                let mut working_set = WorkingSet::<C>::new(ethereum.storage.clone());
 
-                let block =
-                    evm.get_block_by_number(Some(BlockNumberOrTag::Latest), None, &mut working_set);
+                let head_soft_confirmation = ethereum.ledger_db.get_head_soft_confirmation();
 
-                let l2_synced_block_number = match block {
-                    Ok(Some(block)) => block.header.number.unwrap(),
+                let l2_synced_block_number = match head_soft_confirmation {
+                    Ok(Some((height, _))) => height.0,
                     Ok(None) => 0u64,
-                    Err(e) => return Err(e),
+                    Err(e) => return Err(to_jsonrpsee_error_object("LEDGER_DB_ERROR", e)),
                 };
 
                 // handle da service response
                 let l1_head_block_number = match da_response {
                     Ok(header) => header.height(),
-                    Err(e) => return Err(to_jsonrpsee_error_object("DA_ERROR", e)),
+                    Err(e) => return Err(to_jsonrpsee_error_object("DA_SERVICE_ERROR", e)),
                 };
 
                 // get l1 synced block number
                 let l1_synced_block_number = match ethereum.ledger_db.get_last_scanned_l1_height() {
                     Ok(Some(slot_number)) => slot_number.0,
                     Ok(None) => 0u64,
-                    Err(e) => return Err(to_jsonrpsee_error_object("PROVER_ERROR", e)),
+                    Err(e) => return Err(to_jsonrpsee_error_object("LEDGER_DB_ERROR", e)),
                 };
 
                 let l1_status = if l1_synced_block_number < l1_head_block_number {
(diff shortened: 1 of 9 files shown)


crates/ethereum-rpc/src/lib.rs
```diff
@@ -627,20 +635,41 @@ fn register_rpc_methods<C: sov_modules_api::Context, Da: DaService>(
...
-                    Ok::<CitreaStatus, ErrorObjectOwned>(CitreaStatus::Syncing(SyncStatus {
-                        synced_block_number,
-                        head_block_number,
-                    }))
+                let l1_synced_block_number = match ethereum.ledger_db.get_last_scanned_l1_height() {
+                    Ok(Some(slot_number)) => slot_number.0,
+                    Ok(None) => 0,
+                    Err(e) => return Err(to_jsonrpsee_error_object("PROVER_ERROR", e)),
+                };
+                let l1_head_block_number = match ethereum.da_service.get_head_block_header().await {
```
eyusufatik, line 649: i don't think this is updated in ledger db

crates/ethereum-rpc/src/lib.rs
```diff
@@ -627,20 +635,41 @@ fn register_rpc_methods<C: sov_modules_api::Context, Da: DaService>(
...
                     Ok(None) => 0u64,
                     Err(e) => return Err(e),
                 };
 
-                if synced_block_number < head_block_number {
-                    Ok::<CitreaStatus, ErrorObjectOwned>(CitreaStatus::Syncing(SyncStatus {
-                        synced_block_number,
-                        head_block_number,
-                    }))
+                let l1_synced_block_number = match ethereum.ledger_db.get_last_scanned_l1_height() {
```
eyusufatik, line 644: for the sequencer this table is never updated. however we were already considering disabling healthcheck endpoint for sequencer. so do it

crates/ethereum-rpc/src/lib.rs
eyusufatik, line 649: use last finalized height

crates/ethereum-rpc/src/lib.rs
```diff
@@ -613,7 +621,7 @@ fn register_rpc_methods<C: sov_modules_api::Context, Da: DaService>(
                     .block_number()
                     .await;
```
eyusufatik, line 623: we have two requests in this function now.

1. to the sequencer
2. to the da layer.

we can do both async, and await on them, lowering time needed to execute this function

crates/ethereum-rpc/src/lib.rs
```diff
@@ -33,18 +35,23 @@ pub struct SyncStatus {
...
 #[derive(Clone, Debug, serde::Serialize, serde::Deserialize, PartialEq)]
-#[serde(rename_all = "camelCase")]
-pub enum CitreaStatus {
+pub enum LayerStatus {
     Synced(u64),
     Syncing(SyncStatus),
 }
+#[derive(Clone, Debug, serde::Serialize, serde::Deserialize, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct CitreaStatus {
```
eyusufatik, line 45: Can you rename this to SyncStatus?

eyusufatik (changes requested):

crates/ethereum-rpc/src/lib.rs
eyusufatik: @kpp @rakanalh in tokio, what's the equivalent of JavaScript's Promise.all ?

rakanalh: https://docs.rs/futures/0.3.5/futures/future/fn.join_all.html

rakanalh: or:
https://docs.rs/tokio/latest/tokio/macro.join.html

exeokan: done but I cloned ethereum.sequencer_client, if it is bad practice I can find a workaround

crates/ethereum-rpc/src/lib.rs
eyusufatik: I got confused about the healthcheck. this is a different rpc method. 

still sending this to the sequencer will be problematic

eyusufatik: and this is already only registered if not sequencer. cool

eyusufatik (approved):

PR #1197, opened 2024-09-18
use tick interval in sequencer

eyusufatik:
# Description

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/sequencer/src/sequencer.rs
```diff
@@ -866,16 +866,15 @@ where
...
-        loop {
-            let block_production_tick = tokio::time::sleep(
-                target_block_time
-                    .checked_sub(parent_block_exec_time)
-                    .unwrap_or_default(),
-            );
+        let mut block_production_tick = tokio::time::interval(
+            target_block_time
+                .checked_sub(parent_block_exec_time)
+                .unwrap_or_default(),
```
rakanalh, line 872: This is not going to work. Either keep the tick inside the loop to account for the parent block's execution time or set a constant value here.

rakanalh (changes requested):

crates/sequencer/src/sequencer.rs
eyusufatik: Why?

rakanalh: Because outside the loop, `target_block_time` is always going to be 2 seconds. In this case, we should remove `checked_sub` because it doesn't make sense.
Inside the loop, `checked_sub` is going to be useful since `parent_block_exec_time` is going to contain a value larger than 0.

crates/sequencer/src/sequencer.rs
```diff
@@ -977,6 +976,13 @@ where
                             // achieve consistent 2-second block production.
                             parent_block_exec_time = instant.elapsed();
 
+                            block_production_tick = tokio::time::interval(
+                                target_block_time
+                                    .checked_sub(parent_block_exec_time)
+                                    .unwrap_or_default(),
+                            );
+                            block_production_tick.tick().await;
```
rakanalh, line 984: I am curious as to why we would do this here?

crates/sequencer/src/sequencer.rs
eyusufatik: Ah yeah, outside the loop we don’t need checked sub + parent exec time is 0

eyusufatik: I mean at that point

rakanalh (approved):

PR #1201, opened 2024-09-18
Process missed da blocks on start

### LINKED ISSUE #1199: Missed DA blocks might not get L2 blocks created for them
Assuming our DA layer produces 4 blocks: 4,5,6,7.

The sequencer produced L2 blocks for DA block 4 and was stopped. When it restarted, L1 block 7 was produced.
This means that the sequencer did not produce empty blocks for L1 5,6.

We do have code in the sequencer to create empty L2 blocks for missed L1 blocks:
https://github.com/chainwayxyz/citrea/blob/a93b60e4d7229c887720866d22a96cf9ec6d0e60/crates/sequencer/src/sequencer.rs#L946-L962

However, this code is dependent on the `da_monitor` task receiving the fact that we've seen L1 block 7 being created so that it can mark `missed_da_blocks_count` to the number of missed L1 blocks. If that doesn't happen first, we will think that we haven't missed any blocks and produce an L2 block for L1 block 7.

rakanalh:
# Description

## Linked Issues
- Fixes #1199


bin/citrea/tests/e2e/sequencer_behaviour.rs
```diff
@@ -577,10 +570,41 @@ async fn test_sequencer_fills_empty_blocks_for_missed_da_blocks() -> Result<(),
...
+        .ledger_get_head_soft_confirmation_height()
+        .await
+        .unwrap()
+        .unwrap();
+
+    // check that the sequencer has at least one block for each DA block
+    // starting from DA #3 all the way up to DA #13 without no gaps
+    // the first soft confirmation should be on DA #3
+    // the last soft confirmation should be on DA #13
+    for i in 1..=head_soft_confirmation_height {
```
rakanalh, line 589: How about start from 2? You won't need the if statement below.

eyusufatik: I wanted to guarantee sequencer started block production from the last finalized L1 height it first saw on startup.

so if we break it in the future this test will detect it 

like a sanity check let's say

ercecan (approved):

PR #1206, opened 2024-09-19
Use SHORT_PREFIX env in CI

jfldde:
- Uses SHORT_PREFIX env to reduce constant prefixes length.
- Cleanup prefix usage across
- Reduces nonce finding difficulty, speeds up bitcoin_e2e tests and addresses related flakiness.


eyusufatik:
so this change makes sure those prefixes are still a constant, however they are decided on compile time, right?

eyusufatik:
also if you do build with no short prefix, you'll get the same da id probably

eyusufatik:
no bitcoin e2e is flaky it seems. perfect

eyusufatik (approved):

eyusufatik (approved):

PR #1212, opened 2024-09-20
Check compressed state diff size for generating proofs

### LINKED ISSUE #1211: Check compressed state diff size for proof generation
We are not checking compressed state diff size for proof generation, so instead of proving n commitments at the same we prove n-x commitments.

ercecan:
# Description
Fixes the problem where we do not check the compressed state diff before generating proof for commitments

## Linked Issues
- Fixes #1211


kpp (approved):

crates/prover/src/da_block_handler.rs
```diff
@@ -347,10 +348,10 @@ where
                 sequencer_commitment_state_diff.clone(),
             );
 
-            let serialized_state_diff = borsh::to_vec(&cumulative_state_diff)?;
+            let compressed_state_diff = compress_blob(&borsh::to_vec(&cumulative_state_diff)?);
```
rakanalh, line 351: Maybe also explain why we want to check the compressed version in a comment rather than uncompressed one,

crates/primitives/src/constants.rs
```diff
@@ -29,6 +29,6 @@ pub const TEST_PRIVATE_KEY: &str =
     "1212121212121212121212121212121212121212121212121212121212121212";
 
 pub const MAX_STATEDIFF_SIZE_COMMITMENT_THRESHOLD: u64 = 300 * 1024;
-pub const MAX_STATEDIFF_SIZE_PROOF_THRESHOLD: u64 = 400 * 1024;
+pub const MAX_STATEDIFF_SIZE_PROOF_THRESHOLD: u64 = 397000;
```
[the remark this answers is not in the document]
ercecan: Not sure how It is calculated but I took it from bitcoin-da MAX_TXBODY_SIZE
@eyusufatik told me to but it makes sense since there will be other parts of the tx so it should be smaller than 400kb for sure

crates/prover/src/da_block_handler.rs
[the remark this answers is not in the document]
ercecan: We should not do it on the same thread of the sequencer loop as brotli will take too much time to compress data every time we produce a block and might slow down sequencing
We can do this in a separate thread or sth. I will open an issue

crates/primitives/src/constants.rs
kpp, line 32: Can we use the constant from bitcoin-da instead?

ercecan: Sure, I'll make it pub and use it

eyusufatik: can we put this to citrea-primitives so that both the prover and the bitcoin-da can pull it from there?

citrea prover importing something from bitcoin-da does not really make sense

eyusufatik (approved):

crates/prover/Cargo.toml
```diff
@@ -10,6 +10,7 @@ repository.workspace = true
 
 [dependencies]
 # Citrea Deps
+bitcoin-da = { path = "../bitcoin-da" }
```
eyusufatik, line 13: remove

ercecan: that's there because of compress blob func

crates/prover/src/da_block_handler.rs
```diff
@@ -347,10 +348,11 @@ where
...
 
+            // Threshold is checked by comparing compressed state diff size as the data will be compressed before it is written on DA
             let state_diff_threshold_reached =
-                serialized_state_diff.len() as u64 > MAX_STATEDIFF_SIZE_PROOF_THRESHOLD;
+                compressed_state_diff.len() as u64 > MAX_TXBODY_SIZE.try_into().unwrap();
```
eyusufatik, line 355: use saturating_to

PR #1214, opened 2024-09-22
Structured concurrency: Part 1

### LINKED ISSUE #923: Research replacing `tokio::spawn`s with a task manager
We are using `tokio::spawn` to spin up threads all over the codebase. It is hard to track these in case of failure. It could be a good idea to create some sort of `TaskManager` that handles asynchronous works that's going on. It would also enable potentially restarting some critical parts of the code if something goes wrong.

**Achievable**
Research whether this would be truly useful, and try to find some use cases for this. If the conclusion is that this is useful, implement it.

rakanalh:
# Description

From [task_scope's documentation](https://docs.rs/task_scope/latest/task_scope/)

> What is Strucutured Concurrency?

> Structured Concurrency is a programming paradigm that lets asynchronous operations run only within certain scopes so that they form an operation stack like a regular function call stack. As the parent operation waits until all children complete, Structured Concurrency helps local reasoning of concurrent programs.
[Out-of-task concurrency considered harmful](https://docs.rs/task_scope/latest/task_scope/#out-of-task-concurrency-considered-harmful)

> Most of the asynchronous programs are composed of async functions and in-task concurrency primitives such as select and join, and this makes such futures automatically well-structured. As "futures do nothing unless polled," executions of inner (child) operations are very explicit (usually at the point of awaits). Moreover, canceling a future is done by dropping it, which will reclaim resources used for the operation, including the inner futures. This drop-chain propagates the cancellation to the very end of the operation stack.

> Out-of-task concurrencies such as spawn, however, breaks the structure. They allow us to start a new execution unit that can escape from the parent stack. Although frameworks provide a way to join on the spawned task, they don't propagate cancellation properly. If you drop the spawning task, the spawned task may outlive the parent indefinitely.

# Changes in this PR

This PR restructures the ethereum-rpc crate `SubscriptionManager` so that the list of subscriptions is preserved for as long as they are active, while processing new head and new logs notifications in a single worker which notifies all subscribers about new heads and new logs. That is, replacing the functionality of spawning an unmanaged task per subscription.

The PR also keeps track of the handles for the 3 spawned tasks:
1. `new_heads_notifier`
2. `new_logs_notifier`
3. `soft_confirmation_event_handler`

So that when `SubscriptionManager` is dropped, the tasks are also stopped. This could be improved in the future where these components can react the OS signals / cancellation tokens.

# Initiative

Based on this initiative, i'd like to introduce a no `tokio::spawn` rule without storing the handle somewhere, where the spawned task is managed and is cancellable.

## Linked Issues
- Related #923


.github/workflows/checks.yml
```diff
@@ -76,7 +76,7 @@ jobs:
           version: "23.2"
           repo-token: ${{ secrets.GITHUB_TOKEN }}
       - name: Toolchain
-        uses: dtolnay/rust-toolchain@nightly
+        uses: dtolnay/rust-toolchain@nightly-2024-06-13 # 1.79.0 nightly
```
rakanalh, line 79: This maps our nightly to the same stable version we use for our release builds. This is because using `@nightly` updated to 1.83 which changes in the format rules resulting in lint failing.

eyusufatik: lint should be happening on nightly no?

eyusufatik: a138859f6996ba6de5f1a23bbaa8d26fea3cd88d fixed with this commit

crates/ethereum-rpc/src/subscription.rs
```diff
@@ -1,121 +1,116 @@
...
 
+pub async fn new_heads_notifier(
+    mut rx: mpsc::Receiver<RichBlock>,
+    head_subscriptions: Arc<RwLock<Vec<SubscriptionSink>>>,
+) {
+    while let Some(block) = rx.recv().await {
+        // Acquire the read lock here to prevent starving the writes.
+        let subscriptions = head_subscriptions.read().await;
+        let mut send_tasks = vec![];
+        for subscription in subscriptions.iter() {
```
kpp, line 86: This is literally the job of broadcast.

kpp (approved):
That's a nice cleanup. The only thing I question is why you decided to move from broadcast to mpsc.

rakanalh:
> That's a nice cleanup. The only thing I question is why you decided to move from broadcast to mpsc.

broadcast is meant for multiple receivers. In this case, the node (for example sequencer) would send a new soft confirmation event to one receiver which is the `new_heads_worker` and the worker would notify all subscriptions.

crates/ethereum-rpc/src/subscription.rs
```diff
@@ -1,121 +1,116 @@
...
             )
             .unwrap();
-            let Ok(_) = subscription.send(msg).await else {
-                // Connection closed
-                return;
-            };
+            send_tasks.push(subscription.send(msg));
         }
-    });
+        let _ = future::join_all(send_tasks).await;
```
kpp, line 95: Use broadcast

kpp (changes requested):
We may use `for ... { x.send(y) }` in log subscription but new_heads looks better with broadcast.

crates/ethereum-rpc/src/subscription.rs
rakanalh: A subscription has type `SubscriptionSink` it's not a channel.

crates/ethereum-rpc/src/subscription.rs
```diff
@@ -125,12 +120,43 @@ pub async fn handle_logs_subscription(
...
-                        return;
-                    };
+                    send_tasks.push(subscription.send(msg));
                 }
             }
         }
-    });
+        let _ = future::join_all(send_tasks).await;
+        // Drop lock to release the read lock.
+        drop(subscriptions);
```
yaziciahmet, line 129: I am not sure if this is necessary. I think subscriptions is dropped after each iteration automatically.

crates/ethereum-rpc/src/subscription.rs
```diff
@@ -1,121 +1,116 @@
...
-            let Ok(_) = subscription.send(msg).await else {
-                // Connection closed
-                return;
-            };
+            send_tasks.push(subscription.send(msg));
         }
-    });
+        let _ = future::join_all(send_tasks).await;
+        // Drop lock to release the read lock.
+        drop(subscriptions);
```
yaziciahmet, line 97: Same.

yaziciahmet (changes requested):

yaziciahmet:
LGTM other than a small nit I left above.

yaziciahmet:
My only concern is that there is a high possibility that the soft_confirmation_tx will need multiple receivers eventually. Only receiver is SubscriptionManager right now, but I see it more fit to be broadcast as that event channel has no specific context and should notify anyone who wants to subscribe.

This comment might be irrelevant if we were to add another EventManager layer for chain-specific events, but it doesn't seem necessary in the near future as almost all the new incoming information is dependent on a new soft confirmation, and the specifics of the requested notification can be derived from that new soft confirmation.

crates/ethereum-rpc/src/subscription.rs
rakanalh: IMO, it's better to be explicit about it. Since `RwLock` would be waited upon elsewhere for write which means that we want to make sure we don't starve the write task. You're right about it being dropped but again i want to be explicit.

rakanalh:
> My only concern is that there is a high possibility that the soft_confirmation_tx will need multiple receivers eventually. Only receiver is SubscriptionManager right now, but I see it more fit to be broadcast as that event channel has no specific context and should notify anyone who wants to subscribe.

We shouldn't make it use `broadcast` until we have a second receiver. Why use broadcast for something that has 1 receiver, if that's the case then mpsc would be the way to go for now. If we keep assuming future use cases then we build up tech-debt.

PR #1216, opened 2024-09-23
Structured Concurrency: Part 2

### LINKED ISSUE #1200: MockDA spawns a task per DA transaction / commitment the sequencer sends
We've seen this being problematic since spamming the sequencer would prevent tokio from actually running the spawned task. That is, if the spammer doesn't yield control to the runtime.

https://github.com/chainwayxyz/citrea/blob/a93b60e4d7229c887720866d22a96cf9ec6d0e60/crates/sovereign-sdk/adapters/mock-da/src/service.rs#L443-L455

Make this an actual service that runs as a single task that processes sent commitments from sequencer.

rakanalh:
# Description

Related to #1214 so this is to be merged AFTER.

## This PR:
- Primitives:
  - Introduces `TaskManager`: Calling `spawn` using the task manager would force the future-to-be-spawned to accept a cancellation token so that **all tasks using the task manager are cancellable.** 
- DaService:
  - Replaces `tokio::spawn` for each transaction queue call with a single worker.
  - Replaces an additional long running task for logging
- Sequencer
  - Spawns tasks using `TaskManager`.
  - Since most tasks which are running in the sequencer are running in a single tokio task, handling `ctrl-c` event would be sufficient since `tokio::select` is going to execute the `ctrl-c` signal to stop all current tasks. This means that no other branch of `tokio::select` is going to be executed concurrently.
- Full Node
  - Spawns tasks using `TaskManager`.
  - Use cancellation token to shutdown DA handler
- Prover
  - Spawns tasks using `TaskManager`.
  - Use cancellation token to shutdown DA handler

## Task cancellation

For tasks which are spawned by the task manager, all tasks are expected to use the `cancellation_token`. Most of the uses of the token would come inside the `tokio::select`. This means that if all futures inside this select are running in a single task, notifying the task with the cancellation token means that we have to finish processing whatever the task is currently doing in-hand and return control back to `tokio::select` in order for the cancellation token to be seen as `cancelled` and then the component currently running is stopped.

## Linked Issues
- Fixes #1200 
- Related to #923 

Refs:
- [Graceful shutdown](https://tokio.rs/tokio/topics/shutdown)


rakanalh:
> bitcoin-da should be updated too

bitcoin-da does not use `tokio::spawn` anywhere. Are there specific parts you'd like to update?

kpp:
```
 172   │     pub fn spawn_da_queue(
 173   │         self: Arc<Self>,
 174   │         mut rx: UnboundedReceiver<SenderWithNotifier<TxidWrapper>>,
 175   │     ) {
 176   │         // This is a queue of inscribe requests
 177   │         tokio::task::spawn_blocking(|| {
```

rakanalh:
> ```
>  172   │     pub fn spawn_da_queue(
>  173   │         self: Arc<Self>,
>  174   │         mut rx: UnboundedReceiver<SenderWithNotifier<TxidWrapper>>,
>  175   │     ) {
>  176   │         // This is a queue of inscribe requests
>  177   │         tokio::task::spawn_blocking(|| {
> ```

I haven't look at the code here yet and maybe it does need some refactoring. However, based on Tokio's documentation for [spawn_blocking](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html):

> Be aware that tasks spawned using spawn_blocking cannot be aborted because they are not async. If you call [abort](https://docs.rs/tokio/latest/tokio/task/struct.JoinHandle.html#method.abort) on a spawn_blocking task, then this will not have any effect, and the task will continue running normally. The exception is if the task has not started running yet; in that case, calling abort may prevent the task from starting.

kpp:
So there are 2 things a tokio task is doing - to process get_fee_rate (can be a separate task) and to `send_transaction_with_fee_rate` - which is mining a tx prefix inside.

crates/bitcoin-da/src/service.rs
```diff
@@ -168,88 +176,91 @@ impl BitcoinService {
         })
     }
 
-    pub fn spawn_da_queue(
+    pub fn run(
```
kpp, line 179: what is the point of renaming?

crates/prover/src/da_block_handler.rs
```diff
@@ -133,6 +134,10 @@ where
         interval.tick().await;
         loop {
             select! {
+                biased;
+                _ = cancellation_token.cancelled() => {
+                    return;
+                }
                 _ = &mut l1_sync_worker => {},
```
kpp, line 141: Shouldn't `l1_sync_worker` also be spawned with task_manager?

kpp (approved):
LGMT but please note my comments

ercecan (approved):
LGTM overall but I believe l1 and l2 sync can also be managed with task manager? I would like to hear your thoughts on this

crates/bitcoin-da/src/service.rs
```diff
@@ -168,88 +176,91 @@ impl BitcoinService {
...
-                    Ok(None) => {
-                        info!("No pending transactions found");
-                        None
-                    }
-                    Err(e) => {
-                        error!(?e, "Failed to get pending transactions");
-                        None
-                    }
-                };
+        tokio::spawn(async move {
```
eyusufatik, line 183: spawn_blocking might be more suitable for this. 

in prover's case, for instance, mining the wtxid prefix takes a long time. maybe even more than 5 minutes

crates/bitcoin-da/src/service.rs
```diff
@@ -1043,7 +1046,6 @@ mod tests {
         .expect("Error initialazing BitcoinService");
 
         let da_service = Arc::new(da_service);
-        // da_service.clone().spawn_da_queue(_rx);
```
eyusufatik, line 1046: keep this. when we generate test cases we always forget to uncomment this and takes 10 mins to debug :D

crates/fullnode/src/runner.rs
```diff
@@ -325,19 +329,22 @@ where
...
-                sequencer_da_pub_key,
-                prover_da_pub_key,
-                code_commitments_by_spec,
-                accept_public_input_as_proven,
-                l1_block_cache.clone(),
-            );
-            l1_block_handler.run(start_l1_height).await
-        });
+        self.task_manager
+            .spawn(move |cancellation_token| async move {
```
eyusufatik, line 333: if we can also pass the `cancellation_token` into l2 sync, we can make sure we committed db correctly before shutting down and that would be great

crates/primitives/src/lib.rs
```diff
@@ -1,6 +1,10 @@
 pub mod basefee;
 mod constants;
 pub mod forks;
+#[cfg(feature = "native")]
+pub mod tasks;
 pub mod types;
 
 pub use constants::*;
+#[cfg(feature = "native")]
+pub use tasks::*;
```
eyusufatik, line 10: wouldn't tasks make sense in the `common` crate?

crates/bitcoin-da/src/service.rs
rakanalh: `spawn_blocking` would not work for the async stuff going on inside. We could later update the mining part to be executed inside a blocking context.

crates/bitcoin-da/src/service.rs
rakanalh: Restored

rakanalh:
> Shouldn't `l1_sync_worker` also be spawned with task_manager?

> LGTM overall but I believe l1 and l2 sync can also be managed with task manager? I would like to hear your thoughts on this

> if we can also pass the `cancellation_token` into l2 sync, we can make sure we committed db correctly before shutting down and that would be great

The reason we don't spawn `sync_l1` and `sync_l2` by the task manager is because it is already managed by it's own `tokio::select` loop. On our cancellation request, `sync_l2` for example would finish it's work before `ctrl_c` signal is handled if a sync task was already running. For the tasks that we spawn as separate tasks, we would use the task manager to cancel them.

eyusufatik (approved):

PR #1220, opened 2024-09-24
Fix merkle root mismatch

### LINKED ISSUE #1217: Investigate merkle root mismatch error in fullnode
Image

Based on a comment from @eyusufatik 

> wait a few minutes and ask rpc for those l2 heights. they will say finalized
> there is probably a off by one error somewhere, if only a single L2 is not synced yet seq comm processing fails
but it retries so it resolves.

rakanalh:
# Changes in this PR
- [x] Move components which are shared between sequencer / full node and prover to crates/common.
- [x] Add `check_l2_range_exists` from prover to common/utils module.
- [x] Use `check_l2_range_exists` in full node.

## Linked Issues
- Fixes #1217


eyusufatik (approved):
LGTM.

If you have tested this branch against testnet, please merge.

Issue #1190, opened 2024-09-17
Pruning


rakanalh:
Currently, Citrea nodes (Sequencer / Full node / Prover) run in archive mode. Meaning that they store all block data ever since genesis all the way to the tip of the chain.

We'd like to implement pruning so that we can be selective on how much data we retain based on specific configuration parameters.

Based on Reth's [pruning](https://reth.rs/run/pruning.html) documentation, we'd like to implement a similar approach.

There are multiple areas which we should look into:
1. `EVM`
2. `ProverStorageManager`
3. `LedgerDB` tables.

This is a parent issue which outlines the following tasks that need to be done:

- [ ] Research each component and create child issues define what needs to be done
- [ ] Issue 1
- [ ] Issue 2
- [ ] Issue 3
- [ ] Issue 4

PR #1229, opened 2024-09-25
Pruning skeleton

### LINKED ISSUE #1191: Pruning: Implement pruning crate & config
# Description

A new crate needs to be introduced called `pruning` which implements the following:

1. A configuration struct:

```rust
/// Define pruning mode based on configuration and/or CLI arguments
pub enum PruningMode {
   /// Pruner does not run in this case.
   Archive,
   /// Pruner is run based on config.
   Pruned(PruningConfig)
}

/// A configuration type to define the behaviour of the pruner.
pub struct PruningConfig {
  /// Defines the number of blocks from the tip of the chain to remove.
  distance: u64,
}
```

2. Pruner
```rust
pub struct Pruner {
   /// config
   config: PruningConfig,
   /// A channel receiver which gets notified of new L2 blocks.
   receiver: tokio::sync::mpsc::Receiver<u64>,
   /// Components to be pruned.
   evm: Arc<Evm>,
   ledger_db: Arc<LedgerDB>
}

impl Pruner {
   /// Prune everything
   pub fn prune(&self, up_to_block: u64) { 
      self.prune_evm(up_to_block);
      self.prune_ledger(up_to_block);
   }

   /// Prune evm
   pub fn prune_evm(&self, up_to_block: u64) { .. }

   /// Prune ledger
   pub fn prune_ledger(&self, up_to_block: u64) { .. }

    .... 
}

impl Service for Pruner {
   fn run(self) {
      /// Receive new block 
      while let Some(current_l2_block) = self.receiver.next() {
          self.prune(current_l2_block.number - self.config.distance);
      }
   }
}
```

rakanalh:
# Description

This PR implements pruning.

Here we introduce the `Pruner` structure which runs for as long as the node is running to prune blocks every X number of blocks. 
At each L2 block production, we calculate the trigger block number which would invoke pruning `X - distance` blocks. This happens by spawning blocking tasks (since we use sync rocksdb) to delete specific state based on `up_to_block`.

The implementation of the actual pruning is implemented in later tickets referenced in:
https://github.com/chainwayxyz/citrea/issues/1190

## Linked Issues
- Fixes #1191 

## MERGE AFTER
- #1214 
- #1216


crates/fullnode/src/runner.rs
```diff
@@ -320,6 +323,18 @@ where
...
+        if let Some(config) = &self.pruning_config {
+            let pruner = Pruner::<DB>::new(
+                config.clone(),
+                self.ledger_db.get_last_pruned_l2_height()?.unwrap_or(0),
+                self.soft_confirmation_tx.subscribe(),
+                self.ledger_db.clone(),
+            );
+
+            self.task_manager
+                .spawn(|cancellation_token| pruner.run(cancellation_token));
```
kpp, line 335: This is not a blocking task. Is it ok?

rakanalh: Yes that's okay... i am `spawn_blocking` inside the pruner.

resources/configs/mock/sequencer_config.toml
```diff
@@ -5,6 +5,8 @@ deposit_mempool_fetch_limit = 10
 block_production_interval_ms = 1000
 da_update_interval_ms = 2000
 
+# pruning_config.distance = 10
```
kpp, line 8: Why comment?

crates/pruning/src/tests.rs
```diff
@@ -0,0 +1,31 @@
...
+
+use sov_db::ledger_db::LedgerDB;
+use sov_db::rocks_db_config::RocksdbConfig;
+use tokio::sync::broadcast;
+use tokio_util::sync::CancellationToken;
+
+use crate::{Pruner, PruningConfig};
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_pruner_simple_run() {
```
kpp, line 12: What does this test do?

crates/pruning/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+pub struct Pruner<DB>
+where
+    DB: SharedLedgerOps,
+{
+    /// config
+    config: PruningConfig,
+    /// The last block number which was pruned.
+    last_pruned_block: u64,
+    /// A channel receiver which gets notified of new L2 blocks.
+    receiver: broadcast::Receiver<u64>,
```
kpp, line 36: l2_receiver ?

crates/pruning/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+        println!("Finished pruning");
+    }
+
+    pub async fn run(mut self, cancellation_token: CancellationToken) {
+        loop {
+            select! {
+                biased;
+                _ = cancellation_token.cancelled() => {
+                    println!("Stopping pruner");
+                    // Store the last pruned l2 height in ledger DB to be restored in the next initialization.
```
kpp, line 78: Is it required here? You set it when current_l2_block > distance, it should be enough

kpp (approved):

resources/configs/mock/sequencer_config.toml
rakanalh: By default, we don't prune.

crates/pruning/src/tests.rs
rakanalh: Run the pruner with a cleanup distance of 5 L2 blocks. Generate 10 blocks (runs the pruning twice, but we don't implement pruning right now so nothing happens).

And then when the cancellation token is invoked, the pruner should wait for existing tasks to stop and the test will pass.

This will be useful after we implement the actual pruning logic

crates/pruning/src/lib.rs
rakanalh: I am only setting the ledger value when the pruning stops, to decrease the number of writes we do. When current_l2_block > distance, i am only setting a local member variable.

crates/pruning/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+                current_l2_block = self.l2_receiver.recv() => {
+                    if let Ok(current_l2_block) = current_l2_block {
+                        // Calculate the block at which pruning would be triggered.
+                        // This is allowing `self.config.distance` blocks to be produced before we
+                        // decide to prune the previous `self.config.distance` blocks.
+                        let trigger_block = (self.last_pruned_block + self.config.distance) + 1;
+                        if current_l2_block >=trigger_block {
+                            self.prune(current_l2_block - self.config.distance).await;
+                            self.last_pruned_block += self.config.distance;
+                        }
```
yaziciahmet, line 90: I think the naming of last_pruned_block is wrong. Assume distance = 1000 scenario:
1. chain produced 1001th block
2. trigger_block = 0 + 1000 + 1 = 1001, hence triggering a prune
3. prune(1001 - 1000) = prune(1), so nothing is pruned assuming up_to_block is exclusive
4. last_pruned_block = 1000
As you see even though we didn't prune anything, last_pruned_block became 1000. And in the next iteration when chain reaches 2001, 1 to 1000 blocks will be pruned, and last_pruned_block will become 2000. I think more appropriate name is something like next_pruning_end_block.

yaziciahmet (changes requested):

crates/pruning/src/lib.rs
rakanalh: Good observation.

I have changed this to:
```rust
self.prune(self.last_pruned_block + self.config.distance).await;
```

So when we start:
- trigger block = 1001
- prune(up_to_block = 1000)
- last_pruned_block = 1000

Next iteration
- last_pruned_block = 1000
- trigger_block = 2001
- prune(up_to_block = 2000)
- last_pruned_block = 2000

... etc

yaziciahmet: Now it is really last_pruned_block, but I am not fond of this pruning algorithm tbh. Because at trigger points, there will only remain 1 block, which is not good imo because while at 1001th block, its not feasible to not be able to query 1000th block, its still too new and most likely useful. I think if the config is set to 1000, we should always have at least 1000 blocks, and always prune down to last 1000 when a certain condition is met.

I propose the condition to be something like 2x of distance limit. So anytime the stored block count in the database reaches 2000, it can be pruned to its latest 1000 blocks.

yaziciahmet: wdyt @eyusufatik @kpp

crates/prover/src/runner.rs
```diff
@@ -231,6 +234,18 @@ where
...
+                config.clone(),
+                self.ledger_db.get_last_pruned_l2_height()?.unwrap_or(0),
+                self.soft_confirmation_tx.subscribe(),
+                self.ledger_db.clone(),
+            );
+
+            self.task_manager
+                .spawn(|cancellation_token| pruner.run(cancellation_token));
+        }
+
```
ercecan, line 248: Should we prune prover?

rakanalh: For the prover, i don't think we need to keep all the history in order for us to be able to prove sequencer commitments.
For the sequencer, same reasoning applies, as long as we don't use it as an archive node (no full or prover nodes are syncing from there). So basically we can have multiple sequencers running (maybe one or two of which are archive and the rest are pruned).
That is, if we have a high-availability setup

ercecan: talked f2f we might turn off if necessary

crates/pruning/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+                    return;
+                }
+                current_l2_block = self.l2_receiver.recv() => {
+                    if let Ok(current_l2_block) = current_l2_block {
+                        // Calculate the block at which pruning would be triggered.
+                        // This is allowing `self.config.distance` blocks to be produced before we
+                        // decide to prune the previous `self.config.distance` blocks.
+                        let trigger_block = (self.last_pruned_block + self.config.distance) + 1;
+                        if current_l2_block >=trigger_block {
+                            self.prune(self.last_pruned_block + self.config.distance).await;
```
ercecan, line 88: This feels right lmk
```suggestion
                            self.prune(self.last_pruned_block + current_l2_block - trigger_block + 1).await;
```

crates/pruning/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+                            self.last_pruned_block += self.config.distance;
```
ercecan, line 89: Also here

```suggestion
                            self.last_pruned_block +=  current_l2_block - trigger_block + 1;
```

crates/pruning/src/lib.rs
rakanalh: You're totally right. I have not considered the fact that we should keep more than 1 blocks at any given moment. Although i meant to do that given that Reth's and Substrate's implementation do based on my research.

I'll amend

crates/pruning/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+                    }
+                    return;
+                }
+                current_l2_block = self.l2_receiver.recv() => {
+                    if let Ok(current_l2_block) = current_l2_block {
+                        // Calculate the block at which pruning would be triggered.
+                        // This is allowing `self.config.distance` blocks to be produced before we
+                        // decide to prune the previous `self.config.distance` blocks.
+                        let trigger_block = (self.last_pruned_block + self.config.distance) + 1;
+                        if current_l2_block >=trigger_block {
```
kpp, line 87:
```suggestion
                        if current_l2_block >= trigger_block {
```

crates/pruning/src/lib.rs
kpp: Got it. what if the node dies with another signal, lets say `SIGABRT` instead of `SIGINT`?

crates/pruning/src/lib.rs
yaziciahmet: weird linter didn't catch this

crates/pruning/src/lib.rs
rakanalh: Yea good point. I'll attend to that.

crates/pruning/src/lib.rs
rakanalh: Done

ercecan (approved):
So I thought the distance thing was supposed to work like the window size of a sliding window, and prune after each new l2 to keep the distance same, but after we talked with @rakanalh it made sense to at least wait another distance amount of blocks to prune so that it does not have to deal with extra (blocking) reads and writes every new block

crates/pruning/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+                    if let Err(e) = self.ledger_db.set_last_pruned_l2_height(self.last_pruned_block) {
+                        error!("Failed to store last pruned L2 height {}: {:?}", self.last_pruned_block, e);
+                    }
+                    return;
+                }
+                current_l2_block = self.l2_receiver.recv() => {
+                    if let Ok(current_l2_block) = current_l2_block {
+                        if let Some(up_to_block) = self.criteria.should_prune(self.last_pruned_block, current_l2_block) {
+                            self.prune(up_to_block).await;
+                            self.last_pruned_block += self.config.distance;
```
yaziciahmet, line 96: `self.last_pruned_block = up_to_block - 1` seems a bit more solid to me due to the fact that criteria is a trait and you don't really know how it makes its decisions.

crates/pruning/src/pruners/evm.rs
```diff
@@ -0,0 +1,10 @@
+use citrea_evm::Evm;
+use sov_modules_api::default_context::DefaultContext;
+use tracing::debug;
+
+/// Prune evm
+pub(crate) fn prune_evm(up_to_block: u64) {
+    debug!("Pruning EVM, up to L2 block {}", up_to_block);
+    let _evm = Evm::<DefaultContext>::default();
+    // unimplemented!()
+}
```
yaziciahmet, line 10: Is the implementation of pruner methods out of the scope of this PR? If so, why?

yaziciahmet (changes requested):
Great work overall, just have some nitpicks and questions.

crates/pruning/src/pruners/evm.rs
rakanalh: Yes, please refer to: https://github.com/chainwayxyz/citrea/issues/1190

crates/pruning/src/lib.rs
rakanalh: up to block SHOULD be inclusive.

crates/pruning/src/lib.rs
rakanalh: Apparently, `ctrl_c` handles `SIGINT` but `SIGABRT` would only happen if the task panics, in which case, tokio catches the panic and terminates the task as if it finished normally. We may need to handle `SIGTERM`.

Though this is out of scope of this PR IMO.

crates/pruning/src/lib.rs
yaziciahmet: sounds good :)

crates/pruning/src/lib.rs
yaziciahmet: Definitely out of scope of this pr. We definitely need to handle SIGTERM as it is the docker kill signal.

crates/pruning/src/lib.rs
rakanalh, line 96:
```suggestion
                            self.last_pruned_block = up_to_block;
```

yaziciahmet (approved):

bin/citrea/provers/risc0/guest-mock/Cargo.lock
eyusufatik, line 1: are you sure this is supposed to change?

resources/configs/mock/sequencer_config.toml
```diff
@@ -5,6 +5,8 @@ deposit_mempool_fetch_limit = 10
 block_production_interval_ms = 1000
 da_update_interval_ms = 2000
```
eyusufatik, line 7: no pruning in sequencer or prover.

only in full nodes

crates/pruning/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub distance: u64,
+}
+
+impl Default for PruningConfig {
+    fn default() -> Self {
+        Self { distance: 256 }
+    }
+}
+
+pub struct Pruner<DB>
```
eyusufatik, line 31: direct access to NativeStorage, or StorageManager might be helpful here.

eyusufatik: can be added later ofc

eyusufatik (changes requested):

crates/pruning/src/lib.rs
rakanalh: Yea i'd like the later ticket implementer to add this

eyusufatik (approved):

PR #1230, opened 2024-09-26
RPC: Fix max batch check

rakanalh:
# Description

Problem here is that if we set our sequencer's max batch to 10, and full node's `sync_blocks_count` to 10, the full node will receive an error from sequencer's RPC.

The check `end - start < max_batch` would be false for batches


eyusufatik:
assuming 10 is set for max batches constand

I request 1 through 11  and the range is inclusive, I can now get 11 soft confirmations because 11 - 1 <= 10 indeed

but the intended behavior isn't the case

rakanalh:
> assuming 10 is set for max batches constant
> 
> I request 1 through 11 and the range is inclusive, I can now get 11 soft confirmations because 11 - 1 <= 10 indeed
> 
> but the intended behavior isn't the case

11 - 1 = 10 

10 <= 10 is true.

I figured this out when i set `sync_blocks_count` for fullnode to 100 to which the RPC endpoint of the sequencer said that the max was 20. So i set `sync_blocks_count` back to 20 but that was rejected as well. 19 worked which means that we're off by 1.

eyusufatik:
> > assuming 10 is set for max batches constant
> > I request 1 through 11 and the range is inclusive, I can now get 11 soft confirmations because 11 - 1 <= 10 indeed
> > but the intended behavior isn't the case
> 
> 11 - 1 = 10
> 
> 10 <= 10 is true.
> 
> I figured this out when i set `sync_blocks_count` for fullnode to 100 to which the RPC endpoint of the sequencer said that the max was 20. So i set `sync_blocks_count` back to 20 but that was rejected as well. 19 worked which means that we're off by 1.

that's what I'm trying to say, when I can pull 1 through 11 I'm going over the limit.

Probably the bug is where we send the request in full node.

when you set to 20, it tries to fetch 21 blocks is my guess

rakanalh:
Hmm, we call `get_soft_confirmation_range`
https://github.com/chainwayxyz/citrea/blob/4cbbba53bb66f6623adb34d887f5594c64b35111/crates/fullnode/src/runner.rs#L422-L424

Into the sequencer-client:

https://github.com/chainwayxyz/citrea/blob/4cbbba53bb66f6623adb34d887f5594c64b35111/crates/sequencer-client/src/lib.rs#L51-L62

Looks pretty normal.

So requesting with start `l2_height = 1` , end would be `1 + 10 = 11`.

The receiving end would see this request and perform the check.

But we're both right. 

https://github.com/chainwayxyz/citrea/blob/4cbbba53bb66f6623adb34d887f5594c64b35111/crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/rpc.rs#L81-L85

This check is NOT inclusive.

The code right after it IS inclusive:
https://github.com/chainwayxyz/citrea/blob/4cbbba53bb66f6623adb34d887f5594c64b35111/crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/rpc.rs#L86-L89

Which means that we either keep my change or we change the `..=` to `..`

eyusufatik:
the bug here is it is requesting 1, 11 at the end not 1,10 --> which would be interpreted as [1,10] by the sequencer, and since it's inclusive which would correctly yield 10 soft confirmations which is the limit.

Rust's range and range inclusive make this a bit tricky, 

I think we should make `get_soft_confirmation_range`'s param RangeInclusive in sequencer client 

and pass into that function

l2_height..=l2_height + sync_blocks_count - 1.

and revert the check in the ledger rpc side

eyusufatik (approved):
hope it works now :)

PR #1231, opened 2024-09-26
Fix testnet config

rakanalh:
# Description

Fixes for testnet config after trying to run the fullnode on a testnet.

### DIFF
--- a/resources/configs/testnet/rollup_config.toml
+++ b/resources/configs/testnet/rollup_config.toml
@@ -16,7 +16,7 @@ network = "testnet"
 tx_backup_dir = ""
 
 [storage]
-path = "./db"
+path = "resources/dbs"
 
 # comment out if you want to set custom max open files
 # if you leave it like this, it will use the system limit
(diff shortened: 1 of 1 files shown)


resources/configs/testnet/rollup_config.toml
```diff
@@ -16,7 +16,7 @@ network = "testnet"
 tx_backup_dir = ""
 
 [storage]
-path = "./db"
+path = "resources/dbs"
```
eyusufatik, line 19: some users will be downloading this toml and running binaries instead of cloning this repo, that's why I wanted a clearer directory which is just "db"

we can add db to gitignore though

eyusufatik (changes requested):

resources/configs/testnet/rollup_config.toml
rakanalh: The same thing would apply to the devnet file in this case. However, for consistency, we should have it set this way. Since users will download this file, they can update the values as they please.

I have added a comment to clarify.

eyusufatik (approved):

PR #1236, opened 2024-09-27
JSONRPC request-response logging

### LINKED ISSUE #1227: New tracing level for RPC Error responses
When we search for logs on production we see a lot of RPC responses, these were successfuly responded to the client, but the request itself is incorrect. e.g. a send tx with lower nonce than it should be.

a new trace level will help us with filtering the logs

yaziciahmet:
# Description

Added logging to each JSONRPC request and response to be able to have better debuggability of our nodes. Each request and successful response is logged with debug tracing level. Error responses are logged with warn level since it is highly likely that the error is caused by the client-side.

Original idea was to add a new tracing level for rpc request logging. But such behavior is not possible with tracing crate. Instead of that, errors are now logged to warn level, and can be searched with a filter such as level = WARN && message = rpc_error.

I initially logged the request and success responses to info level, but due to full nodes constantly spamming sequencer for new blocks through rpc, sequencer logs became full with useless rpc calls, which is not good. Another potential approach might be to make Sequencer RPC logs debug, but full node rpc logs info.

Also, prover doesn't have any rpc logs as it is unnecessary.

## Linked Issues
- Fixes #1227 

## Testing
I ran sequencer and full node and sent rpc requests and observed logs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1751,6 +1751,7 @@ dependencies = [
  "anyhow",
  "backoff",
  "citrea-primitives",
+ "futures",
  "hyper 1.4.1",
  "jsonrpsee",
  "lru",
@@ -1758,6 +1759,7 @@ dependencies = [
  "sov-rollup-interface",
  "tokio",
  "tower-http",
+ "tracing",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


rakanalh (approved):
Nice!

eyusufatik (approved):

kpp (approved):

eyusufatik:
we might want to delete enter traces to rpcs as well, I think there were some in ledger rpc, wdyt?

PR #1241, opened 2024-09-30
Add enable recovery config

ercecan:
# Description
Currently because of some internal errors in Risc0 bonsai api, we are having troubles on fetching sessions and this causes the l1 handler thread to die because of some panics. When we reopen and try to recover we face the same issues because we are trying to recover the erroneous session and the thread dies again.
 
This pr enables recovery if `enable_recovery` is set to true, we plan on setting this config to false until we have a better handling of internal errors of provers

### DIFF
--- a/bin/citrea/tests/e2e/mod.rs
+++ b/bin/citrea/tests/e2e/mod.rs
@@ -94,6 +94,7 @@ async fn test_all_flow() {
             Some(ProverConfig {
                 proving_mode: sov_stf_runner::ProverGuestRunConfig::Execute,
                 proof_sampling_number: 0,
+                enable_reocvery: true,
             }),
             rollup_config,
             None,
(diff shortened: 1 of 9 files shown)


eyusufatik (approved):

bin/citrea/tests/e2e/mod.rs
```diff
@@ -94,6 +94,7 @@ async fn test_all_flow() {
             Some(ProverConfig {
                 proving_mode: sov_stf_runner::ProverGuestRunConfig::Execute,
                 proof_sampling_number: 0,
+                enable_reocvery: true,
```
otaliptus, line 97: Shall we run grammarly in the repo?

ercecan: Yeah...

PR #1255, opened 2024-10-01
Add sys txs to evm tests

### LINKED ISSUE #1188: Modify all tests to have system txs enabled
In evm tests the system contracts are not enabled. It would be better for us if we enabled them to see possible system tx related errors.
In tests we are using mostly default settings for the config and adding some addresses manually like `DevSigner`
We can continue using the default `ChainId` etc. and use `DevSigner` but we should find a way to integrate system contracts.
For that we can have a simple parser that parses the genesis file under `resources/test-data/integration-tests/evm.json`
The parser should parse the addresses and push them to the config like in `crates/evm/src/tests/sys_tx_tests.rs` function `config_push_contracts`

- [x] Investigate evm tests and how it's config is created
- [x] Investigate sys tx tests
- [x] Write a parser function to parse genesis and get contract addresses and fee vault addresses (the parser can take all the addresses it should not take chain_id spec etc.)
- [x] Integrate system contracts to evm tests (including sys tx tests so we will still call a function like `push_to_config` but the function will read the necessary config from file)

exeokan:
# Description
Enables system contracts in evm tests, see the issue for more detail.

## Linked Issues
- Fixes #1188


crates/evm/src/tests/utils.rs
```diff
@@ -103,3 +106,12 @@ pub(crate) fn commit(
 
     root.0
 }
+
+/// Loads the genesis configuration from the give path and pushes the accounts to the evm config
```
yaziciahmet, line 110:
```suggestion
/// Loads the genesis configuration from the given path and pushes the accounts to the evm config
```

crates/evm/Cargo.toml
```diff
@@ -15,6 +15,7 @@ readme = "README.md"
 sov-modules-api = { path = "../sovereign-sdk/module-system/sov-modules-api", default-features = false, features = ["macros"] }
 sov-prover-storage-manager = { path = "../sovereign-sdk/full-node/sov-prover-storage-manager", optional = true }
 sov-state = { path = "../sovereign-sdk/module-system/sov-state" }
+sov-stf-runner = { path = "../sovereign-sdk/full-node/sov-stf-runner" }
```
yaziciahmet, line 18: If you need this only in tests, make sure to specify it under dev-dependencies

crates/evm/src/tests/sys_tx_tests.rs
```diff
@@ -416,7 +415,7 @@ fn test_bridge() {
     let (mut config, _, _) =
         get_evm_config_starting_base_fee(U256::from_str("1000000").unwrap(), None, 1);
 
-    config_push_contracts(&mut config);
+    config_push_contracts(&mut config, None);
```
eyusufatik, line 418: why not call `config_push_contracts` inside `get_evm_*` functions?

exeokan: makes sense

ercecan (approved):

crates/evm/Cargo.toml
```diff
@@ -59,6 +59,7 @@ revm = { workspace = true, features = ["optional_block_gas_limit", "optional_eip
 sov-modules-api = { path = "../sovereign-sdk/module-system/sov-modules-api", features = ["macros"] }
 sov-prover-storage-manager = { path = "../sovereign-sdk/full-node/sov-prover-storage-manager", features = ["test-utils"] }
 sov-rollup-interface = { path = "../sovereign-sdk/rollup-interface", features = ["testing"] }
+sov-stf-runner = { path = "../sovereign-sdk/full-node/sov-stf-runner" }
```
yaziciahmet, line 62: can you also guard this dependency behind native feature below?

yaziciahmet (changes requested):

crates/evm/Cargo.toml
eyusufatik: if dev dependency no need for future gating, no?

yaziciahmet: I also thought that first, but guest code lock files shouldn't have the dependency, but they still do. Maybe @exeokan disabled his rust-analyzer causing lock files to not update automatically? Try running `cargo check`

exeokan: Yes, apparently rust-analyzer didn't update the lock files, and sov-stf-runner is in fact removed from the lock files. No problem here then.

ercecan (approved):

eyusufatik (approved):

yaziciahmet (approved):

crates/evm/src/tests/call_tests.rs
```diff
@@ -132,50 +134,80 @@ fn call_multiple_test() {
...
+                            ).unwrap()
+                            }]
+                        },
+                        gas_used: 80620,
+                        log_index_start: 0,
+                        l1_diff_size: 561
+                },
+                Receipt { receipt: reth_primitives::Receipt { tx_type: reth_primitives::TxType::Eip1559, success: true, cumulative_gas_used: 213563, logs: vec![] }, gas_used: 132943, log_index_start: 1, l1_diff_size: 567 },
+                Receipt { receipt: reth_primitives::Receipt { tx_type: reth_primitives::TxType::Eip1559, success: true, cumulative_gas_used: 257293, logs: vec![] }, gas_used: 43730, log_index_start: 1, l1_diff_size: 255 },
+                Receipt { receipt: reth_primitives::Receipt { tx_type: reth_primitives::TxType::Eip1559, success: true, cumulative_gas_used: 283923, logs: vec![] }, gas_used: 26630, log_index_start: 1, l1_diff_size: 255 },
```
kpp, line 207: I don't like `log_index_start: 1`. Is it okay?

kpp (approved):
Looks good, but please check `log_index_start: 1`

crates/evm/src/tests/call_tests.rs
ercecan: looking into it

ercecan: there is no problem, sys txs are in a different block

PR #1261, opened 2024-10-02
Fix estiamte gas l1 fee issue when metamask max amount is selected

### LINKED ISSUE #693: Revise L1 fee overhead on `eth_estimateGas` RPC
## Summary
When Metamask send max is used the transaction fails on L1 fee. We probably fail to do a good estimate on the L1 fee overhead for gas, let's dig into this

## Steps to reproduce

1. Switch to Citrea devnet on metamask with an address with balance.
2. Send cBTC to another address. Use the max button on the side.
3. The tx will be stuck on the mempool for sometime and eventually dropped out.

## Expected Result
We can empty the wallet (a few wei is fine)

## Actual Result
Tx is failing

ercecan:
# Description
Metamask initially sends an estimate_gas request with 0 value, due to there is no balance transfer diff size of balance transfer is not calculated and added to the fee, thus results in a rather low gas estimated. This causes an issue when we select the max amount possible to transfer, so this pr updates estimate_gas so that if value is 0 it adds balance transfer diff size l1 fee to the total l1 fee and calculates the estimated gas accordingly. With this we can use the max amount button in metamask

Screenshot 2024-10-02 at 16 13 57

## Linked Issues

- Fixes #693

### DIFF
--- a/crates/evm/src/query.rs
+++ b/crates/evm/src/query.rs
@@ -39,6 +39,10 @@ use crate::evm::db::EvmDb;
 use crate::evm::primitive_types::{BlockEnv, Receipt, SealedBlock, TransactionSignedAndRecovered};
 use crate::evm::DbAccount;
 use crate::handler::TxInfo;
+use crate::handler::{
+    ACCOUNT_DISCOUNTED_PERCENTAGE, DB_ACCOUNT_KEY_SIZE, DB_ACCOUNT_SIZE_EOA,
+    NONCE_DISCOUNTED_PERCENTAGE,
+};
 use crate::rpc_helpers::*;
 use crate::{BloomFilter, Evm, EvmChainConfig, FilterBlockOption, FilterError};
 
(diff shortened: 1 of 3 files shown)


yaziciahmet:
I do understand the motivation, but I have some concerns about the solution. With this, if someone wants to actually send 0 eth, they will be charged extra even though there are no state changes. I am aware that sending 0 eth is not a common operation, and also the fee difference is going to be little, but still wanted to raise this to understand further.

Do we know what does Reth to resolve this?

eyusufatik:
> I do understand the motivation, but I have some concerns about the solution. With this, if someone wants to actually send 0 eth, they will be charged extra even though there are no state changes. I am aware that sending 0 eth is not a common operation, and also the fee difference is going to be little, but still wanted to raise this to understand further.
> 
> Do we know what does Reth to resolve this?

no one is charged extra, they just get a higher estimate for gas

gas fee charged will be the exact usage.


reth does not have a problem with this, we are trying to get wallet's / devs calculate L1 fee without ever knowing about it :D

yaziciahmet (approved):
> no one is charged extra, they just get a higher estimate for gas

For a moment I forgot we are dealing with estimation. LGTM.

crates/evm/src/query.rs
```diff
@@ -71,8 +73,7 @@ impl EstimatedTxExpenses {
     /// Return total estimated gas used including evm gas and L1 fee.
     pub(crate) fn gas_with_l1_overhead(&self) -> U256 {
         // Actually not an L1 fee but l1_fee / base_fee.
-        let l1_fee_overhead =
-            U256::from(1).max(self.l1_fee / (self.base_fee + U256::from(GWEI_TO_WEI))); // assume 1 gwei priority fee
+        let l1_fee_overhead = U256::from(1).max(self.l1_fee.div_ceil(self.base_fee)); // assume 1 gwei priority fee
```
eyusufatik, line 76: remove comment

eyusufatik (approved):

PR #1262, opened 2024-10-02
Storage `get` calls to receive shared reference of working set

### LINKED ISSUE #146: Access storage in read only mode
For #32 we have to implement Revm's DataBaseRef trait. 

Which requires immutable borrowing of self.

However we can only access storage with mutable references (WorkingSet saves every read and write).

We need to access storage without modifying anything.

Checkout branch `orkun/overrides` file `module-system/module-implementations/sov-evm/src/evm/db.rs`.

Cannot implement necessary trait because every StateValue/Map/Vec expects a mutable WorkingSet, while self is borrowed immutably.

TODOs

- [ ] Figure out how to read storage with immutable workingset or something else.

rakanalh:
# Description

## Linked Issues
- Fixes #146


crates/sovereign-sdk/module-system/sov-state/src/witness.rs
```diff
@@ -17,28 +20,54 @@ use sov_modules_core::Witness;
...
+
+impl BorshSerialize for ArrayWitness {
+    fn serialize<W: std::io::prelude::Write>(&self, writer: &mut W) -> std::io::Result<()> {
+        let hints = self.hints.lock().unwrap();
+        borsh::BorshSerialize::serialize(
+            &self.next_idx.load(std::sync::atomic::Ordering::Relaxed),
+            writer,
+        )?;
+        borsh::BorshSerialize::serialize(&*hints, writer)?;
+        Ok(())
```
yaziciahmet, line 62: Lock the hints after serializing next_idx, it will be more efficient as you won't be blocking hints unnecessarily while serializing next_idx.

crates/sovereign-sdk/module-system/sov-state/src/witness.rs
```diff
@@ -17,28 +20,54 @@ use sov_modules_core::Witness;
...
 
-    fn merge(&mut self, rhs: &mut Self) {
-        let rhs_next_idx = rhs.next_idx;
-        let lhs_hints_lock = &mut self.hints;
-        let rhs_hints_lock = &mut rhs.hints;
-        lhs_hints_lock.extend(rhs_hints_lock.drain(rhs_next_idx..))
+    fn merge(&self, rhs: &Self) {
+        let rhs_next_idx = rhs.next_idx.load(std::sync::atomic::Ordering::SeqCst);
+        let mut lhs_hints_lock = self.hints.lock().unwrap();
+        let mut rhs_hints_lock = rhs.hints.lock().unwrap();
```
yaziciahmet, line 49: This is sort of scary. `self.hints` can be locked, and in a completely different operation, the order of locking could be first rhs, and then self, which would cause deadlock. I know there is no such case, but I think it would be safer if we:
1. Create empty vec: merged
2. lock lhs and extend merged vec with lhs
3. lock rhs and extend merged with rhs

This way we won't hold 2 resources locked at the same time, much less risk of deadlock of any sort for possible future iterations

yaziciahmet (changes requested):

crates/sovereign-sdk/module-system/sov-state/src/witness.rs
rakanalh: You forgot the 4th step, lock lhs again and set merged into it.

Which is pretty much what the code is doing. 

> self.hints can be locked, and in a completely different operation, the order of locking could be first rhs, and then self, which would cause deadlock.

I don't think we need to consider this. 

IFF we ever introduce new operations that would lock in the wrong ordering, we'll know and we'll resolve. This PR doesn't need to address imaginary issues right now.

rakanalh: I am opposing your approach because of lock / release, lock / release, lock / release cycle (lhs needs to be locked twice)

crates/sovereign-sdk/module-system/sov-state/src/witness.rs
```diff
@@ -17,28 +20,54 @@ use sov_modules_core::Witness;
 /// assert_eq!(witness.get_hint::<u64>(), 1u64);
 /// assert_eq!(witness.get_hint::<u64>(), 2u64);
 /// ```
-#[derive(Default, BorshDeserialize, BorshSerialize, Debug, Serialize, Deserialize)]
+#[derive(Default, Debug, Serialize, Deserialize)]
 pub struct ArrayWitness {
-    next_idx: usize,
-    hints: Vec<Vec<u8>>,
+    next_idx: AtomicUsize,
+    hints: Mutex<Vec<Vec<u8>>>,
```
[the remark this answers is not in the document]
rakanalh: It won't matter that much since all ops are locking with `write` if you use `RwLock`

crates/sovereign-sdk/module-system/sov-state/src/witness.rs
yaziciahmet: You are right, locking lhs 2 times is not good. Then I modify my suggestion to:

1. Lock rhs and clone it
2. Lock lhs and extend it with cloned rhs

Everything is locked just once, but not at the same time.

kpp:
The thing is: why do we even need this?

rakanalh:
Implemented state and block overrides without needing this. Closing this PR.

rakanalh:
Reopening

crates/sovereign-sdk/module-system/sov-state/src/witness.rs
rakanalh: Done

crates/sovereign-sdk/module-system/sov-state/src/witness.rs
```diff
@@ -17,28 +20,57 @@ use sov_modules_core::Witness;
...
-    fn merge(&mut self, rhs: &mut Self) {
-        let rhs_next_idx = rhs.next_idx;
-        let lhs_hints_lock = &mut self.hints;
-        let rhs_hints_lock = &mut rhs.hints;
-        lhs_hints_lock.extend(rhs_hints_lock.drain(rhs_next_idx..))
+    fn merge(&self, rhs: &Self) {
+        let rhs_next_idx = rhs.next_idx.load(std::sync::atomic::Ordering::SeqCst);
+        let rhs_hints_lock = rhs.hints.lock().unwrap();
+        let mut rhs_hints = rhs_hints_lock.clone();
+        drop(rhs_hints_lock);
```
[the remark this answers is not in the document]
rakanalh: yea but given the previous discussion about it, i intentionally made it explicit.

crates/evm/src/evm/db.rs
```diff
@@ -86,6 +86,45 @@ impl<'a, C: sov_modules_api::Context> Database for EvmDb<'a, C> {
     }
 }
 
+impl<'a, C: sov_modules_api::Context> DatabaseRef for EvmDb<'a, C> {
```
kpp: It's not used anywhere and we don't require this trait, so it's a dead code. This implementation requires `&working_set` instead of `&mut working_set` which is nice, however due to the nature of our codegen we always pass `&mut working_set` into our RPC methods like this:

```
                if method.method_signature.asyncness.is_some() {
                    async_methods_count += 1;
                    quote! {
                        #( #docs )*
                        #signature {
                            <#type_name #ty_generics as ::std::default::Default>::default().#method_name(#(#pre_working_set_args,)* &mut Self::get_working_set(self), #(#post_working_set_args),* ).await
                        }
                    }
                } else {
                    quote! {
                        #( #docs )*
                        #signature {
                            <#type_name #ty_generics as ::std::default::Default>::default().#method_name(#(#pre_working_set_args,)* &mut Self::get_working_set(self), #(#post_working_set_args),* )
                        }
                    }
                }
```

So for now `&working_set` brings nothing but additional mutexes inside of ArrayWitness.

kpp (changes requested):
It's nice to have `StateReaderAndWriter::get(&self, ...)`, but this PR brings little value except for lifting `&mut` to `&`. The full potential of this PR will be unleashed once we fix `build_rpc_impl_trait` to plug in `&` args like `working_set` so we can execute read-only RPC endpoints in parallel. Until then, this PR adds extra no-op mutexes.

PR #1265, opened 2024-10-02
Improve estimate gas and create access list rpcs

### LINKED ISSUE #434: EVM: Don't use `prepare_call_env` in `eth_estimateGas` and `eth_createAccessList`
that function was designed to create `TxEnv` for `eth_call`. We do a bunch of hotfixes to make it work with `eth_estimateGas`, don't need such a thing.

yaziciahmet:
# Description

## Linked Issues
- Fixes #434 

## Testing

## Docs

### DIFF
--- a/crates/evm/src/evm/call.rs
+++ b/crates/evm/src/evm/call.rs
@@ -5,7 +5,7 @@ use std::cmp::min;
 use reth_primitives::{B256, U256};
 use reth_rpc_eth_types::error::{EthApiError, EthResult, RpcInvalidTransactionError};
 use reth_rpc_types::TransactionRequest;
-use revm::primitives::TxEnv;
+use revm::primitives::{CfgEnvWithHandlerCfg, TxEnv};
 
 use crate::caller_gas_allowance;
 use crate::primitive_types::BlockEnv;
@@ -151,9 +151,7 @@ impl CallFees {
     }
 }
 
-// https://github.com/paradigmxyz/reth/blob/d8677b4146f77c7c82d659c59b79b38caca78778/crates/rpc/rpc/src/eth/revm_utils.rs#L201
-// if from_balance is None, gas capping will not be applied
-pub(crate) fn prepare_call_env(
+pub(crate) fn create_txn_env(
     block_env: &BlockEnv,
     request: TransactionRequest,
     cap_to_balance: Option<U256>,
@@ -172,21 +170,6 @@ pub(crate) fn prepare_call_env(
         chain_id,
         ..
     } = request;
-    let mut gas_price = gas_price.map(U256::from);
-    let mut max_fee_per_gas = max_fee_per_gas.map(U256::from);
-    let mut max_priority_fee_per_gas = max_priority_fee_per_gas.map(U256::from);
-    let gas = gas.map(U256::from);
-
-    // TODO: write hardhat and unit tests for this
-    if max_fee_per_gas == Some(U256::ZERO) {
-        max_fee_per_gas = None;
-    }
-    if gas_price == Some(U256::ZERO) {
-        gas_price = None;
-    }
-    if max_priority_fee_per_gas == Some(U256::ZERO) {
-        max_priority_fee_per_gas = None;
-    }
 
     let CallFees {
         max_priority_fee_per_gas,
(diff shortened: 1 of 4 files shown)


crates/evm/src/evm/call.rs
```diff
@@ -245,7 +236,7 @@ pub(crate) fn prepare_call_env(
             .map_err(|_| RpcInvalidTransactionError::GasUintOverflow)?,
         caller: from.unwrap_or_default(),
         gas_priority_fee: max_priority_fee_per_gas,
-        transact_to: to,
+        transact_to: to.unwrap_or_default(),
```
ercecan, line 239: good catch

crates/evm/src/evm/primitive_types.rs
```diff
@@ -12,6 +12,7 @@ pub(crate) struct BlockEnv {
     /// basefee is added in EIP1559 London upgrade
     pub(crate) basefee: u64,
     pub(crate) gas_limit: u64,
+    pub(crate) l1_fee_rate: u128,
```
eyusufatik, line 15: if you do this, method id is going to change and we have to do this carefully with a fork.

pls revert this and releated changes :)

eyusufatik: it might be a smart change tho. make an issue and label with requires fork etc.

yaziciahmet: Reverted and created the issue #1266

kpp (approved):

ercecan (approved):

crates/evm/src/evm/handler.rs
```diff
@@ -619,6 +619,7 @@ fn decrease_caller_balance<EXT, DB: Database>(
     change_balance(context, amount, false, address)
 }
 
+#[allow(unused)]
```
eyusufatik, line 622: this should be used?

crates/evm/src/query.rs
```diff
@@ -500,82 +499,51 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
-            request.clone(),
-            Some(
-                evm_db
-                    .basic(request.from.unwrap_or_default())
-                    .unwrap()
-                    .unwrap_or_default()
-                    .balance,
-            ),
+            &mut cfg_env,
+            &mut request.clone(),
```
eyusufatik, line 539: there might be a logical mistake here:

you are passing request as mutable reference, and in prepare_call_env, it is indeed mutated. 

however, you're passing the clone of request, which you keep using the original


my suggestion would be:


```rust
let cap_balance = evm_db.basic(request.from.unwrap_or_default()).....;

let tx_env = prepare_call_env(....., &mut request, cap_balance)?;
```

crates/evm/src/query.rs
```diff
@@ -616,59 +584,51 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+            let block = self
+                .get_sealed_block_by_number(block_number, working_set)?
+                .ok_or(EthApiError::UnknownBlockNumber)?;
             let block_env = BlockEnv::from(&block);
 
-            (l1_fee_rate, block_env, block.header.number)
+            let cfg = self
+                .cfg
+                .get(working_set)
+                .expect("EVM chain config should be set");
```
eyusufatik, line 596: reading config should happen after setting state to some block.

imagine you make a fork in the future, and change some value in config on the activation block.

with this setup, you are gonna run the latest cfg on a block from the past, which is possibly on an older version of the chain.

now your simulation will be wrong.

eyusufatik (changes requested):

crates/evm/src/evm/handler.rs
eyusufatik: and yes, it is used.

in query.rs line 834

yaziciahmet: it is not used in zk code, gives warning in build. was in between this and putting it behind native feature, chose unused.

crates/evm/src/query.rs
yaziciahmet: It is indeed funky, fixed it by making prepare_call_env consume transaction request

crates/evm/src/query.rs
yaziciahmet: fixed

crates/evm/src/evm/handler.rs
eyusufatik: native makes more sense actually. do it pls

yaziciahmet: done

eyusufatik (approved):

PR #1269, opened 2024-10-04
Derive SequencerClient

jfldde:
# Description

- Use jsonrpsee proc macro to derive SequencerClient trait
- Refactor to use `rpc(client, server)`
- Explicilty mark method behind test_mode as 404

### DIFF
--- a/crates/sequencer/src/rpc.rs
+++ b/crates/sequencer/src/rpc.rs
@@ -158,7 +158,7 @@ impl<C: sov_modules_api::Context, DB: SequencerLedgerOps + Send + Sync + 'static
             }
             Err(e) => {
                 error!("Error processing deposit tx: {:?}", e);
-                Err(e.into())
+                Err(e)
             }
         }
     }
--- a/crates/sequencer/Cargo.toml
+++ b/crates/sequencer/Cargo.toml
@@ -16,6 +16,7 @@ resolver = "2"
 alloy-rlp = { workspace = true }
 alloy-sol-types = { workspace = true }
 anyhow = { workspace = true }
+async-trait = { workspace = true }
 backoff = { workspace = true }
 borsh = { workspace = true }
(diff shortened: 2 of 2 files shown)


crates/sequencer/src/rpc.rs
```diff
@@ -108,39 +131,64 @@ pub(crate) fn create_rpc_module<
...
-
-    rpc.register_blocking_method(
-        "citrea_sendRawDepositTransaction",
-        move |parameters, ctx, _| {
-            let mut params = parameters.sequence();
-            let deposit: Bytes = params.next()?;
+    }
 
-            debug!("Sequencer: citrea_sendRawDepositTransaction");
+    fn citrea_send_raw_deposit_transaction(&self, deposit: Bytes) -> RpcResult<()> {
```
rakanalh, line 136:
```suggestion
    fn send_raw_deposit_transaction(&self, deposit: Bytes) -> RpcResult<()> {
```

crates/sequencer/src/rpc.rs
```diff
@@ -108,39 +131,64 @@ pub(crate) fn create_rpc_module<
...
+                Ok(())
+            }
+            Err(e) => {
+                error!("Error processing deposit tx: {:?}", e);
+                Err(e.into())
             }
+        }
+    }
+
+    async fn citrea_test_publish_block(&self) -> RpcResult<()> {
```
rakanalh, line 166:
```suggestion
    async fn publish_block(&self) -> RpcResult<()> {
```

rakanalh (approved):
NICE!

crates/sequencer/src/rpc.rs
rakanalh: Maybe `publish_test_block` would be okay

ercecan (approved):
🔥

PR #1277, opened 2024-10-07
E2E tests using citrea-e2e

### LINKED ISSUE #1240: E2E Extract framework to dedicated repository

jfldde:
# Description

- Extract e2e-framework to https://github.com/chainwayxyz/citrea-e2e

## Linked Issues
- Fixes #1240

## Notes

- `citrea-e2e` does not default to `./target/debug/citrea` anymore. You'll ever need to set CITREA env or it will try to default to `which citrea`. If citrea is not found through CITREA or in PATH, tests will fail to run
- Citrea configs are re-implemented in `citrea-e2e` and will need to be kept in line.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1217,7 +1217,7 @@ dependencies = [
  "bitcoincore-rpc",
  "borsh",
  "brotli",
- "citrea-primitives 0.5.0-rc.1",
+ "citrea-primitives",
  "futures",
  "hex",
  "pin-project",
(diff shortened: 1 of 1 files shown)


ceyhunsen:
About reimplementing some of stuff, like configs, again in citrea-e2e repo: Because bridge backend is not written in Rust, I had to rewrite configuration options in citrea-e2e. Probably will also write configuration options for Clementine too (because of the cyclic dependencies).

I am pessimistic about finding another way to use both Citrea and Clementine code in that repo, though. Looks to me that our current best bet is rewriting some stuff in citrea-e2e and avoid cyclic dependencies.

Cargo.lock
yaziciahmet, line 1: I would say let's target nightly in citrea-e2e and then merge this in. Lockfile got really ugly with 2 different versions of citrea crates.

Or... Duplicate the config structs and couple of traits we have in citrea-e2e repo so it doesn't have any citrea dependency, and it becomes something like a wrapper around the node APIs and a node process manager.

WDYT?

yaziciahmet (approved):

PR #1280, opened 2024-10-07
Erce/prover generate input rpc

### LINKED ISSUE #1276: Prover rpcs
We need an rpc for generating input given l1 block. It takes l1 height, and creates StateTransitionData based on all the commitments on that l1 block, serializes it with borsh and encodes it in hex.
This is needed for running the same input on different provers

ercecan:
## Linked Issues
- Fixes #1276

### DIFF
--- a/crates/common/src/utils.rs
+++ b/crates/common/src/utils.rs
@@ -4,12 +4,10 @@ use std::sync::Arc;
 use anyhow::anyhow;
 use borsh::de::BorshDeserialize;
 use jsonrpsee::core::DeserializeOwned;
-use sov_db::ledger_db::ProverLedgerOps;
-use sov_db::ledger_db::SharedLedgerOps;
+use sov_db::ledger_db::{ProverLedgerOps, SharedLedgerOps};
 use sov_db::schema::types::BatchNumber;
-use sov_rollup_interface::da::DaSpec;
 use sov_rollup_interface::da::{
-    BlobReaderTrait, BlockHeaderTrait, DaDataBatchProof, SequencerCommitment,
+    BlobReaderTrait, BlockHeaderTrait, DaDataBatchProof, DaSpec, SequencerCommitment,
 };
 use sov_rollup_interface::rpc::SoftConfirmationStatus;
 use sov_rollup_interface::services::da::{DaService, SlotData};
--- a/crates/common/src/utils.rs
+++ b/crates/common/src/utils.rs
@@ -1,10 +1,30 @@
-use std::collections::{HashMap, HashSet};
+use std::collections::{HashMap, HashSet, VecDeque};
+use std::sync::Arc;
 
+use anyhow::anyhow;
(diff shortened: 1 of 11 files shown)


crates/common/src/utils.rs
```diff
@@ -89,3 +95,112 @@ pub fn check_l2_range_exists<DB: SharedLedgerOps>(
...
+    ledger_db: &DB,
+    l1_block_cache: &Arc<Mutex<L1BlockCache<Da>>>,
+) -> Result<CommitmentStateTransitionData<Witness, Da>, anyhow::Error> {
+    let mut state_transition_witnesses: VecDeque<Vec<Witness>> = VecDeque::new();
+    let mut soft_confirmations: VecDeque<Vec<SignedSoftConfirmation>> = VecDeque::new();
+    let mut da_block_headers_of_soft_confirmations: VecDeque<
+        Vec<<<Da as DaService>::Spec as DaSpec>::BlockHeader>,
+    > = VecDeque::new();
+    for sequencer_commitment in sequencer_commitments.to_owned().iter() {
+        // get the l2 height ranges of each seq_commitments
```
eyusufatik, line 143: this is not common?

only the prover is going to use it

eyusufatik (changes requested):

crates/common/src/utils.rs
```diff
@@ -89,3 +79,31 @@ pub fn check_l2_range_exists<DB: SharedLedgerOps>(
...
+    da_data.iter_mut().for_each(|blob| {
+        blob.full_data();
+    });
+    da_data.iter_mut().for_each(|tx| {
+        let data = DaDataBatchProof::try_from_slice(tx.full_data());
+        // Check for commitment
+        if tx.sender().as_ref() == sequencer_da_pub_key {
+            if let Ok(DaDataBatchProof::SequencerCommitment(seq_com)) = data {
+                sequencer_commitments.push(seq_com);
+            } else {
```
rakanalh, line 99: I don't think this else branch is needed. Because since this call is `extract_sequencer_commitment`, if it contains proofs, we don't really care about them. So the warn message is going to be misleading.

rakanalh: We had it from before but the same concept applies.

crates/prover/src/rpc.rs
```diff
@@ -0,0 +1,288 @@
...
+                )
+            })?;
+
+        let mut da_data: Vec<<<Da as DaService>::Spec as DaSpec>::BlobTransaction> =
+            self.context.da_service.extract_relevant_blobs(&l1_block);
+
+        // if we don't do this, the zk circuit can't read the sequencer commitments
+        da_data.iter_mut().for_each(|blob| {
+            blob.full_data();
+        });
```
rakanalh, line 93: We're already doing this inside `extract_sequencer_commitments. Isn't this redundant?

eyusufatik (approved):

PR #1286, opened 2024-10-07
Move node configs to a separate crate

### LINKED ISSUE #1275: Move node config structs from sov-stf-runner to citrea/node-configs

exeokan:
# Description
Moves `FullNodeConfig`, `ProverConfig` and related config structs to a new `node-configs` crate.

## Linked Issues
- Fixes #1275


crates/node-configs/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+serde = { workspace = true }
+toml = { workspace = true }
+tracing = { workspace = true }
+
+[dev-dependencies]
+sov-mock-da = { path = "../sovereign-sdk/adapters/mock-da", features = ["native"] }
+tempfile = { workspace = true }
+
+[features]
+native = []
```
eyusufatik, line 30: you don't need native feature.

realize there is nothing that's not behind native feature in the crate

so no zk code is importing anything from here. only already-"native" crates importing from here. so no need for the feature.

crates/sequencer/src/config.rs
eyusufatik, line 1: let's move this as well

eyusufatik (changes requested):

crates/node-configs/Cargo.toml
exeokan: got it

eyusufatik (approved):

rakanalh (changes requested):
Instead of introducing a new crate called `node-configs` which i disagree with, we can move those configs into `common`.

eyusufatik:
> Instead of introducing a new crate called `node-configs` which i disagree with, we can move those configs into `common` or `primitives`. Use `native` feature where appropriate.

esp for things like #1291, I think having a crate is useful

eyusufatik:
discussed with @rakanalh , common looks like a better crate for this

eyusufatik (approved):

rakanalh (approved):

PR #1302, opened 2024-10-08
Get fee recommendation from mempool space

### LINKED ISSUE #1274: Switch to mempool.space fee estimation

ercecan:
# Description
Mempool space fee estimation

## Linked Issues
- Fixes #1274


kpp (approved):
Lgtm but maybe this url should be set via config?

crates/bitcoin-da/src/service.rs
```diff
@@ -56,6 +56,9 @@ use crate::REVEAL_OUTPUT_AMOUNT;
 pub const FINALITY_DEPTH: u64 = 8; // blocks
 const POLLING_INTERVAL: u64 = 10; // seconds
 
+const MEMPOOL_SPACE_RECOMMENDED_FEE_ENDPOINT: &str =
+    "https://mempool.space/testnet4/api/v1/fees/recommended";
+
```
yaziciahmet, line 61: IMO this should be set via config and it should be optional. If the url is not passed in config, then estimate_smart_fee should still be used

eyusufatik, line 61: you should construct the url with the network field in daservice

eyusufatik (changes requested):

crates/bitcoin-da/src/service.rs
ercecan: talked with @eyusufatik and decided it was not necessary, check the new impl after last commit

eyusufatik (approved):

PR #1312, opened 2024-10-09
Fetch smart fee only if mempool returns None

yaziciahmet:
# Description

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/bitcoin-da/src/service.rs
+++ b/crates/bitcoin-da/src/service.rs
@@ -548,11 +548,10 @@ impl BitcoinService {
     #[instrument(level = "trace", skip_all, ret)]
     pub async fn get_fee_rate_as_sat_vb(&self) -> Result<u64, anyhow::Error> {
         // If network is regtest or signet, mempool space is not available
-        let smart_fee = get_fee_rate_from_mempool_space(self.network).await?.or(self
-            .client
-            .estimate_smart_fee(1, None)
-            .await?
-            .fee_rate);
+        let smart_fee = match get_fee_rate_from_mempool_space(self.network).await? {
+            Some(fee_rate) => Some(fee_rate),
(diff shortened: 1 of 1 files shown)


crates/bitcoin-da/src/service.rs
```diff
@@ -548,11 +548,10 @@ impl BitcoinService {
     #[instrument(level = "trace", skip_all, ret)]
     pub async fn get_fee_rate_as_sat_vb(&self) -> Result<u64, anyhow::Error> {
         // If network is regtest or signet, mempool space is not available
-        let smart_fee = get_fee_rate_from_mempool_space(self.network).await?.or(self
-            .client
-            .estimate_smart_fee(1, None)
-            .await?
-            .fee_rate);
+        let smart_fee = match get_fee_rate_from_mempool_space(self.network).await? {
```
eyusufatik, line 551: isn't this the same thing?

is the change to make the or part lazy maybe? if so there should be another function for this i believe

yaziciahmet: previous one is not lazy and calculated immediately even before getting fee from the mempool. there is a method called `or_else`which is the lazy version of `or`, but `or_else` does not accept async code.

kpp: https://docs.rs/futures/latest/futures/future/trait.TryFutureExt.html#method.or_else

yaziciahmet: that's a cool feat I just learned :) but I think this produces something less readable for the same job:
```
        let smart_fee = get_fee_rate_from_mempool_space(self.network)
            .or_else(|_| async move {
                Ok::<_, anyhow::Error>(self.client.estimate_smart_fee(1, None).await?.fee_rate)
            })
            .await?;
```

imo match statement is better

PR #1320, opened 2024-10-10
Retrieve configs from env

### LINKED ISSUE #1291: Retreive configs from env if .toml is not present
TODOs:
- [x] Create `FromEnv` trait, which has function `from_env() -> Result<T>`. When implemented for a struct, the function reads field of a struct form environment and returns it.
- [x] For `SequencerConfig`, `FullNodeConfig`, `BatchProverConfig`, and `LightClientProverConfig`, implement `FromEnv`
- [x] In main, if a config file is not present, try calling `from_env()` for configs.

exeokan:
# Description
Creates `FromEnv` trait, which has function `from_env() -> Result<T>`. When implemented for a struct, the function reads field of a struct form environment and returns it.
Implements it for `SequencerConfig`, `FullNodeConfig`, `ProverConfig`, and related configs.
Modifies CLI options such that configs can be read from file or the environment.
Note: In CLI options, `--sequencer-config-path` and `--prover-config-path` are renamed to `--sequencer` and `--prover`

## Linked Issues
- Fixes #1291 
## Testing
Wrote new tests: `test_correct_full_node_config_from_env`, `test_correct_sequencer_config_from_env`, `test_correct_prover_config_from_env`

### DIFF
--- a/bin/citrea/src/main.rs
+++ b/bin/citrea/src/main.rs
@@ -29,9 +29,9 @@ struct Args {
     #[arg(long, default_value = "mock")]
     da_layer: SupportedDaLayer,
 
-    /// The path to the rollup config.
+    /// The path to the rollup config, if a string is provided, it will be used as the path to the rollup config, otherwise environment variables will be used.
     #[arg(long, default_value = "resources/configs/mock/rollup_config.toml")]
-    rollup_config_path: String,
+    rollup_config_path: Option<Option<String>>,
 
     /// The option to run the node in sequencer mode, if a string is provided, it will be used as the path to the sequencer config, otherwise environment variables will be used.
     #[arg(long, conflicts_with = "prover")]
@@ -72,7 +72,7 @@ async fn main() -> Result<(), anyhow::Error> {
     };
     initialize_logging(logging_level);
 
-    let rollup_config_path = args.rollup_config_path.as_str();
+    let rollup_config_path = args.rollup_config_path.unwrap();
 
     let sequencer_config = match args.sequencer {
         Some(Some(path)) => Some(
@@ -133,7 +133,7 @@ async fn start_rollup<S, DaC>(
         <S as RollupBlueprint>::NativeContext,
         <S as RollupBlueprint>::DaSpec,
     >>::GenesisPaths,
-    rollup_config_path: &str,
+    rollup_config_path: Option<String>,
     prover_config: Option<ProverConfig>,
     sequencer_config: Option<SequencerConfig>,
 ) -> Result<(), anyhow::Error>
@@ -142,12 +142,12 @@ where
     S: CitreaRollupBlueprint<DaConfig = DaC>,
     <<S as RollupBlueprint>::NativeContext as Spec>::Storage: NativeStorage,
 {
-    let rollup_config: FullNodeConfig<DaC> =
-        from_toml_path(rollup_config_path).unwrap_or_else(|_| {
-            FullNodeConfig::<DaC>::from_env()
-                .context("Failed to read rollup configuration")
-                .unwrap()
-        });
+    let rollup_config: FullNodeConfig<DaC> = match rollup_config_path {
+        Some(path) => from_toml_path(path)
+            .context("Failed to read rollup configuration from the config file")?,
+        None => FullNodeConfig::from_env()
+            .context("Failed to read rollup configuration from the environment")?,
+    };
     let rollup_blueprint = S::new();
 
     if let Some(sequencer_config) = sequencer_config {
--- a/bin/citrea/src/main.rs
+++ b/bin/citrea/src/main.rs
@@ -3,7 +3,7 @@ use core::fmt::Debug as DebugTrait;
 use anyhow::Context as _;
 use bitcoin_da::service::BitcoinServiceConfig;
 use citrea::{initialize_logging, BitcoinRollup, CitreaRollupBlueprint, MockDemoRollup};
-use citrea_common::{from_toml_path, FullNodeConfig, ProverConfig, SequencerConfig};
+use citrea_common::{from_toml_path, FromEnv, FullNodeConfig, ProverConfig, SequencerConfig};
 use citrea_stf::genesis_config::GenesisPaths;
 use clap::Parser;
 use sov_mock_da::MockDaConfig;
@@ -76,15 +76,18 @@ async fn main() -> Result<(), anyhow::Error> {
 
     let sequencer_config: Option<SequencerConfig> =
         args.sequencer_config_path.clone().map(|path| {
-            from_toml_path(path)
-                .context("Failed to read sequencer configuration")
-                .unwrap()
+            from_toml_path(path).unwrap_or_else( |_| {
+                SequencerConfig::from_env()
+                    .context("Failed to read sequencer configuration")
+                    .unwrap()
+            })
         });
-
     let prover_config: Option<ProverConfig> = args.prover_config_path.clone().map(|path| {
-        from_toml_path(path)
-            .context("Failed to read prover configuration")
-            .unwrap()
+        from_toml_path(path).unwrap_or_else(|_| {
+            ProverConfig::from_env()
+                .context("Failed to read prover configuration")
+                .unwrap()
+        })
     });
 
     if prover_config.is_some() && sequencer_config.is_some() {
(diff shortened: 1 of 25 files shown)


crates/common/src/config.rs
```diff
@@ -49,7 +69,32 @@ pub struct RpcConfig {
...
+            toml_string.push_str(&format!("max_request_body_size = {}\n", val));
+        }
+        if let Ok(val) = std::env::var("RPC_MAX_RESPONSE_BODY_SIZE") {
+            toml_string.push_str(&format!("max_response_body_size = {}\n", val));
+        }
+        if let Ok(val) = std::env::var("RPC_BATCH_REQUESTS_LIMIT") {
+            toml_string.push_str(&format!("batch_requests_limit = {}\n", val));
+        }
+        if let Ok(val) = std::env::var("RPC_ENABLE_SUBSCRIPTIONS") {
+            toml_string.push_str(&format!("enable_subscriptions = {}\n", val));
```
eyusufatik, line 90: why go this way instead of just returning Self {} with std::env::var's

if you can't figure our how to handle optional fields of configs here's what you can do:

```rust
let optional_env_var = std::env::var("SOME_OPTIONAL_ENV_VAR").ok();
```

crates/common/src/config.rs
```diff
@@ -21,7 +25,23 @@ pub struct RunnerConfig {
...
+            include_tx_body: std::env::var("INCLUDE_TX_BODY")?.parse()?,
+            accept_public_input_as_proven: std::env::var("ACCEPT_PUBLIC_INPUT_AS_PROVEN")
+                .ok()
+                .map(|val| val.parse().ok())
+                .flatten(),
+            sync_blocks_count: toml::from_str(&std::env::var("SYNC_BLOCKS_COUNT").unwrap_or("".to_string()))?,
+            pruning_config: std::env::var("PRUNING_CONFIG")
+                .ok()
+                .map(|val| toml::from_str(&val).ok())
+                .flatten(),
```
eyusufatik, line 41: like my last command, 

don't expect the whole config struct to be in env, but assume individual fields of that config is in env

you can implement FromEnv for all configs used here, 

then when constructing ConfigA which has a field ConfigB

you would do 

```rust
impl FromEnv for ConfigA {
    fn from_env() -> ... {
        Self {
            ...
            config_b: ConfigB::from_env()
        }
    }
}
```

eyusufatik (changes requested):

crates/common/src/config.rs
exeokan: I wanted to use serde default values and not duplicate const function names, but we discussed with Erce to move on as you said

bin/citrea/src/main.rs
eyusufatik, line 1: now we need two new args 

--sequencer and --prover

the reasoning is:
- we normally assumed that if --sequencer-config-path is passed to the binary, we were launching a sequencer, and if reading that toml failed, the program would also fail.
- now we support launching a sequencer without passing --seqeuncer-config-path but setting environment variables.
- if the env vars were set wrong somehow, maybe due to a misspelling, `SequencerConfig::from_env()` will also fail. But we cannot panic on this, so the binary will start as a full node :D

So I suggest these changes:
- remove sequencer_config_path and prover_config_path from args.
- add `sequencer: Option<String>` and `prover: Option<String>` to args.
- modify existing code such that:
  - if one of sequencer or prover CLI argument is passed, we launch a sequencer or a prover.
  - we first check if a String is passed along with the --sequencer or --prover, if not passed, we do `::from_env()` if that also fails, we panic, or return Err on main.
  - we assert that at least one of --sequencer and --prover is None, and only launch a full node when both is None.
  - we always panic or return Err if FullNode cannot be constructed from rollup_config_path or from_env

crates/common/src/config.rs
```diff
@@ -124,7 +222,20 @@ pub struct FullNodeConfig<BitcoinServiceConfig> {
...
+    fn from_env() -> anyhow::Result<Self> {
+        Ok(Self {
+            rpc: RpcConfig::from_env()?,
+            storage: StorageConfig::from_env()?,
+            runner: RunnerConfig::from_env().ok(),
+            da: BitcoinServiceConfig::from_env()?,
+            public_keys: RollupPublicKeys::from_env()?,
+        })
+    }
+}
```
eyusufatik, line 238: this can be implemented as such so that it supports all DA adapters:

```rust
impl<T FromEnv for FullNodeConfig<T>
where
    T: FromEnv + DeserializeOwned, // not sure if DeserializeOwned is needed
{
        Ok(Self {
            rpc: RpcConfig::from_env()?,
            storage: StorageConfig::from_env()?,
            runner: RunnerConfig::from_env().ok(),
            da: T::from_env()?,
            public_keys: RollupPublicKeys::from_env()?,
        })
    }
}
```

crates/common/src/config.rs
```diff
@@ -6,6 +6,39 @@ use citrea_pruning::PruningConfig;
...
+impl FromEnv for sov_mock_da::MockDaConfig {
+    fn from_env() -> anyhow::Result<Self> {
+        Ok(Self {
+            sender_address: std::env::var("SENDER_ADDRESS")?.parse()?,
+            db_path: std::env::var("DB_PATH")?.into(),
+        })
+    }
+}
```
eyusufatik, line 27: should be implemented in mock da

crates/common/src/config.rs
```diff
@@ -6,6 +6,39 @@ use citrea_pruning::PruningConfig;
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use sov_stf_runner::ProverGuestRunConfig;
```
eyusufatik, line 8: this would also be implemented in ProverGuestRunConfig

eyusufatik (changes requested):

eyusufatik:
should be merged after `v0.5.4` is out

so putting hold merge label

crates/common/src/config.rs
exeokan: I actually implemented it as generic but the naming was confusing, renaming it to `DaC`

crates/common/src/config.rs
exeokan: kept this version since implementing it that way causes cyclic dependency issues in the tests

crates/common/src/config.rs
exeokan: similarly, implementing `FromEnv` for `ProverGuestRunConfig` and `PruningConfig` in their own crate causes cyclic dependency here, lmk if there is a better way of implementing these

bin/citrea/src/main.rs
```diff
@@ -33,13 +33,13 @@ struct Args {
     #[arg(long, default_value = "resources/configs/mock/rollup_config.toml")]
     rollup_config_path: String,
```
eyusufatik: you can also change rollup config path to Option<String> and pull from env if no path is given

keep the name as is

eyusufatik (changes requested):

rakanalh (approved):

bin/citrea/src/main.rs
```diff
@@ -29,17 +29,17 @@ struct Args {
     #[arg(long, default_value = "mock")]
     da_layer: SupportedDaLayer,
 
-    /// The path to the rollup config.
+    /// The path to the rollup config, if a string is provided, it will be used as the path to the rollup config, otherwise environment variables will be used.
     #[arg(long, default_value = "resources/configs/mock/rollup_config.toml")]
-    rollup_config_path: String,
+    rollup_config_path: Option<Option<String>>,
```
eyusufatik: can you remove the default value for this btw?

bin/citrea/src/main.rs
```diff
@@ -72,20 +72,30 @@ async fn main() -> Result<(), anyhow::Error> {
     };
     initialize_logging(logging_level);
 
-    let rollup_config_path = args.rollup_config_path.as_str();
+    let rollup_config_path = args.rollup_config_path.unwrap();
```
eyusufatik: i think no unwrap here, if no --rollup-config-path is passed, try to do from_env() directly, if that fails, regardless of launch mode, return Err

so rollup config path is Option<String> and not forced as an argument.

docs/run-dev.md
eyusufatik, line 1: also update other docs

eyusufatik (changes requested):

bin/citrea/src/main.rs
exeokan: got it

eyusufatik (approved):

jfldde (approved):

PR #1321, opened 2024-10-10
Only spawn_blocking for tx creation in DA queue

yaziciahmet:
# Description

It doesn't make much sense to spawn the whole DA queue in a spawn_blocking, due to the fact that it waits empty for messages most of the time, hence, it doesn't need a whole thread to itself. Instead, with this PR, queue is spawned as a tokio runtime worker, so it doesn't occupy any resources while waiting, but only acquires a thread with spawn_blocking when it's time to create the transaction and find the right prefix (or nonce).

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


rakanalh (approved):
👍

PR #1323, opened 2024-10-10
fix bugs regarding tx gas limit

### LINKED ISSUE #1322: Transaction never included in block if gas limit > 8m
Tx not included if gas usage < 8m but tx was signed with limit of > 8m.

Should not happen.

eyusufatik:
# Description
The real "bug" turned out to be something different from my initial thoughts in the issue.

## Linked Issues
Fixes #1322 

## Testing

## Docs

### DIFF
--- a/crates/evm/src/query.rs
+++ b/crates/evm/src/query.rs
@@ -57,7 +57,7 @@ const ESTIMATE_GAS_ERROR_RATIO: f64 = 0.015;
 #[derive(Copy, Clone, Eq, PartialEq, Debug)]
 pub(crate) struct EstimatedTxExpenses {
     /// Evm gas used.
-    gas_used: U64,
+    pub gas_used: U64,
     /// Base fee of the L2 block when tx was executed.
     base_fee: U256,
     /// L1 fee.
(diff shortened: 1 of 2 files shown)


crates/sequencer/src/mempool.rs
```diff
@@ -85,6 +85,9 @@ impl<C: sov_modules_api::Context> CitreaMempool<C> {
         let validator = TransactionValidationTaskExecutor::eth_builder(Arc::new(chain_spec))
             .no_cancun()
             .no_eip4844()
+            // TODO: if we ever increase block gas limits, we need to pull this from
+            // somewhere else
+            .set_block_gas_limit(evm_config.block_gas_limit)
```
eyusufatik, line 90: bc we forgot to add this, the mempool thought our block gas limit is 30m, then accepted txs with over 8m gas.

but they can never be included in  a block

crates/evm/src/query.rs
```diff
@@ -735,7 +760,36 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+        // case the mempool will reject the transaction and even if it didn't, the sequencer
+        // wouldn't put it into a block since it's against EVM rules to hava tx that
+        // has more gas limit than the block.
+        //
+        // But in the case where the gas used is already bigger than the block gas limit
+        // we want to return the gas estimation as is since the mempool will reject it
+        // anyway.
+        if estimated.gas_used > block_gas_limit {
+            Ok(estimated.gas_with_l1_overhead())
+        } else {
```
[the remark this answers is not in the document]
eyusufatik: yes

since these are two seperate rpc methods

crates/evm/src/query.rs
```diff
@@ -735,7 +735,17 @@ impl<C: sov_modules_api::Context> Evm<C> {
         working_set: &mut WorkingSet<C>,
     ) -> RpcResult<reth_primitives::U256> {
         let estimated = self.estimate_tx_expenses(request, block_number, working_set)?;
-        Ok(estimated.gas_with_l1_overhead())
+
+        // TODO: this assumes all blocks have the same gas limit
+        // if gas limit ever changes this should be updated
+        let last_block = self
```
kpp, line 741: Why is it last_block, not a block by block_number?

eyusufatik: all blocks have the same gas limit on citrea.

until we do a fork and increase or decrease the gas limit, it doesnt matter which block we use there.

but once we do, we have to change that part, hence the todo

crates/evm/src/query.rs
```diff
@@ -735,7 +735,17 @@ impl<C: sov_modules_api::Context> Evm<C> {
         working_set: &mut WorkingSet<C>,
     ) -> RpcResult<reth_primitives::U256> {
         let estimated = self.estimate_tx_expenses(request, block_number, working_set)?;
-        Ok(estimated.gas_with_l1_overhead())
+
+        // TODO: this assumes all blocks have the same gas limit
```
ercecan, line 739: maybe open an issue and ref here if you haven't already

ercecan (approved):

kpp (approved):

PR #1324, opened 2024-10-10
Use spawn_blocking in bitcoin service not in tx builders

kpp:
# Description

Follow up https://github.com/chainwayxyz/citrea/pull/1321

`create_zkproof_transactions` and `create_seqcommitment_transactions` are pure functions, no need to wrap them in async to spawn_blocking inside. We may spawn_blocking in BitcoinService, to block while running those fns.


rakanalh (approved):
Also makes sense. As long as we wrap blocking tasks only with `spawn_blocking` rather than async + sync code.

yaziciahmet (approved):

PR #1326, opened 2024-10-11
Remove unused risc0 host & introduce local proving

yaziciahmet:
# Description

Removes a Risc0Host that is not used in the code other than tests.

Also introduce local proving to Risc0BonsaiHost. Added a test for it as well, but it timed-out even with 3 hours wait in my local machine. Hence I marked it as ignore. In the future when we add metal or cuda support to local proving, it still could be useful to run locally.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


kpp (approved):
Less code > more code

eyusufatik:
tbf, I would keep the local risc0 host :D

how are we gonna do local proving if we delete it?

yaziciahmet:
> tbf, I would keep the local risc0 host :D
> 
> how are we gonna do local proving if we delete it?

can't we use a single host (existing Risc0BonsaiHost) for that? if there is no bonsai-key, and specified proving mode is `prove`, we can configure it to do local proving.

eyusufatik (changes requested):
requesting changes so that this doesnt get merged accidentally.

eyusufatik:
One more thing to improve can be how Risc0Host was implemented. 

I guess the code was written for older versions of risc0. Checkout the newer examples from risc0 and try to see if we can make better use of it

rakanalh (approved):

bin/citrea/tests/bitcoin_e2e/prover_test.rs
```diff
@@ -307,14 +314,101 @@ impl TestCase for SkipPreprovenCommitmentsTest {
...
+    fn sequencer_config() -> SequencerConfig {
+        SequencerConfig {
+            // Made this 1 or-else proving takes forever
+            min_soft_confirmations_per_commitment: 1,
+            ..Default::default()
+        }
+    }
+
+    async fn run_test(&mut self, f: &mut TestFramework) -> Result<()> {
+        citrea::initialize_logging(tracing::Level::INFO);
```
eyusufatik: comment out

bin/citrea/tests/bitcoin_e2e/prover_test.rs
```diff
@@ -307,14 +314,101 @@ impl TestCase for SkipPreprovenCommitmentsTest {
...
+                ("BONSAI_API_URL", ""),
+                ("BONSAI_API_KEY", ""),
+            ],
+            ..Default::default()
+        }
+    }
+
+    fn batch_prover_config() -> BatchProverConfig {
+        BatchProverConfig {
+            proving_mode: ProverGuestRunConfig::Prove,
```
eyusufatik, line 350: very very against proving on CI

eyusufatik (changes requested):

crates/risc0-bonsai/src/host.rs
```diff
@@ -87,7 +88,6 @@ pub struct Risc0BonsaiHost<'a> {
     env: Vec<u8>,
     image_id: Digest,
     client: Option<Client>,
-    last_input_id: Option<String>,
```
eyusufatik, line 90: good change :)

eyusufatik (changes requested):

bin/citrea/tests/bitcoin_e2e/prover_test.rs
yaziciahmet: agreed 100%, that's why I ignored. but still didn't wanna delete an existing test case in case we make it work with cuda and at least it is somewhat feasible to wait.

bin/citrea/tests/bitcoin_e2e/prover_test.rs
yaziciahmet: done

bin/citrea/tests/bitcoin_e2e/prover_test.rs
eyusufatik: didnt realize it was ignored

eyusufatik (approved):

PR #1330, opened 2024-10-11
Fix mempool space inf loop bug

ercecan:
Fixes bug where if mempool space returns error causes inf loop in sequencer

### DIFF
--- a/crates/bitcoin-da/src/service.rs
+++ b/crates/bitcoin-da/src/service.rs
@@ -560,9 +560,25 @@ impl BitcoinService {
     #[instrument(level = "trace", skip_all, ret)]
     pub async fn get_fee_rate_as_sat_vb(&self) -> Result<u64, anyhow::Error> {
         // If network is regtest or signet, mempool space is not available
-        let smart_fee = match get_fee_rate_from_mempool_space(self.network).await? {
-            Some(fee_rate) => Some(fee_rate),
-            None => self.client.estimate_smart_fee(1, None).await?.fee_rate,
+        let smart_fee = match get_fee_rate_from_mempool_space(self.network).await {
+            Ok(Some(fee_rate)) => Some(fee_rate),
+            Err(e) => {
+                tracing::warn!(?e, "Failed to get fee rate from mempool.space");
+                match self.client.estimate_smart_fee(1, None).await {
+                    Ok(res) => res.fee_rate,
+                    Err(e) => {
+                        error!(?e, "Failed to get fee rate from Bitcoin node");
+                        None
+                    }
+                }
+            }
+            Ok(None) => match self.client.estimate_smart_fee(1, None).await {
+                Ok(res) => res.fee_rate,
+                Err(e) => {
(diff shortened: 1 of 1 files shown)


yaziciahmet:
On a second thought, maybe this is not the right approach to fix this issue. Technically, anything within that loop can fail, db file may be corrupted, or any other rpc call to Bitcoin might fail. And all of these errors within that loop would result in an infinite loop. I don't think the best solution is to try to ignore the error or trying to replace it with a default value. But rather have some sort of exponential backoff retry mechanism within the loop, and have great logging + monitoring + alert systems.

crates/bitcoin-da/src/service.rs
```diff
@@ -560,9 +560,25 @@ impl BitcoinService {
     #[instrument(level = "trace", skip_all, ret)]
     pub async fn get_fee_rate_as_sat_vb(&self) -> Result<u64, anyhow::Error> {
         // If network is regtest or signet, mempool space is not available
-        let smart_fee = match get_fee_rate_from_mempool_space(self.network).await? {
-            Some(fee_rate) => Some(fee_rate),
-            None => self.client.estimate_smart_fee(1, None).await?.fee_rate,
+        let smart_fee = match get_fee_rate_from_mempool_space(self.network).await {
+            Ok(Some(fee_rate)) => Some(fee_rate),
+            Err(e) => {
```
eyusufatik, line 565: why not Err(e) | Ok(None) here?

crates/bitcoin-da/src/service.rs
```diff
@@ -560,9 +560,25 @@ impl BitcoinService {
...
+                        error!(?e, "Failed to get fee rate from Bitcoin node");
+                        None
+                    }
+                }
+            }
+            Ok(None) => match self.client.estimate_smart_fee(1, None).await {
+                Ok(res) => res.fee_rate,
+                Err(e) => {
+                    error!(?e, "Failed to get fee rate from Bitcoin node");
+                    None
```
eyusufatik: also on estimate smart fee Err case, we have to return Err at that momen instead of defaulting to 1 sat/vb

crates/bitcoin-da/src/service.rs
ercecan: In order to log the error I need to use the inner value of Err, in that case I cannot use the inner value, it raises a compiler error saying sth like it may not be Err so you cannot use the inner value

eyusufatik:
> On a second thought, maybe this is not the right approach to fix this issue. Technically, anything within that loop can fail, db file may be corrupted, or any other rpc call to Bitcoin might fail. And all of these errors within that loop would result in an infinite loop. I don't think the best solution is to try to ignore the error or trying to replace it with a default value. But rather have some sort of exponential backoff retry mechanism within the loop, and have great logging + monitoring + alert systems.

That's correct.

For the fee stuff, it's already retried whenever it's used though.

And chances of `estimatesmartfee` failing is very low.

eyusufatik (approved):

PR #1335, opened 2024-10-14
Introduce prove RPC method

### LINKED ISSUE #1285: RPC for manual proof production, in case we need to hand trigger stuff
We are already implementing `generate_input` RPC for prover, which with a given L1 block height/hash, the prover returns the input to be given to the guest code.

We might also benefit from manual proof production trigger on the prover.

This will come handy esp. on testnet where we have proof sampling.

rakanalh:
# Description

- [x] Add the RPC method
- [x] Write a test 

## Linked Issues
- Fixes #1285


crates/prover/src/rpc/mod.rs
```diff
@@ -0,0 +1,301 @@
...
+    pub da_service: Arc<Da>,
+    pub prover_service: Arc<Ps>,
+    pub ledger: DB,
+    pub sequencer_da_pub_key: Vec<u8>,
+    pub sequencer_pub_key: Vec<u8>,
+    pub l1_block_cache: Arc<Mutex<L1BlockCache<Da>>>,
+    pub code_commitments_by_spec: HashMap<SpecId, Vm::CodeCommitment>,
+    pub(crate) c: PhantomData<fn() -> C>,
+    pub(crate) vm: PhantomData<fn() -> Vm>,
+}
```
ercecan: Instead of adding stateroot and witness generics, would 
`StateRoot = <<<C as Spec>::Storage as Storage>::Root>` and `Witness = <<<C as Spec>::Storage as Storage>::Witness>`
work just the same?
I believe it would be cleaner and you wouldn't need to add static bounds to stf

crates/prover/src/rpc/utils.rs
```diff
@@ -0,0 +1,234 @@
...
+use sov_rollup_interface::zk::{StateTransitionData, ZkvmHost};
+use sov_stf_runner::ProverService;
+use tracing::error;
+
+use super::RpcContext;
+use crate::da_block_handler::{
+    break_sequencer_commitments_into_groups, get_state_transition_data_from_commitments,
+};
+
+pub(super) async fn get_sequencer_commitments_for_proving<C, Da, Ps, Vm, DB, StateRoot, Witness>(
```
eyusufatik: we should also use this function in the da block handler too imo

or else whenever we change anything regarding the order of processing sequencer commitments etc.

now we have to update these places:
- prover da block handler
- full node da block handler
- the function the batch proof zk guest calls (cant remember the name now)
- *and this function*

eyusufatik (changes requested):
everything else LGTM

crates/prover/src/rpc/mod.rs
rakanalh: For some reason, i could not do that. Some parts expect `<Ps as ProverService>::StateRoot` and witness and others expect from `C`... this was the way to unify them.

crates/prover/src/rpc/utils.rs
rakanalh: Done

crates/common/src/da.rs
```diff
@@ -37,3 +42,64 @@ pub async fn get_da_block_at_height<Da: DaService>(
...
+
+    if sequencer_commitments.is_empty() {
+        return vec![];
+    }
+
+    // Make sure all sequencer commitments are stored in ascending order.
+    // We sort before checking ranges to prevent substraction errors.
+    sequencer_commitments.sort();
+
+    sequencer_commitments
```
yaziciahmet: If sequencer commitments is empty, sort will be no-op anyways. No need for the if statement.

crates/prover/src/runner.rs
```diff
@@ -164,14 +164,17 @@ where
...
             ledger: self.ledger_db.clone(),
             da_service: self.da_service.clone(),
             sequencer_da_pub_key: self.sequencer_da_pub_key.clone(),
             sequencer_pub_key: self.sequencer_pub_key.clone(),
             l1_block_cache: self.l1_block_cache.clone(),
-            phantom: std::marker::PhantomData,
+            prover_service: self.prover_service.clone(),
+            code_commitments_by_spec: self.code_commitments_by_spec.clone(),
+            c: std::marker::PhantomData,
+            vm: std::marker::PhantomData,
```
yaziciahmet: This might be a bit confusing. maybe phantom_c and phantom_vm?

crates/prover/src/da_block_handler.rs
```diff
@@ -171,121 +170,89 @@ where
...
+                    L1ProcessingError::NoSeqCommitments { l1_height } => {
+                        info!("No sequencer commitment found at height {}", l1_height,);
+                        self.ledger_db
+                                .set_last_scanned_l1_height(SlotNumber(l1_height))
+                                .unwrap_or_else(|_| {
+                                    panic!(
+                                        "Failed to put prover last scanned l1 height in the ledger db {}",
+                                        l1_height
+                                    )
+                                });
```
yaziciahmet: `.expect(format!("..."))`

crates/prover/src/da_block_handler.rs
```diff
@@ -171,121 +170,89 @@ where
...
+                            });
+
+                        self.pending_l1_blocks.pop_front();
+                        continue;
+                    }
+                    L1ProcessingError::L2RangeMissing {
+                        start_block_number,
+                        end_block_number,
+                    } => {
+                        error!("L2 range of commitments is not synced yet: {start_block_number} - {end_block_number}");
```
yaziciahmet: Maybe warn here?

crates/prover/src/proving.rs
```diff
@@ -0,0 +1,403 @@
...
+        return Err(L1ProcessingError::L2RangeMissing {
+            start_block_number,
+            end_block_number,
+        });
+    }
+
+    let (sequencer_commitments, preproven_commitments) =
+        filter_out_proven_commitments(&ledger, &sequencer_commitments).map_err(|e| {
+            error!("Error filtering out proven commitments: {:?}", e);
+            L1ProcessingError::Other(format!("{}", e))
```
yaziciahmet: Already logged where it's called. Can put the "Error filtering out..." message into the error as well.

crates/prover/src/rpc.rs
```diff
@@ -102,226 +157,126 @@ impl<C: sov_modules_api::Context, Da: DaService, DB: ProverLedgerOps + Send + Sy
...
-                ErrorObjectOwned::owned(
-                    INTERNAL_ERROR_CODE,
-                    INTERNAL_ERROR_MSG,
-                    Some(format!("{e}",)),
-                )
-            })?,
-            _ => vec![(0..=sequencer_commitments.len() - 1)],
-        };
+    async fn prove(&self, l1_height: u64, group_commitments: Option<bool>) -> RpcResult<()> {
+        debug!("Prover: prover_prove");
```
yaziciahmet: Should already be logged by the rpc middleware.

crates/prover/src/proving.rs
```diff
@@ -0,0 +1,403 @@
...
+{
+    let l1_height = l1_block.header().height();
+
+    let mut da_data: Vec<<<Da as DaService>::Spec as DaSpec>::BlobTransaction> =
+        da_service.extract_relevant_blobs(&l1_block);
+
+    // if we don't do this, the zk circuit can't read the sequencer commitments
+    da_data.iter_mut().for_each(|blob| {
+        blob.full_data();
+    });
```
yaziciahmet, line 65: I don't think this is needed as extract_sorted_sequencer_commitments already does this.

yaziciahmet (changes requested):
Why do we need this functionality? It seems pretty risky and not so useful to me tbh.

With the current rpc implementation, anyone can invoke a proving process. If someone is set out to bankrupt us, he can invoke proving the moment they see a sequencer commitment proof in DA, and force us to prove only 1 commitment, causing us a lot of money from both proving and paying fee to DA side.

Even if this problem was to be solved, I can't see the use of this functionality.

**UPDATE:**

I learned that the prover is not public to the internet. So a random person can not invoke this rpc. But I am still skeptical of this feature as this still relies on the human (us) intervention, and a small mistake can cause a lot. If we want to prove a certain commitment right away without waiting for sampling, it makes more sense to lower the sampling config number and restart the prover for one time.

yaziciahmet:
Also a question, what happens if prover is busy proving a commitment at l1 block 100, but there are commitments at 150 and 200. Someone invoked the commitment at l1 block 200? Do we break the assumption of ordered proofs on DA? If so it might be problematic on the light client and bridge side.

eyusufatik:
> Also a question, what happens if prover is busy proving a commitment at l1 block 100, but there are commitments at 150 and 200. Someone invoked the commitment at l1 block 200? Do we break the assumption of ordered proofs on DA, if so it might be problematic on the light client and bridge side.

prove rpc is useful when we do proof sampling and we find some interesting sequencer commitments that we want to prove.

So if there is no proof sampling we wont use the RPC

bin/citrea/tests/e2e/proving.rs
```diff
@@ -187,3 +187,178 @@ async fn full_node_verify_proof_and_store() {
...
+        .ledger_get_sequencer_commitments_on_slot_by_hash(third_block_hash.0)
+        .await
+        .unwrap()
+        .unwrap();
+    assert_eq!(commitments_hash, commitments);
+
+    let prover_proof = prover_node_test_client
+        .ledger_get_proofs_by_slot_height(3)
+        .await[0]
+        .clone();
```
eyusufatik, line 318: from this I understand that we record proofs and related info to ledger db, right? we should do it so good.

yaziciahmet:
> prove rpc is useful when we do proof sampling and we find some interesting sequencer commitments that we want to prove.
> So if there is no proof sampling we wont use the RPC

@eyusufatik 

> **UPDATE:**
> I learned that the prover is not public to the internet. So a random person can not invoke this rpc. But I am still skeptical of this feature as this still relies on the human (us) intervention, and a small mistake can cause a lot. If we want to prove a certain commitment right away without waiting for sampling, it makes more sense to lower the sampling config number and restart the prover for one time.

eyusufatik (approved):
great refactor in the latest commits.

eyusufatik:
> I learned that the prover is not public to the internet. So a random person can not invoke this rpc. But I am still skeptical of this feature as this still relies on the human (us) intervention, and a small mistake can cause a lot. If we want to prove a certain commitment right away without waiting for sampling, it makes more sense to lower the sampling config number and restart the prover for one time.

the thing is, that's also something we have to implement :) and it's not worth the time spent on shutting down the prover and restarting etc. 

having this RPC is really helpful while we still have proof sampling.

bin/citrea/tests/e2e/proving.rs
```diff
@@ -187,3 +187,178 @@ async fn full_node_verify_proof_and_store() {
     prover_node_task.abort();
     full_node_task.abort();
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_prover_prove_rpc() {
+    citrea::initialize_logging(tracing::Level::DEBUG);
```
kpp: Will panic if it was already init in this OS thread

bin/citrea/tests/e2e/proving.rs
```diff
@@ -187,3 +187,178 @@ async fn full_node_verify_proof_and_store() {
...
+
+    // Full node sync commitment block
+    test_client.send_publish_batch_request().await;
+    wait_for_l2_block(&full_node_test_client, 5, None).await;
+
+    // Trigger proving via the RPC endpoint
+    prover_node_test_client.prover_prove(3, Some(true)).await;
+
+    // wait here until we see from prover's rpc that it finished proving
+    wait_for_prover_l1_height(&prover_node_test_client, 4, None)
```
kpp, line 290: Looks suspicious. L1 height != we successfully proved

crates/common/src/da.rs
```diff
@@ -37,3 +42,64 @@ pub async fn get_da_block_at_height<Da: DaService>(
...
+        if tx.sender().as_ref() == sequencer_da_pub_key {
+            if let Ok(DaDataBatchProof::SequencerCommitment(seq_com)) = data {
+                sequencer_commitments.push(seq_com);
+            }
+        }
+    });
+    sequencer_commitments
+}
+
+pub async fn extract_zk_proofs<Da: DaService>(
```
kpp, line 75: There is no need for this function because DA::extract_relevant_proofs returns `Vec<Proof>`.

crates/common/src/da.rs
```diff
@@ -37,3 +42,64 @@ pub async fn get_da_block_at_height<Da: DaService>(
         .put(l1_block.header().height(), l1_block.clone());
     Ok(l1_block)
 }
+
+pub fn extract_sorted_sequencer_commitments<Da>(
```
kpp: I am not sure we need a separate `extract_sorted_sequencer_commitments` if we can have `extract_sequencer_commitments` and then sort if afterwards.

crates/prover/src/proving.rs
```diff
@@ -0,0 +1,403 @@
...
+use crate::errors::L1ProcessingError;
+
+type TxId<Da> = <Da as DaService>::TransactionId;
+
+pub(crate) async fn data_to_prove<Da, DB, StateRoot, Witness>(
+    da_service: Arc<Da>,
+    ledger: DB,
+    sequencer_pub_key: Vec<u8>,
+    sequencer_da_pub_key: Vec<u8>,
+    l1_block_cache: Arc<Mutex<L1BlockCache<Da>>>,
```
kpp, line 35: `l1_block_cache` goes into `get_da_block_at_height` only. So we can add `l1_block_cache` to BitcoinDa and add a common method for DaService trait so we don't have to pass that block_cache around.

eyusufatik: well actually it can be in the da adapter implementations, however, that's a seperate issue imo

crates/prover/src/proving.rs
rakanalh: It's needed. We don't pass `da_data` as mut to any function here so the version we pass won't be affected by the `full_data` called internally. This is why we do it here to make the zk circuit able to read the commitment as the comment states.

bin/citrea/tests/e2e/proving.rs
rakanalh: Yea that was what the test in the same file is doing so i did the same.

crates/prover/src/da_block_handler.rs
rakanalh: Clippy:
```
warning: use of `expect` followed by a function call
   --> crates/prover/src/da_block_handler.rs:190:30
    |
190 |   ...                   .expect(&format!(
    |  ________________________^
191 | | ...                       "Failed to put prover last scanned l1 height in the ledger db {}",
192 | | ...                       l1_height
193 | | ...                   ));
    | |________________________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#expect_fun_call
    = note: `#[warn(clippy::expect_fun_call)]` on by default
help: try
    |
190 ~                             .unwrap_or_else(|_| panic!("Failed to put prover last scanned l1 height in the ledger db {}",
191 ~                                 l1_height));
    |
```

bin/citrea/tests/e2e/proving.rs
rakanalh: Yes, this makes sure that the prover has synced up to that block. Later we check for the proof itself.

crates/common/src/da.rs
rakanalh: No it returns `Vec<DaDataLightClient>`

yaziciahmet (approved):

PR #1347, opened 2024-10-16
SP1 adapter & zkvm guest implementation

### LINKED ISSUE #1287: SP1 adapter

### LINKED ISSUE #1288: SP1 guest codes
After #1287 is implemented, we need to create `bin/citrea/provers/sp1` and then create guests for bitcoin da and mock da.

yaziciahmet:
# Description

I implemented SP1 zkvm guest code, with SP1Host and SP1Guest adapters. Some things to be careful about before running:
1. All the SP1 code usage is commented out, and is currently unused.
2. To build sp1-sdk, the build has to be in release mode.
3. New sp1 dev tools are added to the makefile install-dev-tools command. Make sure to install them.
4. Currently, to build and test the proving using SP1, you should
  a. make install-dev-tools
  b. Comment out sp1 related lines in BitcoinRollup, and comment risc0 lines
  c. BUILD_SP1_GUEST=1 make build-release
  d. ./target/release/citrea ... whatever you want to run

I did some tests on a real world data. I ran a prover in our testnet in execute mode for both RISC0 and SP1. Used `prover_prove` rpc to invoke proving. @eyusufatik pointed me to block 45949 which contains 3 zk proofs and 12 seq commitments. Here are the cycle comparison:

Proof 1:
- RISC0 => total cycles: 6_804_733_952   user cycles: 5_291_119_399
- SP1 => cycles: 4_406_830_562

Proof 2:
- RISC0 => total cycles: 6_728_712_192	user cycles: 5_235_187_308
- SP1 => cycles: 4_366_318_784

Proof 3:
- RISC0 => total cycles: 2_691_170_304	user cycles: 2_110_365_034
- SP1 => cycles: 1_758_320_756

**BELOW IS SYNTHETIC DATA TESTS. KEEPING THEM AS A HISTORY**

I did some cycle comparison tests comparing sp1 and risc0 and here are the results:

-------------------------

SP1

422_969 => 45_803_073 cycles

-------------------------

RISC0

422_494 =>
  total: 100_794_368 cycles
  user: 63_306_694 cycles

-------------------------

Left hand-side is the data size in bytes. Right hand-side is the cycle counts. Even though I am not 100% sure, but SP1 cycles most likely correspond to user cycles in risc0. Hence, risc0 takes 40% more cycles compared to sp1. Keep in mind that these tests are done with 20 empty blocks, it is relatively small data, and does not include any transaction validation etc.. More sophisticated benchmarking will need to be done.

**UPDATE:**

I did a more sophisticated benchmarking. Produced 200 blocks with 5 transactions in each block. Input to the guest turned out to be 4 MB. Results:

-------------------------

SP1

200 blocks with 5 txs each (4 MB) => 664_201_037 cycles

-------------------------

RISC0

200 blocks with 5 txs each (4 MB) =>
  total: 3_778_019_328 cycles
  user: 2_834_160_041 cycles

-------------------------

RISC0 proving takes 4x cycles of SP1. The big difference is most likely caused by missing patches in Risc0. Similar benchmarking should be repeated once we upgrade to latest Risc0 and they provide more patches

## Linked Issues
- Fixes #1287 and #1288


crates/sp1/src/guest.rs
```diff
@@ -0,0 +1,49 @@
...
+        _serialized_proof: &[u8],
+        _code_commitment: &Self::CodeCommitment,
+    ) -> Result<sov_rollup_interface::zk::StateTransition<Da, Root>, Self::Error> {
+        unimplemented!()
+    }
+}
+
+impl ZkvmGuest for SP1Guest {
+    fn read_from_host<T: BorshDeserialize>(&self) -> T {
+        let buf = io::read_vec();
```
eyusufatik, line 43: btw, checkout how sp1's API for committing / reaading is, we might work around without needing borsh serialize / deserialize. low chance but possible.

also create issue so that we benchmark serialization methods for sp1, i don't think we'll have different results from risc0 but we should try.

crates/sp1/src/host.rs
```diff
@@ -0,0 +1,219 @@
...
+            let data = bincode::serialize(&proof_with_public_values)
+                .expect("SP1 zk proof serialization must not fail");
+            Ok(Proof::Full(data))
+        } else {
+            let (public_values, report) = self.client.execute(self.elf, stdin).run()?;
+            info!("Number of cycles: {}", report.total_instruction_count());
+
+            let data = bincode::serialize(&public_values)
+                .expect("SP1 zk public values serialization must not fail");
+            Ok(Proof::PublicInput(data))
```
eyusufatik: no local proving?

eyusufatik (changes requested):
we've already talked about creating a guest in this pr. can you also add a new command to our make file called "dev-install-sp1" so that when we want to experiment with sp1, we can but we don't have to unnecessarily install it

crates/sp1/src/host.rs
yaziciahmet: If you check the initialization of the ProverClient, internally it decides on whether to use LocalProver or NetworkProver (SP1's Bonsai correspondence which is called Succinct) based on the environment variables. So if there are no env variables present, and client initializes a LocalProver, calls to prove will do local proving.

crates/sp1/src/guest.rs
yaziciahmet: sp1 does not have reader implementation as risc0 does, so the whole vector should be allocated anyways. the internal io::read::<T>() method uses bincode internally, so we might potentially ditch borsh and just go with the bincode everywhere.

eyusufatik: ah that's a shame. 

put a todo comment for this in the files pls

bin/citrea/provers/sp1/guest-bitcoin/src/main.rs
```diff
@@ -0,0 +1,32 @@
...
+        stf,
+        BitcoinVerifier::new(RollupParams {
+            reveal_batch_prover_prefix: REVEAL_BATCH_PROOF_PREFIX.to_vec(),
+            reveal_light_client_prefix: REVEAL_LIGHT_CLIENT_PREFIX.to_vec(),
+        }),
+    );
+
+    stf_verifier
+        .run_sequencer_commitments_in_da_slot(guest, storage)
+        .expect("Prover must be honest");
```
eyusufatik: no commit to env?

eyusufatik (changes requested):

bin/citrea/src/rollup/bitcoin.rs
```diff
@@ -98,7 +98,13 @@ impl RollupBlueprint for BitcoinRollup {
...
-        map.insert(SpecId::Genesis, Digest::new(citrea_risc0::BITCOIN_DA_ID));
+        // map.insert(SpecId::Genesis, Digest::new(citrea_risc0::BITCOIN_DA_ID));
+        map.insert(
+            SpecId::Genesis,
+            VerifyingKey::from_elf(include_bytes!(
+                "../../provers/sp1/guest-bitcoin/elf/zkvm-elf"
+            )),
+        );
         map
     }
```
eyusufatik: also comment out the sp1 insert, so we can merge this

bin/citrea/provers/sp1/guest-bitcoin/src/main.rs
yaziciahmet: commit to env happens in `run_sequencer_commitments_in_da_slot`, same as risc0 guest code.

bin/citrea/src/rollup/bitcoin.rs
yaziciahmet: pr not ready yet, just finalized the guest code and about to test and compare. will revert back to risc0 citrea once i have some numbers in hand.

bin/citrea/provers/sp1/guest-bitcoin/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+sov-modules-stf-blueprint = { path = "../../../../../crates/sovereign-sdk/module-system/sov-modules-stf-blueprint" }
+sov-rollup-interface = { path = "../../../../../crates/sovereign-sdk/rollup-interface" }
+sov-state = { path = "../../../../../crates/sovereign-sdk/module-system/sov-state" }
+sp1-zkvm = { git = "https://github.com/succinctlabs/sp1", tag = "v3.0.0-rc3", default-features = false, features = ["lib"] }
+
+# Have to put this here to enable features for the patch crate even though we don't use this crate explicitly
+curve25519-dalek = { version = "4.1.3", default-features = true }
+
+[patch.crates-io]
+# sha2-v0-9-8 = { git = "https://github.com/sp1-patches/RustCrypto-hashes", package = "sha2", branch = "patch-sha2-v0.9.8" }
```
kpp: Please put down a comment why we don't use these patches

crates/sp1/src/guest.rs
```diff
@@ -0,0 +1,51 @@
+//! This module implements the `ZkvmGuest` trait for the SP1 VM.
+use borsh::{BorshDeserialize, BorshSerialize};
+use sov_rollup_interface::zk::{Zkvm, ZkvmGuest};
+use sp1_zkvm::io;
+
+/// A guest for the SP1 VM. Implements the `ZkvmGuest` trait
+///  in terms of Risc0's env::read and env::commit functions.
```
kpp: Not risc0

crates/sp1/src/host.rs
```diff
@@ -0,0 +1,236 @@
...
+        Ok(BorshDeserialize::try_from_slice(public_values.as_slice())?)
+    }
+
+    fn recover_proving_sessions(&self) -> Result<Vec<Proof>, anyhow::Error> {
+        // We can only recover if prover is configured to be Succinct
+        if !self.is_succinct_prover() {
+            return Ok(vec![]);
+        }
+
+        let request_ids = self.ledger_db.get_pending_proving_sessions()?;
```
kpp, line 159: Probably we have to separate risc0 and sp1 sessions

kpp (approved):

crates/sp1/src/host.rs
yaziciahmet: the fact that we store sessions as Vec<u8> helped a lot in using both type of sessions without needing to add another ledgerdb functionality. IMO this is fine at least for now since we technically don't ever use SP1 in the codebase.

crates/sp1/src/guest.rs
yaziciahmet: fixed

bin/citrea/provers/sp1/guest-bitcoin/Cargo.toml
yaziciahmet: done

crates/sp1/src/host.rs
kpp: We can store (kind, session) in one get_pending_proving_sessions. Otherwise it will be painful to debug in the future

yaziciahmet: You are correct. But this is only going to be a problem when we are switching from risc0 to sp1. We will never have pending sessions from both, except, potentially when we restart the prover through deployment, and right at the restart, there was a risc0 proof happening, and with that deployment we switched to sp1.

I think this should be handled within the pr where we actually switch sp1, if we ever.

bin/citrea/build.rs
```diff
@@ -12,4 +16,47 @@ fn main() {
...
+    println!("cargo:rerun-if-env-changed=BUILD_SP1_GUEST");
+
+    let bitcoin_program_path = "provers/sp1/guest-bitcoin";
+
+    let build_args = sp1_helper::BuildArgs {
+        elf_name: "zkvm-elf".to_string(),
+        ..Default::default()
+    };
+
+    match std::env::var("BUILD_SP1_GUEST") {
```
rakanalh, line 33: Shouldn't this be the default? If we want to be consistent, we should have a flag `SKIP_GUEST_BUILD` similar to the risc0 one.

rakanalh (commented):
For me, instead of introducing an `sp1` crate, i would try to gather:
- `crates/risc0`
- `crates/sp1`
- `bin/citrea/provers`
into a single `citrea-zk` crate

crates/sp1/src/host.rs
rakanalh: Yea i agree. We know that we won't use both sessions simultaneously so we don't have to support both in storage.

As for how we handle storage related to risc0 when we switch to SP1 (if ever), would be in a ticket i am working on:
https://github.com/chainwayxyz/citrea/issues/1338

bin/citrea/build.rs
yaziciahmet: The thing is, we are not using SP1 currently in any code-paths of the citrea program. And building SP1 code adds unnecessary build time. Hence, I made the building opt-in rather than opt-out. If you wanna play around with it you have to specify it.

yaziciahmet: Once we select one of the zkvm providers, we will still use SKIP for that ofc.

yaziciahmet:
> For me, instead of introducing an `sp1` crate, i would try to gather:
> 
> * `crates/risc0`
> * `crates/sp1`
> * `bin/citrea/provers`
>   into a single `citrea-zk` crate

I do agree we can put risc0 and sp1 together, but I don't think we should add bin/citrea/provers to this citrea-zk crate. Because all provers has is a main function with 3-4 function calls. I don't see any benefit to separating that into another crate.

Putting risc0 and sp1 is useful. We can put the sp1 code behind a sp1 feature, so if it is enabled it will use the sp1 guest and host, but by default it will use the risc0. Maybe we can throw in some stuff that are common in zk and native as well.

deny.toml
```diff
@@ -18,7 +18,7 @@ allow = [
   "Unicode-3.0",
   "Unicode-DFS-2016",
   # "Unlicense",
-  # "Zlib",
+  "Zlib",
```
eyusufatik, line 21: revert?

.gitignore
```diff
@@ -28,3 +28,6 @@ adapters/solana/libyellowstone_grpc_geyser.dylib
 reveal_*.tx
 
 *.txs
+
+elf/
+dbs
```
eyusufatik: I think putting dbs in gitignore broke something, because we recently added and removed it

crates/sp1/src/guest.rs
```diff
@@ -0,0 +1,51 @@
...
+
+impl SP1Guest {
+    /// Constructs a new SP1Guest
+    pub fn new() -> Self {
+        Self::default()
+    }
+}
+
+impl Zkvm for SP1Guest {
+    #[cfg(feature = "native")]
```
eyusufatik, line 19: i don't think you need native flag here

bin/citrea/build.rs
```diff
@@ -12,4 +16,47 @@ fn main() {
             .args(["install", "cargo-risczero"])
             .status();
     }
+
+    build_sp1_guest();
+}
+
+fn build_sp1_guest() {
+    println!("cargo:rerun-if-env-changed=BUILD_SP1_GUEST");
```
eyusufatik, line 24: why do we have to explicitly build sp1 guests?

risc0 guests build just fine?

crates/sp1/src/guest.rs
yaziciahmet: VerifyingKey is a wrapper of SP1VerifyingKey which exists under sp1-sdk. And when sp1-sdk is imported into the guest code, very very very bad things happen. This is going to be something we have to resolve to do recursive verification, because we need VK for verification.

bin/citrea/build.rs
yaziciahmet: this is a choice, not a must. Most likely nothing will not need SP1 guest built, unless we decide to switch to sp1, so I made it opt-in to build instead of opt-out. So we skip the building to not prolong the builds and development even though we don't need it.

eyusufatik (approved):

PR #1336, opened 2024-10-14
Light client starter

yaziciahmet:
# Description

Added initial support for light client, and an example test. This PR is intended to give skeleton to light client both file structure-wise and behavior-wise. It is in no way complete functionality of light client.

**IMPORTANT:**
Before merging this pr, citrea-e2e pr https://github.com/chainwayxyz/citrea-e2e/pull/20 should be merged, and rev in Cargo.toml should be updated accordingly.

## Linked Issues
- Fixes # (issue, if applicable)
- Related to # (issue) 

## Testing
Describe how these changes were tested. If you've added new features, have you added unit tests?

## Docs
Describe where this code is documented. If it changes a documented interface, have the docs been updated?

PR #1349, opened 2024-10-16
Check compressed sequencer commitments

### LINKED ISSUE #1213: Check compressed state diff size before sending seq_commtiments
Is this only applicable to prover? because we also calculate the state diff size in the sequencer?

https://github.com/chainwayxyz/citrea/blob/7a077f0cdf471a9871db6b7a6cd1ae64d016c8d1/crates/sequencer/src/sequencer.rs#L550-L553

_Originally posted by @rakanalh in https://github.com/chainwayxyz/citrea/pull/1212#discussion_r1768725654_

rakanalh:
# Description

This PR uses `TaskManager` to spawn a new task for running `CommitmentService` which in turn handles everything related to submitting sequencer commitments.

# Detailed changes in this PR

- Move commitment controller logic & some parts of the state diff threshold into `CommitmentStrategy` structs.
- Move logic for running commitment strategies into `CommitmentController`
- Use `CommitmentService` for submitting commitments if `CommitmentController` advises to.
- Move `sequencer.rs` into `runner.rs` for consistency with prover / full node.
- Simplify the sequencer runner code base.

# How it works

- Sequencer produces L2 block
- Sequencer sends height & state diff to `CommitmentService`
- `CommitmentService` consults with `CommitmentController` whether it should commit or not.
- `CommitmentController` runs each `CommitmentStrategy` and takes whichever returns `CommitmentInfo`
- `CommitmentService` submits commitment if advised to do so.

One more thing to pay attention to, is that the `StateDiffThreshold` commitment strategy now uses `brotli` for compressing the state diff before deciding on the threshold. This is a CPU heavy operation which should run in it's own task.

## Linked Issues
- Fixes #1213


crates/sequencer/src/commitment/strategy.rs
```diff
@@ -0,0 +1,177 @@
...
+            .get_pending_commitments_l2_range()?
+            .iter()
+            .map(|(_, end)| *end)
+            .max()
+            .unwrap_or(BatchNumber(0));
+        let last_committed_l2_height = cmp::max(last_finalized_l2_height, last_pending_l2_height);
+
+        let merged_state_diff =
+            merge_state_diffs(self.last_state_diff.clone(), l2_state_diff.clone());
+        let compressed_state_diff = compress_blob(&borsh::to_vec(&merged_state_diff)?);
```
kpp, line 121: Can you bench how much it takes to compress? If thats a long cpu task, maybe we should spawn it in a blocking task?

eyusufatik: yes for big state diffs it might be slow

rakanalh: It's done. We execute commitment controller strategies in the blocking thread.

crates/common/src/compression.rs
eyusufatik, line 1: this is going to sound very weird

but compress here, decompress in primitives work better.

usually primitives is for everywhere (both zk and node level stuff)

common is for non-zk parts of the repo right?

with #1336, decompress is needed in ZK context 

so maybe move both to primitives so they are together?

eyusufatik (commented):
very solid PR

crates/common/src/compression.rs
rakanalh: Moved

eyusufatik (approved):

PR #1356, opened 2024-10-18
Implement boundless adapter

### LINKED ISSUE #1308: Risc0 Boundless adapter

ercecan:
# Description
Implements risc0 boundless adapter
**cannot merge yet since boundless repos require ssh access**

## Linked Issues
- Fixes #1308


crates/risc0-boundless/Cargo.toml
```diff
@@ -0,0 +1,51 @@
+[package]
```
rakanalh, line 1: Sorry but i am very opinionated about introducing crates. Especially this when we have `risc0-bonsai` and `risc0-boundless` where both can live in a single `proving-clients` crate or something like that.

crates/risc0-boundless/src/host.rs
```diff
@@ -0,0 +1,539 @@
...
+};
+use sov_db::ledger_db::LedgerDB;
+use sov_risc0_adapter::guest::Risc0Guest;
+use sov_rollup_interface::zk::{Proof, Zkvm, ZkvmHost};
+use tokio::sync::mpsc::{self, UnboundedSender};
+use tokio::time::{sleep, Duration};
+use tracing::{debug, error, info, instrument, trace, warn};
+use url::Url;
+
+// TODO: Remove blocking wrapper once the boundless client has blocking api
```
rakanalh, line 34: This TODO seems to be in the wrong place...

crates/risc0-boundless/src/host.rs
```diff
@@ -0,0 +1,539 @@
...
+}
+
+impl BoundlessClient {
+    async fn from_parts(
+        requestor_private_key: PrivateKeySigner,
+        rpc_url: Url,
+        proof_market_address: Address,
+        set_verifier_address: Address,
+    ) -> Self {
+        macro_rules! unwrap_boundless_response {
```
rakanalh, line 86: Move macro outside the the impl block.

crates/risc0-boundless/src/host.rs
```diff
@@ -0,0 +1,539 @@
...
+                                error!(?e, "Got unrecoverable error from Boundless");
+                                panic!("Boundless API error: {}", e);
+                            }
+                        }
+                    }
+                }
+            );
+        }
+        let (queue, mut rx) = tokio::sync::mpsc::unbounded_channel();
+        let join_handle = tokio::spawn(async move {
```
rakanalh, line 119: Move this async block into it's own function... 
```suggestion
        let join_handle = tokio::spawn(handle_response(client))
```

crates/risc0-boundless/src/host.rs
```diff
@@ -0,0 +1,539 @@
...
+            let client = Client::from_parts(
+                requestor_private_key.clone(),
+                rpc_url.clone(),
+                proof_market_address,
+                set_verifier_address,
+            )
+            .await
+            .unwrap();
+
+            'queue: loop {
```
rakanalh, line 131: 1. Is it a good idea to retry indefinitely?
2. Wouldn't it be better if somehow we return an error that the call-site can handle in a backoff-retry fashion or maybe keep `pending` list of requests and retry later? 
3. Handle errors in the call-site for more control over what happens if this keeps failing.

crates/risc0-boundless/src/host.rs
```diff
@@ -0,0 +1,539 @@
...
+        }
+    }
+
+    #[instrument(level = "trace", skip(self), ret)]
+    fn upload_img(&self, elf: Vec<u8>) -> String {
+        let (notify, mut rx) = mpsc::unbounded_channel();
+        self.queue
+            .send(BoundlessRequest::UploadImg { elf, notify })
+            .unwrap();
+        tokio::task::block_in_place(|| rx.blocking_recv().unwrap())
```
rakanalh, line 204: Why use `block_in_place` when you can make this method async + you're calling this method from an async block?

crates/risc0-boundless/src/host.rs
```diff
@@ -0,0 +1,539 @@
...
+            client,
+            last_input_url: None,
+            _ledger_db: ledger_db,
+        }
+    }
+
+    fn upload_input_to_boundless(&mut self, input: &[u8]) {
+        let client = self.client.clone();
+
+        // Retry backoff
```
rakanalh, line 312: Is this a TODO? Otherwise, this is not retry-backoff

rakanalh (changes requested):

ercecan:
Superseded by #2391

PR #1358, opened 2024-10-20
Ledger db migrations

### LINKED ISSUE #1338: LedgerDB supports migrations

rakanalh:
# Description

This PR implements a migrator which atomically executes ledger DB migrations. The process runs as follows:
1. Create a temporary directory which contains rocksdb files from the current ledger DB.
2. Execute all migrations on temporary directory
3. If all are successful, mark migrations as such
4. Move successful temporary files into the original ledger DB directory.

This happens of course, in case where no ledger DB instance holds the lock on the directory. Otherwise, the process will fail.

This also implements the possibility to execute different migrations for each of sequencer, full node and prover.

# TODO
- [x] Implement an example migration to update table data
- [x]  Implement an example migration to update table fields
- [x] Implement an example migration to delete table
- [x] Add tests

## Linked Issues
- Fixes #1338


crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/migrations.rs
```diff
@@ -0,0 +1,148 @@
...
+        let temp_db_path = tempfile::tempdir()?;
+        copy_db_dir_recursive(original_path, temp_db_path.path())?;
+
+        let new_ledger_db = Arc::new(LedgerDB::with_config(&RocksdbConfig::new(
+            temp_db_path.path(),
+            max_open_files,
+        ))?);
+
+        for migration in self.migrations {
+            if !executed_migrations.contains(&migration.identifier()) {
```
eyusufatik, line 74: i think erroring out on failed migration is reasonable. bc if the migration couldn't be executed successfully, then the node will fail once it needs to use one of the tables in those migrations.

eyusufatik: also before erroring out, it can revert to the old db

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/tests.rs
```diff
@@ -0,0 +1,122 @@
...
+        ("OldToNew".to_owned(), 1)
+    }
+
+    fn execute(&self, ledger_db: sov_rollup_interface::RefCount<LedgerDB>) -> anyhow::Result<()> {
+        let Some(values) = ledger_db.db.get::<TestTableOld>(&())? else {
+            return Ok(());
+        };
+        for (index, value) in values.into_iter().enumerate() {
+            ledger_db.put_value(index as u64, (index as u64, value))?;
+        }
```
eyusufatik, line 36: we could also create the new table here right?

crates/fullnode/src/db_migrations/mod.rs
eyusufatik, line 1: are you planning to also implement absolute seperation of ledger db tables by node type?

If not, let's have this migrations in a common place pls

eyusufatik (changes requested):

crates/fullnode/src/db_migrations/mod.rs
rakanalh: > are you planning to also implement absolute seperation of ledger db tables by node type?

Most probably not in this PR.

> If not, let's have this migrations in a common place pls

I don't agree with this. Since each node type deals with different tables, they will in result have different migrations that tamper those different tables. I'd rather keep this separated for now.

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/migrations.rs
rakanalh: I am not sure i understand the change request here. But in general, those lines are running migrations on the temporary directory database. If any of those migrations fail, we don't have to roll back anything we just discard the temporary directory DB updates. This is why i copy the temporary directory into the ledger DB directory as the last step.

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/tests.rs
rakanalh: Yes, `TestTableNew` is a new table and the call to `put_value` here is adding entries into this new table.

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/migrations.rs
eyusufatik: I mean on `let Err(e) = migration.execute(new_ledger_db.clone())` case we should return err and terminate the node imo

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/tests.rs
eyusufatik: but isn't that table already in the schema when you start the process? I'm talking about creating a new table that we haven't defined in `LEDGER_TABLES` previously, but we do now, with the migration

rakanalh: So tables are not "created" when you start the process. They are created as soon as you write to them. 

When we create the `inner: DB` part of the `LedgerDB`, we pass the schema of those tables but it won't actually create them until they're written to.

rakanalh: The idea behind the `TestTableNew` is to demonstrate how we can define the new table's schema and we add it to the code and we get the code to use it. However, it doesn't mean that it has data to be used. That's what migrations are for.

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/migrations.rs
rakanalh: Done

ercecan (approved):
LGTM

eyusufatik (approved):

PR #1376, opened 2024-10-24
Fix get_prev_utxo logic

### LINKED ISSUE #1370: E2E flaky: test_sequencer_transaction_chaining
## Summary

```
Error: assertion `left == right` failed: TX6 should reference TX5's output
  left: 02f76e3ee9af4b61e1c7df3d3b123ebf9cdfdd160c586c4497ce06ff8e2679ff
 right: 23e31578a08505ae283919166fdbb0b2202ab8f86066b7bb0e8ea0c8a7d324e5
 ```
 
 ```
 --- TRY 2 STDERR:        citrea::all_tests bitcoin_e2e::tx_chain::test_sequencer_transaction_chaining ---
Error: assertion `left == right` failed: TX6 should reference TX5's output
  left: 33fb584f5a841e0344d7d2f89427cf567e36514ba48ef305447649f4425a7331
 right: d20d4077c309374b81262c8253dd6dfc851c7d1257e66452152f1b582ce0e19d
 ```
 
 https://github.com/chainwayxyz/citrea/actions/runs/11461889179/job/31892011158#step:11:331

### PR DESCRIPTION
**Pull request by jfldde**
# Description

-  Make sure to retain only utxos from reveal TXs by matching on REVEAL_OUTPUT_AMOUNT
-  Sort by (confirmations - ancestor_count). If any tx in mempool, confirmation would be 0 and ancestor count Some(_), giving us a negative value and prioritising in-mempool TXs. If no tx in mempool, confirmation would be >= 0 and ancestor count None

## Linked Issues
- Fixes #1370 

## Testing
- Add a test restarting sequencer with commit/reveal txs in mempool

### DIFF
--- a/crates/bitcoin-da/src/service.rs
+++ b/crates/bitcoin-da/src/service.rs
@@ -276,6 +276,11 @@ impl BitcoinService {
         });
     }
 
+    /// Retrieves the most recent spendable UTXO from the transaction chain on startup.
+    /// `get_prev_utxo` is called on service startup in `spawn_da_queue`.
+    /// On very first startup, return any most recent UTXO.
+    /// Any subsequent startup would return latest reveal TX first output.
+    /// Sorts by (confirmations - ancestor_count), where lower values indicate more recent transactions.
     #[instrument(level = "trace", skip_all, ret)]
     async fn get_prev_utxo(&self) -> Result<Option<UTXO>, anyhow::Error> {
         let mut previous_utxos = self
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: crates/bitcoin-da/src/service.rs [hunk 1]
```diff
@@ -276,22 +276,31 @@ impl BitcoinService {
...
     async fn get_prev_utxo(&self) -> Result<Option<UTXO>, anyhow::Error> {
         let mut previous_utxos = self
             .client
             .list_unspent(Some(0), None, None, Some(true), None)
             .await?;
 
-        previous_utxos.retain(|u| u.spendable && u.solvable);
+        // Make sure to retain only utxos from reveal TXs by matching on REVEAL_OUTPUT_AMOUNT
+        previous_utxos.retain(|u| {
+            u.spendable && u.solvable && u.amount == Amount::from_sat(REVEAL_OUTPUT_AMOUNT)
```
eyusufatik, line 293: u.amount >= DUST_VALUE

eyusufatik (changes requested):

crates/bitcoin-da/src/service.rs
jfldde: That would also match commit change output (vout 1), since both commit/reveal are included in same block and have the same amount of confirmations.
If we want the match the behaviour of the running node, we need to use reveal's first vout.
This creates a clean tx chain where we have commit tx using previous reveal first output, in line with the behaviour in `spawn_da_queue` loop :
```
                                                prev_utxo = Some(UTXO {
                                                    tx_id: tx.id,
                                                    vout: 0,
                                                    script_pubkey: tx.tx.output[0].script_pubkey.to_hex_string(),
                                                    address: None,
                                                    amount: tx.tx.output[0].value.to_sat(),
                                                    confirmations: 0,
                                                    spendable: true,
                                                    solvable: true,
                                                });
```
This makes it much cleaner to follow along an explorer imo

eyusufatik: since reveal is using commit vout0, and commit vout1 is free

wouldn't that mean in the sort below, we'll  choose the reveal output?

my concern is what happens when we don't find an output to use

was that taken care of where this func is called? cant remember

jfldde: Yes we pass down prev_utxo as Option all the way down to `choose_utxos`.

The sort would be flaky when restarting against an empty (without any of our tx) mempool. Commit and reveal outputs will have the same confirmations. It will then select one of commit vout 1 or reveal vout 0 sorted on Txid (sorting return by listunspent itself).
The old logic was only querying from the mempool so the `ancestor_count` was taking care of prioritising reveal over commit

eyusufatik: yeah that is fine

eyusufatik (approved):

PR #1386, opened 2024-10-29
Fix rocksdb config

yaziciahmet:
# Description

Rocksdb configs were not being set properly. The main db options passed only sets configuration of the global database variables (e.g. max_open_files or max_total_wal_size), and the configuration of the "default" column family, which is not used at all.

This PR only sets the global configurations on the core database options, and propagates the column family based options down to be applied for each column family.

Changes include:
- Lowered `set_compression_options_parallel_threads` to 2 from 4. We don't need 4 threads just for db compression. We didn't allow multiple thread usage through `increase_parallelism` option, so I doubt it was ever parallelized in the first place.
- Set `increase_parallelism` to 2. This allows rocksdb to use multiple threads for background jobs such as compaction or flushing.
- Remove `set_write_buffer_size` config. In the comment it says we are increasing the buffer size to 30MB, but the default is 64MB, which is a sufficient number.
- Remove `set_block_based_table_factory` config from the global config. This only effects the default column family, which is unused. Hence, there is no need to use this.
- Remove `set_atomic_flush` config. This is only useful when used with no-WAL config. With WAL, every flush is already atomic.
- Added block based options to the column families. This allows to set each column family block configuration, including caching, increase block size (lower index count, lower disk space)

These are pretty well-known wins, but I am still currently conducting a benchmark. I will post the results here

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/sovereign-sdk/full-node/db/sov-db/src/rocks_db_config.rs
+++ b/crates/sovereign-sdk/full-node/db/sov-db/src/rocks_db_config.rs
@@ -63,7 +63,7 @@ impl<'a> RocksdbConfig<'a> {
         block_options.set_optimize_filters_for_memory(true);
         // By default our block size is 4KB, we set this to 32KB.
         // Increasing the size of the block decreases the number of blocks,
-        // therefore, less memory consumption for indicies.
+        // therefore, less memory consumption for indices.
         block_options.set_block_size(32 * 1024);
         // Default is Snappy but Lz4 is recommend
         // https://github.com/facebook/rocksdb/wiki/Compression
--- a/crates/sovereign-sdk/full-node/db/sov-db/src/rocks_db_config.rs
+++ b/crates/sovereign-sdk/full-node/db/sov-db/src/rocks_db_config.rs
@@ -70,7 +70,8 @@ impl<'a> RocksdbConfig<'a> {
         db_options.set_compression_type(rocksdb::DBCompressionType::Lz4);
         db_options.set_compression_options_parallel_threads(2);
 
-        db_options.increase_parallelism(2);
+        let allowed_cores = std::cmp::max(1, num_cpus::get() / 2) as i32;
+        db_options.increase_parallelism(allowed_cores);
 
         db_options.set_max_open_files(self.max_open_files);
         db_options.set_max_total_wal_size(self.max_total_wal_size);
--- a/crates/sovereign-sdk/full-node/db/sov-db/src/rocks_db_config.rs
+++ b/crates/sovereign-sdk/full-node/db/sov-db/src/rocks_db_config.rs
@@ -4,6 +4,7 @@ use std::path::Path;
 
 use rlimit::{getrlimit, Resource};
 use rocksdb::{BlockBasedOptions, Cache, Options};
+use sov_schema_db::RawRocksdbOptions;
 use tracing::warn;
 
 /// Port selected RocksDB options for tuning underlying rocksdb instance of our state db.
@@ -41,11 +42,11 @@ impl<'a> RocksdbConfig<'a> {
         }
     }
 
-    /// Build [`rocksdb::Options`] from [`RocksdbConfig`]
-    pub fn as_rocksdb_options(&self, readonly: bool) -> (Options, BlockBasedOptions) {
-        let mut db_opts = Options::default();
+    /// Build [`RawRocksdbOptions`] from [`RocksdbConfig`]
+    pub fn as_raw_options(&self, readonly: bool) -> RawRocksdbOptions {
+        let mut db_options = Options::default();
 
-        let mut block_based_options = BlockBasedOptions::default();
+        let mut block_options = BlockBasedOptions::default();
         /*
          * The following settings are recommended in:
          * https://github.com/facebook/rocksdb/wiki/memory-usage-in-rocksdb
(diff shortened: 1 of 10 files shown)


yaziciahmet:
Benchmark results on testnet synced fullnode:
- `nightly`: 30.2 GB / synced to 1_800_00 in 9h 12m
- `yaziciahmet/fix-rocksdb-config`: 27.4 GB / synced to 1_800_00 in 8h 42m

So it is 10% drop on disk size, and 5% sync speed up.

I am pretty sure we can further improve this by investigating more advanced configurations, as rocksdb has infinitely many of them. But these should suffice as easy-wins, and further investigations should be done througly.

I am running one more test where I set the max_open_file configuration to infinity. This doesn't affect the code, it is just a config change, so it shouldn't block this PR from being merged.

rakanalh (changes requested):
All these removed parts were set based on memory profiling of the code. Blocking the PR so that I can take a deeper look.

crates/sovereign-sdk/full-node/db/sov-db/src/rocks_db_config.rs
```diff
@@ -67,22 +67,19 @@ impl<'a> RocksdbConfig<'a> {
...
+        db_opts.set_compression_options_parallel_threads(2);
+
+        db_opts.increase_parallelism(2);
 
         db_opts.set_max_open_files(self.max_open_files);
         db_opts.set_max_total_wal_size(self.max_total_wal_size);
         db_opts.set_max_background_jobs(self.max_background_jobs);
         if !readonly {
-            // Increase write buffer size to reduce allocations.
-            db_opts.set_write_buffer_size(30 * 1024 * 1024); // 30 MB
```
rakanalh, line 77: Reasonable, i checked the default it was indeed 64... i must have mistakenly found the default to be way less than that when i set this.

crates/sovereign-sdk/full-node/db/sov-schema-db/src/lib.rs
```diff
@@ -65,6 +66,7 @@ impl DB {
             column_families.into_iter().map(|cf_name| {
                 let mut cf_opts = rocksdb::Options::default();
                 cf_opts.set_compression_type(rocksdb::DBCompressionType::Lz4);
+                cf_opts.set_block_based_table_factory(block_opts);
```
rakanalh: I am curious, where is it documented that setting the block-based factory results in the factory being used for default column families? I would tend to think that setting it globally means that it's used for all CFs.

rakanalh (approved):
After further examining the change, this looks reasonable.

crates/sovereign-sdk/full-node/db/sov-schema-db/src/lib.rs
rakanalh: ahh i think i see. It's a sov-schema-db thing not a rocksdb thing. Here we're iterating the tables and creating default config not containing the additional global stuff we've set.

I think we can apply additional configs other than just block based table factory.

crates/sovereign-sdk/full-node/db/sov-db/src/rocks_db_config.rs
```diff
@@ -67,22 +67,19 @@ impl<'a> RocksdbConfig<'a> {
...
-            db_opts.set_block_based_table_factory(&block_based_options);
-
             db_opts.create_if_missing(true);
             db_opts.create_missing_column_families(true);
-            db_opts.set_atomic_flush(true);
         }
 
-        db_opts
+        (db_opts, block_based_options)
```
rakanalh: How about we introduce a struct that contains the values configured and just return it here?

This way, we can implement `into()` which converts it to either `Options` or `BlockBasedOptions` based on the `Into<T>` implementation.

Also, this enables us to have the configuration unified where we can also use this sov-db-schema.

crates/sovereign-sdk/full-node/db/sov-schema-db/src/lib.rs
yaziciahmet: what other options that are column family specific are you suggesting? all the options we currently set for the db options are global options. I looked into it a bit and couldn't see an option that strike me as potentially useful. There are few fields that I want us to look into related to bloom filtering, but those are also in BlockBasedOptions.

crates/sovereign-sdk/full-node/db/sov-db/src/rocks_db_config.rs
```diff
@@ -67,22 +67,19 @@ impl<'a> RocksdbConfig<'a> {
         // Default is Snappy but Lz4 is recommend
         // https://github.com/facebook/rocksdb/wiki/Compression
         db_opts.set_compression_type(rocksdb::DBCompressionType::Lz4);
-        db_opts.set_compression_options_parallel_threads(4);
+        db_opts.set_compression_options_parallel_threads(2);
+
+        db_opts.increase_parallelism(2);
```
jfldde: Why 2 and not something like 50/75% of physical cores ?

yaziciahmet: That is a very valid question ser. I made up a number to test quickly on a machine to see the difference. 50% makes sense for sure.

crates/sovereign-sdk/full-node/db/sov-db/src/rocks_db_config.rs
```diff
@@ -57,29 +58,33 @@ impl<'a> RocksdbConfig<'a> {
...
         // is reached.
         let cache = Cache::new_lru_cache(100 * 1024 * 1024); // 100 MB
-        block_based_options.set_block_cache(&cache);
+        block_options.set_block_cache(&cache);
         // jemalloc friendly bloom filter sizing
-        block_based_options.set_optimize_filters_for_memory(true);
+        block_options.set_optimize_filters_for_memory(true);
         // By default our block size is 4KB, we set this to 32KB.
         // Increasing the size of the block decreases the number of blocks,
         // therefore, less memory consumption for indicies.
```
jfldde: Small typo here, should be `indices`

jfldde (approved):

crates/sovereign-sdk/full-node/db/sov-db/src/rocks_db_config.rs
yaziciahmet: fixed

PR #1388, opened 2024-10-29
Ignore regtest in get_fee_rate_from_mempool_space

kpp:
# Description

So that e2e tests are not spammed when RUST_BACKTRACE=1. 

```
WARN bitcoin_da::service: Failed to get fee rate from mempool.space e=Unsupported network for mempool space fee estimation

Stack backtrace:
   0: anyhow::error::<impl anyhow::Error>::msg
```

### DIFF
--- a/crates/bitcoin-da/src/service.rs
+++ b/crates/bitcoin-da/src/service.rs
@@ -1707,11 +1707,17 @@ mod tests {
         let _fee_rate = get_fee_rate_from_mempool_space(bitcoin::Network::Testnet)
             .await
             .unwrap();
-        assert!(get_fee_rate_from_mempool_space(bitcoin::Network::Regtest)
-            .await
-            .is_err());
-        assert!(get_fee_rate_from_mempool_space(bitcoin::Network::Signet)
-            .await
-            .is_err());
+        assert_eq!(
+            None,
+            get_fee_rate_from_mempool_space(bitcoin::Network::Regtest)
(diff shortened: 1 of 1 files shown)


crates/bitcoin-da/src/service.rs
```diff
@@ -569,7 +569,7 @@ impl BitcoinService {
     pub async fn get_fee_rate_as_sat_vb(&self) -> Result<u64, anyhow::Error> {
         // If network is regtest or signet, mempool space is not available
         let smart_fee = match get_fee_rate_from_mempool_space(self.network).await {
-            Ok(fee_rate) => Some(fee_rate),
+            Ok(fee_rate) => fee_rate,
             Err(e) => {
                 tracing::warn!(?e, "Failed to get fee rate from mempool.space");
```
yaziciahmet: With this change, it might be more appropriate to log error here maybe? Not exactly sure, wdyt?

yaziciahmet (approved):
Perfecto!

crates/bitcoin-da/src/service.rs
kpp: Probably yes, you are right

kpp:
```suggestion
                tracing::error!(?e, "Failed to get fee rate from mempool.space");
```

jfldde (approved):

PR #1389, opened 2024-10-29
Use unwrap() in e2e tests instead of operator ?

kpp:
# Description

We don't need to propagate error. We need an early panic & stop.

### DIFF
--- a/bin/citrea/tests/bitcoin_e2e/batch_prover_test.rs
+++ b/bin/citrea/tests/bitcoin_e2e/batch_prover_test.rs
@@ -61,28 +61,29 @@ impl TestCase for BasicProverTest {
         };
 
         // Generate confirmed UTXOs
-        da.generate(120, None).await?;
+        da.generate(120, None).await.unwrap();
 
         let min_soft_confirmations_per_commitment =
             sequencer.min_soft_confirmations_per_commitment();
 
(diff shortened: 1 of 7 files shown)


jfldde:
What do we gain from this ? We're already early exiting on any error + displaying backtrace if RUST_BACKTRACE is toggled

yaziciahmet:
Out of curiosity, what was the previous behavior when an error is returned? It could be better to propagate the error to e2e-framework so it can dump all the relevant information about the current test-state. I think e2e-framework also captures panics but I am not exactly sure.

kpp:
> What do we gain from this ? We're already early exiting on any error + displaying backtrace if RUST_BACKTRACE is toggled

I don't need a backtrace of the error, I need a backtrace where it failed in the test + error.

Now this info is lost.

yaziciahmet:
@jfldde it might be cool to remove Result return value from the run_test method, and force panic, and handle the panic on the e2e framework. This way, in the first step we can see the actual stack trace, and e2e-framework can do a graceful shutdown of resources. In the following steps, we can even add metadata information to these panics, such as ran resources, their latest block height etc.

kpp:
Seems like we already propagate backtrace for ? errors. The issue is in asserts

PR #1399, opened 2024-10-30
Offchain state map: reduce state diff size by removing contract code.

### LINKED ISSUE #1351: Offchain contracts
I haven't figured out how to do this yet.

The aim of this issue is to make smart contract codes inputs to the ZK circuit.  

Anything we use regarding the state is already an input to the circuit. But it's verified against the state root by verifying a merkle proof.

But on code case, the code hash is already committed to an account.

So we don't need merkle verificaiton, we can just check against the code hash.

Another benefit is now we don't need to include the account codes to our state diff.

One possible way to accomplish this is to create a new type of "State" something in between "State" and "AccessoryState" it's accessible in ZK but not merkle verified.

Another way is to modify accessory state read and writes to accept a new input `add_to_witness` which just adds the value to the hint, like the "State"  operations are done, however, it does not put the merkle path as a hint. 

In both of the methods, we'll need to hash the code and check against account.

rakanalh:
# Description

This PR introduces:
### Working Set
2. `OffchainDelta`
3. `OffchainArchivalWorkingSet`
4. `OffchainWorkingSet`

### State Data types
6. `OffchainStateMap`

The `code` member of the EVM now uses `OffchainStateMap` which basically just:
1. Reads from storage or cache
2. Does NOT append `code` writes to state diff.

### TODO
- [x] `OffchainStateMap`
- [x] Add check for keccack of code
- [ ] Check state diff size decrease rate

## Linked Issues
- Fixes #1351


crates/evm/src/evm/db.rs
```diff
@@ -88,7 +88,7 @@ impl<'a, C: sov_modules_api::Context> Database for EvmDb<'a, C> {
         // TODO move to new_raw_with_hash for better performance
         Ok(self
```
eyusufatik: need to check if keccak(code_from_offchain) == code_hash

crates/evm/src/evm/db.rs
```diff
@@ -88,7 +88,7 @@ impl<'a, C: sov_modules_api::Context> Database for EvmDb<'a, C> {
...
             .code
-            .get(&code_hash, self.working_set)
+            .get(&code_hash, &mut self.working_set.offchain_state())
```
eyusufatik: once more this should be like this:

```rust
let code = self
            .code
            .get(&code_hash, &mut self.working_set.offchain_state()

assert_eq!(keccak256(code), code_hash);

Ok(code)
```

or you can return Err as well.

There is nothing preventing us from supplying the wrong code to the offchain witness right now (for zk proving)

crates/evm/src/evm/db_commit.rs
```diff
@@ -53,11 +53,17 @@ impl<'a, C: sov_modules_api::Context> DatabaseCommit for EvmDb<'a, C> {
...
                         .is_some();
                     if !exists_in_db {
-                        self.code
-                            .set(&account_info.code_hash, code, self.working_set);
+                        self.code.set(
+                            &account_info.code_hash,
+                            code,
+                            &mut self.working_set.offchain_state(),
+                        );
                     }
```
eyusufatik: create the offchain working set once and use it?

eyusufatik (changes requested):
great pr

evm db check is VERY IMPORTANT though

eyusufatik (approved):

kpp:
I am not quite sure about this PR. Is this PR equal to

```
--- a/crates/sovereign-sdk/module-system/sov-state/src/prover_storage.rs
+++ b/crates/sovereign-sdk/module-system/sov-state/src/prover_storage.rs
@@ -144,6 +144,7 @@ where
         let batch = state_accesses
             .ordered_writes
             .into_iter()
+            .filter(|(key, _value)| !key.key.starts_with(b"Evm/c"))
             .map(|(key, value)| {
                 let key_hash = KeyHash::with::<DefaultHasher>(key.key.as_ref());
```

?

eyusufatik:
@kpp that works for not including it in the diff but the code would still end up in the state tree. your solution lacks two important things from the current impl.:

1. The code is still committed to the tree, that means when we use contract code in zk, we still do merkle proof verification, but in this pr, comparing with a code hash is sufficient -> better cycles (esp. when keccak256 accelerator is ready)
2. The code is still committed to the tree, that means when we reconstruct tree from DA, we can't end up in the same state root. But in this pr since we only put the code hash to the state tree, we don't have that problem.

Clever thinking though :)

crates/evm/src/evm/handler.rs
```diff
@@ -559,7 +560,8 @@ fn calc_diff_size<EXT, DB: Database>(
             if let Some(code) = account.info.code.as_ref() {
                 // Don't charge for account code if it is already in DB.
                 let db_code = db.code_by_hash(account.info.code_hash)?;
-                if db_code.is_empty() {
+                // This would only add code's size to the diff size in case CANCUN is NOT activated.
+                if db_code.is_empty() && !SPEC::enabled(SpecId::CANCUN) {
```
kpp, line 564: Why?

PR #1400, opened 2024-10-30
Attempt to make seq comm threshold test faster

yaziciahmet:
# Description

test_sequencer_commitment_threshold takes 16 minutes, while all other tests end in 6-8 minutes. So if we can make it somehow faster, we can drop the CI time quite a lot.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


jfldde (approved):
Cheeky. Will be useful for testing TX chunking as well

crates/primitives/src/constants.rs
```diff
@@ -31,4 +31,7 @@ pub const TEST_PRIVATE_KEY: &str =
 pub const MIN_BASE_FEE_PER_GAS: u128 = 10_000_000; // 0.01 gwei
 
 /// Maximum size of a bitcoin transaction body in bytes
+#[cfg(not(feature = "test-utils"))]
```
kpp: Although this works it introduces additional feature flag. We already have prover rpc for this to force it to prove, no?

kpp (changes requested):
Very good but questionable.

crates/primitives/src/constants.rs
yaziciahmet: We do have rpc for forcing to prove. But we don't have rpc to force sequencer to submit a commitment to DA. I don't like too much feat flags either, but `test-utils` is used across our repository widely and might be the one that's actually useful.

kpp: How about adding an rpc method to force seq comm?

yaziciahmet: It might not be so useful tbh.

In all other tests we just keep min_seq_commitments config very low and easily force commitment. In this specific test, the aim is to force a seq comm by making it go above a size threshold. So we don't even test the commitment, we just test that the threshold triggers the action.

eyusufatik: why not go SHORT_PREFIX way?

eyusufatik: I mean instead of adding a feature

yaziciahmet: makes more sense. done.

eyusufatik (approved):

jfldde:
Few places where we need to adapt if we change `SHORT_PREFIX` since we use that in `citrea-e2e` and for `nightly_build_push.yml` workflow

kpp (approved):

Issue #892, opened 2024-07-16
Bitcoin DA: post-submit tx monitoring


eyusufatik:
Track a transaction in the mempool and use CPFP or RBF to bump up the fees if needed.

I'll add more details soon

Issue #1364, opened 2024-10-21
Replace patch.bitcoincore-rpc with an actual replacement 


kpp:
## Summary

If someone is using our crates, `[patch]` section will not be applied, resulting in a broken dep tree (code will not compile)

PR #1407, opened 2024-11-01
Da monitoring service

jfldde:
# Description

- Monitor txs send in `spawn_da_queue`
- Monitor chain re-org and resubmit txs
- Add `DaRpc` server/client
- Add cpfp fee bumping endpoint: `da_bumpFeeCpfp` which takes an Optional<Txid> and a new fee_rate. Bumps latest TX in chain if txid is None
- Add monitoring endpoints : `da_getPendingTransactions`, `da_getTxStatus` and `da_getLastMonitoredTx`
- Revive TX chain on `MonitoringService` initialization and keep as single source of truth for `prev_utxo` and `pending_txs`

## Linked Issues
- Related to #892 

## Tests
- Add a re-org test asserting that txs are properly re-submitted and commitments are part of the updated height post re-org
- Add a cpfp fee bumping test that assert that TX chain is followed.

## Missing
- Add test again full mempool
- Add RBF fee bumping
- Add auto fee bumping trigger if needed


crates/bitcoin-da/src/service.rs
```diff
@@ -242,8 +264,11 @@ impl BitcoinService {
                                                     spendable: true,
                                                     solvable: true,
                                                 });
-
                                                 let _ = request.notify.send(Ok(tx_id));
+
+                                                if let Err(e) = self.monitoring.monitor_transaction_chain(txids).await {
+                                                    error!(?e, "Failed to monitored tx chain");
```
eyusufatik: monitor*

crates/bitcoin-da/src/service.rs
```diff
@@ -358,7 +383,7 @@ impl BitcoinService {
         prev_utxo: Option<UTXO>,
         da_data: DaData,
         fee_sat_per_vbyte: u64,
-    ) -> Result<TxWithId, anyhow::Error> {
+    ) -> Result<(Vec<Txid>, TxWithId), anyhow::Error> {
```
eyusufatik: Just make it Vec<TxWithId>?

crates/bitcoin-da/src/monitoring.rs
```diff
@@ -0,0 +1,502 @@
...
+use bitcoincore_rpc::json::GetTransactionResult;
+use bitcoincore_rpc::{Client, RpcApi};
+use thiserror::Error;
+
+use crate::service::FINALITY_DEPTH;
+
+// Todo pass down lower value for test
+// const DEFAULT_CHECK_INTERVAL: Duration = Duration::from_secs(60);
+const DEFAULT_CHECK_INTERVAL: Duration = Duration::from_millis(10);
+const DEFAULT_HISTORY_LIMIT: usize = 1_000; // Keep track of last 1k txs
```
eyusufatik, line 23: putting size limitation on the history is probably helpful as well.

crates/bitcoin-da/src/monitoring.rs
```diff
@@ -0,0 +1,502 @@
...
+        confirmations: u64,
+    },
+    Finalized {
+        block_hash: BlockHash,
+        block_height: BlockHeight,
+    },
+    Replaced {
+        by_txid: Txid,
+    },
+    Evicted,
```
eyusufatik, line 49: what is evicted

crates/bitcoin-da/src/service.rs
eyusufatik, line 1: can put monitoring service config under bitcoin service config, that way we can set it through our config toml's

eyusufatik (changes requested):

crates/bitcoin-da/src/monitoring.rs
jfldde: That would be the TX status if it ever goes from `Pending` to not found in mempool. 
Could be a good trigger for us to bump fee as eviction can happen in time of congestion when a mempool gets full.

crates/bitcoin-da/src/service.rs
jfldde: You mean a Vec of `Txid` or `TxWithId` ? 
We need the full Transaction from `TxWithId` to build the `prev_utxo`.
We could hold a Vec of Transaction but we still need to call `gettransaction <txid>` in the monitoring service to fetch the number of confirmations and to assert that it's held in mempool

eyusufatik: Vec of TxWithId

crates/bitcoin-da/src/monitoring.rs
jfldde: Other than the DEFAULT_HISTORY_LIMIT you mean ?

crates/bitcoin-da/src/service.rs
jfldde: The TxWithId is used to build the next prev_utxo whereas the Vec<Txid> is used to monitor all the txs.
We don't really need to hold the Transaction itself alongside them since we need to do a gettransaction call anyway to check confirmations.

crates/bitcoin-da/src/monitoring.rs
eyusufatik: yeah the size of all txs in history, might be unnecessary though.

but for the prover for example

assume at high traffic for a long time 

400kb for 2 tx groups (commit + reveal)

totals at 195 MB

hmm, that's fine actually

or is the history just txid?

crates/bitcoin-da/src/service.rs
```diff
@@ -183,13 +215,16 @@ impl BitcoinService {
...
 
     pub fn spawn_da_queue(
         self: Arc<Self>,
         mut rx: UnboundedReceiver<Option<SenderWithNotifier<TxidWrapper>>>,
     ) {
+        self.monitoring.clone().spawn();
+
         tokio::spawn(async move {
             let mut prev_utxo = match self.get_prev_utxo().await {
                 Ok(Some(prev_utxo)) => Some(prev_utxo),
```
eyusufatik: should we get the prev_utxo from the monitoring service now?

and update it from there so if tx is bumped, we use an output from the bump tx

crates/bitcoin-da/src/service.rs
```diff
@@ -357,17 +395,25 @@ impl BitcoinService {
...
         // get all available utxos
         let utxos = self.get_utxos().await?;
 
+        // If Utxo has already been spent, fallback to get_prev_utxo
+        // Would happen in case of cpfp spending reveal output
+        if let Some(prev) = &prev_utxo {
+            if !utxos.iter().any(|utxo| utxo == prev) {
+                prev_utxo = self.get_prev_utxo().await?;
+            }
+        }
```
eyusufatik: (referring to my last comment) and if we do it this way, now we don't need to search in the utxo list, the monitoring chain becomes the source of truth for `prev_utxo`

crates/bitcoin-da/src/service.rs
```diff
@@ -582,6 +628,128 @@ impl BitcoinService {
...
+            for (idx, utxo) in parent_tx.output.iter().enumerate().skip(1) {
+                if total_input > required_fee {
+                    break;
+                }
+                child_tx.input.push(create_tx_input(OutPoint {
+                    txid: parent_txid,
+                    vout: idx as u32,
+                }));
+                total_input += utxo.value;
+            }
```
eyusufatik: as you keep adding this inputs, you have to recalculate the fee needed to pay since the size of the child increases :)

eyusufatik (changes requested):

crates/bitcoin-da/src/service.rs
jfldde: Sounds good, I'll amend

crates/bitcoin-da/src/service.rs
```diff
@@ -684,45 +651,33 @@ impl BitcoinService {
...
-                    break;
-                }
-
                 child_tx.input.push(create_tx_input(OutPoint {
                     txid: utxo.txid,
                     vout: utxo.vout,
                 }));
 
                 total_input += utxo.amount;
+                required_fee = calculate_child_fee(&child_tx);
```
eyusufatik: wouldn't you need total_input >= required_fee + output amount? or smth like that?

crates/bitcoin-da/src/monitoring.rs
jfldde: Sounded good and future proof, added it

Cargo.toml
```diff
@@ -156,4 +156,4 @@ tower = { version = "0.4.13", features = ["full"] }
 hyper = { version = "1.4.0" }
 
 [patch.crates-io]
-bitcoincore-rpc = { version = "0.18.0", git = "https://github.com/chainwayxyz/rust-bitcoincore-rpc.git", rev = "ede8097" }
+bitcoincore-rpc = { version = "0.18.0", git = "https://github.com/chainwayxyz/rust-bitcoincore-rpc.git", rev = "5ce1bed" }
```
eyusufatik, line 159: if this closes the bitcoin core rpc bump issue pls add it to the pr sidebar

bin/citrea/tests/bitcoin_e2e/bitcoin_test.rs
```diff
@@ -0,0 +1,366 @@
...
+        assert!(mempool0.contains(&pending_txs[1].txid));
+
+        sequencer.restart(None).await?;
+
+        // Assert that txs are properly monitored after a restart
+        let pending_txs = sequencer
+            .client
+            .http_client()
+            .da_get_pending_transactions()
+            .await?;
```
eyusufatik, line 230: assert fee is really bumped to where it should be

eyusufatik (changes requested):

Cargo.toml
jfldde: This is not addressing issue #1364 , only bumping to include https://github.com/chainwayxyz/rust-bitcoincore-rpc/pull/6 RPC typing fix

bin/citrea/src/rollup/bitcoin.rs
```diff
@@ -162,6 +166,7 @@ impl RollupBlueprint for BitcoinRollup {
         if require_wallet_check {
             // run only for sequencer and prover
             Arc::clone(&service).spawn_da_queue(rx);
+            service.monitoring.clone().spawn().await?;
```
rakanalh: Please think about how we can use a cancellation token with `ctrl_c` from tokio to support future cancellation for both the da queue service and the monitoring service.

crates/bitcoin-da/src/monitoring.rs
```diff
@@ -0,0 +1,510 @@
...
+        let txids = unspent.into_iter().map(|utxo| utxo.txid).collect();
+
+        self.monitor_transaction_chain(txids).await
+    }
+
+    /// Spawn a tokio task to keep track of TX status and chain re-orgs
+    pub async fn spawn(self: Arc<Self>) -> Result<()> {
+        self.init().await?;
+
+        tokio::spawn(async move {
```
rakanalh: Accept a cancellation token as a param to `spawn` and use it here to stop if the cancellation token is triggered.

crates/bitcoin-da/src/monitoring.rs
```diff
@@ -0,0 +1,510 @@
...
+    // }
+
+    // pub async fn get_monitored_transactions(&self) -> Vec<(Txid, MonitoredTx)> {
+    //     self.monitored_txs
+    //         .read()
+    //         .await
+    //         .iter()
+    //         .map(|(txid, tx)| (*txid, tx.clone()))
+    //         .collect()
+    // }
```
rakanalh: Remove?

bin/citrea/src/rollup/bitcoin.rs
jfldde: [efc1ba1](https://github.com/chainwayxyz/citrea/pull/1407/commits/efc1ba1ffe9b49505ceb610d89a54f7dca03d406) Sharing the task manager and handling it via the main loop select

eyusufatik (approved):

PR #1411, opened 2024-11-04
simplify risc0 host

### LINKED ISSUE #1381: Use risc0's BonsaiProver
Instead of handling local and bonsai proving ourselves, we can use existing environment variables for risc0 to use Bonsai / prove locally.

See [relevant code](https://github.com/risc0/risc0/blob/89f5c830aaf273e45d0e534cee003df78387a51e/risc0/zkvm/src/host/client/prove/mod.rs#L351)

eyusufatik:
# Description
Uses env vars for local / ipc / bonsai prover selection as risc0 intended. 

the checks in `Risc0BonsaiHost::new` is not necessary actually but makes our program fail faster.

## Linked Issues
Closes #1381


crates/risc0-bonsai/src/host.rs
```diff
@@ -39,242 +33,63 @@ pub struct RecoveredBonsaiSession {
...
+        match std::env::var("RISC0_PROVER") {
+            Ok(prover) => match prover.as_str() {
+                "bonsai" => {
+                    if std::env::var("BONSAI_API_URL").is_err()
+                        || std::env::var("BONSAI_API_KEY").is_err()
+                    {
+                        panic!("Bonsai API URL and API key must be set when RISC0_PROVER is set to bonsai");
+                    }
+                }
+                "local" => {}
```
rakanalh:
```suggestion
                "local" => { unimplemented!() }
```

crates/risc0-bonsai/src/host.rs
```diff
@@ -39,242 +33,63 @@ pub struct RecoveredBonsaiSession {
...
+                    panic!("Invalid prover specified: {}", prover);
+                }
+            },
+            Err(_) => {
+                debug!("No prover specified.");
+
+                if std::env::var("BONSAI_API_URL").is_ok()
+                    && std::env::var("BONSAI_API_KEY").is_ok()
+                {
+                    panic!(
```
rakanalh: Can we not assume intent instead of panicking?

crates/risc0-bonsai/src/host.rs
```diff
@@ -39,242 +33,63 @@ pub struct RecoveredBonsaiSession {
...
-                tracing::debug!("Image uploaded");
-
-                Some(client)
-            })
-            .join()
-            .unwrap()
-        } else {
-            None
-        };
+        match std::env::var("RISC0_PROVER") {
```
rakanalh: nit: I would implement

```rust
enum ProverType {}

impl From<&'static str> for ProverType
```
and match prover types... cleaner.

crates/risc0-bonsai/src/host.rs
eyusufatik: we just don't do anything on that case

crates/risc0-bonsai/src/host.rs
eyusufatik: I wanted to be very explicit so that we don't run into a case where we accidentally run some mode until it fails on proving

crates/risc0-bonsai/src/host.rs
eyusufatik: it's not gonna be used anywhere else so I'm not gonna skip this one

jfldde (approved):

Cargo.lock
```diff
@@ -268,7 +297,7 @@ dependencies = [
  "alloy-json-rpc",
  "alloy-network",
  "alloy-network-primitives",
- "alloy-primitives",
+ "alloy-primitives 0.7.7",
```
kpp: This is a duplicate dep alongside with alloy-primitives 0.8.10

kpp (approved):
Nice! Then it's not a risc0-bonsai anymore, it's just risc0

Cargo.lock
eyusufatik: i don't know how that happened tbh

eyusufatik: came with merging nightly

yaziciahmet:
Roll alloy-chains and async-compression back and your lockfile should be fixed.

yaziciahmet (approved):
LGTM

PR #1416, opened 2024-11-04
Parallel proving

### LINKED ISSUE #1283: Parallel proof production
We use `ParallelProverService` which supports initiating multiple proving sessions underneath.

However, we only use it sequentially.

This will not be applicable for light client proofs since they depend on the previous light client proof but the batch proofs can make use of it.

For example when we devide sequencer commitments on DA because of DA TX size limitations, we don't need to prove group by group waiting on the previous proof to end.

We can spawn multiple proof generation tasks for the groups.

yaziciahmet:
# Description

Pr enables parallel proving.

Batch prover always proves in parallel per l1 block. So it is parallelized for each proof in a single l1 block, and does not move to next l1 block until all finishes.

Light client prover proves always one-by-one.

If there are proofs in a block more than the provided env `PARALLEL_PROOF_LIMIT`, then the extra ones are queued, and blocked until one of the proofs finishes.

## Linked Issues
- Fixes #1283 

## Testing

## Docs


crates/prover-services/src/parallel/new.rs
```diff
@@ -0,0 +1,329 @@
...
+            ProofGenMode::Execute => {
+                tracing::info!("Prover is configured to execute proving");
+            }
+            ProofGenMode::Prove => {
+                tracing::info!("Prover is configured to prove");
+            }
+        };
+
+        let thread_pool = rayon::ThreadPoolBuilder::new()
+            .num_threads(thread_pool_size)
```
rakanalh: If we have `thread_pool_size` as option, don't set this value if `None. 

>> If num_threads is 0, or you do not call this function, then the Rayon runtime will select the number of threads automatically. At present, this is based on the RAYON_NUM_THREADS environment variable (if set), or the number of logical CPUs (otherwise).

crates/prover-services/src/parallel/new.rs
```diff
@@ -0,0 +1,329 @@
...
+            num_cpus - 1,
+            ledger_db,
+        )
+    }
+
+    pub fn set_current_da_hash(&mut self, da_hash: <Da::Spec as DaSpec>::SlotHash) {
+        assert!(
+            self.current_da_hash.is_none(),
+            "Da hash to prove should never be set twice"
+        );
```
rakanalh: `assert` here causes panic if there was a bug. Why not just skip setting the value or return an Err? I would avoid `assert` and `unwrap()` and `expect` in production code.

crates/prover-services/src/parallel/new.rs
```diff
@@ -0,0 +1,329 @@
...
+        self.current_da_hash = Some(da_hash);
+    }
+
+    pub fn add_proof_data(&mut self, proof_data: ProofData) {
+        assert!(
```
rakanalh: Same as above

crates/prover-services/src/parallel/new.rs
```diff
@@ -0,0 +1,329 @@
...
+                self.current_da_hash
+            );
+
+            self.current_da_hash = None;
+            self.proof_queue.clear();
+
+            return Ok(vec![]);
+        }
+
+        assert!(
```
rakanalh: Assert = panic

crates/prover-services/src/parallel/new.rs
```diff
@@ -0,0 +1,329 @@
...
+            !self.proof_queue.is_empty(),
+            "Prove should never be called before setting some proofs"
+        );
+
+        // Clear current proof data
+        self.current_da_hash = None;
+        let proof_queue = std::mem::take(&mut self.proof_queue);
+
+        // Prove all
+        let proofs = self.prove(proof_queue).await;
```
rakanalh: What if `proof_queue` had more items than configured number of threads?

crates/prover-services/src/parallel/new.rs
```diff
@@ -0,0 +1,329 @@
...
+    /// Creates a new `ParallelProverService` with the default number of thread pool size.
+    /// Default thread pool size is num_cpus - 1.
+    pub fn new_with_default_workers(
+        da_service: Da,
+        vm: Vm,
+        proof_mode: ProofGenMode<Da, Vm, Stf>,
+        zk_storage: Stf::PreState,
+        ledger_db: LedgerDB,
+    ) -> anyhow::Result<Self> {
+        let num_cpus = num_cpus::get();
```
eyusufatik: we are either gonna prover on gpus or we are going to have an external service do it for us. 
so having an env var instead of num_cpu's call makes sense here. and also for the light client proofs we won't have concurrent proof production


call it 

PROVER_SERVICE_JOB_LIMIT or smth like that.

crates/prover-services/src/parallel/new.rs
```diff
@@ -0,0 +1,329 @@
...
+            ledger_db,
+        )
+    }
+
+    pub fn set_current_da_hash(&mut self, da_hash: <Da::Spec as DaSpec>::SlotHash) {
+        assert!(
+            self.current_da_hash.is_none(),
+            "Da hash to prove should never be set twice"
+        );
+        self.current_da_hash = Some(da_hash);
```
eyusufatik: if we are taking the queue order of the proofs as absolute order why would we need the da hash in prover srevice?

eyusufatik (changes requested):

crates/prover-services/src/parallel/new.rs
```diff
@@ -0,0 +1,329 @@
...
+        let num_threads = self.thread_pool.current_num_threads();
+
+        // Future buffer to keep track of ongoing provings
+        let mut ongoing_proofs = Vec::with_capacity(num_threads);
+        let mut proofs = vec![Proof::default(); proof_queue.len()];
+        // Initialize proof workers
+        for (idx, proof_data) in proof_queue.into_iter().enumerate() {
+            if ongoing_proofs.len() == num_threads {
+                // If no available threads, wait for one of the proofs to finish
+                let ((idx, proof), _, remaining_proofs) = future::select_all(ongoing_proofs).await;
```
rakanalh: I like this :+1:

rakanalh (changes requested):

crates/prover-services/src/parallel/new.rs
yaziciahmet: done

crates/prover-services/src/parallel/new.rs
yaziciahmet: we dont. removed

crates/prover-services/src/parallel/new.rs
yaziciahmet: asserted that thread_pool_size > 0

crates/prover-services/src/parallel/new.rs
yaziciahmet: handled with future::select_all. waits for one of the proofs to end, and starts the  new proof. once all the proofs are started, it waits for all of them at once.

crates/prover-services/src/parallel/new.rs
yaziciahmet: Removed this method completely as its unnecessary.

crates/prover-services/src/parallel/new.rs
yaziciahmet: This is also removed.

crates/prover-services/src/parallel/new.rs
yaziciahmet: I think it is fine to panic, if it is a bug. Because there is no recovery back from a bug. And I think it makes it easier to actually figure the bug out instead of proving being finished immediately and trying to debug why is that.

crates/prover-services/src/parallel/new.rs
yaziciahmet: Removed this method completely as its unnecessary.

bin/citrea/src/rollup/mock.rs
```diff
@@ -124,26 +126,30 @@ impl RollupBlueprint for MockDemoRollup {
...
-        .expect("Should be able to instantiate prover service")
+        let proof_mode = match prover_config.proving_mode {
+            ProverGuestRunConfig::Skip => ProofGenMode::Skip,
+            ProverGuestRunConfig::Simulate => {
+                let stf_verifier = StateTransitionVerifier::new(zk_stf, da_verifier);
+                ProofGenMode::Simulate(stf_verifier)
+            }
+            ProverGuestRunConfig::Execute => ProofGenMode::Execute,
+            ProverGuestRunConfig::Prove => ProofGenMode::Prove,
+        };
```
eyusufatik, line 142: why not modify the proofguestrunconfig instead of mapping it to this new struct

eyusufatik (approved):
looks ok. I didn't see if you implemented it this way, however, if you can take the sending of the proofs from prover service to the prover, it'd be great

bin/citrea/src/rollup/mock.rs
yaziciahmet: This mapping to new structure existed before. I just carried it to here. And renamed ProofGenConfig to ProofGenMode. And the reason we have this is ProverGuestRunConfig is used in the ProverConfig read from the config file, so it doesn't make sense for it to have a variant with Simulate(stf_verifier).

rakanalh (approved):

PR #1417, opened 2024-11-04
Don't hash da headers more than once

### LINKED ISSUE #1414: Cache computed header hash in bitcoin service
every time we invoke `DaService::BlockHeader::hash` on bitcoin da we recompute the header hash, instead we can calculate it once and use that computation on the next invocations.

eyusufatik:
# Description

I realized whenever we called `BlockHeaderTrait::hash`, we computed bitcoin header hash again and again.

Added capability to supply a precomputed hash, which can be verified once and used from the precomputation

## Results

*Nightly*: SessionStats { segments: 5292, total_cycles: 5548277760, user_cycles: 3819206639 }
*This branch*: SessionStats { segments: 5211, total_cycles: 5464129536, user_cycles: 3774273877 }

This is against Bitcoin testnet4 block #52685 's first sequencer commitment group, it has very few DA blocks and very high evm gas consumption, so look at the raw reduction not in percentage wise :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,7 @@ test-legacy: ## Runs test suite with output from tests printed
 	@cargo test -- --nocapture -Zunstable-options --report-time
 
 test: build $(EF_TESTS_DIR) ## Runs test suite using next test
-	@cargo nextest run --workspace --all-features --no-fail-fast $(filter-out $@,$(MAKECMDGOALS))
+	RISC0_DEV_MODE=1 cargo nextest run --workspace --all-features --no-fail-fast $(filter-out $@,$(MAKECMDGOALS))
 
 install-dev-tools:  ## Installs all necessary cargo helpers
 	cargo install --locked dprint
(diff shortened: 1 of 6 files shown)


eyusufatik:
up to %1.7 reduction in another proof 😓

Makefile
```diff
@@ -35,7 +35,7 @@ test-legacy: ## Runs test suite with output from tests printed
 	@cargo test -- --nocapture -Zunstable-options --report-time
 
 test: build $(EF_TESTS_DIR) ## Runs test suite using next test
-	@cargo nextest run --workspace --all-features --no-fail-fast $(filter-out $@,$(MAKECMDGOALS))
+	RISC0_DEV_MODE=1 cargo nextest run --workspace --all-features --no-fail-fast $(filter-out $@,$(MAKECMDGOALS))
```
kpp, line 38: RISC0_DEV_MODE=1 - is it okay? It will not produce proofs so `make test` will fail on basic_prover_test

kpp (approved):

Makefile
eyusufatik: no dev mode produces "fake proofs" which are verifiable by processes that also have risc0_dev_mode=1 :)

ercecan (approved):

yaziciahmet (approved):

PR #1428, opened 2024-11-05
Check batch proofs stateroots and ensures integrity in light client

### LINKED ISSUE #1422: Stitch and check pre and post state roots from batch proofs
The order of batch proofs might be wrong in the da layer, the light client proof should handle accordingly.

Assume they are almost always be in order though, so you can optimize accordingly.

ercecan:
# Description
This pr chains the state roots and updates the light client proof state root accordingly, if there are unchainable state roots in the output they are moved to the next light client proof
This pr also adds one field to batch proof circuit output: last l2 height

## Linked Issues
- Fixes #1422


crates/light-client-prover/src/circuit.rs
```diff
@@ -61,23 +62,55 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+
+    // Deserialize batch proof journals
+    let mut deserialized_journals: Vec<_> = input
+        .batch_proof_journals
+        .iter()
+        .map(|journal| {
+            StateTransition::<DaV::Spec, [u8; 32]>::try_from_slice(journal)
+                .expect("Should have deserialized the journal")
+        })
+        .collect();
```
eyusufatik: i'd go for verify_and_extract here too

crates/light-client-prover/src/circuit.rs
```diff
@@ -61,23 +62,55 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
-    // complete proof has raw bytes inside
-    // to extract *and* verify the proof you need to use the zk guest
-    // can be passed from the guest code to this function
+    // For the output state root, update with the latest state root if a batch proof journal is present
+    // Otherwise, use the state root from the previous light client proof
+    // If it is the first light client proof and there are no journals zero it out
+    let final_state_root = deserialized_journals.last().map_or_else(
+        || {
+            deserialized_previous_light_client_proof_journal
+                .map_or([0u8; 32], |journal| journal.state_root)
```
eyusufatik: smart

crates/light-client-prover/src/circuit.rs
```diff
@@ -61,23 +62,55 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+        .map(|journal| {
+            StateTransition::<DaV::Spec, [u8; 32]>::try_from_slice(journal)
+                .expect("Should have deserialized the journal")
+        })
+        .collect();
+
+    // Ensure the journals are sorted and check state root order
+    if !deserialized_journals.is_empty() {
+        // Sort the journals first
+        deserialized_journals.sort_unstable_by_key(|journal| journal.sequencer_commitments_range.0);
```
eyusufatik: is this sort O(n) if the list is already sorted?

eyusufatik: if not use a sorting algo that has this

crates/light-client-prover/src/circuit.rs
```diff
@@ -61,23 +62,55 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+
+        // Now safely access the first journal after sorting
+        if let Some(first_journal) = deserialized_journals.first() {
+            if let Some(previous_journal) = &deserialized_previous_light_client_proof_journal {
```
eyusufatik: name this previous light client output

crates/light-client-prover/src/circuit.rs
eyusufatik, line 1: instead of naming journal, we can use proof output, or output

eyusufatik (changes requested):

eyusufatik:
now that I think about it sorting on sequencer commitment range doesn't work

they can be batch proofs for different da blocks

crates/light-client-prover/src/circuit.rs
```diff
@@ -61,23 +67,69 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+                i += 1;
+            } else {
+                // Use the map to find the correct journal quickly
+                if let Some(&pos) = journal_map.get(&current_final_state_root) {
+                    if pos > i {
+                        // Swap the found matching journal to the current position
+                        deserialized_outputs.swap(i, pos);
+                        // Update the map to reflect the swap
+                        journal_map.insert(deserialized_outputs[pos].initial_state_root, pos);
+                        journal_map.insert(deserialized_outputs[i].initial_state_root, i);
```
yaziciahmet: I think this is completely unnecessary. We will have at most like 10, maybe 15 batch proofs in the array, and I am pretty sure that iteration over list is going to be much faster for such a small list.

yaziciahmet (dismissed):

crates/light-client-prover/src/circuit.rs
ercecan: I agree as well, I have both versions check https://github.com/chainwayxyz/citrea/pull/1428/commits/791f44138235dc6bd4cb0a15879c47eb7501c3cb 
We will benchmark and see
But personally I would also prefer the the one in https://github.com/chainwayxyz/citrea/pull/1428/commits/791f44138235dc6bd4cb0a15879c47eb7501c3cb because 99% it will be ordered

ercecan:
After further discussions with @ekrembal and @eyusufatik we have decided to change how we do things. Now we will keep a set of state roots that are not chained together because a previous state root might not be in the current l1 block for example. We will still generate light client proof even if the last state does not change, we will pass the unverified stateroots to the next light client proof as an input and update the last state root accordingly.
In order to block replay attacks we will add another field to the batch proof: Last L2 Height

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
```diff
@@ -221,6 +251,10 @@ pub struct LightClientCircuitOutput {
     /// The method id of the light client proof
     /// This is used to compare the previous light client proof method id with the input (current) method id
     pub light_client_proof_method_id: [u32; 8],
+    /// Batch proof info from current or previous light client proofs that were not changed and unable to update the state root yet
+    pub unverified_batch_proofs_info: Vec<BatchProofInfo>,
```
eyusufatik: I'd call this something else

crates/light-client-prover/src/circuit.rs
```diff
@@ -61,23 +66,107 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+                continue;
+            }
+            if let Some((final_root, last_l2)) =
+                initial_to_final.remove(&unverified_info.final_state_root)
+            {
+                initial_to_final
+                    .entry(unverified_info.initial_state_root)
+                    .or_insert((final_root, last_l2));
+            }
+        }
```
eyusufatik: for loop does nothing

crates/light-client-prover/src/circuit.rs
```diff
@@ -61,23 +66,107 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+
+        for output in deserialized_outputs.iter() {
+            if output.last_l2_height <= previous_output.last_l2_height {
+                continue;
+            }
+
+            if let Some((final_root, last_l2)) = initial_to_final.remove(&output.final_state_root) {
```
eyusufatik: have to check if final_root is also used as key in the map here

eyusufatik: so what you pop, you also check that

crates/light-client-prover/src/circuit.rs
eyusufatik, line 1: functionize the traversal, we have to write a lot of unit tests

crates/light-client-prover/src/circuit.rs
```diff
@@ -61,23 +66,107 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+        let entries: Vec<(_, (_, _))> = initial_to_final
+            .iter()
+            .map(|(k, v)| (*k, (v.0, v.1)))
+            .collect();
+
+        for (initial_root, (final_root, last_l2)) in entries {
+            // Directly remove the entry to ensure there is no simultaneous mutable and immutable borrow
+            if let Some((final_root_value, last_l2_value)) = initial_to_final.remove(&final_root) {
+                // Insert the new entry after removing to avoid mutable borrow issues
+                initial_to_final.insert(initial_root, (final_root_value, last_l2_value));
```
eyusufatik: again the popped final root should be checked

eyusufatik (commented):
might want to write this in a recursive style to make it much much easier

crates/light-client-prover/src/circuit.rs
```diff
@@ -61,23 +66,231 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+    // If not, skip for now
+    // TODO: Once we have a manually planted light client proof use that and assume prev light client proof always exists
+    // So there will be no need for all these if lets
+    if let Some(previous_output) = &deserialized_previous_light_client_proof_journal {
+        last_l2_height = previous_output.last_l2_height;
+        last_state_root = previous_output.state_root;
+        for unchained_info in previous_output.unchained_batch_proofs_info.iter() {
+            if unchained_info.last_l2_height <= previous_output.last_l2_height {
+                continue;
+            }
```
eyusufatik: shouldn't be necessary

crates/light-client-prover/src/circuit.rs
```diff
@@ -61,23 +66,231 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+                ),
+            );
+        }
+
+        // Do recursive matching for previous state root
+        recursive_match_state_roots(
+            &mut initial_to_final,
+            &BatchProofInfo::new(
+                previous_output.state_root,
+                previous_output.state_root,
```
eyusufatik: initial sr should be our genesis state root actually.

we can make this an optional input to the circuit.

we supply it in the first ever LC proof ourselves. then we output it on every proof.

eyusufatik (commented):
didnt read tests. rest lgtm

crates/light-client-prover/src/circuit.rs
```diff
@@ -61,23 +66,231 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+        for output in deserialized_outputs.iter() {
+            // Do not add if last l2 height is smaller or equal to previous output
+            // This is to defend against replay attacks, for example if somehow there is the script of batch proof 1 we do not need to go through it again
+            if output.last_l2_height <= previous_output.last_l2_height {
+                continue;
+            }
+            recursive_match_state_roots(
+                &mut initial_to_final,
+                &BatchProofInfo::new(
+                    previous_output.state_root,
```
ercecan: shouldn't this be output.initial_state_root?

ercecan: yes

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
```diff
@@ -221,6 +251,10 @@ pub struct LightClientCircuitOutput {
     /// The method id of the light client proof
     /// This is used to compare the previous light client proof method id with the input (current) method id
     pub light_client_proof_method_id: [u32; 8],
+    /// Batch proof info from current or previous light client proofs that were not changed and unable to update the state root yet
+    pub unchained_batch_proofs_info: Vec<BatchProofInfo>,
+    /// Last l2 height the light client proof verifies
+    pub last_l2_height: u64,
```
eyusufatik, line 254: you might think we don't need to output the l2_genesis_root

but we have to, or else we can generate a different state root starting from a different genesis

eyusufatik: the genesis root output of LCP will be asserted in the bridge itself

crates/light-client-prover/src/circuit.rs
```diff
@@ -61,23 +68,260 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+
+    let fourth = initial_to_final.get(&[45u8; 32]).unwrap();
+    assert_eq!(fourth.0, [46u8; 32]);
+    assert_eq!(fourth.1, 46);
+
+    let bp_info_prev = BatchProofInfo::new([1u8; 32], [1u8; 32], 1);
+
+    recursive_match_state_roots(&mut initial_to_final, &bp_info_prev);
+
+    assert_eq!(initial_to_final.len(), 4);
```
eyusufatik: we need way more tests

crates/light-client-prover/src/circuit.rs
```diff
@@ -61,23 +68,260 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+            recursive_match_state_roots(
+                &mut initial_to_final,
+                &BatchProofInfo::new(
+                    output.initial_state_root,
+                    output.final_state_root,
+                    output.last_l2_height,
+                ),
+            );
+        }
+
```
eyusufatik: this cant happen. assert deserialized outputs is empty

crates/light-client-prover/src/circuit.rs
```diff
@@ -61,23 +68,260 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+        // Do recursive matching for previous/genesis state root
+        recursive_match_state_roots(
+            &mut initial_to_final,
+            &BatchProofInfo::new(genesis_state_root, genesis_state_root, 0),
+        );
```
eyusufatik: what the hell man?  😩
you should insert batch proofs from da to the initial_to_final

and then maybe you can pop recursive

eyusufatik (changes requested):
looks like you got lost at some point

ercecan:
> looks like you got lost at some point

talked f2f to ensure I did not :D

eyusufatik (approved):

PR #1431, opened 2024-11-05
Return reference from SignedSoftConfirmation fields

### LINKED ISSUE #1413: `SignedSoftConfirmation::txs` always clones
It could return reference to its txs field and it could be up to the invoker to clone or not. there is many places it's cloned unnecessarily.

yaziciahmet:
# Description

Return reference from SignedSoftConfirmation fields, and clone them whenever the caller needs to. Also changes the txs input to apply_soft_confirmation_txs be a slice so we can avoid cloning it in couple of places. This could be useful specifically for txs since it potentially can be really big.

## Linked Issues
- Fixes #1413 

## Testing

## Docs


crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/stf_blueprint.rs
```diff
@@ -63,15 +63,16 @@ where
...
         mut sc_workspace: WorkingSet<C>,
     ) -> (WorkingSet<C>, Vec<TransactionReceipt<TxEffect>>) {
         let mut tx_receipts = Vec::with_capacity(txs.len());
         for raw_tx in txs {
-            let raw_tx_hash = <C as Spec>::Hasher::digest(&raw_tx).into();
+            let raw_tx_hash = <C as Spec>::Hasher::digest(raw_tx).into();
             // Stateless verification of transaction, such as signature check
             // TODO: https://github.com/chainwayxyz/citrea/issues/1061
-            let tx = Transaction::<C>::deserialize_reader(&mut &*raw_tx)
+            let mut reader = std::io::Cursor::new(raw_tx);
```
eyusufatik, line 74: why reader?

yaziciahmet: deserialize_reader requires a mutable reader, and I can not get mutable slice as method argument since it would require making tons of methods and structs mutable all the way up. So I have to create a slice reader.

eyusufatik (approved):

PR #1433, opened 2024-11-05
Don't clone SignedSoftConfirmation::txs

### LINKED ISSUE #1413: `SignedSoftConfirmation::txs` always clones

kpp:
# Description

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/1413
- Related to https://github.com/chainwayxyz/citrea/pull/1431

### DIFF
--- a/crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
+++ b/crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
@@ -3,6 +3,7 @@
 
 extern crate alloc;
 
+use alloc::borrow::Cow;
 use alloc::vec::Vec;
 use core::fmt::Debug;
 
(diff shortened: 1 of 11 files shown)


crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
```diff
@@ -82,23 +83,23 @@ impl UnsignedSoftConfirmation {
...
+pub struct SignedSoftConfirmation<'txs> {
     l2_height: u64,
     hash: [u8; 32],
     prev_hash: [u8; 32],
     da_slot_height: u64,
     da_slot_hash: [u8; 32],
     da_slot_txs_commitment: [u8; 32],
     l1_fee_rate: u128,
-    txs: Vec<Vec<u8>>,
+    txs: Cow<'txs, [Vec<u8>]>,
```
yaziciahmet, line 95: why not Arc? I think we don't mutate txs at all, and we pass it around quite a lot.

yaziciahmet (approved):
Perfekto! Just have one question.

crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
kpp: Because we either own the data (from db/rpc request) or we borrow it (from UnsignedSoftConfirmation). So it's an Rc with strong ref = 1, and no weak refs at all. We don't need Arc, because we don't share it between threads.

PR #1439, opened 2024-11-06
Implement header chain proof for light client

### LINKED ISSUE #1421: Add header chain proof to our light client proofs
See https://github.com/chainwayxyz/risc0-to-bitvm2/blob/main/circuits/src/header_chain.rs

We can add this to our light client proofs easily, either by porting code from that repo or copy pasting :)

One problem we'll have is that the circuit has access to header's through our DA abstractions.

Might need some changes to those.

But I think the validity condition abstractions can support this.

yaziciahmet:
# Description

Implements header chain proof for light client. Here are the core verifications that are being done:

1. previous light client proof is valid, and has matching method id.
2. da block hash is valid.
3. block heights of prev da block and current da block are consecutive.
4. block hash of prev da block matches current da block's prev hash.
5. target bits of prev da block matches current da block's bits.
6. current da block hash satisfies the current blockchain target.
7. timestamp of current da block is valid.

Some updates to expected values for the next block are:

1. If this is an epoch starting block, set the epoch start time
2. If this is the block one before the epoch start block, update the expected target bits

PR also removes the journals from the light client proof input, as it can be extracted from the da_data.

NOTE: After hardcoding the starting da block into the program, there are still couple of things that needs to be done for header chain proof.

## Linked Issues
- Fixes #1421 

## Testing

## Docs


crates/light-client-prover/src/circuit.rs
```diff
@@ -61,23 +85,208 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+        da_block_height: block_updates.height,
+        da_total_work: block_updates.total_work,
+        da_current_target_bits: block_updates.current_target_bits,
+        da_epoch_start_time: block_updates.epoch_start_time,
+        da_prev_11_timestamps: block_updates.prev_11_timestamps,
     })
 
     // First
 }
+
```
eyusufatik: move pow verify related things to another file

crates/sovereign-sdk/full-node/db/sov-db/src/schema/types.rs
```diff
@@ -72,13 +72,35 @@ pub struct StoredSlot {
...
+/// The on-disk format for a light client proof output
+#[derive(Debug, PartialEq, BorshDeserialize, BorshSerialize)]
+pub struct StoredLightClientProofOutput {
+    /// State root of the node after the light client proof
+    pub state_root: [u8; 32],
+    /// The method id of the light client proof
+    /// This is used to compare the previous light client proof method id with the input (current) method id
+    pub light_client_proof_method_id: [u32; 8],
+    /// Proved DA block's header hash
+    /// This is used to compare the previous DA block hash with first batch proof's DA block hash
```
eyusufatik, line 84: good catch

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -217,26 +211,36 @@ where
             da_block_header: l1_block.header().clone(),
             batch_prover_da_pub_key: self.batch_prover_da_pub_key.clone(),
             batch_proof_method_id: batch_proof_method_id.clone().into(),
-            batch_proof_journals,
```
eyusufatik, line 228: good that we finally removed this

eyusufatik (changes requested):
do the verification in a function of the da verifier trait

crates/light-client-prover/src/circuit.rs
yaziciahmet: done

eyusufatik (approved):

Issue #1045, opened 2024-08-27
Don't reserialize sov-tx


eyusufatik:
see comment in code

https://github.com/chainwayxyz/citrea/blob/4126c9cdb542fa7eeb4b1f60991624a8e413101f/crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/stf_blueprint.rs#L141

kpp:
Fixed in https://github.com/chainwayxyz/citrea/pull/1440

PR #1440, opened 2024-11-07
remove TransactionReceipt::body_to_save from stf

### LINKED ISSUE #1412: remove `TransactionReceipt::body_to_save` from stf
body to save is a field which holds serialized sov-tx's.

it's used by nodes to keep record of sov-tx's in soft confirmations.

however, this is filled in STF, so used in zk :D 

this unfortunately very stupid and at the time of initial implementation we never thought of just pulling it from the soft confirmation on our own at the node level, not stf

kpp:
# Description

Don't keep TransactionReceipt::body_to_save in STF, instead pass it from outside.

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/1412  https://github.com/chainwayxyz/citrea/issues/1045

### DIFF
--- a/crates/batch-prover/src/runner.rs
+++ b/crates/batch-prover/src/runner.rs
@@ -19,7 +19,7 @@ use sequencer_client::{GetSoftConfirmationResponse, SequencerClient};
 use sov_db::ledger_db::BatchProverLedgerOps;
 use sov_db::schema::types::{BatchNumber, SlotNumber};
 use sov_modules_api::storage::HierarchicalStorageManager;
-use sov_modules_api::{Context, SlotData};
+use sov_modules_api::{Context, SignedSoftConfirmation, SlotData};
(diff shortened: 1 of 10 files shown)


crates/batch-prover/src/runner.rs
```diff
@@ -388,6 +388,7 @@ where
             .storage_manager
             .create_storage_on_l2_height(l2_height)?;
 
+        let mut signed_soft_confirmation: SignedSoftConfirmation = soft_confirmation.clone().into();
```
eyusufatik, line 391: just clone txs?

eyusufatik (commented):
small looking but very important pr

crates/batch-prover/src/runner.rs
kpp: Later we need signed_soft_confirmation too. 5 lines below

eyusufatik (approved):

PR #1445, opened 2024-11-08
Gradual contract code migration to offchain storage

### LINKED ISSUE #1441: Migrate EVM's code state map to the offchain state map

rakanalh:
# Description

## Linked Issues
- Fixes #1441

### DIFF
--- a/crates/evm/src/evm/db.rs
+++ b/crates/evm/src/evm/db.rs
@@ -2,7 +2,8 @@
 use std::collections::HashMap;
 
 use reth_primitives::{keccak256, Address, B256};
-use revm::primitives::{AccountInfo as ReVmAccountInfo, Bytecode, U256};
+use revm::primitives::SpecId::CANCUN;
+use revm::primitives::{AccountInfo as ReVmAccountInfo, Bytecode, SpecId, U256};
 use revm::Database;
 use sov_modules_api::{StateMapAccessor, WorkingSet};
 use sov_state::codec::BcsCodec;
@@ -34,23 +35,29 @@ impl std::error::Error for DBError {
 
 pub(crate) struct EvmDb<'a, C: sov_modules_api::Context> {
     pub(crate) accounts: sov_modules_api::StateMap<Address, AccountInfo, BcsCodec>,
-    pub(crate) code: sov_modules_api::OffchainStateMap<B256, Bytecode, BcsCodec>,
+    pub(crate) code: sov_modules_api::StateMap<B256, Bytecode, BcsCodec>,
+    pub(crate) offchain_code: sov_modules_api::OffchainStateMap<B256, Bytecode, BcsCodec>,
     pub(crate) last_block_hashes: sov_modules_api::StateMap<U256, B256, BcsCodec>,
     pub(crate) working_set: &'a mut WorkingSet<C>,
+    pub(crate) current_spec: SpecId,
 }
 
 impl<'a, C: sov_modules_api::Context> EvmDb<'a, C> {
     pub(crate) fn new(
         accounts: sov_modules_api::StateMap<Address, AccountInfo, BcsCodec>,
-        code: sov_modules_api::OffchainStateMap<B256, Bytecode, BcsCodec>,
+        code: sov_modules_api::StateMap<B256, Bytecode, BcsCodec>,
+        offchain_code: sov_modules_api::OffchainStateMap<B256, Bytecode, BcsCodec>,
         last_block_hashes: sov_modules_api::StateMap<U256, B256, BcsCodec>,
         working_set: &'a mut WorkingSet<C>,
+        current_spec: SpecId,
     ) -> Self {
         Self {
             accounts,
             code,
+            offchain_code,
             last_block_hashes,
             working_set,
+            current_spec,
         }
     }
 
@@ -80,6 +87,17 @@ impl<'a, C: sov_modules_api::Context> EvmDb<'a, C> {
             );
         }
     }
+
+    pub(crate) fn check_against_code_hash(
+        &self,
+        code: &Bytecode,
+        code_hash: &B256,
+    ) -> Result<(), DBError> {
+        if *code_hash != keccak256(code.original_bytes()) {
+            return Err(DBError::CodeHashMismatch);
+        }
+        Ok(())
+    }
 }
 
 impl<'a, C: sov_modules_api::Context> Database for EvmDb<'a, C> {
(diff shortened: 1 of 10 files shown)


crates/evm/src/evm/db_commit.rs
```diff
@@ -51,13 +51,13 @@ impl<'a, C: sov_modules_api::Context> DatabaseCommit for EvmDb<'a, C> {
...
-                    let offchain_state = &mut self.working_set.offchain_state();
                     let exists_in_db = self
                         .code
-                        .get(&account_info.code_hash, offchain_state)
+                        .get(&account_info.code_hash, &mut self.working_set)
                         .is_some();
                     if !exists_in_db {
-                        self.code.set(&account_info.code_hash, code, offchain_state);
+                        self.code
+                            .set(&account_info.code_hash, code, &mut self.working_set);
```
eyusufatik: for backwards compatability you also have to set according to spec :)

crates/evm/src/evm/db_init.rs
```diff
@@ -20,8 +20,7 @@ impl<'a, C: sov_modules_api::Context> InitEvmDb for EvmDb<'a, C> {
     }
 
     fn insert_code(&mut self, code_hash: B256, code: Bytecode) {
-        self.code
-            .set(&code_hash, &code, &mut self.working_set.offchain_state())
+        self.code.set(&code_hash, &code, &mut self.working_set)
```
eyusufatik: again spec

crates/evm/src/genesis.rs
```diff
@@ -162,7 +162,29 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+        spec.sort_by(|a, b| a.0.cmp(&b.0));
+
+        if spec.is_empty() {
+            spec.push((0, SpecId::SHANGHAI));
+        } else if spec[0].0 != 0u64 {
+            panic!("EVM spec must start from block 0");
+        }
+
+        let (_, current_spec) = spec.last().expect("Spec should be set");
+        let mut evm_db = self.get_db(working_set, *current_spec);
```
eyusufatik, line 187: you can also impl. this according to spec, so that we can make genesis of mainnet cancun

crates/evm/src/query.rs
```diff
@@ -352,9 +352,7 @@ impl<C: sov_modules_api::Context> Evm<C> {
 
         let account = self.accounts.get(&address, working_set).unwrap_or_default();
         let code = if let Some(code_hash) = account.code_hash {
-            self.code
-                .get(&code_hash, &mut working_set.offchain_state())
-                .unwrap_or_default()
+            self.code.get(&code_hash, working_set).unwrap_or_default()
```
eyusufatik: would need to query both if cancun is activated

crates/evm/src/evm/db.rs
```diff
@@ -92,13 +110,31 @@ impl<'a, C: sov_modules_api::Context> Database for EvmDb<'a, C> {
...
+            if self
+                .offchain_code
+                .get(&code_hash, &mut self.working_set.offchain_state())
+                .is_none()
+                && self.current_spec.is_enabled_in(CANCUN)
+            {
+                self.offchain_code
+                    .set(&code_hash, &code, &mut self.working_set.offchain_state());
             }
+            self.check_against_code_hash(&code, &code_hash)?;
```
eyusufatik: if reading from the old state map, no need to verify with hash

crates/evm/src/evm/db.rs
```diff
@@ -92,13 +110,31 @@ impl<'a, C: sov_modules_api::Context> Database for EvmDb<'a, C> {
...
+                return Ok(code);
+            }
+        }
+        let code = self.code.get(&code_hash, &mut self.working_set);
         if let Some(code) = code {
-            if code_hash != keccak256(code.original_bytes()) {
-                return Err(DBError::CodeHashMismatch);
+            // Gradually migrate contract codes into the offchain code state map.
+            if self
+                .offchain_code
```
eyusufatik: the first part of the if is unnecessary, since we check it in the if before actually

eyusufatik (changes requested):

crates/evm/src/genesis.rs
rakanalh: This will change based on your PR. There's nothing to check here based on the spec (contract code wise)

eyusufatik: there is

you push directly to offchain if first activated spec is Fork1

rakanalh: `insert_code` does the check internally.

crates/evm/src/lib.rs
```diff
@@ -80,7 +80,13 @@ pub struct Evm<C: sov_modules_api::Context> {
 
     /// Mapping from code hash to code. Used for lazy-loading code into a contract account.
     #[state(rename = "c")]
-    pub(crate) code: sov_modules_api::OffchainStateMap<
+    pub(crate) code:
+        sov_modules_api::StateMap<reth_primitives::B256, revm::primitives::Bytecode, BcsCodec>,
+
+    /// Mapping from code hash to code. Used for lazy-loading code into a contract account.
```
eyusufatik, line 86: let's add to comment smth like this

actvated after fork1

crates/evm/src/query.rs
```diff
@@ -348,13 +349,23 @@ impl<C: sov_modules_api::Context> Evm<C> {
         block_id: Option<BlockId>,
         working_set: &mut WorkingSet<C>,
     ) -> RpcResult<reth_primitives::Bytes> {
+        let cfg = self
+            .cfg
+            .get(working_set)
+            .expect("EVM chain config should be set");
+        let (_, current_spec) = cfg.spec.last().expect("Spec should be set");
```
eyusufatik, line 357: this won't work actually, we don't update cfg, it was a mistake to impl. it like that. see the same file in my pr

crates/evm/src/query.rs
```diff
@@ -348,13 +349,23 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
 
         let account = self.accounts.get(&address, working_set).unwrap_or_default();
         let code = if let Some(code_hash) = account.code_hash {
-            self.code
-                .get(&code_hash, &mut working_set.offchain_state())
-                .unwrap_or_default()
+            if current_spec.is_enabled_in(CANCUN) {
+                self.offchain_code
+                    .get(&code_hash, &mut working_set.offchain_state())
+                    .unwrap_or_default()
```
eyusufatik: code might have not been migratted, so if cancun is enabled but the code cant be found in offchain code, try to get it from old code (you cant migrate it here tho, so just get)

crates/evm/src/query.rs
eyusufatik: put todo, I'll do it in my pr

eyusufatik (changes requested):

eyusufatik (approved):

PR #1449, opened 2024-11-08
Remove final state root from batch proof input

### LINKED ISSUE #1345: Assert proof outputs
On full nodes we verify the ZK proofs on DA, however, some of the outputs has to be checked against known data as these are trusted inputs to the ZK circuit.

These are:

- `initial_state_root`
- `initial_batch_hash`
- `validity condition` -- i still don't know if this is relevant to our rollup system

yaziciahmet:
# Description

We don't need to pass in final state root and verify it in the zk code, since the path to the final state root is already what is verified.

## Linked Issues
- Related to #1345 

## Testing

## Docs


crates/batch-prover/src/proving.rs
```diff
@@ -247,7 +240,7 @@ where
 }
 
 pub(crate) fn state_transition_already_proven<StateRoot, Witness, Da>(
-    input: &BatchProofCircuitInput<StateRoot, Witness, Da::Spec>,
+    input: &BatchProofCircuitInputV2<StateRoot, Witness, Da::Spec>,
```
eyusufatik, line 230: create issue for this function

we don't need to prepare the whole input to understand the same proof has been processed before

eyusufatik (changes requested):

crates/batch-prover/src/proving.rs
yaziciahmet: #1456

eyusufatik (approved):

PR #1453, opened 2024-11-08
Don't serialize apply_soft_confirmation_txs txs

### LINKED ISSUE #1452: StfBlueprintTrait::apply_soft_confirmation_txs - accept Vec<Transaction> instead of Vec<Vec<u8>>
So we don't need to borsh::from_slice for Transaction<C: Context>.

kpp:
# Description

Pass them unborshed into zk context.

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/1452


crates/sovereign-sdk/module-system/sov-modules-api/src/transaction.rs
```diff
@@ -71,6 +72,16 @@ impl<C: Context> Transaction<C> {
     }
 }
 
+impl<C: Context> TransactionTrait for Transaction<C> {
```
kpp: I am not sure if I can calc hash this way or I need to add a hash field to Transaction struct

crates/batch-prover/src/da_block_handler.rs
```diff
@@ -66,9 +66,10 @@ where
     pending_l1_blocks: VecDeque<<Da as DaService>::FilteredBlock>,
     _state_root: PhantomData<StateRoot>,
     _witness: PhantomData<Witness>,
+    _tx: PhantomData<fn() -> Tx>,
```
rakanalh:
```suggestion
    _tx: PhantomData<Tx>,
```
Why not this?

crates/batch-prover/src/da_block_handler.rs
```diff
@@ -82,6 +83,7 @@ where
         + AsRef<[u8]>
         + Debug,
     Witness: Default + BorshDeserialize + BorshSerialize + Serialize + DeserializeOwned,
+    Tx: Clone + BorshDeserialize + BorshSerialize + 'static,
```
rakanalh: You're using `'static` anyway so i don't see the point of `fn() ->` part above.

crates/sovereign-sdk/rollup-interface/src/state_machine/stf.rs
```diff
@@ -182,6 +188,14 @@ pub struct SoftConfirmationResult<S, Cs, T, W, Da: DaSpec> {
 ///  - batch: Set of transactions grouped together, or block on L2
 ///  - blob: Non serialised batch or anything else that can be posted on DA layer, like attestation or proof.
 pub trait StateTransitionFunction<Da: DaSpec> {
+    /// The type of rollup transaction
+    type Transaction: TransactionTrait
```
rakanalh:
```suggestion
    type Transaction: TransactionDigest
```

crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
```diff
@@ -13,18 +13,19 @@ use serde::{Deserialize, Serialize};
...
 /// Contains raw transactions and information about the soft confirmation block
 #[derive(Debug, PartialEq, BorshSerialize)]
-pub struct UnsignedSoftConfirmation<'txs> {
+pub struct UnsignedSoftConfirmation<'txs, Tx> {
     l2_height: u64,
     da_slot_height: u64,
     da_slot_hash: [u8; 32],
     da_slot_txs_commitment: [u8; 32],
     txs: &'txs [Vec<u8>],
+    txs_new: &'txs [Tx],
```
rakanalh: We can probably find a better name for this.

We can name `txs` as `serialized_txs` and leave `txs` as the unserialized ones.

rakanalh: Or `txs` becomes `blobs` and `txs_new` becomes `txs`

crates/batch-prover/src/da_block_handler.rs
kpp: https://www.reddit.com/r/rust/comments/1cktu0g/comment/l2pwii0/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
kpp: I mean that's a temporary name until we get rid of Fork1

crates/sovereign-sdk/module-system/sov-modules-api/src/transaction.rs
kpp: @eyusufatik ^

eyusufatik: depends on how many times we invoke the hash func.

if only once, we don't need it in the struct. else, it's better to have some form of calculate and then reuse the calculated value.

eyusufatik (approved):

PR #1436, opened 2024-11-05
Activate Dencun/Cancun upgrade

eyusufatik:
# Description
Adds dencun support to our evm. however, we won't let users submit eip4844 transactions at the mempool level, since there is no DA guarantees of blobs on a centralized sequencer. However, we want to keep compatability so all blob stuff are implemented. The main benefit we get is new opcodes and `seldestruct` change

TODOs:

- [x] Block blob transaction from the rpc
- [x] ~Pass on blob transactions inside evm~ instead we'll trust the sequencer to not include eip4844 txs.
- [ ] Benchmark kzg precompile and decide on enabling it
- [x] ~Make sure blob gas fee settings = 0 does not break anything~ Instead we'll correctly calculate eip4844 fee rates. Since we don't accept any eip4844 txs, this won't have any effect on anything.
- [ ] Activate ef tests for cancun -- though they might be problematic, instead test opcodes individually 
- [x] Write a test where we activate Fork1 hence cancun
- [ ] Update `selfdestruct` in `db_commit.rs`

## Linked Issues
Closes #1299

PR #1459, opened 2024-11-11
Offchain contract test

### LINKED ISSUE #1458: Write a test for offchain storage
# TODO

- [x] Deploy contract prior to a fork
- [x] Activate fork
- [x] First `get_code` RPC returns from state storage
- [x] Second `get_code` RPC returns from offchain
- [x] Deploy a contract post-fork
- [x] Execute a transaction on deployed contract

rakanalh:
# Description

Adds a test to make sure that a contracts code is correct and that the contract is executing transactions properly pre and post forks.

## Linked Issues
- Fixes #1458 
- Needs #1436


bin/citrea/tests/e2e/mod.rs
```diff
@@ -574,3 +582,139 @@ async fn execute_blocks(
...
+        .await
+        .unwrap();
+    assert_eq!(code.to_vec()[..runtime_code.len()], runtime_code.to_vec());
+
+    let code = sequencer_client
+        .eth_get_code(contract_address, None)
+        .await
+        .unwrap();
+    assert_eq!(code.to_vec()[..runtime_code.len()], runtime_code.to_vec());
+
```
eyusufatik: i think you meant to do the second assert after you do a transaction after the fork

kpp (approved):

bin/citrea/tests/e2e/mod.rs
```diff
@@ -574,3 +574,129 @@ async fn execute_blocks(
...
+            .await
+            .unwrap()
+            .contract_address
+            .unwrap();
+
+        (contract_address, contract)
+    };
+
+    {
+        let set_value_req = sequencer_client
```
eyusufatik, line 695: we should get the value and see if the tx actually worked. if the code couldn't be reached for example, the tx would go through but value wont be set

PR #1471, opened 2024-11-13
Implement EIP-6780: SELFDESTRUCT only in same transaction

### LINKED ISSUE #1461: Update `selfdestruct` according to Cancun upgrade
selfdestruct behaviour changed in the dencun / cancun upgrade. we also have to update our DbCommit implementation to reflect the changes. keep in mind we still have to support activation prior to cancun.

see EIP describing the change https://eips.ethereum.org/EIPS/eip-6780

keep in mind, the same tx code deletion is probably handled by revm itself. but pls double check.

kpp:
# Description

Mostly fixing issues prior to cancun.

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/1461

### DIFF
--- a/crates/evm/src/evm/handler.rs
+++ b/crates/evm/src/evm/handler.rs
@@ -485,7 +485,29 @@ fn calc_diff_size<EXT, SPEC: Spec, DB: Database>(
                 // So we need to only force the nonce change.
                 account.account_info_changed = true;
             }
-            JournalEntry::AccountDestroyed { address, .. } => {
+            JournalEntry::AccountDestroyed {
+                address,
+                target,
+                was_destroyed,
+                had_balance,
+            } => {
+                // This event is produced only if acc.is_created() || !is_cancun_enabled
+                // State is not changed:
+                // * if we are after Cancun upgrade and
+                // * Selfdestruct account that is created in the same transaction and
+                // * Specify the target is same as selfdestructed account. The balance stays unchanged.
+
+                if *was_destroyed {
+                    // It was already destroyed before in the log, no need to do anything.
+                    continue;
+                }
+
+                if address != target && !had_balance.is_zero() {
+                    // mark changes to the target account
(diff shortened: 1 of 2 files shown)


crates/evm/src/evm/handler.rs
```diff
@@ -485,7 +485,29 @@ fn calc_diff_size<EXT, SPEC: Spec, DB: Database>(
...
-            JournalEntry::AccountDestroyed { address, .. } => {
+            JournalEntry::AccountDestroyed {
+                address,
+                target,
+                was_destroyed,
+                had_balance,
+            } => {
+                // This event is produced only if acc.is_created() || !is_cancun_enabled
+                // State is not changed:
+                // * if we are after Cancun upgrade and
```
rakanalh, line 496:
```suggestion
                // * if  Cancun fork is activated and
```

kpp: https://github.com/bluealloy/revm/blob/bbc8d81dbe2a6a4d184e32fa540e1c4e248a65c9/crates/revm/src/journaled_state.rs#L530

rakanalh: Pasting the text as-is doesn't mean that the wording is correct.

okkothejawa:
> Note that if the target is the same as the contract calling SELFDESTRUCT there is no net change in balances. Unlike the prior specification, Ether will not be burnt in this case.

Should this be handled by this PR or does Revm already handle it?

kpp:
> does Revm already handle it?

it does all the magic. Actually it doesnt generate any log in this case so for an observer nothing changes -> no state diff

crates/evm/src/evm/handler.rs
eyusufatik: resolving this as both wording work

eyusufatik (approved):

PR #1472, opened 2024-11-14
Don't clone last journal Vec in calc_diff_size

kpp:
# Description

To reduce mem consumption.

### DIFF
--- a/crates/evm/src/evm/handler.rs
+++ b/crates/evm/src/evm/handler.rs
@@ -439,7 +439,7 @@ fn calc_diff_size<EXT, SPEC: Spec, DB: Database>(
     } = &mut context.evm.inner;
 
     // Get the last journal entry to calculate diff.
-    let journal = journaled_state.journal.last().cloned().unwrap_or(vec![]);
+    let journal = journaled_state.journal.last().into_iter().flatten();
     let state = &journaled_state.state;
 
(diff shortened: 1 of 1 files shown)


crates/evm/src/evm/handler.rs
```diff
@@ -457,7 +457,7 @@ fn calc_diff_size<EXT, SPEC: Spec, DB: Database>(
     let from = account_changes.entry(&env.tx.caller).or_default();
     from.account_info_changed = true;
 
-    for entry in &journal {
+    for entry in journal {
```
jfldde, line 460: What about using `if let Some(journal) = journaled_state.journal.last() {` instead ?

kpp: That will work too but it would require 1 extra tab and it will increase cyclomatic complexity

jfldde: It's the same cyclomatic complexity since we won't go into the code path when journal is empty though, in an implicit way

jfldde: Oh yeah the last journal can be empty itself, my bad

jfldde (approved):

rakanalh (approved):

eyusufatik (approved):
good catch

PR #1473, opened 2024-11-14
FORKS are always sorted, use binary_search to find current fork

### LINKED ISSUE #1460: ForkManager: Trust input is sorted and use binary search
in the `ForkManager::new()` we sort the described forks by height.

assume they are always passed sorted so remove the sorting.

In the same function, use binary search to find which fork we are on.

kpp:
# Description

- Check if FORKS are sorted at compile time
- Use binary search in fork_from_block_number
- Don't cast FORKS to vec
- Don't pass forks as arg to .apply_soft_confirmations_from_sequencer_commitments

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/1460

### DIFF
--- a/crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
+++ b/crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
@@ -2,6 +2,7 @@
 #![doc = include_str!("../README.md")]
 
 use borsh::BorshDeserialize;
+use citrea_primitives::forks::FORKS;
 use itertools::Itertools;
 use rs_merkle::algorithms::Sha256;
 use rs_merkle::MerkleTree;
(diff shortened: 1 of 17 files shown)


crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
```diff
@@ -625,7 +625,7 @@ where
         let mut previous_batch_hash = soft_confirmations[0][0].prev_hash();
         let mut last_commitment_end_height: Option<u64> = None;
 
-        let mut fork_manager = ForkManager::new(forks, sequencer_commitments_range.0 as u64);
+        let mut fork_manager = ForkManager::new(FORKS, sequencer_commitments_range.0 as u64);
```
rakanalh, line 628: You did this now, but at some point i know i will undo your version because after discussing this with @eyusufatik  we want to be more flexible by passing the forks as some tests might want to use a different set of forks compared to the production ones. But for us to do this, we have to have a planned fork because passing `forks` into the verifier would require a fork (i know it's confusing)

rakanalh (approved):
Great PR. :+1:

crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
kpp: Only this one? I can fix it, not a big deal.

rakanalh: No it's actually everywhere FORKS is used.

eyusufatik (approved):

PR #1481, opened 2024-11-18
Optimize DA verify_transactions

yaziciahmet:
# Description

Optimizes verify_transactions method of DA verifier.

Previously:
1. Iterate over completeness proof (Verifying each tx in completeness also exist in inclusion)
  1.1. Validate wtxid
  1.2. Ensure corresponding tx exists in inclusion proof
  1.3. Collect completeness proof wtxids into a BTreeSet
2. Iterate over inclusion proof (Verifying each relevant tx in inclusion also exist in completeness)
  2.1 If prefix matches, verify it also exists in collected completeness proof wtxids BTreeSet and remove it
3. If completeness proof wtxids BTreeSet not empty, there is extra tx in completeness, hence invalid

New:
1. Filter out nonrelevant transactions from inclusion proof
2. ZipEq filtered out inclusion and completeness, they must be of same length
  1.1 Verify wtxids match

With the new version, collecting and later removing wtxids into an BTreeSet is removed, and there is now only 1 iteration over proofs.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/bitcoin-da/src/verifier.rs
+++ b/crates/bitcoin-da/src/verifier.rs
@@ -90,8 +90,7 @@ impl DaVerifier for BitcoinVerifier {
             .wtxids
             .iter()
             .filter(|wtxid| wtxid.starts_with(prefix));
-        let completeness_iter = completeness_proof.iter();
-        for (wtxid, tx) in relevant_wtxid_iter.zip_eq(completeness_iter) {
+for (wtxid, tx) in relevant_wtxid_iter.zip_eq(&completeness_proof) {
             // ensure completeness proof tx matches the inclusion tx
             if tx.compute_wtxid().as_byte_array() != wtxid {
                 return Err(ValidationError::RelevantTxNotInProof);
--- a/crates/bitcoin-da/src/verifier.rs
+++ b/crates/bitcoin-da/src/verifier.rs
@@ -1044,10 +1044,33 @@ mod tests {
             to_light_client_prefix: vec![2, 2],
         });
 
-        let (block_header, inclusion_proof, mut completeness_proof, mut txs) = get_mock_data();
+        let (block_header, inclusion_proof, mut completeness_proof, txs) = get_mock_data();
 
         completeness_proof.swap(2, 3);
-        txs.swap(2, 3);
+
+        assert_eq!(
+            verifier.verify_transactions(
+                &block_header,
+                txs.as_slice(),
+                inclusion_proof,
+                completeness_proof,
+                DaNamespace::ToBatchProver,
+            ),
+            Err(ValidationError::RelevantTxNotInProof)
+        );
+    }
+
+    #[test]
+    fn break_rel_tx_order() {
+        let verifier = BitcoinVerifier::new(RollupParams {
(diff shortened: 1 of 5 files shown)


crates/bitcoin-da/src/verifier.rs
```diff
@@ -1116,30 +1098,6 @@ mod tests {
         );
     }
 
-    #[test]
```
eyusufatik, line 1119: why remove test

yaziciahmet: if you check closely, you will see that it is the exact same test with: break_completeness_proof_order test

yaziciahmet:
I did some benchmarks locally, and on average optimized version consumes 0.5% less cycles. This test was only with 1 sequencer commitment, hence, I can safely say that more commitments will result in a greater cycle difference.

crates/bitcoin-da/src/verifier.rs
```diff
@@ -82,29 +80,26 @@ impl DaVerifier for BitcoinVerifier {
...
 
-            // make sure completeness txs are ordered same in inclusion proof
-            // this logic always start seaching from the last found index
-            // ordering should be preserved naturally
-            let is_found_in_block =
-                inclusion_iter.any(|wtxid_inc| wtxid_inc == wtxid.as_byte_array());
-            if !is_found_in_block {
-                return Err(ValidationError::RelevantTxNotFoundInBlock);
+            // completeness proof tx for the relevant tx must exist
+            let Some(tx) = completeness_iter.next() else {
```
kpp: https://docs.rs/itertools/0.8.0/itertools/trait.Itertools.html#method.zip_longest

yaziciahmet: with zip_longest we would have to check if the values are some, imo panicing with zip_eq is better. As there is no recovery we can perform if we gave the wrong input.

crates/bitcoin-da/src/verifier.rs
eyusufatik: let's remove `txs.swap(2, 3);` from break_completeness_proof_order and restore this test then

crates/bitcoin-da/src/verifier.rs
```diff
@@ -82,29 +81,20 @@ impl DaVerifier for BitcoinVerifier {
...
-            let is_found_in_block =
-                inclusion_iter.any(|wtxid_inc| wtxid_inc == wtxid.as_byte_array());
-            if !is_found_in_block {
-                return Err(ValidationError::RelevantTxNotFoundInBlock);
+        let relevant_wtxid_iter = inclusion_proof
+            .wtxids
+            .iter()
+            .filter(|wtxid| wtxid.starts_with(prefix));
+        let completeness_iter = completeness_proof.iter();
+        for (wtxid, tx) in relevant_wtxid_iter.zip_eq(completeness_iter) {
```
jfldde:
```suggestion
for (wtxid, tx) in relevant_wtxid_iter.zip_eq(&completeness_proof) {
```

jfldde (approved):

PR #1490, opened 2024-11-19
Reorganize citrea rollup files

rakanalh:
# Description

bin/citrea and internal provers are nested more levels that i care for. IMO, the currently suggested reorg of files could improve things a bit.


eyusufatik:
I like bin/citrea/provers -> guests/

however, I'd like to keep bin/citrea naming. Soon we'll have bin/citrea-l1-syncer for example. We might have different binaries.

What do you think?

eyusufatik (approved):

PR #1495, opened 2024-11-20
Embedded elf guests for forks

### LINKED ISSUE #1476: New compilation and read ELF from files system
checkout esad/new-compilation-system branch

The goal we have with this issue is to read ELfs from files instead of by building and putting it as a constant in the binary. With this change, our provers will be able to prove with any ELF at any point in time. (If we don't do this forking becomes very time dependent for deployment). However, for test builds, we would like to use the newly compiled ELF, this is already implemented in the branch. So for non-testnet builds, we'd like to not see risc0 guests get compiled.

Ideally, we would like a new directory in project root, `elfs/`, which would have two subdirectories `risc0/` and `sp1`. Each will have subdirectories `testnet/` `mainnet/` and the ELFs will be inside these subdirectories. The naming convention can be 0.elf 1.elf ... for different forks. However, it's ok to implement them differently if implementor wishes.

Now, notice that for the production builds, we now don't depend on `citrea-risc0`, the crate that holds our ZK guests, and we can use the same binary on both testnet and mainnet. (This would become a problem, let's say, if we put genesis state root of the rollup as a constant in the light client proof guest. We would need to build two guests, hence two binaries. But we can do 2 ELFs and one binary, which we point at which elfs to load. see #1444 ) 

A new CLI arg --testnet is needed to indicate that the node is going to be run on Citrea testnet. (absence of --testnet indicates mainnet run.). This would load ELFs from `elfs/../testnet`.

Now that we get our provers to load multiple ELFs, we'd also need to change prover service trait so that the prover runner can signal which ELF will be used by proving. To actually decide on which ELF to use for proving see below:

- the batch prover will look at the L2 range it is about to prove, if at any point fork x is activated, it will use the elf for fork x.
- the light client prover will always use the latest ELF, we can change this behaviour later. I am not sure at this moment.

The full node behaviour of loading up method id's as constants are now also gone, the full nodes will load ELF's at the corresponding directory and calculate method id's on their own. They can discard the ELF's afterwards.

In the PR, please also go back to our latest release, do a reproducible build of the batch proof guest, and put it under testnet's first elf.

rakanalh:
# Description
`make build` now builds the ELF guest codes separately using `make build-risc0` and `make build-sp1`.

This basically removes the fact that we build the guest code along with the binary. Now, the guest code is loaded from `resources/guests` based on what we want to load.

That being said, each fork should add the ELF to the designated directory with `1.elf`, `2.elf` .. etc for each fork we end up introducing. The names can change of course but that's how it goes for now.

# Development workflow

Previously, we used to have the guest code already built within the binary. If you run:
- `make build` 
- `make test`

Both will act the same way by building the guest code and provide a binary that has the guest codes needed.

If you are running testing a specific build for the guest code, you can use:
- `make -C guests/risc0 batch-proof-bitcoin`
- `make -C guests/risc0 batch-proof-mock`
- `make -C guests/risc0 light-client-bitcoin`
- `make -C guests/risc0 light-client-mock`

# TODO
- [x] Prevent `build.rs` for risc0 from running.
- [x] Every time a fork activates, we should upload the newly activated guest code to Bonsai
- [x] Override `network` if `--testnet` is used.
- [x] Pass `Network` argument

## Linked Issues
- Fixes #1476


Makefile
```diff
@@ -9,8 +9,17 @@ PARALLEL_PROOF_LIMIT := 1
 help: ## Display this help message
 	@awk 'BEGIN {FS = ":.*?## "} /^[a-zA-Z_-]+:.*?## / {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST)
 
+.PHONY: build-risc0
+build-risc0:
+	cargo risczero build --manifest-path guests/risc0/batch-proof-bitcoin/Cargo.toml
+	cargo risczero build --manifest-path guests/risc0/light-client-proof-bitcoin/Cargo.toml
```
rakanalh: Seems like the build command has `--target-dir` to specify where the elfs should go... but i couldn't find it when running `cargo risczero build` locally.
https://github.com/risc0/risc0/blob/v1.1.3/risc0/cargo-risczero/src/commands/build.rs#L56

Seems this is unreleased yet?

bin/citrea/src/main.rs
```diff
@@ -23,6 +23,16 @@ mod test_rpc;
...
 struct Args {
+    /// The mode in which the node runs.
+    /// This determines which guest code to use.
+    /// Default is Mainnet.
+    #[clap(short, long, default_value_t, value_enum)]
+    network: Network,
+
+    /// Overrides the run mode to use testnet
+    #[arg(long)]
+    testnet: bool,
```
rakanalh, line 34: TODO: Pass this around

Makefile
eyusufatik: are you sure you're on 1.1.3?

rakanalh:
```
cargo risczero --version
cargo-risczero 1.1.3
```

bin/citrea/src/utils.rs
```diff
@@ -0,0 +1,50 @@
...
+        .get_last_commitment_l2_height()
+        .ok()
+        .flatten()
+        .expect("Should be able to fetch last l2 height");
+    let fork = fork_from_block_number(FORKS, last_l2_height.into());
+    let guest = guests
+        .get(&fork.spec_id)
+        .cloned()
+        .expect("A fork should have a guest code attached");
+    guest
```
eyusufatik: didn't read so carefully but just a heads up

the batch prover will choose which elf to use with this algo:

1- get sequencer commitments last l2 height.
2- get spec at that height
3- use corresponding elf

the light client prover will need use this algo:

1- get last l2 height in batch proof output
2- get spec at that height
3- use the elf for that spec
4- set the batch proof method id input from the batch proof elf at that spec 

(this is a bit due to change, we might change how we handle the light client case, even inputs might change)

eyusufatik:
> Every time a fork activates, we should upload the newly activated guest code to Bonsai

btw just pass down elf from runner > prover service > zkvm host

I think bonsai doesnt upload all the time since it checks if elf exists on remote

crates/risc0/src/host.rs
```diff
@@ -44,7 +43,7 @@ pub struct Risc0BonsaiHost {
 
 impl Risc0BonsaiHost {
     /// Create a new Risc0Host to prove the given binary.
-    pub fn new(image_id: Digest, elf: Vec<u8>, ledger_db: LedgerDB) -> Self {
+    pub fn new(image_id: Digest, ledger_db: LedgerDB) -> Self {
```
eyusufatik: image_id is irrelevant at this point if we are passing elfs on `run`

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -277,32 +289,27 @@ where
...
-
-        da_data.iter_mut().for_each(|tx| {
-            // Check for commitment
-            if tx.sender().as_ref() == self.batch_prover_da_pub_key.as_slice() {
-                let data = DaDataLightClient::try_from_slice(tx.full_data());
-
-                if let Ok(proof) = data {
-                    batch_proofs.push(proof);
-                } else {
-                    tracing::warn!(
```
eyusufatik: can you put this log back with debug! ?

crates/batch-prover/src/proving.rs
```diff
@@ -201,8 +202,18 @@ where
...
+    let last_l2_height = sequencer_commitments
+        .last()
+        .expect("Should have at least 1 commitment")
+        .l2_end_block_number;
+    let current_spec = fork_from_block_number(FORKS, last_l2_height).spec_id;
+    let elf = elfs_by_spec
+        .get(&current_spec)
+        .expect("Every fork should have an elf attached")
+        .clone();
+
```
eyusufatik, line 234: the grouping that'll be applied to these sequencer commitments are important.

let's say there is 3 items i n`sequencer_commmitments` vec, and only the last one has the height which the next fork activates.

down in execution, if these guys are grouped like [(1,2), (3)] than we can prove the first group  elf n-1 and the last group with elf 3. 

or else there will be confusion when we need to pull batch proof method ids based on `output.last_l2_height` in light client circuit

eyusufatik (approved):
looks good. I'll need another round of reviewing since I've only reviewed since my last review this time. needs a small change though

eyusufatik (changes requested):
I meant to request changes

crates/risc0/src/host.rs
rakanalh:
```
        receipt.verify(self.image_id)?;
```
In host, we use the image ID to verify receipt apparently.

eyusufatik: yes but image id can be calculated from the elf passed wherever needed.

bin/citrea/Cargo.toml
```diff
@@ -91,15 +92,14 @@ rustc_version_runtime = { workspace = true }
 # bitcoin-e2e dependencies
 bitcoin.workspace = true
 bitcoincore-rpc.workspace = true
-citrea-e2e = { git = "https://github.com/chainwayxyz/citrea-e2e", rev = "5baaef3" }
+citrea-e2e = { git = "https://github.com/chainwayxyz/citrea-e2e", branch = "rakanalh/use-dev-mode" }
```
rakanalh: Place correct revision after companion PR is merged.

crates/batch-prover/src/proving.rs
rakanalh: I have implemented a change in `break_sequencer_commitments` to not only account for state diff threshold, but also forks. Commitments will be grouped by fork.

rakanalh: One thing to note, is that with:
`GroupCommitment::SingleShot`: we don't account for forks
`GroupCommitment::OneByOne`: Each commitment is split by default so no need to account for forks
`break_sequencer_commitment`: Accounts for forks.

bin/citrea/src/rollup/bitcoin.rs
```diff
@@ -225,10 +281,7 @@ impl RollupBlueprint for BitcoinRollup {
         da_service: &Arc<Self::DaService>,
         ledger_db: LedgerDB,
     ) -> Self::ProverService {
-        let vm = Risc0BonsaiHost::new(
-            citrea_risc0::LIGHT_CLIENT_PROOF_BITCOIN_ELF,
-            ledger_db.clone(),
-        );
+        let vm = Risc0BonsaiHost::new(ledger_db.clone());
         let zk_stf = StfBlueprint::new();
```
eyusufatik: (can be a new issue) 

since the zkvm adapters now don't need elfs on ::new()

we can have the same create_prover_service for light client prover and batch prover, where the da verifier is an input

crates/batch-prover/src/da_block_handler.rs
```diff
@@ -456,14 +469,15 @@ pub(crate) fn break_sequencer_commitments_into_groups<DB: BatchProverLedgerOps>(
         // Threshold is checked by comparing compressed state diff size as the data will be compressed before it is written on DA
         let state_diff_threshold_reached = compressed_state_diff.len() > MAX_TXBODY_SIZE;
 
-        if state_diff_threshold_reached {
-            // We've exceeded the limit with the current commitments
-            // so we have to stop at the previous one.
-            result_range.push(range);
+        let commitment_spec =
+            fork_from_block_number(FORKS, sequencer_commitment.l2_start_block_number).spec_id;
```
eyusufatik: check l2_end_block_number here

crates/batch-prover/src/proving.rs
```diff
@@ -221,8 +222,18 @@ where
...
     // Prove all proofs in parallel
-    let proofs = prover_service.prove().await?;
+    let proofs = prover_service.prove(elf).await?;
```
eyusufatik, line 236: this is good, however, we'll need to generate V1 input if we prove for older elfs on testnet or mainnet

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -153,7 +159,7 @@ where
         // Do any kind of ordering etc. on batch proofs here
         // If you do so, don't forget to do the same inside zk
         let batch_proof_method_id = self
-            .batch_proof_code_commitments_by_spec
+            .batch_proof_code_commitments
             .get(&SpecId::Genesis)
             .expect("Batch proof code commitment not found");
```
eyusufatik: we'll need to get method id by batch_proof_output.last_l2_height's spec id

crates/prover-services/src/parallel/mod.rs
```diff
@@ -113,7 +113,7 @@ where
         )
     }
 
-    async fn prove_all(&self, proof_queue: Vec<ProofData>) -> Vec<Proof> {
+    async fn prove_all(&self, elf: Vec<u8>, proof_queue: Vec<ProofData>) -> Vec<Proof> {
```
eyusufatik, line 116: what if different proofs in the queue require different elfs

crates/batch-prover/src/da_block_handler.rs
eyusufatik: I meant get with l2_end_block_number

crates/prover-services/src/parallel/mod.rs
rakanalh: The batch prover should break up commitments and therefore we'll have different inputs in which the inputs are related to non-forking commitments.

bin/citrea/src/rollup/bitcoin.rs
rakanalh: Done

crates/light-client-prover/src/da_block_handler.rs
rakanalh: Done

crates/batch-prover/src/proving.rs
rakanalh: Will be done in: https://github.com/chainwayxyz/citrea/issues/1538

resources/guests/risc0/testnet/batch-0.elf
eyusufatik, line 1: please make sure image id is the same with our old image id

.github/workflows/checks.yml
```diff
@@ -419,6 +423,7 @@ jobs:
           CITREA_E2E_TEST_BINARY: ${{ github.workspace }}/target/debug/citrea
           PARALLEL_PROOF_LIMIT: 1
           TEST_OUT_DIR: ${{ runner.temp }}/test
+          TAIL_N_LINES: 500
```
eyusufatik: might want to remove this

eyusufatik (approved):

resources/guests/risc0/testnet/batch-0.elf
rakanalh: Built the code on 0.5.4 release using:
```
REPR_GUEST_BUILD=1 make build
```

Calculated the image ID for bitcoin prover:
```
> cp target/riscv-guest/riscv32im-risc0-zkvm-elf/docker/citrea_bitcoin_prover/bitcoin_da resources/guests/risc0/testnet/batch-0.elf

> r0vm --elf resources/guests/risc0/testnet/batch-0.elf --id
3631d90630a3f0deb47f3a3411fe6e7ede1b0d86ad4216c75041e1a2020f009f
```
Checked against calculated value in `guests.rs`:

```rust
    println!(
        "{:?}",
        citrea::guests::BATCH_PROOF_TESTNET_GUESTS
            .get(&SpecId::Genesis)
            .unwrap()
            .0
    );
```

Result: 
```
Digest(3631d90630a3f0deb47f3a3411fe6e7ede1b0d86ad4216c75041e1a2020f009f)
```

Makefile
```diff
@@ -9,11 +9,19 @@ PARALLEL_PROOF_LIMIT := 1
...
+build-risc0:
+	$(MAKE) -j 4 -C guests/risc0 all
+
+.PHONY: build-sp1
+build-sp1:
+	$(MAKE) -C guests/sp1 all
+
 .PHONY: build
-build: ## Build the project
+build: build-risc0 build-sp1 ## Build the project
```
yaziciahmet, line 21: I think we should skip building sp1 for now.

bin/citrea/src/main.rs
```diff
@@ -23,6 +25,20 @@ mod test_rpc;
...
+    #[clap(short, long, default_value_t, value_enum)]
+    network: NetworkArg,
+
+    /// Override network to run testnet directly.
+    #[arg(long)]
+    testnet: bool,
+
+    /// Run the development chain
+    #[arg(long, conflicts_with_all = ["testnet"])]
+    dev: bool,
```
yaziciahmet, line 40: I don't see dev being used anywhere

bin/citrea/src/rollup/bitcoin.rs
```diff
@@ -175,15 +160,94 @@ impl RollupBlueprint for BitcoinRollup {
...
+            Network::Nightly => LIGHT_CLIENT_LATEST_BITCOIN_GUESTS
+                .iter()
+                .map(|(k, (id, _))| (*k, *id))
+                .collect(),
+        }
+    }
+
     #[instrument(level = "trace", skip_all)]
-    async fn create_batch_prover_service(
+    async fn create_prover_service(
```
yaziciahmet, line 243: nice!

yaziciahmet (commented):
LGTM! not blocking the pr, just have 1 comment.

bin/citrea/src/main.rs
rakanalh: https://github.com/chainwayxyz/citrea/pull/1495/files/36bf252ef863ee0e0a07cf9251a83266b7ddcabc#diff-c1783a446ab178e59e9ad736b1364049f88a3efac218f37793c2ca081b5489ffR153

yaziciahmet: ok im blind, ty :D

yaziciahmet (approved):

eyusufatik (approved):

PR #1491, opened 2024-11-19
Light client proof bitcoin e2e test

ercecan:
# Description
Tests light client proof with bitcoin e2e framework

## Linked Issues
- Fixes #1466

PR #1502, opened 2024-11-21
DA Verifier light client proof tests

### LINKED ISSUE #1020: Write BitcoinDA Verifier tests against light client namespace

yaziciahmet:
# Description

Added light client proof namespace tests to DAVerifier. I only added tests that are exclusive to light client proofs. For instance, I didn't add a new `missing_tx_in_completeness_proof_lcp` test, as the purpose of this test is ensure the inclusion and completeness proof txs match, irrelevant of their content.

## Linked Issues
- Fixes #1020 

## Testing

## Docs

### DIFF
--- a/crates/bitcoin-da/src/helpers/test_utils.rs
+++ b/crates/bitcoin-da/src/helpers/test_utils.rs
@@ -4,11 +4,12 @@ use bitcoin::block::{Header, Version};
 use bitcoin::hash_types::{TxMerkleNode, WitnessMerkleNode};
 use bitcoin::hashes::Hash;
 use bitcoin::{BlockHash, CompactTarget, Transaction};
+use citrea_primitives::compression::decompress_blob;
 use sov_rollup_interface::da::{DaSpec, DaVerifier};
 
+use super::merkle_tree;
 use super::parsers::{parse_batch_proof_transaction, parse_light_client_transaction, ParserError};
-use super::{calculate_sha256, merkle_tree};
-use crate::helpers::parsers::parse_hex_transaction;
+use crate::helpers::parsers::{parse_hex_transaction, VerifyParsed};
 use crate::spec::blob::BlobWithSender;
 use crate::spec::header::HeaderWrapper;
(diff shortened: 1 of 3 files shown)


crates/bitcoin-da/src/service.rs
eyusufatik, line 1: can you add extract_relevant test for light client namespace?

I found a bug there today :) 

solved it and committed to #1502

crates/bitcoin-da/src/helpers/test_utils.rs
```diff
@@ -23,26 +30,50 @@ pub(crate) fn get_mock_txs() -> Vec<Transaction> {
...
+                    let blob = decompress_blob(&complete.body);
+                    (blob, complete.public_key, hash)
+                }
+                super::parsers::ParsedLightClientTransaction::Aggregate(aggregate) => {
+                    let hash = aggregate
+                        .get_sig_verified_hash()
+                        .expect("Invalid sighash on aggregate zk proof");
+                    (aggregate.body, aggregate.public_key, hash)
+                }
+                _ => unimplemented!(),
```
eyusufatik, line 65: why unimplemented? we can get chunk in a block

eyusufatik (changes requested):

crates/bitcoin-da/src/service.rs
eyusufatik: I meant #1491

crates/bitcoin-da/src/helpers/test_utils.rs
yaziciahmet: we should ignore chunked transactions in here as this function is used for retrieving relevant prefixed transactions, and requesting chunks is later going to be performed on da service.

this also means that we can't write a unit test currently for extract_relevant_zk_proof as it requires local running regtest bitcoin. I left it as ignored for now to have a reference when we carry the regtest needing tests to bitcoin-e2e.

eyusufatik (approved):
aproving on accounts that a follow up of #1503 pr will be implemented

PR #1515, opened 2024-11-25
Carry bitcoin-da tests to e2e

### LINKED ISSUE #1503: Move bitcoin service and verifier tests to bitcoin-e2e
so we don't have to work on preconstructed .txt transactions.

yaziciahmet:
# Description

Carry bitcoin-da tests to e2e test framework to avoid needing to maintain mock tx files existing under test_data directory, and being able to properly test all methods.

## Linked Issues
- Fixes #1503 

## Testing

## Docs


crates/bitcoin-da/src/helpers/parsers.rs
```diff
@@ -469,7 +469,7 @@ mod batch_proof {
     }
 }
 
-#[cfg(all(test, feature = "native"))]
+#[cfg(feature = "native")]
```
eyusufatik, line 472: this is only used in tests are you sure you want to expose this to non-test code?

crates/bitcoin-da/src/verifier.rs
```diff
@@ -78,6 +79,10 @@ impl DaVerifier for BitcoinVerifier {
         completeness_proof: <Self::Spec as DaSpec>::CompletenessProof,
         namespace: DaNamespace,
     ) -> Result<(), Self::Error> {
+        if block_header.tx_count as usize != inclusion_proof.wtxids.len() {
+            return Err(ValidationError::HeaderInclusionTxCountMismatch);
+        }
```
eyusufatik, line 84: this comparisoon is ok to have but beware that the tx_count field is not something actually committed to it's arbitrarily set when creating the struct.

crates/bitcoin-da/tests/verifier.rs
```diff
@@ -0,0 +1,593 @@
...
+
+struct BitcoinVerifierTest;
+
+#[async_trait]
+impl TestCase for BitcoinVerifierTest {
+    fn test_config() -> TestCaseConfig {
+        // Only run bitcoin regtest
+        TestCaseConfig {
+            with_sequencer: false,
+            ..Default::default()
```
eyusufatik, line 34: then shouldn't you do with_fullnode: false etc.?

crates/bitcoin-da/tests/service.rs
```diff
@@ -0,0 +1,208 @@
...
+
+        // Extracts relevant batch proof blobs with proof correctly
+        {
+            let (txs, inclusion_proof, completeness_proof) =
+                service.extract_relevant_blobs_with_proof(&block, DaNamespace::ToBatchProver);
+            assert_eq!(inclusion_proof.wtxids.len(), 29);
+            assert_eq!(inclusion_proof.wtxids[1..], block_wtxids[1..]);
+            // 3 valid commitments, and 1 invalid commitment with wrong public key
+            assert_eq!(txs.len(), 4);
+            // it is >= due to the probability that one of commit transactions ended up
```
eyusufatik, line 66: good thinking :)

crates/bitcoin-da/tests/verifier.rs
```diff
@@ -0,0 +1,593 @@
...
+
+            assert!(verifier
+                .verify_transactions(
+                    &block.header,
+                    &b_txs,
+                    inclusion_proof,
+                    b_completeness_proof.clone(),
+                    DaNamespace::ToBatchProver,
+                )
+                .is_err());
```
eyusufatik: don't you want to assert error types?

crates/bitcoin-da/src/helpers/merkle_tree.rs
```diff
@@ -144,12 +143,6 @@ mod tests {
         compare_merkle_tree_against_bitcoin_impl(vec![[255; 32]; 33]);
         compare_merkle_tree_against_bitcoin_impl(vec![[200; 32]; 2]);
         compare_merkle_tree_against_bitcoin_impl(vec![[99; 32]; 1]);
-
-        let txs = get_mock_txs()
-            .iter()
-            .map(|tx| tx.compute_wtxid().to_byte_array())
-            .collect();
-        compare_merkle_tree_against_bitcoin_impl(txs);
```
eyusufatik, line 152: why did you remove this?

crates/bitcoin-da/tests/verifier.rs
yaziciahmet: the default is with_sequencer: true and all others false

crates/bitcoin-da/src/helpers/parsers.rs
yaziciahmet: since tests/test_helpers.rs is not considered to be a test itself, we are unable to use it (at least i couldn't find a nice way). hence, i just removed it. i can carry this method to tests-utils, or if you have any other idea i am definitely open to suggestions

crates/bitcoin-da/tests/verifier.rs
yaziciahmet: asserted errors: https://github.com/chainwayxyz/citrea/pull/1515/commits/341755599bebd53a7fa33afeb2d8e458f51a411a

crates/bitcoin-da/src/verifier.rs
yaziciahmet: merkle root mismatch will catch this type of error eventually, but it seemed useful to me to have an early return for this. if you think this would cause potential problems, its definitely ok to remove.

crates/bitcoin-da/src/helpers/merkle_tree.rs
yaziciahmet: initially removed because get_mock_txs method were no longer existing. added it back (https://github.com/chainwayxyz/citrea/pull/1515/commits/99154fc0d6fc0fa94f2d0df4c7bcb42a62753317) by directly reading the file in this test since this is the only remaining usage.

also didn't want to bother carrying this into e2e, as it seemed too localized to be residing in e2e.

eyusufatik (approved):

PR #1517, opened 2024-11-25
Convert dyn to generics in `verified_blob_content

yaziciahmet:
# Description

Replaces dyn with generics as its slightly more efficient. 

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


kpp:
It's questionable since &dyn generates less code and generics monomorphize code for every unique T.

I would really love to have a test for prover/verifier to calculate exact cycle count for a block with a fixed content inside. It will help us a lot in cases like this.

Anyway, I am not against this PR.

rakanalh:
> It's questionable since &dyn generates less code and generics monomorphize code for every unique T.
> 
> I would really love to have a test for prover/verifier to calculate exact cycle count for a block with a fixed content inside. It will help us a lot in cases like this.
> 
> Anyway, I am not against this PR.

This would firstly affect the compile time aspect of the code & binary size. Secondly, I think that using generics would allow the compiler to better optimize the code after monomorphization takes place.

eyusufatik (approved):
since this function will only actually take in bitcoin blob txs, we won't be generating more code with templates. only reason we went dyn or T is we are implementing a trait.

so I guess a little bit of better runtime performance is ok.

yaziciahmet:
I also optimized the verified_blob_content method in this pr:
1. removed clone, as the blob is already owned byte iterator, we can just return the reference
2. no need for advance, as full_data method is already called in native, and the verified data is populated in CountedBufReader.

Step 2 has something unnecessary that is happenning right now. current impl of BlobWithSender is that it has its blob as CountedBufReader, which holds 1 reader, and 1 vec which is being read into, and anytime advance (or full_data) is called, it reads from the reader into the vec. the thing is, we always use another vector as the reader into this CountedBufReader, meaning that we just pass 2x the same data unnecessarily into the zk, and we have no partial reads whatsoever. we can just replace CountedBufReader with vec. but this is relatively tricky to do since BlobWithSender is part of the trait type definition of the BitcoinSpec, and versioning is going to be very uncomfortable.

PR #1526, opened 2024-11-26
Prometheus telemetry

### LINKED ISSUE #1293: Profiling / metrics stuff for all nodes — enable in testnet

rakanalh:
# Description

From the referenced issue, the following was the list of metrics to collect:

### All
- [ ] Storage size ( I think this can be fed into prometheus outside the scope of citrea)
- [ ] RPC served in last 5 min (Same as this one, nginx?)
### Sequencer
- [x] Current number txs in mempool
- [ ] Mempool inbound tx/s (Can be calculated from the current number)
- [x] dry_run_transactions execution time
- [x] produce_block transactions execution time
- [ ] (this one will be hard to track so if we cant no problem) avg. wait time for tx in mempool (Can be calculated i believe)
- [x] Send sequencer commitment execution time
- [x] Sequencer commitment number of blocks
- [x] Current L1 block number used in block production
### Full node
- [ ] Sync progress (Not applicable since we have to know sequencer latest L2 vs fullnode or prover current L2)
- [x] Current L2 block number
- [x] Current L1 block number
- [ ] Soft confirmation pull rate (important while syncing) 
- [x] Soft confirmaiton process durations
- [x] L1 scan duration per L1 block
### Prover(s)
- [x] Current L2 block number
- [x] Current L1 blcok number
- [x] Cycle count for each proof
- [x] DA tx mining time

# TODO
- [ ] Implement / Export graphana dashboards to display these metrics.

## Linked Issues
- Fixes #1293


crates/batch-prover/src/telemetry.rs
```diff
@@ -0,0 +1,135 @@
...
+    ) -> Self {
+        Self {
+            config,
+            registry,
+            targets,
+        }
+    }
+}
+
+pub struct TelemetryTargets {
```
ercecan: Move above telemetry struct

eyusufatik:
> (Cycle count is NOT part of the output, cannot be reported)

in risc0 adapter, here you can get cycle counts from stats. change return types if you wish, but we need this information. probably the most important telemetry both provers can have. also why don't you have any telemetry implemented for light client prover 😅?



```rust
        let ProveInfo { receipt, stats } =
            prover.prove_with_opts(env, &elf, &ProverOpts::groth16())?;
```

crates/batch-prover/src/telemetry.rs
```diff
@@ -0,0 +1,135 @@
...
+    registry.register(
+        "current_l1_block",
+        "The current L1 block number which is used to produce L2 blocks",
+        current_l1_block.clone(),
+    );
+
+    registry.register(
+        "current_l2_block",
+        "The current L2 block number",
+        current_l1_block.clone(),
```
ercecan: current_l2_block

crates/sovereign-sdk/rollup-interface/src/node/telemetry.rs
```diff
@@ -0,0 +1,18 @@
+//! Provide the data structure for telemetry targets
```
ercecan, line 1: I believe these also can be in common- or it's own crate

crates/common/src/telemetry.rs
```diff
@@ -0,0 +1,75 @@
...
+/// Boxed HTTP body for responses
+type BoxBody = combinators::BoxBody<Bytes, hyper::Error>;
+
+pub async fn start_telemetry_server(
+    addr: SocketAddr,
+    registry: Arc<Registry>,
+    cancellation_token: CancellationToken,
+) -> anyhow::Result<()> {
+    let tcp_listener = TcpListener::bind(addr).await.unwrap();
+    let server = hyper::server::conn::http1::Builder::new();
```
ercecan: out of curiosity, why http1?

crates/common/src/telemetry.rs
```diff
@@ -0,0 +1,75 @@
+use std::net::SocketAddr;
```
ercecan: I suggest separate crate

crates/fullnode/src/telemetry.rs
```diff
@@ -0,0 +1,135 @@
...
+    registry.register(
+        "current_l1_block",
+        "The current L1 block number which is used to produce L2 blocks",
+        current_l1_block.clone(),
+    );
+
+    registry.register(
+        "current_l2_block",
+        "The current L2 block number",
+        current_l1_block.clone(),
```
ercecan: current_l2_block

crates/primitives/src/utils.rs
```diff
@@ -0,0 +1,8 @@
+use std::time::Duration;
+
+/// `duration_to_seconds` converts Duration to seconds.
+#[inline]
+pub fn duration_to_seconds(d: Duration) -> f64 {
+    let nanos = f64::from(d.subsec_nanos()) / 1e9;
+    d.as_secs() as f64 + nanos
+}
```
ercecan: You can use [`as_secs_f64`]

crates/sequencer/src/runner.rs
```diff
@@ -591,6 +625,10 @@ where
                 txs_to_remove.extend(l1_fee_failed_txs);
 
                 self.mempool.remove_transactions(txs_to_remove.clone());
+                self.telemetry
+                    .targets
+                    .mempool_txs
+                    .dec_by(txs_to_remove.len() as i64);
```
ercecan: I would suggest using pools len function to check the size of the pool, it might be discarding some txs on it's own as well after a while
You'll need to add that function to CitreaMemppol first as it does not exist there

crates/sovereign-sdk/full-node/db/sov-schema-db/src/iterator.rs
```diff
@@ -109,7 +108,11 @@ where
         let raw_value = self.db_iter.value().expect("db_iter.value() failed.");
         let value_size_bytes = raw_value.len();
         SCHEMADB_ITER_BYTES
-            .with_label_values(&[S::COLUMN_FAMILY_NAME])
+            .histogram
+            .get_or_create(&vec![(
+                "cf_name".to_owned(),
```
ercecan, line 113: why do we need cf_name here, asking because I did not understand

crates/sovereign-sdk/rollup-interface/src/node/telemetry.rs
ercecan: To not import anything in rollup-interface

ercecan (changes requested):
LGTM small fixes and couple of suggestions

crates/sovereign-sdk/rollup-interface/src/node/telemetry.rs
rakanalh: The types defined here are returned by the new method defined in `DaService`. This trait lives in rollup interface which cannot import primitives or common because both of the mentioned crates import rollup-interface. This is preventing cyclic dependencies. 

TBH, i was thinking that we merge rollup interface with our own primitives into a single crate.

crates/sovereign-sdk/full-node/db/sov-schema-db/src/iterator.rs
rakanalh: 1. This was porting telemetry already defined in sov SDK.
2. "cf_name" and "db_name" are labels provided for prometheus. I believe this is the open-metrics format or something like that.

crates/common/src/telemetry.rs
rakanalh: HTTP 1.1 i believe is what is common and widely supported. HTTP 2 is this (relatively) new protocol. Not sure if it's supported by the prometheus client that is going to pull data.

crates/common/src/telemetry.rs
rakanalh: Common is the crate which groups code that is shared between different node types. I don't think a separate crate is needed for this.

rakanalh:
> > (Cycle count is NOT part of the output, cannot be reported)
> 
> in risc0 adapter, here you can get cycle counts from stats. change return types if you wish, but we need this information. probably the most important telemetry both provers can have. also why don't you have any telemetry implemented for light client prover 😅?
> 
> ```rust
>         let ProveInfo { receipt, stats } =
>             prover.prove_with_opts(env, &elf, &ProverOpts::groth16())?;
> ```

Added... thanks for the hint.

crates/sovereign-sdk/rollup-interface/src/node/telemetry.rs
ercecan: > TBH, i was thinking that we merge rollup interface with our own primitives into a single crate.

That also would be nice

resources/configs/bitcoin-regtest/batch_prover_rollup_config.toml
yaziciahmet, line 1: can you make telemetry ports of sequencer, fullnode, and prover for mock and regtest different? it causes problems when running different nodes locally.

guests/risc0/light-client-proof-mock/Cargo.lock
```diff
@@ -716,6 +716,7 @@ dependencies = [
  "bincode",
  "borsh",
  "hex",
+ "prometheus-client",
```
yaziciahmet, line 719: it would be great if you can guard this dep behind native feature.

docker/docker-compose.telemetry.yaml
yaziciahmet, line 1: With this, do we have to spawn up a prometheus container for working locally?

crates/sovereign-sdk/rollup-interface/src/node/services/da.rs
```diff
@@ -127,6 +129,9 @@ pub trait DaService: Send + Sync + 'static {
         &self,
         sequencer_da_pub_key: &[u8],
     ) -> Vec<SequencerCommitment>;
+
+    /// Returns the telemetry targets type for the DA service.
+    fn telemetry_targets(&self) -> &DaTelemetryTargets;
```
yaziciahmet, line 134: This should be native only as well

crates/sovereign-sdk/full-node/db/sov-schema-db/src/lib.rs
```diff
@@ -416,6 +441,13 @@ fn default_write_options() -> rocksdb::WriteOptions {
     opts
 }
 
+/// `duration_to_seconds` converts Duration to seconds.
+#[inline]
+pub fn duration_to_seconds(d: Duration) -> f64 {
+    let nanos = f64::from(d.subsec_nanos()) / 1e9;
+    d.as_secs() as f64 + nanos
+}
```
yaziciahmet, line 449: IMO nanos is a bit too much precision. We can use micros and just do `d.as_micros() as u64` with no overflow no floating point whatsoever.

crates/sovereign-sdk/full-node/db/sov-schema-db/src/iterator.rs
```diff
@@ -120,6 +123,15 @@ where
             ScanDirection::Backward => self.db_iter.prev(),
         }
 
+        let v = Instant::now().saturating_duration_since(start);
+        SCHEMADB_ITER_LATENCY_SECONDS
+            .histogram
+            .get_or_create(&vec![(
+                "cf_name".to_owned(),
+                S::COLUMN_FAMILY_NAME.to_owned(),
+            )])
```
yaziciahmet, line 132: get_or_create is used everywhere in sov-schema-db, but we already know the dbs, and column families. can we somehow create these at the startup as well?

yaziciahmet (changes requested):
Good PR. Left some comments as in it's current state. I will re-review once PR is not draft anymore.

rakanalh:
Closing in favor of #1589 1589

PR #1528, opened 2024-11-27
L1 fee throttle

### LINKED ISSUE #1357: Sequencer DA limit throttling

jfldde:
# Description

- L1 fee throttle

Default throttle values :

```
            (0.0, 1.0),   // No usage
            (0.25, 1.0),  // Below threshold
            (0.5, 1.0),   // At threshold
            (0.6, 1.016), // Above threshold, start increasing fee
            (0.7, 1.256),
            (0.8, 2.296),
            (0.85, 3.40),
            (0.9, 4.0), // Max multiplier hit
            (0.95, 4.0),
            (1.0, 4.0),
```
No fee increase below 0.5 threshold, then slowly increase towards 0.7 with a 1.25~ multiplier then a sharp increase to the max multiplier of 4.0.

These values were arbitrarily chosen to slowly disincentivize tx in block between 0.5 and 0.7 L1 usage and then sharply increase.
This is configurable via sequencer config https://github.com/chainwayxyz/citrea/pull/1528/files#diff-7fd132bbf4762e829ab08aac6cfc52089b4674fac321c2faf1dc961f9fbc1237

## Linked Issues
- Fixes #1357


eyusufatik:
There are three different fees on Citrea:

- Base fee
- Priority fee
- L1 fee

Out of all three, L1 fee stands out. The reasons being: 

- Only fee rate which is not trustless -- decided by the sequencer.
- Only fee rate which is not signed by the user.

The second property becomes very problematic when we get the sequencer to adjust fee rates however it wants. A scenario:

- Somebody with a lot of cBTC in their account sends a tx.
- At the same time sequencer raises L1 fee rates dramatically since it's reaching the DA throughput limit.
- The user gets accepted after some time, at which point the L1 fee rates 100xed.
- Since the address has a lot of cBTC, the tx goes through and user ends up paying way more than what he wishes for.

We have plans to make L1 fee rates trustless in the future, however, since fee rates are not a part of the Bitcoin consensus rules, there is no way to make it %100 trustless today.

So instead we can construct throttling on priority fees instead.

We use a similar curve and when we get closer to the DA throughput we start not including txs below some threshold.

The difference here is that the priority fee is already a widely implemented feature in Ethereum (to the point that I don't know a wallet that doesn't have EIP-1559 support) and it would prevent people from overspending by sending txs when DA throughput is being reached.

We already have support for tx priority fee bumping in the mempool so that's good.

One possible against-argument for using priority fees in this sense is that while we are trying to limit DA usage, hence the state diff size, raising L1 fee rates made sense. Because we disincentivize increasing the state diff more. Rasiing the priority fees on the other hand, now we disincentivize regular gas usage.

So an even more dramatic increase with a later start can still work I believe.

PR #1535, opened 2024-11-28
Use HexOrNum in ledger RPCs

### LINKED ISSUE #1530: Use hex on ledger_getSoftConfirmationStatus
**Is your feature request related to a problem? Please describe.**

To check if transaction is finalized, one must
a) Get the block number of the transaction from the RPC
b) Use `ledger_getSoftConfirmationStatus` endpoint to check if the corresponding block is finalized

A tiny problem here is that option a) returns block number in hex, while option b) accepts decimals. 
While it's easy to do the conversion, we can unify such usage - or can decide on all endpoints how the usage should be for final.

**Describe the solution you'd like**
I'd probably make this hex or allowing both strings and numbers and then auto-deciding which data type is used

**Describe alternatives you've considered**
None, it's not something super important.

jfldde:
# Description

- Accept u64 or hex string as input in ledger RPCs

## Linked Issues
- Fixes #1530 

## Testing

```
curl -s -X POST -H 'accept: application/json' -H 'content-type: application/json' -d '
{
  "id": 1,
  "jsonrpc": "2.0",
  "method": "eth_getTransactionReceipt",       
  "params": [
    "0x9c56cd995a89812aa575d691e50a791c83ad62a64416c8e4c64e706b89862b5b"
  ]
}' localhost:12345  | jq ' .result.blockNumber'
"0x1"

curl -s -X POST -H 'accept: application/json' -H 'content-type: application/json' -d '
{
  "id": 1,
  "jsonrpc": "2.0",
  "method": "ledger_getSoftConfirmationStatus",
  "params": [
    1                                                                   
  ]
}' localhost:12345  | jq                       
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": "trusted"
}

curl -s -X POST -H 'accept: application/json' -H 'content-type: application/json' -d '
{
  "id": 1,
  "jsonrpc": "2.0",
  "method": "ledger_getSoftConfirmationStatus",
  "params": [
    "0x1"                                                               
  ]
}' localhost:12345  | jq                       
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": "trusted"
}
```


crates/sovereign-sdk/full-node/sov-ledger-rpc/src/lib.rs
eyusufatik, line 1: we can import something like alloy::U64 btw. already has all kind of hex ser/de

jfldde: Much better

jfldde: Does implement `From<u64>` though, makes it tedious.

crates/sovereign-sdk/full-node/sov-ledger-rpc/src/client.rs
```diff
@@ -70,7 +71,7 @@ pub trait Rpc {
     #[method(name = "getBatchProofsBySlotHeight")]
     async fn get_batch_proofs_by_slot_height(
         &self,
```
eyusufatik, line 73: there is a [u8; 32] in the above function we forgot to update, can you also update that?

also the (u64, u64) in get_soft_confirmation_range

eyusufatik (changes requested):

eyusufatik (approved):

PR #1539, opened 2024-11-29
Use V1/V2 circuit inputs/outputs based on fork

### LINKED ISSUE #1538: Use V1/V2 circut input while proving where appropriate
In `data_to_prove`, we always return V2 circuit input. For proving data prior up to testnet's launch, we have been using V1.

Update proving to use V1/V2 based on L2 height / current spec.

rakanalh:
# Description

Defines:
`BatchProofCircuitInput` and `BatchProofCircuitOutput` as enum with variants `V1` and `V2` for being able to deserialize into the proper structure when passed between host and guest.

# TODO
- [ ] If deserialize is unsuccessful, deserialize back to V1.

## Linked Issues
- Fixes #1538 
- Depends on #1495


crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
```diff
@@ -242,6 +242,29 @@ pub struct BatchProofCircuitInputV2<'txs, StateRoot, Witness, Da: DaSpec, Tx: Cl
...
+pub enum BatchProofCircuitInput<'txs, StateRoot, Witness, Da: DaSpec, Tx: Clone> {
+    /// V1 or genesis
+    V1(BatchProofCircuitInputV1<'txs, StateRoot, Witness, Da, Tx>),
+    /// V2
+    V2(BatchProofCircuitInputV2<'txs, StateRoot, Witness, Da, Tx>),
+}
+
+/// Grouping circuit output to determine which to use based on current fork.
+#[derive(Clone, Debug, Serialize, Deserialize, BorshSerialize, BorshDeserialize, PartialEq, Eq)]
+pub enum BatchProofCircuitOutput<Da: DaSpec, StateRoot> {
```
eyusufatik, line 261: one problem I know you'll have is that you are adding a byte for the outter enum now, you wont be able to parse testnet zk proofs

eyusufatik: I mean the ones we already sent

eyusufatik: maybe since we are only in testnet, we can forget about backwards compatability for these kind of things 🤔 

but it'll be hard to initialize light client proof circuit nevertheless

crates/citrea-stf/src/verifier.rs
```diff
@@ -40,26 +41,75 @@ where
...
+            sequencer_public_key,
+            initial_state_root,
+            prev_soft_confirmation_hash,
+            sequencer_commitments_range,
+            state_transition_witnesses,
+            soft_confirmations,
+            preproven_commitments,
+        ) = match data {
+            BatchProofCircuitInput::V1(input) => (
+                1,
```
kpp, line 64: A little bit confusing. A local enum would be more friendly to grasp. But that's not a blocker note for this PR

kpp (approved):

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
rakanalh: I stopped yesterday because of this exact reason. Problem is, unless we are able to know what we're parsing, v1/v2, we won't have access to the last l2 height from which we can figure out the fork and therefore the output type. I have to find an alternative.

eyusufatik: one solution is to try parsing V2, if fails, go with v1

but that's not really a always working solution 😓

rakanalh: Yea... and in this case we have to change the way `zkvm.read_from_host();` behaves since we consume stdin bytes once and try to parse them.

rakanalh:
In a conversation with @eyusufatik, we've concluded that it's not worth having V1/V2 structs where V2 only removes 1 field that is 32 bytes. It's better to keep passing this redundantly compared to the complexity this PR adds.

Closing this PR in favor of restoring 1 input/output structs.

PR #1540, opened 2024-11-29
Always prove to produce proofs for BatchProofs

### LINKED ISSUE #1477: new proof sampling with fake receipts
The thing is we do proof sampling this way: prover works once a day, we send the proof to da.

It works. But the problem is we need to test da and proving system that we can actually works.

Instead we can run every seqcom on l1 block in execute mode (no proving but producing fake receipts).

So we have to implement producing fake proof for all batch proofs but with sampling to produce a real proof once a day.

### LINKED ISSUE #1478: Testnet light client proofs without proving
followup of #1477

kpp:
# Description

We run every seqcom on l1 block in execute mode (no proving but producing fake receipts).
So we have to implement producing fake proof for all batch proofs but with sampling to produce a real proof once a day.

In order to do that there is a new proving mode: `prove-with-fakes`.

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/1477
- Fixes https://github.com/chainwayxyz/citrea/issues/1478

## Testing

I tested locally for:

```
index e4bb86e5..37228afa 100644
--- a/bin/citrea/tests/bitcoin_e2e/light_client_test.rs
+++ b/bin/citrea/tests/bitcoin_e2e/light_client_test.rs
@@ -45,6 +45,7 @@ impl TestCase for LightClientProvingTest {
     fn batch_prover_config() -> BatchProverConfig {
         BatchProverConfig {
             enable_recovery: false,
+            proof_sampling_number: 999999999usize,
             ..Default::default()
         }
     }
@@ -52,6 +53,7 @@ impl TestCase for LightClientProvingTest {
     fn light_client_prover_config() -> LightClientProverConfig {
         LightClientProverConfig {
             enable_recovery: false,
+            proof_sampling_number: 0,
             ..Default::default()
         }
     }
@@ -453,7 +455,7 @@ impl TestCase for LightClientProvingTestMultipleProofs {
 }
```

So light client proving for fake batch proofs works.


eyusufatik (changes requested):
explained in DMs

eyusufatik (approved):
sorry I misunderstood

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
```diff
@@ -349,9 +349,13 @@ impl SharedLedgerOps for LedgerDB {
         match commitments {
             // If there were other commitments, upsert
             Some(mut commitments) => {
-                commitments.push(commitment);
-                self.db
-                    .put::<CommitmentsByNumber>(&SlotNumber(height), &commitments)
+                if !commitments.contains(&commitment) {
```
kpp, line 352: I couldn't find any other way to fix duplicates in db. See https://github.com/chainwayxyz/citrea/actions/runs/12084846708/job/33700913239#step:13:2544

crates/prover-services/src/lib.rs
```diff
@@ -15,10 +21,33 @@ where
     /// Skips proving.
     Skip,
     /// The simulator runs the rollup verifier logic without even emulating the zkVM
-    Simulate(StateTransitionVerifier<Stf, Da::Verifier, Vm::Guest>),
+    Simulate(Simulator<Stf, Da::Verifier, Vm::Guest>),
```
yaziciahmet, line 24: Very nice!

crates/prover-services/src/parallel/mod.rs
```diff
@@ -234,22 +240,31 @@ fn make_proof<Da, Vm, Stf>(
...
 ) -> Result<Proof, anyhow::Error>
 where
     Da: DaService,
     Vm: ZkvmHost,
     Stf: StateTransitionFunction<Da::Spec> + Send + Sync,
     Stf::PreState: Send + Sync,
 {
-    let mut proof_mode = proof_mode.blocking_lock();
-    match proof_mode.deref_mut() {
+    let proof_mode = proof_mode;
```
yaziciahmet: not needed anymore

crates/prover-services/src/parallel/mod.rs
```diff
@@ -234,22 +240,31 @@ fn make_proof<Da, Vm, Stf>(
...
+                    anyhow::anyhow!("Guest execution must succeed but failed with {:?}", e)
+                })
+        }
         ProofGenMode::Execute => vm.run(elf, false),
-        ProofGenMode::Prove => vm.run(elf, true),
+        ProofGenMode::Prove(proof_sampling_number) => {
+            let with_prove = proof_sampling_number == 0
+                || rand::thread_rng().gen_range(0..proof_sampling_number) == 0;
+            vm.run(elf, with_prove)
+        }
```
yaziciahmet: I have one concern about doing this here. We are doing parallel proving, which will cause some proofs in a da block to be proved, while some of them to be executed. I am not sure if that's ok?

yaziciahmet (changes requested):

crates/prover-services/src/parallel/mod.rs
yaziciahmet: I think we should carry this decision up to ParallelProverService, and either execute the whole block, or prove the whole block.

kpp: > some proofs in a da block to be proved, while some of them to be executed

Not really. With ProveWithSamplingWithFakeProofs some proofs will be faked, some proofs will be proved for real.

eyusufatik (approved):
are you sure light client proofs with work with fake proofs and real proofs? (running in dev mode ofc)

kpp:
Yes. Before we added a new proving mode `prove-with-fakes` I tested it with:

```
index e4bb86e5..37228afa 100644
--- a/bin/citrea/tests/bitcoin_e2e/light_client_test.rs
+++ b/bin/citrea/tests/bitcoin_e2e/light_client_test.rs
@@ -45,6 +45,7 @@ impl TestCase for LightClientProvingTest {
     fn batch_prover_config() -> BatchProverConfig {
         BatchProverConfig {
             enable_recovery: false,
+            proof_sampling_number: 999999999usize,
             ..Default::default()
         }
     }
@@ -52,6 +53,7 @@ impl TestCase for LightClientProvingTest {
     fn light_client_prover_config() -> LightClientProverConfig {
         LightClientProverConfig {
             enable_recovery: false,
+            proof_sampling_number: 0,
             ..Default::default()
         }
     }
@@ -453,7 +455,7 @@ impl TestCase for LightClientProvingTestMultipleProofs {
 }
 ```
 
 I can't do it now with `prover-with-fakes` though, because BatchProverConfig is in citrea-e2e repo.

File: crates/sovereign-sdk/full-node/sov-stf-runner/src/prover_service/mod.rs [hunk 5]
```diff
@@ -29,6 +31,7 @@ impl<'de> Deserialize<'de> for ProverGuestRunConfig {
             "simulate" => Ok(ProverGuestRunConfig::Simulate),
             "execute" => Ok(ProverGuestRunConfig::Execute),
             "prove" => Ok(ProverGuestRunConfig::Prove),
+            "prove-with-fakes" => Ok(ProverGuestRunConfig::ProveWithFakeProofs),
```
yaziciahmet, line 34: "fake-prove" might be more concise?

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
yaziciahmet: the best way to do this would be to actually store commitments as with key `{slot_height}-{comm_start}-{comm_end}`, and then use the prefix iterator functionality of rocksdb to iterate over slot heights, which would ensure uniqueness of commitments in db level and would be still quite efficient since the keys will be ordered by nature.

yaziciahmet:
I don't quite understand the motivation behind this PR. As far as I observed:
- if prove specified, we don’t call prove-l1 at all unless random sampling satisfies
- if prove-with-fakes specified, we always call prove-l1, but we randomly decide in prover-service to generate either fake or real proofs in a per-proof basis

What is the effect of this to fullnodes? How will fullnodes know whether the proof is fake or not when verifying if some of them can be fake and some of them can be real?

How will this be used in conjunction with the `RISC0_DEV_MODE` env variable?

What does this bring us in general?

eyusufatik:
> I don't quite understand the motivation behind this PR. As far as I observed:
> 
> * if prove specified, we don’t call prove-l1 at all unless random sampling satisfies
> * if prove-with-fakes specified, we always call prove-l1, but we randomly decide in prover-service to generate either fake or real proofs in a per-proof basis
> 
> What is the effect of this to fullnodes? How will fullnodes know whether the proof is fake or not when verifying if some of them can be fake and some of them can be real?
> 
> How will this be used in conjunction with the `RISC0_DEV_MODE` env variable?
> 
> What does this bring us in general?

we get to test da consumption and gather logs on testnet without any additional costs.

looks like our testnet zk aproach will be based on this

crates/prover-services/src/parallel/mod.rs
```diff
@@ -234,22 +243,38 @@ fn make_proof<Da, Vm, Stf>(
...
+                })
+        }
         ProofGenMode::Execute => vm.run(elf, false),
-        ProofGenMode::Prove => vm.run(elf, true),
+        ProofGenMode::ProveWithSampling => {
+            // `make_proof` is called with a probability in this case.
+            // When it's called, we have to produce a real proof.
+            vm.run(elf, true)
+        }
+        ProofGenMode::ProveWithSamplingWithFakeProofs(proof_sampling_number) => {
```
yaziciahmet: Only nit I have is that ProveWithSampling is a bit confusing to me, because in the eyes of ProverService it is just Prove, and sampling happens above in the call stack in da_block_handler. So i think we should just call them Prove and ProveWithSampling respectively, so its obvious from the ProofGenMode whether ProverService should do any sampling itself.

I am not going to block the PR for this nit, but would love to see this change.

yaziciahmet (approved):

crates/prover-services/src/parallel/mod.rs
yaziciahmet: Also, what if instead of sampling on per proof basis for ProveWithSamplingWithFakeProofs, we sample per DA block. Meaning, we again decide the sampling on da_block_handler level, but what we decide is to whether proofs in current DA should be run in execute or prove mode. It would simplify the flow quite a bit. Just throwing this out as a thought. WDYT?

kpp: We talked with Esad, the approach doesn't matter much in this PR. We may open a new issue as a followup

yaziciahmet:
I verified that it is possible to verify real Groth16 proofs even in RISC0_DEV_MODE. If the receipt is of type Groth16, no dev mode check is done and it just does the verification. Hence, we are good to go with this PR!

PR #1543, opened 2024-12-02
Apply brotli compression statistic

### LINKED ISSUE #1298: Calculate statistic for brotli compression and charge user accordingly
Currently we overcharge L1 fee on users since using brotli compression on a single tx's state diff would not yield good results.

However as we did with other L1 fee statistics, we can work on real world data and calculate a "brotli compression" statistic and charge users accordingly.

exeokan:
# Description

## Linked Issues
- Fixes #1298


crates/evm/src/evm/handler.rs
```diff
@@ -401,8 +405,12 @@ impl<SPEC: Spec, EXT: CitreaExternalExt, DB: Database> CitreaHandler<SPEC, EXT,
         context: &mut Context<EXT, DB>,
         result: FrameResult,
     ) -> Result<ResultAndState, EVMError<<DB as Database>::Error>> {
-        let diff_size =
-            calc_diff_size::<EXT, SPEC, DB>(context).map_err(EVMError::Database)? as u64;
+        let uncompressed_size =
+            calc_diff_size::<EXT, SPEC, DB>(context).map_err(EVMError::Database)?;
```
eyusufatik, line 409: you should check if Spec::Cancun is enabled here. 

On Fork1, we'll set cancun to enabled in revm, so internally actions can be branched on that.


why? imagine you're syncing to testnet with this code. you'll apply different balance changes to all txs, causing you to be on a completely different state than all other nodes

eyusufatik (changes requested):

crates/evm/src/evm/handler.rs
exeokan: made changes

kpp (approved):

rakanalh (approved):
LGTM

crates/evm/src/evm/handler.rs
```diff
@@ -401,8 +405,16 @@ impl<SPEC: Spec, EXT: CitreaExternalExt, DB: Database> CitreaHandler<SPEC, EXT,
...
+
+        let diff_size = if SPEC::enabled(SpecId::CANCUN) {
```
eyusufatik: I'd just go

```rust
let mut uncompressed_size = ...

if SPEC::enabled(SpecId::CANCUN) {
  uncompressed_size = (uncompressed_size * BROTLI_COMPRESSION_PERCENTAGE / 100) as u64
}
```

kpp: Well, better avoid muts if they can be avoided.

```
let uncompressed_size = ...
let compression_ratio = if SPEC::enabled(SpecId::CANCUN) {
    BROTLI_COMPRESSION_PERCENTAGE / 100
} else {
    1.0
}
let diff_size = (uncompressed_size * compression_ratio) as u64;
```

yaziciahmet (approved):

eyusufatik:
ege is writing tests for the pr btw that's why we are not merging

crates/evm/src/evm/handler.rs
exeokan: I wanted to avoid using floats and muts at the same time, that's why I used percentage like other discount constants. If using floats is fine I can go with Roman's suggestion?

eyusufatik (approved):

crates/evm/src/evm/handler.rs
exeokan: made changes

PR #1547, opened 2024-12-03
Restore input/output

### LINKED ISSUE #1538: Use V1/V2 circut input while proving where appropriate

rakanalh:
# Description

## Linked Issues
- Fixes #1538


eyusufatik (approved):
aproving but let's not merge until we reach a final decision

PR #1548, opened 2024-12-03
No production guest builds in tests

rakanalh:
# Description

Leftover from the previous #1495 PR where we build production guests for tests. This should NOT happen since we use citrea-risc0 to build latest.


eyusufatik (approved):

kpp:
Thank you

PR #1514, opened 2024-11-25
Invalidate Soft Confirmation on sov-module errors

eyusufatik:
# Description
any failing (I don't mean reverting here, failing) evm tx will make the soft confirmation invalid.

The way I handled the hash bug fix should be changed into a better one imo.

At the beginning of the project we thought when we enabled forced transactions we would need to run them along soft confirmations. However, they can be treated differently. All the workaround and ignoring errored transactions makes a lot of stuff much harder. e.g. we have to have weird workaround accessory maps for block bulding etc.

Another goal of this pr is to fix error handling and propogation in the STF   

Closes #1438 
Closes #1061
Closes #1048

TODO
- [x] before merging: Double check working_set reverts in stf blueprint. one thing to notice is that a sov-tx erroring now means reverting the whole soft confirmation. Resolved: we don't checkpoint in stf now, becuase we invalidate on error. way less memcpy in general.
- [x] #1493 
- [x] Check ZK side checkpointing in stf

however, in evm unit tests, we keep working on a working set after a failing evm call. such as going over a block gas limit.  now the working set reverting must happen **explicitly**.

PR #1551, opened 2024-12-03
Add missing migrations

### LINKED ISSUE #1542: Add missing ledger db migrations
## Summary
We have some name changes on ledger db tables after we implemented light client proofs, we added namespaces to those tables like `...BatchProofs` or `...LightClientProofs`. In order to be able to sync from the beginning on the current nightly version we need to implement necessary migrations

ercecan:
# Description
Adds migration of ledger db table name change
VerifiedProofsBySlotNumber -> VeriifiedBatchProofsBySlotNumber

Also adds an api to help make this easier for the future migrations

## Linked Issues
- Fixes #1542


crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
```diff
@@ -25,6 +25,19 @@ pub struct UnsignedSoftConfirmation<'txs, Tx> {
     timestamp: u64,
 }
 
+/// TODO: doc or find workaround
+#[derive(BorshSerialize)]
+pub struct OldFormat<'txs> {
```
rakanalh:
```suggestion
pub struct UnsignedSoftConfirmationV1<'txs> {
```

crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
```diff
@@ -109,9 +122,34 @@ impl<'txs, Tx: BorshSerialize> UnsignedSoftConfirmation<'txs, Tx> {
...
+            l1_fee_rate: self.l1_fee_rate,
+            timestamp: self.timestamp,
+        };
+
+        let raw = borsh::to_vec(&old).unwrap();
         D::digest(raw.as_slice())
     }
+
+    /// TODO: Documentation
+    pub fn get_old_format(&self) -> OldFormat<'txs> {
```
rakanalh: This can be in an 
```
impl Into<UnsignedSoftConfirmation> for UnsignedSoftConfirmationV1 
```

crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
```diff
@@ -866,7 +866,7 @@ fn pre_fork1_verify_soft_confirmation_signature<C: Context, Tx: BorshSerialize>(
     signature: &[u8],
     sequencer_public_key: &[u8],
 ) -> Result<(), anyhow::Error> {
-    let message = borsh::to_vec(&unsigned_soft_confirmation).unwrap();
+    let message = borsh::to_vec(&unsigned_soft_confirmation.get_old_format()).unwrap();
```
rakanalh:
```suggestion
    let message = borsh::to_vec(&unsigned_soft_confirmation.into()).unwrap();
```

.gitignore
```diff
@@ -16,6 +16,8 @@ resources/dbs/sequencer-db/
 resources/dbs/full-node-db/
 resources/dbs/batch-prover-db/
 resources/dbs/light-client-prover-db/
+data/
```
rakanalh: Anything we write to disk should go into `resources`. We should keep the root clean of dynamically created files / directories.

rakanalh: I also searched where we use this directory and could not find any usage of this.

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
```diff
@@ -103,17 +103,18 @@ impl<S: SlotData, B, T> SlotCommit<S, B, T> {
...
 impl LedgerDB {
     /// Open a [`LedgerDB`] (backed by RocksDB) at the specified path.
-    /// The returned instance will be at the path `{path}/ledger-db`.
+    /// Will take optional column families, used for migration purposes.
+    /// The returned instance will be at the path `{path}/ledger`.
     #[instrument(level = "trace", skip_all, err)]
-    pub fn with_config(cfg: &RocksdbConfig) -> Result<Self, anyhow::Error> {
+    pub fn with_config(
+        cfg: &RocksdbConfig,
+        column_families: Option<Vec<String>>,
```
rakanalh: I believe column families are rocksdb related. Please include them in `RocksdbConfig`

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/migrations.rs
```diff
@@ -107,27 +121,58 @@ impl<'a> LedgerDBMigrator<'a> {
...
+        assert!(!ledger_path.exists(), "Failed to clear original path");
+
         // Copy new DB into original path
-        copy_db_dir_recursive(temp_db_path.path(), original_path)?;
+        copy_db_dir_recursive(&temp_ledger_path, &ledger_path)?;
 
         Ok(())
     }
+
+    fn merge_column_families(column_families_in_db: Vec<String>) -> Vec<String> {
```
rakanalh: This can be moved outside `LedgerDbMigrator` above `clear_db_dir`

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
```diff
@@ -130,6 +131,60 @@ impl LedgerDB {
...
+    }
+
+    /// List all column families in the database
+    pub fn list_column_families(path: &Path) -> Vec<String> {
+        rocksdb::DB::list_cf(
+            &rocksdb::Options::default(),
+            path.join(LEDGER_DB_PATH_SUFFIX),
+        )
+        .unwrap()
+    }
```
rakanalh: These don't belong to `LedgerDB`... maybe add them as migration helpers / utils?

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/migrations.rs
```diff
@@ -107,27 +121,58 @@ impl<'a> LedgerDBMigrator<'a> {
...
         drop(new_ledger_db);
+
+        // Now that the lock is gone drop the tables that were migrated
+        for table in tables_to_drop {
+            LedgerDB::drop_cf(
+                &RocksdbConfig::new(temp_db_path.path(), max_open_files),
+                Some(all_column_families.clone()),
+                &table,
+            )?;
+        }
```
rakanalh: I don't believe that dropping tables is the responsibility of the migration manager. Each migration should decide what it wants to do whether it's to migrate old tables to new or remove tables .. etc. Please call `drop_column_family` inside the migration itself.

rakanalh (changes requested):

.gitignore
ercecan: Yeah my bad, it is something else will remove

crates/fullnode/src/runner.rs
```diff
@@ -420,6 +423,13 @@ where
                     self.task_manager.abort().await;
                     return Ok(());
                 }
+                // Use this when SIGINT is not accepted
+                // Get the process id of the node and run `kill -15 <process_id>`
+                _ = shutdown_stream.recv() => {
+                    info!("Shutting down");
+                    self.task_manager.abort().await;
+                    return Ok(());
+                }
```
rakanalh: This could have belonged to it's own PR since it's unrelated to the main context of the PR.

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/migrations.rs
ercecan: I cannot do that because it is still locked outside of the migration itself
I agree that it is not the responsibility but the migration it self decides whether to drop it or not by adding them to a vector.
To do it as you say I will need to:
1-  pass the temp db path to the execute function
2- create ledger db
3- Do the migration stuff
4- Drop temp db
5- Drop cf
6- Repeat this process on every single migration

crates/fullnode/src/runner.rs
ercecan: Yeah, did not want to bother tbh :D

ercecan: Should I?

crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
ercecan: These changes are cherry-picked from #1514 
I believe I should wait for him to change these parts

crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
ercecan: These changes are cherry-picked from #1514 
I believe I should wait for him to change these parts

crates/fullnode/src/runner.rs
rakanalh: yes please

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/migrations.rs
rakanalh: I don't understand that.

- The migration `execute` receives an instance of the `new_ledger_db` which is already using the one in tempdir.
- Do step number 3
- Drop cf

crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
rakanalh: Commented there on this.

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
ercecan: On second thought, I am not sure if it should be in the config, because we can already get it with a function just by giving the path

ercecan: I'll think on this a bit more

ercecan: It also kinda makes sense as well :D

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/migrations.rs
ercecan: I cannot drop cf without dropping all instances of a rocksdb

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
ercecan: How about I put them in LedgerDBMigrator instead of utils or sth like that?

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/migrations.rs
rakanalh: What happens if you don't drop it?

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
ercecan: Okay, so I think let's skip that one for now because the column_families should be optional and if it is None it should be either LEDGER_TABLES STATE_TABLES or NATIVE_STATE_TABLES, to extract which to use we can use the path or pass another parameter but I feel like that's just extra code

I recommend implementing this if we really need to

WDYT?

crates/fullnode/src/runner.rs
ercecan: #1560

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/migrations.rs
ercecan: 1) The ledger db will not open with the usual LEDGER_TABLES and cry missing column family etc etc because the struct is removed from LEDGER_TABLES
2) It stays there and occupies extra space

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
rakanalh: I am not sure i follow. In either way, we are passing `column_families` here... why not move this param to `RocksdbConfig` and use it the exact same way you do in the PR?

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/migrations.rs
rakanalh: Okay. Let me take a stab at it tomorrow morning and i'll let you know the result.

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
ercecan: I can do that for ledger db but for example here I will need to create two configs just for that, there are other examples to this as well, I think this will just generate more code + all of the column families are accessible publicly by all crates importing sov-db as they are pub consts
Screenshot 2024-12-05 at 22 48 48

rakanalh: In this example, we pass `None` and then `unwrap_or_else` when the default list of CFs are needed.

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/migrations.rs
rakanalh: > I cannot drop cf without dropping all instances of a rocksdb

I see your point now. The problem is that `DB` used by ledger is behind an `Arc` and `drop_cf` needs to borrow a mutable reference. Which is why we need to call `drop_cf` with the instance of `mut DB`

rakanalh:
Should be merged after: https://github.com/chainwayxyz/citrea/pull/1514

To fix the cherry-picked stuff from the merged version.

rakanalh (approved):

eyusufatik (approved):
after merging #1514 you need to remove OldFormat stuff and we are good to go

eyusufatik:
I reverted

PR #1554, opened 2024-12-04
Restore short-prefix behind cargo flag

jfldde:
# Description

- Restore short-prefix behind cargo `short-prefix` flag

Main rationale for the removal in https://github.com/chainwayxyz/citrea/pull/1537 was that compile time env was not properly propagated to risczero docker build. This restores it by propagating via cargo `short-prefix` flag, albeit in a convoluted way.

- Automatically derive it when in running `make test` and `make coverage`
- Remove the need to keep `CI_TEST_MODE` in line between build and test profiles.

Note that if for whatever reason, one is running via `cargo test` (not advised, `make test` is the recommended way) directly, additional `--features short-prefix` is needed in order not to have a mismatch in prefix length

### DIFF
--- a/.github/workflows/nightly_build_push.yml
+++ b/.github/workflows/nightly_build_push.yml
@@ -35,11 +35,11 @@ jobs:
 
       - name: Build Project
         run: |
-          cargo build --release
+          cargo build --features short-prefix
 
       - name: Copy binary to build-push/nightly
         run: |
-          cp target/release/citrea docker/build-push/nightly/citrea
+          cp target/debug/citrea docker/build-push/nightly/citrea
           chmod +x docker/build-push/nightly/citrea
 
       - name: Set up Docker Buildx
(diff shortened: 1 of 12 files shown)


.github/workflows/nightly_build_push.yml
```diff
@@ -35,11 +35,11 @@ jobs:
 
       - name: Build Project
         run: |
-          cargo build --release
+          cargo build --features short-prefix
```
eyusufatik, line 38: since we are calling this nightly build it might be confusing to use short prefix bc some people might try to use the docker image in prod. let's maybe rename to test build or smth

jfldde: The image is called citrea-test, isn't that enough ?

jfldde: `tags: ${{ vars.DOCKERHUB_USERNAME }}/citrea-test:${{ env.IMAGE_TAG }}`

eyusufatik: sorry missed that.

eyusufatik (approved):

rakanalh (approved):

PR #1556, opened 2024-12-04
Add rlp codec for Block and Header

kpp:
# Description

RLP encoding is the ethereum spec encoding. if we want to keep upgrading reth and alloy in the future without breaking backwards compatability, we should abide by the ethereum rules for encoding decoding stuff.

right now we cant upgrade reth because they changed some minor thing in the header field :)

This PR implements forkable switch between bcs and rlp codecs.


eyusufatik:
for AccessoryState, I believe we can always find a solution. The problem will be much more problematic in State though. Look at evm.head it's read first in the hook, then set. How can we achieve upgrading reth for that?

rakanalh (commented):
- What issue does this PR close?
- What is the purpose of having RLP encoded blocks & headers?

Would be nice to have a description for the PR & mention of the issue we're trying to resolve.

ercecan:
Added rlp codec for all reth types, including accessory state, for state only Head needed changing so added head_rlp field.
All nodes will need to resync but it will work

crates/evm/src/lib.rs
```diff
@@ -64,6 +66,24 @@ impl PendingTransaction {
...
+impl Decodable for PendingTransaction {
+    fn decode(rlp: &mut &[u8]) -> Result<Self, alloy_rlp::Error> {
+        let transaction = TransactionSignedAndRecovered::decode(rlp)?;
+        let receipt = Receipt::decode(rlp)?;
+        Ok(Self {
+            transaction,
+            receipt,
+        })
+    }
+}
```
eyusufatik: are we sure we can just stitch together two rlp's like this?

crates/evm/src/hooks.rs
```diff
@@ -157,11 +177,26 @@ where
...
+                        .into();
+                    block.seal()
+                }
+            }
+        } else {
+            self.head
+                .get(working_set)
+                .expect("Head block should always be set")
+                .seal()
+        };
```
eyusufatik, line 199: want to make a function for this? same stuff happens in two places

crates/evm/src/call.rs
```diff
@@ -135,6 +135,7 @@ impl<C: sov_modules_api::Context> Evm<C> {
     ) -> Result<CallResponse> {
         // use of `self.block_env` is allowed here
 
+        // TODO: should not include non deserilizable transactions
```
eyusufatik, line 138: open issue for this

eyusufatik (changes requested):

eyusufatik:
> * What issue does this PR close?
> * What is the purpose of having RLP encoded blocks & headers?
> 
> Would be nice to have a description for the PR & mention of the issue we're trying to resolve.

RLP encoding is the ethereum spec encoding. if we want to keep upgrading reth and alloy in the future without breaking backwards compatability, we should abide by the ethereum rules for encoding decoding stuff.

right now we cant upgrade reth because they changed some minor thing in the header field :)

eyusufatik:
@ercecan pls confirm you can sync to current testnet with this pr

crates/evm/src/lib.rs
ercecan: why not?

crates/evm/src/hooks.rs
ercecan: I tried it but it made worse, created more copying because of having to do more from into conversions, can stay as is

crates/evm/src/call.rs
ercecan: #1617

crates/evm/src/evm/primitive_types.rs
```diff
@@ -133,22 +133,69 @@ impl From<Header> for AlloyHeader {
...
     /// Signer of the transaction
     pub(crate) signer: Address,
     /// Signed transaction
-    #[serde_as(as = "reth_tx_serde_bincode_compat::TransactionSigned")]
     pub(crate) signed_transaction: TransactionSigned,
     /// Block the transaction was added to
     pub(crate) block_number: u64,
 }
 
+impl Encodable for TransactionSignedAndRecovered {
```
jfldde: We can use alloy-rlp derive feature and #[derive(RlpEncodable, RlpDecodable)] here and across the PR

crates/evm/src/lib.rs
eyusufatik: please make sure this is not a problem. read about how RLP encoding / decoding work. make sure this is not a problem.

jfldde: Let's have the derive macro take care of it

crates/evm/src/evm/primitive_types.rs
ercecan: Range<u64> or [u64; 2] does not implement Encodable unfortunately

eyusufatik (approved):

PR #1557, opened 2024-12-04
Fix sequencer commitments

yaziciahmet:
# Description

Current sequencer commitment creation had certain bugs and potential improvements.

Fixes:

1. When state diff threshold exceeded the limit, last L2 block was included into the commitment, which it shouldn't because it would make the tx body bigger than the limit.
2. When state diff threshold exceeded the limit, state diff is set to the last l2 block's state diff, which is the correct behavior, but in `CommitmentService::commit` method, this state diff was overwritten with an empty vector.
3. Soft confirmation threshold was prioritized over state diff threshold, which shouldn't be the case since if the soft confirmation threshold is 50, and at 50th block both state diff and soft confirmation thresholds satisfy, we should create the commitment for blocks 1-49, and not 1-50.

General Improvements:

1. Instead of iterating over trait and running each threshold no matter what, a bit more fine grained controlled threshold checks are made.
2. If incoming blocks state diff is empty, no state diff operation is made as it's guaranteed to not change anything, and not meet state diff threshold condition.
3. In state diff threshold check, first uncompressed state diff is checked against the limit. Because if uncompressed state diff is already smaller the limit, there is no need to do compression.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


yaziciahmet:
One thing open to discussion is that, we know from our statistics that on average brotli has 66% compression rate. Knowing this, we can even return early for uncompressed data if its size is less than `MAX_TXBODY_SIZE * 3`. Of course this would be a risky move since we are talking average and a more sane number would be smh. like `MAX_TXBODY_SIZE * 1.5` which would assume that brotli compression has at least 33% compression rate.

Do we have such statistics with brotli, lowest compression rate(s) @eyusufatik ?

crates/sequencer/src/commitment/controller.rs
```diff
@@ -0,0 +1,161 @@
...
+            ledger_db,
+            min_soft_confirmations,
+            last_state_diff,
+        }
+    }
+
+    pub fn should_commit(
+        &mut self,
+        l2_height: u64,
+        l2_state_diff: StateDiff,
```
kpp, line 43: Why not &StateDiff?

crates/sequencer/src/commitment/controller.rs
```diff
@@ -0,0 +1,161 @@
...
+    min_soft_confirmations: u64,
+    last_state_diff: StateDiff,
+}
+
+impl<Db> CommitmentController<Db>
+where
+    Db: SequencerLedgerOps,
+{
+    pub fn new(ledger_db: Db, min_soft_confirmations: u64) -> Self {
+        let last_state_diff = ledger_db.get_state_diff().unwrap_or_default();
```
kpp, line 32: I am not sure if this logic is correct. What if we launch controller with sdiff from L2=1000, but .should_commit is called with L2=900? Their merging does not make sense.

kpp (approved):

yaziciahmet:
I did some manual runs with as much randomized data as I can, and the minimum compression rate I saw from fully empty blocks to many tx blocks was 58%. Which gives me confidence on the number 33% as the safe bet.

crates/sequencer/src/commitment/controller.rs
yaziciahmet: It is the responsibility of Sequencer runner to ensure that each new L2 block is sent through the channel in order. If somehow there is a mismatch in between blocks and statediff, then we have a very big problem.

The only place where we produce multiple L2 blocks, but do send only signal is when we have missed DA blocks. Which is then fine, because those are guaranteed to be empty blocks, so there will not be any state diff for sure.

It might actually be safe to store the last l2 height in CommitmentController, and assert that the next incoming request to should_commit is 1 increment of it.

WDYT?

crates/sequencer/src/commitment/controller.rs
yaziciahmet: where we call should_commit, state diff is only needed once, so we can just give the ownership to this method

rakanalh (commented):
I like the strategy approach better. I am not sure why you felt you needed to delete that approach and rewrite it while you could have fixed issues in the code as-is.

yaziciahmet:
> I like the strategy approach better. I am not sure why you felt you needed to delete that approach and rewrite it while you could have fixed issues in the code as-is.

The main reason I removed the strategy approach is that CommitmentController needs fine-grained control over which condition to call first, and have the capability to early return without calling the second one. Also, this allows optimizing to code further as we can get away with less clones, early returns, and single db calls.

Other than that I don't think the strategy trait abstraction was a good abstraction. It was something like a "Rule" trait. I don't think each rule deserves an encapsulation of their own. It was unnecessarily complicating a very simple flow and opening gates for overriding each other's states.

crates/sequencer/src/commitment/controller.rs
kpp: Good idea

rakanalh:
> CommitmentController needs fine-grained control over which condition to call first

Whats the difference between returning a list of strategies which are executed in the order they are defined vs hard coding which condition to call first? I don't see your point here.

> and have the capability to early return without calling the second one

This could have been fixed in the controller.

> Other than that I don't think the strategy trait abstraction was a good abstraction. It was something like a "Rule" trait. I don't think each rule deserves an encapsulation of their own. It was unnecessarily complicating a very simple flow and opening gates for overriding each other's states.

I disagree, encapsulating logic separately is always a better approach. I don't think the added complexity is that big vs the readability of each component in this case.

> opening gates for overriding each other's states.

State is not shared between them anyway. One deals with the min_soft_confirmations and the other deals with state diffs. Your argument is invalid.

I am not against refactoring code which was refactored before for additional improvement as long as there is a valid point behind it.

yaziciahmet:
It is meaningless to have them as list of traits if you know exactly what they are going to be, and expect a certain order of them to be provided from the CommitmentService in the first place, and also exactly know what the strategies are going to be. It just creates indirection, makes understanding, following and fixing code harder. It also causes unnecessary extra clonings and extra db calls.

> State is not shared between them anyway. One deals with the min_soft_confirmations and the other deals with state diffs. Your argument is invalid.

Ledger db's last state diff is shared between them, and both should write to it.

I don't think abstracting code is good unless proven to be necessary, and I don't see a good argument for strategy trait here. Our current codebase is already filled with traits and indirections, no need to add more to it for 10 lines of consecutive if statements.

eyusufatik:
when we first created commitment strategies we thought there would be different commitment strategies we'd implement. but looks like it's just going to be diff size or num of l2 blocks.

so having a trait does not really bring anything extra. so I support removing it.

yaziciahmet:
> when we first created commitment strategies we thought there would be different commitment strategies we'd implement. but looks like it's just going to be diff size or num of l2 blocks.
> 
> so having a trait does not really bring anything extra. so I support removing it.

Even if we were to add a new strategy, we would also most likely have to update should_commit signature to also encapsulate the newly needed variables. And if we have to change the signature of the trait and only use 1 of the arguments per strategy implementation and ignore others, it is an indication that this trait actually is not a good encapsulation because it has to change its signature per implementation.

yaziciahmet:
Merging this into v0.5.5 because it is blocking #1546.

@rakanalh if you still have strong feelings about this PR, open a discussion and we can continue discussing there with more people included.

Issue #1513, opened 2024-11-25
Remove StfBlueprintTrait


eyusufatik:
there is only one implementation let's just get rid of it.

PR #1558, opened 2024-12-05
Remove HashStf

jfldde:
# Description

- Remove `HashStf` which was only used for `init_and_restart` full node test
- Fixes [get_l2_state_root](https://github.com/chainwayxyz/citrea/pull/1558/commits/ba5766cea309722e4dbb2bd98789a60e5de1ce1e) return type which was failing serialization

## Linked Issues
- Pre-req for #1513

## Testing
- Replaced with e2e full_node restart test


eyusufatik:
one thing `init_and_restart` did different was that it tested if full node tried to run genesis in the second start. (bc it never synced to the sequencer). any chance you can capture the same in your test also?

rakanalh (approved):
LGTM

yaziciahmet (approved):
LGTM

PR #1560, opened 2024-12-05
Alternative to unresponsive SIGINT signal

ercecan:
# Description
On my macbook m1 pro the ctrl_c (SIGINT) signal won't stop the full node when syncing, however SIGTERM does, so I added listener for that signal as well


crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
```diff
@@ -25,6 +25,19 @@ pub struct UnsignedSoftConfirmation<'txs, Tx> {
     timestamp: u64,
 }
 
+/// TODO: doc or find workaround
```
yaziciahmet: forgotten I assume :D

yaziciahmet (changes requested):

crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
yaziciahmet: not the TODO: but the Old blah blah stuff in general

ercecan: Cherrypicked those, lemme cherry pick the new ones 🍒

ercecan: or just revert lol

yaziciahmet:
can you actually combine sigint and sigterm signals into one helper function like `shutdown_signal` with a select statement on sigint and sigterm?

it could be nice to have that wrapped as a helper method.

crates/fullnode/src/runner.rs
```diff
@@ -420,6 +423,13 @@ where
                     self.task_manager.abort().await;
                     return Ok(());
                 }
+                // Use this when SIGINT is not accepted
+                // Get the process id of the node and run `kill -15 <process_id>`
+                _ = shutdown_stream.recv() => {
+                    info!("Shutting down");
+                    self.task_manager.abort().await;
+                    return Ok(());
```
rakanalh: Refactor this into:
```rust
fn shutdown(&self) {
    info!("Shutting down");
    self.task_manager.abort().await;
}
```
and call it from both ctrl_c and shutdown stream handlers.

rakanalh (approved):
LGTM with one nit.

crates/common/src/utils.rs
```diff
@@ -77,3 +81,29 @@ pub fn check_l2_range_exists<DB: SharedLedgerOps>(
...
+    tokio::spawn(async move {
+        if let Ok(mut term_signal) = signal(SignalKind::terminate()) {
+            while term_signal.recv().await.is_some() {
+                let _ = tx_term.send(()).await;
+            }
+        }
+    });
+
+    ReceiverStream::new(rx)
+}
```
yaziciahmet: you can listen to both of them in a single spawn using tokio::select, and you should be able to just return the rx without needing to create a stream, as rx.recv() is already a future.

crates/common/src/utils.rs
```diff
@@ -77,3 +81,29 @@ pub fn check_l2_range_exists<DB: SharedLedgerOps>(
...
+    tokio::spawn(async move {
+        if signal::ctrl_c().await.is_ok() {
+            let _ = tx_ctrl_c.send(()).await;
+        }
+    });
+
+    // Listen for SIGTERM
+    tokio::spawn(async move {
+        if let Ok(mut term_signal) = signal(SignalKind::terminate()) {
+            while term_signal.recv().await.is_some() {
```
yaziciahmet: why loop here?

yaziciahmet (changes requested):

crates/common/src/utils.rs
ercecan: Done

crates/common/src/utils.rs
```diff
@@ -77,3 +80,26 @@ pub fn check_l2_range_exists<DB: SharedLedgerOps>(
...
+            }
+            _ = async {
+                if let Some(ref mut term_signal) = term_signal {
+                    term_signal.recv().await;
+                }
+            } => {
+                let _ = tx.send(()).await;
+            }
+        }
+    });
```
yaziciahmet:
```suggestion
    tokio::spawn(async move {
        let term_signal = signal(SignalKind::terminate()).ok();

        if let Some(mut term_signal) = term_signal {
            tokio::select! {
                _ = signal::ctrl_c() => {
                    let _ = tx.send(()).await;
                }
                _ = term_signal.recv() => {
                    let _ = tx.send(()).await;
                }
            }
        } else {
            let _ = signal::ctrl_c().await;
            let _ = tx.send(()).await;
        }
    });
```

if term_signal is None, this will immediately send signal to shutdown as the async block will finish executing. need to change it as above

yaziciahmet (changes requested):
Last fix and finito

yaziciahmet (approved):

PR #1561, opened 2024-12-05
Suggest 100 wei constant as tip in transactions

### LINKED ISSUE #1486: Revisit priority fee RPC

yaziciahmet:
# Description

As a rollup, tips do not make much sense for us, as we will have congestions very rarely, and we already have dynamic basefee algorithm in place. After discussions with @eyusufatik , we decided to suggest 100 wei as tip always.

First we thought about returning 0, but then we would lose the ability to order EIP-1559 transactions completely since there won't be any chance to speed up in rare times of congestion. Hence, we settled on 100 wei, which should almost make no difference to price in regular times, but also should allow people to bump the tips by 10% in case of congestion.

## Linked Issues
- Fixes #1486 

## Testing

## Docs

### DIFF
--- a/crates/ethereum-rpc/src/ethereum.rs
+++ b/crates/ethereum-rpc/src/ethereum.rs
@@ -87,8 +87,6 @@ impl<C: sov_modules_api::Context, Da: DaService> Ethereum<C, Da> {
 
     #[instrument(level = "trace", skip_all)]
     pub(crate) fn max_fee_per_gas(&self, working_set: &mut WorkingSet<C>) -> (U256, U256) {
-        // let suggested_tip = self.gas_price_oracle.suggest_tip_cap(working_set).unwrap();
-
         let evm = Evm::<C>::default();
         let base_fee = evm
             .get_block_by_number(None, None, working_set)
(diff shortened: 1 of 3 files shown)


crates/ethereum-rpc/src/ethereum.rs
```diff
@@ -86,7 +87,7 @@ impl<C: sov_modules_api::Context, Da: DaService> Ethereum<C, Da> {
 
     #[instrument(level = "trace", skip_all)]
     pub(crate) fn max_fee_per_gas(&self, working_set: &mut WorkingSet<C>) -> (U256, U256) {
-        let suggested_tip = self.gas_price_oracle.suggest_tip_cap(working_set).unwrap();
+        // let suggested_tip = self.gas_price_oracle.suggest_tip_cap(working_set).unwrap();
```
rakanalh: Remove if we don't need this anymore.

rakanalh (approved):

crates/ethereum-rpc/src/gas_price/cache.rs
```diff
@@ -21,6 +21,7 @@ impl<C: sov_modules_api::Context> BlockCache<C> {
     }
 
     /// Gets block from cache or from provider
+    #[allow(unused)]
```
kpp, line 24: Why is that?

yaziciahmet: It is not used, but I think it is too early to remove unused parts related to gas price oracle. We should keep them temporarily until we see that current suggestion mechanism works fine.

eyusufatik (approved):

PR #1562, opened 2024-12-05
Remove hierarchical storage manager

eyusufatik:
Removes hierchical storage manager. so we get a cleaner stf blueprint usage. (and many more changes that make so much sense)


crates/sovereign-sdk/module-system/sov-modules-core/tests/working_set_tests.rs
```diff
@@ -18,14 +12,14 @@ fn test_workingset_get() {
...
-    let mut working_set = WorkingSet::<DefaultContext>::new(storage.clone());
+    let mut working_set = WorkingSet::new(storage.clone());
     working_set.set(&storage_key, storage_value.clone());
 
     assert_eq!(Some(storage_value), working_set.get(&storage_key));
 }
 
 #[test]
-fn test_versioned_workingset_get() {
+fn test_workingset_get2() {
```
yaziciahmet: exact same test with test_workingset_get()

crates/sovereign-sdk/module-system/sov-modules-api/src/lib.rs
```diff
@@ -213,7 +212,7 @@ pub mod da {
 }
 
 pub mod storage {
```
yaziciahmet: Completely removable

crates/sovereign-sdk/full-node/sov-prover-storage-manager/src/lib.rs
yaziciahmet, line 1: Remove `HierarchicalStorageManager` from the doc of ProverStorageManager

crates/evm/src/query.rs
```diff
@@ -1759,7 +1768,7 @@ fn update_estimated_gas_range(
 #[inline]
 fn set_state_to_end_of_evm_block<C: sov_modules_api::Context>(
```
yaziciahmet, line 1760: generic can be storage instead of context here

crates/evm/src/lib.rs
```diff
@@ -172,15 +172,19 @@ impl<C: sov_modules_api::Context> sov_modules_api::Module for Evm<C> {
 
     type Event = ();
 
-    fn genesis(&self, config: &Self::Config, working_set: &mut WorkingSet<C>) {
+    fn genesis(
+        &self,
+        config: &Self::Config,
+        working_set: &mut WorkingSet<<Self::Context as Spec>::Storage>,
```
yaziciahmet: just use C as Spec

crates/evm/src/lib.rs
```diff
@@ -172,15 +172,19 @@ impl<C: sov_modules_api::Context> sov_modules_api::Module for Evm<C> {
...
+    ) {
         self.init_module(config, working_set)
     }
 
     fn call(
         &mut self,
         msg: Self::CallMessage,
         context: &Self::Context,
-        working_set: &mut WorkingSet<C>,
+        working_set: &mut WorkingSet<<Self::Context as Spec>::Storage>,
```
yaziciahmet: C as Spec

crates/citrea-stf/src/hooks_impl.rs
```diff
@@ -61,7 +61,7 @@ impl<C: Context, Da: DaSpec> ApplySoftConfirmationHooks<Da> for Runtime<C, Da> {
     fn begin_soft_confirmation_hook(
         &mut self,
         soft_confirmation_info: &HookSoftConfirmationInfo,
-        working_set: &mut WorkingSet<Self::Context>,
+        working_set: &mut WorkingSet<<Self::Context as Spec>::Storage>,
```
yaziciahmet: C as Spec

bin/citrea/src/main.rs
```diff
@@ -241,13 +249,16 @@ where
...
-        let rollup = CitreaRollupBlueprint::create_new_rollup(
+        let (mut rollup, rpc_methods) = CitreaRollupBlueprint::create_new_rollup(
             &rollup_blueprint,
             rt_genesis_paths,
             rollup_config,
         )
         .await
         .expect("Could not start full-node");
+
+        rollup.start_rpc_server(rpc_methods, None).await;
```
yaziciahmet, line 260: Is it on purpose that we don't return errors and panic from FullNode::start_rpc_server?

yaziciahmet (changes requested):
Have only nits. Great cleanup PR.

crates/evm/src/query.rs
eyusufatik: yes that's practically true for everywhere. BUT that's so much change everywhere, I don't know if I should continue :D 

basically %90 places where we pass Context, should be Storage actually

bin/citrea/src/main.rs
eyusufatik: tbh, I don't even know

yaziciahmet (approved):

PR #1576, opened 2024-12-09
Devnet guests

### LINKED ISSUE #1575: Add devnet as a `--network` arg
Use testnet's guest code for Devnet

rakanalh:
# Description

Upon @eyusufatik 's request.

## Linked Issues
- Fixes #1575


bin/citrea/src/main.rs
```diff
@@ -35,6 +35,10 @@ struct Args {
     #[arg(long)]
     testnet: bool,
 
+    /// Override network to run devnet directly.
+    #[arg(long)]
+    devnet: bool,
+
     /// Run the development chain
     #[arg(long, conflicts_with_all = ["testnet"])]
     dev: bool,
```
eyusufatik, line 36: should we get rid of this now? or rename it to something like local-testing etc

rakanalh: I don't think it's necessary. The development version vs devnet is known to blockchain savvy people.

IMO, If any update is to be made, is to remove `--devnet` and `--testnet` shortcuts and explicity use `--network testnet` or `--network devnet` but `--dev` is very convenient during development.

eyusufatik: well actually that makes sense. --network mainnet/testnet/devnet (default to mainnet)

eyusufatik (approved):

PR #1579, opened 2024-12-09
Remove eip4844 calculations

ercecan:
# Description
Removes blob gas calculations since  we will not be accepting eip4844 transactions

### DIFF
--- a/crates/evm/src/evm/executor.rs
+++ b/crates/evm/src/evm/executor.rs
@@ -1,4 +1,3 @@
-use alloy_eips::eip4844::MAX_DATA_GAS_PER_BLOCK;
 use reth_primitives::TransactionSignedEcRecovered;
 use revm::primitives::{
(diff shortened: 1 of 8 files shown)


crates/evm/src/evm/executor.rs
eyusufatik, line 1: blob_gas_used: &mut u64, argument of function execute_multiple_tx should be removed

eyusufatik (changes requested):
```rust
    #[memory]
    pub(crate) blob_gas_used: u64,
```


should also be removed

eyusufatik (changes requested):
in query.rs 

`get_pending_block_env` function, you can simply the blob related field calculations to be Some(0) if cancun is activated and None if not

eyusufatik (approved):

yaziciahmet (approved):

PR #1582, opened 2024-12-09
Guest Dockerfile for reproducible builds

### LINKED ISSUE #1565: In repo dockerfile for reproducible builds
looks like it's going to take a while for compile time env vars to get support in risc0 docker builds.

we can put dockerfiles in the repo and just use them.

we should do this urgently to not block #1444

yaziciahmet:
# Description

Added a Dockerfile for guest reproducible builds. Can be easily invoked via a Makefile script. 

Added 2 make scripts to build bitcoin guests to `guests/risc0/Makefile`.
- ```make batch-proof-bitcoin-docker OUT_PATH=<elf_path>```
- ```make light-client-bitcoin-docker OUT_PATH=<elf_path>```

If OUT_PATH is not specified, `resources/guests/risc0/<guest_name>` is used as default path.

Added 1 make script to `Makefile`.
- ```make build-risc0-docker BATCH_OUT_PATH=<batch_elf_path> LIGHT_OUT_PATH=<light_elf_path>```

If BATCH_OUT_PATH or LIGHT_OUT_PATH is not specified, same defaults are used.

Also added a dummy env passing example to Dockerfile for future reference.

## Linked Issues
- Fixes #1565

## Testing

- Ran `make build-risc0` and `make build-risc0-docker` and verified that elfs match.
- Added a dummy compile-time env to guest, ran `make build-risc0-docker` with and without it, and verified that elfs differ only by 1 line, which is the compile-time constant value.

## Docs


resources/guests/risc0/docker/batch.elf
eyusufatik, line 1: we won't be putting elfs in a `docker/` directory. once we know we are ready for pinning a fork's elf, we are just going to put it under `resources/guests/risc0/` so delete these elf files pls

guests/risc0/Dockerfile
eyusufatik, line 1: can you add how to pass env vars to the docker file?

so we know how to use them

eyusufatik (changes requested):

eyusufatik (approved):

PR #1587, opened 2024-12-11
Remove StfBlueprintTrait

### LINKED ISSUE #1513: Remove StfBlueprintTrait

jfldde:
# Description

- Removes `StfBlueprintTrait` and implement methods directly on `StfBlueprint`
- Passes concrete `StfBlueprint` type to `CitreaFullnode` and `CitreaSequencer`.

## Linked Issues
- Fixes #1513


crates/fullnode/src/runner.rs
eyusufatik, line 1: maybe we should go the batch prover way in full node?

eyusufatik: wdyt

jfldde: In what sense ?

jfldde: Fixed in https://github.com/chainwayxyz/citrea/pull/1587/commits/9e4552a50da1c0a992d8f622502ba02d4bcfb565

eyusufatik (approved):

PR #1589, opened 2024-12-11
Prometheus metrics

### LINKED ISSUE #1293: Profiling / metrics stuff for all nodes — enable in testnet

rakanalh:
# Description

**IMPORTANT**: For the record, we have to use the exact same version for `metrics` crate as Reth's because different crates will create different global recorders (We won't be able to collect reth's metrics in this case).

From the referenced issue, the following was the list of metrics to collect:

### All
- [ ] Storage size ( I think this can be fed into prometheus outside the scope of citrea)
- [ ] RPC served in last 5 min (Same as this one, nginx?)
### Sequencer
- [x] Current number txs in mempool
- [x] Mempool inbound tx/s (Can be calculated from the current number)
- [x] dry_run_transactions execution time
- [x] produce_block transactions execution time
- [ ] (this one will be hard to track so if we cant no problem) avg. wait time for tx in mempool (Can be calculated i believe)
- [x] Send sequencer commitment execution time
- [x] Sequencer commitment number of blocks
- [x] Current L1 block number used in block production
### Full node
- [ ] Sync progress (Not applicable since we have to know sequencer latest L2 vs fullnode or prover current L2)
- [x] Current L2 block number
- [x] Current L1 block number
- [ ] Soft confirmation pull rate (important while syncing) 
- [x] Soft confirmaiton process durations
- [x] L1 scan duration per L1 block
### Prover(s)
- [x] Current L2 block number
- [x] Current L1 blcok number
- [x] Cycle count for each proof
- [x] DA tx mining time

# TODO
- [x] Implement / Export graphana dashboards to display these metrics.

## Linked Issues
- Fixes #1293


resources/configs/devnet/rollup_config.toml
eyusufatik, line 1: let's not put any config here. this is considered to be freezed

crates/light-client-prover/src/metrics.rs
```diff
@@ -0,0 +1,16 @@
...
+pub struct ProverMetrics {
+    #[metric(describe = "The current L1 block number which is used to produce L2 blocks")]
+    pub current_l1_block: Gauge,
+}
+
+/// Light client metrics
+pub static PROVER_METRICS: Lazy<ProverMetrics> = Lazy::new(|| {
+    ProverMetrics::describe();
+    ProverMetrics::default()
+});
```
eyusufatik: rename this and batch prover metrics different names pls

eyusufatik (changes requested):
apart from the comments, do we also get any metric recorded by `reth-transaction-pool`?

rakanalh:
> apart from the comments, do we also get any metric recorded by `reth-transaction-pool`?

Yes. Metrics from our own code & reth will self register and be included in telemetry information we provide for Prometheus.

bin/citrea/src/main.rs
```diff
@@ -201,6 +205,25 @@ where
...
+    )
+    .parse()
+    .map_err(|_| anyhow!("Invalid telemetry address"))?;
+
+    debug!("Starting telemetry server on: {}", telemetry_addr);
+
+    let builder = PrometheusBuilder::new().with_http_listener(telemetry_addr);
+    builder
+        .idle_timeout(
+            MetricKindMask::GAUGE | MetricKindMask::HISTOGRAM,
```
eyusufatik: not everyone's going to run telemetry. let's make it optional in the config. if telemetry config is given, we start the server, if not, we don't.

eyusufatik (changes requested):

eyusufatik (approved):

PR #1597, opened 2024-12-12
make db commit better

eyusufatik:
# Description
No unnecessary moves

Better (and maybe even fixed) code set


eyusufatik:
addin @rakanalh as a reviewer manually because he implemented offchain code

crates/evm/src/evm/db_commit.rs
```diff
@@ -55,22 +51,35 @@ impl<'a, C: sov_modules_api::Context> DatabaseCommit for EvmDb<'a, C> {
 
             if let Some(ref code) = account_info.code {
                 if !code.is_empty() {
-                    let exists_in_db = self
```
eyusufatik, line 58: even after Fork1 we were checking if code exists in self.code, which means an extra nonexistence proof. which is bad as you can guess in zk :)

rakanalh (approved):
LGTM

ercecan (approved):

Issue #1596, opened 2024-12-11
Remove sequencer-client crate


jfldde:
Can be replaced by a simple `HttpClient` with `SequencerRpcClient` trait or `LedgerRpcClient` trait in scope as needed.

`GetSoftConfirmationResponse` removal is a plus as it is a duplicate of SoftConfirmationResponse and error prone

PR #1598, opened 2024-12-12
Derive rpc(server) for LedgerRpc

### LINKED ISSUE #1568: Use #[rpc(server)] in LedgerRpc for stricter typing

jfldde:
# Description

Rationale for this PR was some type mismatch between loosely typed LedgerRpcServer not in line with the types declared in the LedgeRrpc trait. Caught in https://github.com/chainwayxyz/citrea/pull/1558 using get_l2_genesis_state_root in https://github.com/chainwayxyz/citrea/pull/1558/commits/ff9ba26dfcc6ef4150ec8392f2b77be6743f15b0

- Derive rpc(server) for LedgerRpc
- Cleanup unused generics
- Use LedgerRpcClient across and fix types mismatch.

## Linked Issues
- Fixes #1568 
- Pre-req for https://github.com/chainwayxyz/citrea/issues/1596


crates/sequencer-client/src/lib.rs
```diff
@@ -59,7 +59,7 @@ impl SequencerClient {
             .client
             .request(
                 "ledger_getSoftConfirmationRange",
-                rpc_params![range.start(), range.end()],
+                rpc_params![(range.start(), range.end())],
```
jfldde: Not addressed here as this whole crate will be removed by #1596

eyusufatik: just curious, does this still work?

```rust
rpc_params![range.start(), range.end()]
```

or do we have to make it a tuple?

because if not, we should make sure v0.5.5 can still request ranges from us.

if they can't, every node will need to be updated at the same time lol. since this is a testnet it's fine however, can be problematic.

eyusufatik (approved):
looking good

eyusufatik:
idk why comment was both a reply and a comment

crates/sequencer-client/src/lib.rs
jfldde: Good catch, I'll restore it so that it does not break backward compatibility

ercecan (approved):
clean 🙏

Issue #1464, opened 2024-11-11
Fork1 activation tests


eyusufatik:

ercecan:
- [x] Test dencun activation, tstore tload operations, test before and after fork to see if it works as it is supposed to
- [x] Activate dencun ef tests and see which are possible to fix/adjust to our rules

eyusufatik:
> Activate dencun ef tests and see which are possible to fix/adjust to our rules

this is represented in #1463 

see here for different things activated in dencun:

https://wazirx.com/blog/a-breakdown-of-eips-included-in-the-dencun-upgrade-2024/

(not all might apply to us)

and then write here what tests need to be written

ercecan:
According to eips in dencun taken from [here](https://wazirx.com/blog/a-breakdown-of-eips-included-in-the-dencun-upgrade-2024/)

- ~~1 EIP-7514: “The Gatekeeper”~~
- [x] 2 EIP-1153: “The Memory Magician” Test this => transient storage
- ~~3 EIP-4788: “The Bridge Builder”~~
- ~~4 EIP-4844: “The Luggage Carrier”~~
- [x] 5 EIP-6780: “The Code Keeper” Test this => this proposal restricts SELFDESTRUCT usage to the creating contract
- [x] 6 EIP-5656: “The Data Dynamo” Test this => introduces the MCOPY
- ~~7 EIP-7044: “The Exit Maestro”~~
- ~~8 EIP-7045: “The Time Extender”~~
- ~~9 EIP-7516: “The Price Scanner”~~

eyusufatik:
For 9, unfortunately we cant change the behaviour of revm crate, so that opcode will return 1. That's also what optimism does so imo we are ok.

eyusufatik:
- [x] (implemented in e2e rn) test genesis -> fork1 self.code -> self.offchain_code transition
- [x] test if call to point eval precompile reverts

eyusufatik:
I think there is no more tests we should write. except for maybe prover switch ELFs test but I'm going to create an issue for that.

PR #1604, opened 2024-12-12
Fork1 tests v2

ercecan:
Tests these => https://github.com/chainwayxyz/citrea/issues/1464#issuecomment-2537022638


crates/evm/src/evm/test_data/KZGPointEvaluation.sol
eyusufatik, line 1: rename this to something like KZGPointEvaluationCaller

eyusufatik: and also other related structs.

crates/evm/src/tests/fork_tests.rs
```diff
@@ -535,3 +556,239 @@ fn test_blob_base_fee_should_return_1() {
...
+        da_slot_txs_commitment: [42u8; 32],
+        pre_state_root: [10u8; 32].to_vec(),
+        current_spec: SovSpecId::Fork1,
+        pub_key: vec![],
+        deposit_data: vec![],
+        l1_fee_rate,
+        timestamp: 0,
+    };
+
+    // Deploy transient storage contract
```
eyusufatik: fix comment

crates/evm/src/tests/fork_tests.rs
```diff
@@ -535,3 +556,239 @@ fn test_blob_base_fee_should_return_1() {
...
+        da_slot_txs_commitment: [42u8; 32],
+        pre_state_root: [10u8; 32].to_vec(),
+        current_spec: SovSpecId::Genesis,
+        pub_key: vec![],
+        deposit_data: vec![],
+        l1_fee_rate,
+        timestamp: 0,
+    };
+
+    // Deploy transient storage contract
```
eyusufatik: comment

crates/evm/src/tests/fork_tests.rs
```diff
@@ -535,3 +556,239 @@ fn test_blob_base_fee_should_return_1() {
...
+    evm.finalize_hook(&[99u8; 32].into(), &mut working_set.accessory_state());
+    sleep(std::time::Duration::from_secs(2));
+    l2_height += 1;
+
+    let offchain_code = evm
+        .offchain_code
+        .get(&code_hash, &mut working_set.offchain_state());
+
+    assert!(offchain_code.is_none());
+
```
eyusufatik, line 754: make tx on the contract that was deployed before fork1 and see that you can read it from offchain storage afterwards

eyusufatik (changes requested):

rakanalh (approved):

eyusufatik:
holding on to this pr until we get our upgrade prs merged. because disabling point evaluation precompile needs some changes unfortunately.

crates/evm/src/evm/handler.rs
```diff
@@ -314,6 +314,29 @@ struct CitreaHandler<SPEC, EXT, DB> {
...
+
+            pub fn kzg(_input: &Bytes, _gas_limit: u64, _env: &Env) -> PrecompileResult {
+                pub const GAS_COST: u64 = 0;
+                pub const RETURN_VALUE: &[u8; 64] = &[0; 64];
+                Ok(PrecompileOutput::new(GAS_COST, RETURN_VALUE.into()))
+            }
+
+            let mut precompiles = revm::handler::mainnet::load_precompiles::<SPEC, DB>();
+            let precompiles_inner = precompiles.to_mut();
+            if let Some(kzg_point_evaluation) = precompiles_inner.get_mut(&u64_to_address(0x0A)) {
```
eyusufatik: since we know at what point kzg comes into play, we can remove only then with the index :)

eyusufatik (changes requested):

crates/evm/src/evm/handler.rs
eyusufatik: also the costs should not be zero. it should be the same as call gas cost to a regular address with no precompile

eyusufatik (approved):

PR #1607, opened 2024-12-13
make testnet4 reorg depth 30

eyusufatik:
# Description
ifykyk


rakanalh (approved):

ercecan (approved):

crates/batch-prover/src/runner.rs
```diff
@@ -354,6 +354,7 @@ where
                                 // This block failed to process, add remaining L2 blocks to queue including this one.
                                 let remaining_l2s = l2_blocks[index..].to_vec();
                                 pending_l2_blocks.extend(remaining_l2s);
+                                break;
```
kpp, line 357: It will stop syncing in case of error. Is it ok?

eyusufatik: no that's just the for above

kpp (approved):

PR #1611, opened 2024-12-16
Improve light client initialization

rakanalh:
# Description


jfldde (approved):

crates/light-client-prover/src/runner.rs
```diff
@@ -186,7 +149,7 @@ where
         let last_l1_height_scanned = match self.ledger_db.get_last_scanned_l1_height()? {
             Some(l1_height) => l1_height,
             // If not found, start from the first L2 block's L1 height
-            None => SlotNumber(get_initial_da_height(&self.sequencer_client).await),
+            None => SlotNumber(get_initial_slot_height(&self.sequencer_client).await),
```
eyusufatik: tbh now that I think about it, it's not a good idea to pull the init height like this :D

I've thought about how we are going to enable light client proofs on testnet (we didn't prove everything from the beginning) and it looks like we need to put this information in the light client prover config

eyusufatik (changes requested):

eyusufatik (approved):

PR #1612, opened 2024-12-16
Fix ci tests - metrics related

rakanalh:
# Description

Web3py and uniswap tests are failing because fullnode doesn't start. The failure to start is because `Address already in use` since they both use the same port (8001) by default. This fixes it by specifying different ports for each node.

Although this changes config files which are "frozen" based on @eyusufatik 's recommendation. But for CI to pass, these files HAVE to be changed.

## Linked Issues
- Related to #1293

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,7 @@ clean: ## Cleans compiled
 	@cargo clean
 
 clean-node: ## Cleans local dbs needed for sequencer and nodes
+	rm -rf resources/dbs/da-db
 	rm -rf resources/dbs/sequencer-db
 	rm -rf resources/dbs/batch-prover-db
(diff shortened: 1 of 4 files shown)


eyusufatik:
> Although this changes config files which are "frozen" based on @eyusufatik 's recommendation. But for CI to pass, these files HAVE to be changed.

changing mock toml's is fine  

> Web3py and uniswap tests are failing because fullnode doesn't start.

how? I thought since we didnt put telemetry config in those tomls, the telemetry server shouldn't have started?

rakanalh:
You're right. I now just disable telemetry by default.

eyusufatik (approved):

PR #1613, opened 2024-12-16
Set ZK constants in compile time

### LINKED ISSUE #672: MockDA: Separate the sender addresses of sequencer and prover
Currently in mockda the sender addresses of sequencer and prover are the same, this causes some unnecessary checks when parsing da data and creates confusion

### LINKED ISSUE #1371: E2E flaky: prover_skips_preproven_commitments_test
## Summary

```
Error: called `Result::unwrap()` on an `Err` value: FullNode failed to get zkproofs within the specified timeout
```

https://github.com/chainwayxyz/citrea/actions/runs/11381368018/job/31662616665#step:11:716

### LINKED ISSUE #1444: Take some of the inputs as constants in ZK circuits
https://github.com/chainwayxyz/citrea/issues/1566#issuecomment-2527984758

see this comment please.

Things that are mentioned in that comment are inputs to the ZK circuits and they are always outputted so that the next upper level of the verification pipeline can verify inputs were passed correctly or the said proofs are produced for the chain. (cant pass testnet proof in mainnet for example)

we want to make these constants in the guest codes itself and remove them from the outputs.

### LINKED ISSUE #1626: Remove Simulate proof mode

### LINKED ISSUE #1635: Build pipeline -- light client depends on batch proof method id

yaziciahmet:
# Description

Made certain variables constant in zk. Only environment variable needed is `CITREA_NETWORK`, and all the constant values are decided based on the network env value. Valid `CITREA_NETWORK` values are:
- mainnet
- testnet
- devnet
- nightly

If no `CITREA_NETWORK` env variable is found, it defaults to nightly. And to make things more testable, if the network is nightly, the exact values for each constant value can be provided through env variable.

Batch Proof:
- FORKS
- SEQUENCER_PUBLIC_KEY
- SEQUENCER_DA_PUBLIC_KEY

Light Client Proof:
- L2_GENESIS_ROOT
- BATCH_PROOF_METHOD_ID
- BATCH_PROVER_DA_PUBLIC_KEY

TODO:
- [x] Make needed variables constant in guest
- [x] Update circuit methods to take constants as arguments
- [x] Separate network fork lists and use one of them based on the provided network (through CLI in native, and through env in guest)
- [x] Update guest Dockerfile & Makefile
- [x] Update release CI files

## Linked Issues
- Fixes #1444 
- Fixes #672 
- Fixes #1626 
- Fixes #1371 
- Fixes #1635 

## Testing

## Docs


guests/risc0/batch-proof-bitcoin/src/bin/batch_proof_bitcoin.rs
```diff
@@ -6,27 +6,78 @@ use citrea_risc0_adapter::guest::Risc0Guest;
...
+            if count == 0 {
+                panic!("FORKS can not be empty");
+            }
+            Some((forks, count))
+        }
+        None => panic!("FORKS must be valid comma separated list"),
+    },
+    // TODO: what to do here?
+    None => None,
+};
```
eyusufatik: make these env!

not option_env!

guests/risc0/batch-proof-bitcoin/Cargo.toml
```diff
@@ -15,6 +15,7 @@ bitcoin-da = { path = "../../../crates/bitcoin-da", default-features = false }
 citrea-primitives = { path = "../../../crates/primitives" }
 citrea-risc0-adapter = { path = "../../../crates/risc0" }
 citrea-stf = { path = "../../../crates/citrea-stf" }
+const-hex = "1"
```
eyusufatik: go with a specific version

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
```diff
@@ -173,8 +173,10 @@ pub struct BatchProofCircuitInput<'txs, StateRoot, Witness, Da: DaSpec, Tx: Clon
     /// DA block headers the soft confirmations was constructed on.
     pub da_block_headers_of_soft_confirmations: VecDeque<Vec<Da::BlockHeader>>,
     /// Sequencer soft confirmation public key.
+    /// **DO NOT USE THIS FIELD**
```
eyusufatik: do not use in post fork 1 guests*

eyusufatik (changes requested):

eyusufatik:
> If using VSCode, set the following key-value into your settings.json:

you can put a setting.json under .vscode under this pr I think.

Also create a test_builds.env file unders guests/ so that people can see the env vars in the repo

crates/citrea-stf/src/verifier.rs
```diff
@@ -1,45 +1,34 @@
...
-            phantom: Default::default(),
-        }
+        Self { app, da_verifier }
     }
 
     /// Verify the next block
     pub fn run_sequencer_commitments_in_da_slot(
         &mut self,
-        zkvm: Zk,
+        data: BatchProofCircuitInput<Stf::StateRoot, Stf::Witness, Da::Spec, Stf::Transaction>,
```
kpp, line 29: These are very nice and neat refactoring changes I would love to see in a separate PR.

kpp (dismissed):
option_env!, please.

crates/citrea-stf/src/verifier.rs
yaziciahmet: The reason I did this is I will be changing that trait to take all the env variables into this method's args as well. So I thought while doing a change to the method signature, let's change this one as well.

crates/sovereign-sdk/rollup-interface/src/spec.rs
```diff
@@ -34,6 +34,18 @@ mod spec {
         /// 3. Don't use borsh when signing SoftConfirmation's
         Fork1 = 1,
     }
+
+    impl SpecId {
+        /// Const fn to convert u8 to corresponding SpecId. Valid values are
+        /// 0 and 1. Panics otherwise.
```
jfldde:
```suggestion
        /// 0 and 1. Returns None otherwise.
```

crates/sovereign-sdk/rollup-interface/src/spec.rs
```diff
@@ -66,4 +78,18 @@ mod spec {
         /// Third fork
         Fork3 = 3,
     }
+
```
jfldde: Might be better as a single

```
pub enum SpecId {
    /// Genesis spec
    #[default]
    Genesis = 0,
    Fork1 = 1,
    #[cfg(feature = "testing")]
    /// Second fork
    Fork2 = 2,
    #[cfg(feature = "testing")]
    /// Third fork
    Fork3 = 3,
}

impl SpecId {
    /// Const fn to convert u8 to corresponding SpecId.
    pub const fn from_u8(n: u8) -> Option<SpecId> {
        match n {
            0 => Some(SpecId::Genesis),
            1 => Some(SpecId::Fork1),
            #[cfg(feature = "testing")]
            2 => Some(SpecId::Fork2),
            #[cfg(feature = "testing")]
            3 => Some(SpecId::Fork3),
            _ => None,
        }
    }
}
```

eyusufatik:
> option_env!, please.

see my above comments for not using option_env!

kpp:
> see my above comments for not using option_env!

Yes, I can see you requested to make it env instead of option_env, but I don't see why.

```
    "rust-analyzer.cargo.extraEnv": {
         "SEQUENCER_PUBLIC_KEY": "0000000000000000000000000000000000000000000000000000000000000000",
         "SEQUENCER_DA_PUBLIC_KEY": "000000000000000000000000000000000000000000000000000000000000000000",
         "FORKS": "0:0",
         "L2_GENESIS_ROOT": "0000000000000000000000000000000000000000000000000000000000000000",
         "BATCH_PROOF_METHOD_ID": "0000000000000000000000000000000000000000000000000000000000000000",
         "BATCH_PROVER_DA_PUBLIC_KEY": "000000000000000000000000000000000000000000000000000000000000000000",
    },
```

If these can be default values for vscode, they can be default values in case option_env didn't find env variable. I predict we will have an outrage once it stops compiling out of the box without having to feed the shell/IDE environment with a correct list of env variables. Looks like a big no-no.

yaziciahmet:
It might be annoying to have it as env, but we can get around it with ide options and makefile. It would be a CRUCIAL problem if we were to build the guest without env variables at any point of time.

IMO it is more important to secure the prod builds, and find other ways to get around the problem during development.

crates/sovereign-sdk/rollup-interface/src/fork/mod.rs
```diff
@@ -27,4 +186,153 @@ impl Fork {
...
+
+        Some(Fork {
+            spec_id,
+            activation_height: height,
+        })
+    }
+}
+
+/// Verifies the order of forks. `size` is needed here due to being in const environment,
+/// size of the fork might not be known beforehand.
```
jfldde: Shouldn't we know the size of fork at compile time and have stricter/simpler assertions using arrays instead ?

yaziciahmet: I first thought of having a CITREA_ENV environment variable, in which all the constant values decide their value on, which would allow us to hardcode the values for testnet, mainnet, local development etc. into the binary. This is also needed to be able to hardcode the forks as in your suggestion since forks will be environment dependent.

But after discussions with @eyusufatik, it seemed like having all these variables independently would give us maximum flexibility, which would allow us to see whether this flexibility could potentially be useful for some cases. I don't know if this is better, since too much flexibility is confusing and more complex to manage, but IMO it might be good to start like this, and over time we can narrow the usage down.

guests/risc0/light-client-proof-bitcoin/src/bin/light_client_proof_bitcoin.rs
```diff
@@ -6,9 +6,80 @@ use citrea_primitives::{TO_BATCH_PROOF_PREFIX, TO_LIGHT_CLIENT_PREFIX};
...
+    match const_hex::const_decode_to_array::<32>(hex_method_id.as_bytes()) {
+        Ok(method_id) => constmuck::cast(method_id),
+        Err(_) => panic!("BATCH_PROOF_METHOD_ID must be valid 32-byte hex string"),
+    }
+};
+
+const BATCH_PROVER_DA_PUBLIC_KEY: [u8; 33] = {
+    let hex_pub_key = match NETWORK {
+        // TODO: Update mainnet pub key
+        Network::Mainnet => "0357d255ab93638a2d880787ebaadfefdfc9bb51a26b4a37e5d588e04e54c60a42",
```
eyusufatik: laeve as 000...0 pls

crates/prover-services/src/parallel/mod.rs
eyusufatik, line 1: all the changes in this pr is related to simulate mode

is that ever used in the repo? not even in testing I assume. we can remove it if you wish

crates/common/src/config.rs
```diff
@@ -193,8 +193,8 @@ impl FromEnv for RollupPublicKeys {
     fn from_env() -> anyhow::Result<Self> {
         Ok(Self {
             sequencer_public_key: hex::decode(std::env::var("SEQUENCER_PUBLIC_KEY")?)?,
-            sequencer_da_pub_key: hex::decode(std::env::var("SEQUENCER_DA_PUB_KEY")?)?,
-            prover_da_pub_key: hex::decode(std::env::var("PROVER_DA_PUB_KEY")?)?,
+            sequencer_da_pub_key: hex::decode(std::env::var("SEQUENCER_DA_PUBLIC_KEY")?)?,
```
eyusufatik: don't change this one because they are the same with config field name

eyusufatik (changes requested):

crates/prover-services/src/parallel/mod.rs
yaziciahmet: not at all. lets remove it.

crates/sovereign-sdk/rollup-interface/src/network.rs
```diff
@@ -0,0 +1,56 @@
...
+        } else if const_compare_str(s, "testnet") {
+            Some(Network::Testnet)
+        } else if const_compare_str(s, "devnet") {
+            Some(Network::Devnet)
+        } else if const_compare_str(s, "nightly") {
+            Some(Network::Nightly)
+        } else {
+            None
+        }
+    }
```
jfldde:
```suggestion
    pub const fn const_from_str(s: &str) -> Option<Network> {
        match s.as_bytes() {
            b"mainnet" => Some(Network::Mainnet),
            b"testnet" => Some(Network::Testnet),
            b"devnet" => Some(Network::Devnet),
            b"nightly" => Some(Network::Nightly),
            _ => None,
        }
    }
```

yaziciahmet: thank you very much ser

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
```diff
@@ -173,8 +173,10 @@ pub struct BatchProofCircuitInput<'txs, StateRoot, Witness, Da: DaSpec, Tx: Clon
     /// DA block headers the soft confirmations was constructed on.
     pub da_block_headers_of_soft_confirmations: VecDeque<Vec<Da::BlockHeader>>,
     /// Sequencer soft confirmation public key.
+    /// **DO NOT USE THIS FIELD IN POST FORK1 GUEST**
```
rakanalh, line 176: Use this:
```suggestion
    /// **DO NOT USE THIS FIELD IN POST FORK1 GUEST**
    #[deprecated]
```

crates/sovereign-sdk/adapters/mock-da/src/service.rs
```diff
@@ -221,7 +221,7 @@ impl MockDaService {
         let blob = MockBlob::new_with_zkp_proof(
             blob.to_vec(),
             zkp_proof,
-            self.sequencer_da_address,
+            self.sequencer_da_address.clone(),
```
kpp, line 224: I don't quite understand why our addresses (mock/regular) turned from [u8; 32] into Vec<u8>.

yaziciahmet: compressed bitcoin da addresses are 33 bytes (first byte for parity), so 32 bytes was incorrect. but when I changed it to [u8; 33] all hell broke loose, hence, temporarily I just changed it to be a Vec, so current 32 byte addresses would work, and I can use 33 bytes wherever I need to.

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
yaziciahmet: We actually shouldnt because we are still using it until we are officially past Fork1.

But still great suggestion because this allowed me to see one part I missed: https://github.com/chainwayxyz/citrea/pull/1613/commits/b9a02174a2e4cba129ffff9b339b3377381f132c

.github/workflows/release.yml
```diff
@@ -40,7 +40,8 @@ jobs:
 
       - name: Build Project
         env:
-          REPR_GUEST_BUILD: 1
+          # TODO: make this dynamic somehow
```
eyusufatik: this does not matter for reasons explained on huddle, remove comment and SKIP_GUEST_BUILD in this build

crates/evm/src/tests/queries/estimate_gas_tests.rs
```diff
@@ -110,7 +110,7 @@ fn test_tx_request_fields_gas() {
     );
     assert_eq!(
         contract_diff_size.unwrap(),
-        serde_json::from_value::<EstimatedDiffSize>(json![{"gas":"0x6601","l1DiffSize":"0x60"}])
+        serde_json::from_value::<EstimatedDiffSize>(json![{"gas":"0x6601","l1DiffSize":"0x1f"}])
```
eyusufatik, line 113: we need identical test that tests genesis diff size vs fork1 diff size

eyusufatik (changes requested):
update all evm tests to use fork1 spec unless we test for specific behaviour going from genesis to fork1

guests/risc0/batch-proof/bitcoin/src/bin/batch_proof_bitcoin.rs
```diff
@@ -0,0 +1,94 @@
...
+        }
+    };
+
+    match const_hex::const_decode_to_array(hex_pub_key.as_bytes()) {
+        Ok(pub_key) => pub_key,
+        Err(_) => panic!("SEQUENCER_PUBLIC_KEY must be valid 32-byte hex string"),
+    }
+};
+
+const SEQUENCER_DA_PUBLIC_KEY: [u8; 33] = {
```
jfldde, line 48: Can use `PUBLIC_KEY_SIZE` from bitcoin (`pub const PUBLIC_KEY_SIZE: usize = 33;`)

crates/evm/src/tests/queries/estimate_gas_tests.rs
yaziciahmet: adjusted all evm tests accordingly. there is already a test for it: `test_l1_fee_compression_discount`

guests/risc0/batch-proof/bitcoin/src/bin/batch_proof_bitcoin.rs
yaziciahmet: I would have to add another crate for this. Even though it is already in lockfile, I don't think its worth just for this. wdyt?

eyusufatik (approved):

PR #1618, opened 2024-12-17
Verify chunked txs

### LINKED ISSUE #1591: Light client proof verify chunked txs
right now we don't do anything about chunked bitcoin transactions.

a solution for verifying batch proofs that were sent to DA in a chunk-aggregator set up would look like this:

1 - To `LightClientCircuitOutput` we add a HashMap of wtxid => data.
2 - In the circuit, whenever we see a chunk, we push to this HashMap.
3 - Since the HashMap acts as a light client state now, when we encounter an aggregator we search the map for the chunks. If we cant find any of them we discard the aggregator. (if the aggregator somehow lands on Bitcoin before the chunks, we can send the aggregator again)
4- If we have all the cunks, we concat verify and go on with our normal light client execution.

Further optimizations can be done regarding the size of this newly added light client circuit state, such as instead of having a mapping of wtxid => data, we merkleize the chunks and once we know we'll need the chunks, we provide the circuit with the chunks + their merkle proof.

rakanalh:
# Description

## Linked Issues
- Fixes #1591


crates/light-client-prover/src/circuit.rs
```diff
@@ -99,35 +105,52 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+                        // If we have all the chunks, perform verification
+                        if aggregate_tx_ids.is_subset(&existing_tx_ids) {
+                            // Concatenate complete proof
+                            let complete_proof = tx_ids
+                                .iter()
+                                .filter_map(|k| wtxid_data.get(k).cloned())
+                                .fold(vec![], |mut p, c| {
+                                    p.extend(c);
+                                    p
+                                });
```
jfldde: Small nit but it looks nicer with flatten imo
```suggestion
                            let complete_proof =  tx_ids
            .iter()
            .filter_map(|k| wtxid_data.get(k).cloned())
            .flatten()
            .collect::<Vec<_>>();
```

crates/light-client-prover/src/circuit.rs
```diff
@@ -99,35 +105,50 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
-                                batch_proof_output.initial_state_root,
-                                batch_proof_output.final_state_root,
-                                batch_proof_output.last_l2_height,
-                            ),
-                        );
                     }
-                    DaDataLightClient::Aggregate(_) => todo!(),
-                    DaDataLightClient::Chunk(_) => todo!(),
+                    DaDataLightClient::Aggregate(tx_ids) => {
+                        let existing_tx_ids: HashSet<[u8; 32]> =
```
eyusufatik: us btreeset if it exists :) 

they perform better in risc0

crates/light-client-prover/src/circuit.rs
```diff
@@ -99,35 +105,50 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+                                continue;
+                            }
+
+                            for tx_id in &aggregate_tx_ids {
+                                wtxid_data.remove(tx_id);
+                            }
+                        }
+                    }
+                    DaDataLightClient::Chunk(chunk) => {
+                        wtxid_data.insert(blob.hash(), chunk);
```
eyusufatik: unfortunately this blob.hash() is not the txs wtxid...


@kpp any solutions?

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
```diff
@@ -261,4 +263,6 @@ pub struct LightClientCircuitInput<Da: DaSpec> {
     pub previous_light_client_proof_journal: Option<Vec<u8>>,
     /// L2 Genesis state root
     pub l2_genesis_state_root: Option<[u8; 32]>,
+    /// A map from tx hash to chunk data
+    pub wtxid_data: BTreeMap<[u8; 32], Vec<u8>>,
```
eyusufatik: don't need additional field for this actually.

previous_light_client_proof_journal has this information ;)

eyusufatik: bc it's the borsh serialization of LightClientCircuitOutput

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -256,6 +256,7 @@ where
             light_client_proof_method_id: light_client_proof_code_commitment.clone().into(),
             previous_light_client_proof_journal: light_client_proof_journal,
             l2_genesis_state_root,
+            wtxid_data,
```
eyusufatik: not an input like this, it's already in previous_light_client_proof_journal

eyusufatik (changes requested):
You'll also need to implement

BitcoinService::extract_relevant_blobs_with_proof line 993

and BitcoinVerifier::verify_transactions line 153

crates/light-client-prover/src/circuit.rs
eyusufatik: thinking about how to get wtxid up here 🤔 

In the mean time let's not allow this to block us, we can focus on other issues

rakanalh: If we don't resolve it, the tx ids passed in the aggregator will not match against the wtxid map

kpp: > thinking about how to get wtxid up here 🤔

```
pub fn parse_light_client_transaction(
    tx: &Transaction,
) -> Result<ParsedLightClientTransaction, ParserError> {
    let script = get_script(tx)?;
    let instructions = script.instructions().peekable();
    // Map all Instructions errors into ParserError::ScriptError
    let mut instructions = instructions.map(|r| r.map_err(ParserError::from));

    parse_relevant_lightclient(&mut instructions)
}
```

This should be changed to include wtxid for chunk

kpp: Oh, wait. In Aggregates we store txids only to be able to request them in DaService for BitcoinService in extract_relevant_zk_proofs. So actually Aggregates must contain both txids (to request with rpc) and wtxids (for this prover).

eyusufatik: > Oh, wait. In Aggregates we store txids only to be able to request them in DaService for BitcoinService in extract_relevant_zk_proofs. So actually Aggregates must contain both txids (to request with rpc) and wtxids (for this prover).

ah, that's ugly...

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
```diff
@@ -236,6 +236,8 @@ pub struct LightClientCircuitOutput<Da: DaSpec> {
     pub last_l2_height: u64,
     /// Genesis state root of Citrea
     pub l2_genesis_state_root: [u8; 32],
+    /// A map from tx hash to chunk data
+    pub wtxid_data: BTreeMap<[u8; 32], Vec<u8>>,
```
eyusufatik: let's call this something else. chunks found or smth

crates/bitcoin-da/src/verifier.rs
```diff
@@ -307,6 +313,35 @@ impl DaVerifier for BitcoinVerifier {
...
+fn verified_blob_chunk<'a, T, I>(
+    tx: &T,
+    blobs_iter: &mut I,
+    wtxid: &[u8; 32],
+) -> Result<Option<&'a [u8]>, ValidationError>
+where
+    T: VerifyParsed,
+    I: Iterator<Item = &'a BlobWithSender>,
+{
+    if let Some(blob_hash) = tx.get_sig_verified_hash() {
```
eyusufatik: get sig verified hash will fail because blob sigs and pubkeys are 0

impl. a new method like get_hash_sig_unverified 😂

crates/light-client-prover/src/circuit.rs
```diff
@@ -99,35 +104,52 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+                                continue;
+                            }
+
+                            for tx_id in &aggregate_tx_ids {
+                                wtxid_data.remove(tx_id);
+                            }
+                        }
+                    }
+                    DaDataLightClient::Chunk(chunk) => {
+                        wtxid_data
```
eyusufatik: realized there is a dos attack vector with our aproach. let me think of a solution

eyusufatik (changes requested):

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -178,11 +188,25 @@ where
...
+            } else if let DaDataLightClient::Aggregate(_txids, wtxids) = batch_proof {
+                for wtxid in wtxids {
+                    if !current_block_wtxids.contains(&wtxid) {
+                        let (chunk_from_db, proof) = self.mmr_db.generate_proof(wtxid);
+                        mmr_hints.push((chunk_from_db, proof));
+                    }
+                }
+            } else if let DaDataLightClient::Chunk(body) = batch_proof {
+                // if chunk not used by any aggregate in block {
+                //     self.mmr_native.add_leaf(chunk); // TODO: this should happen after circuit execution
```
kpp: TODO?

crates/bitcoin-da/src/service.rs
```diff
@@ -984,13 +983,17 @@ impl DaService for BitcoinService {
...
                                     );
 
                                     relevant_txs.push(relevant_tx);
                                 }
                             }
-                            ParsedLightClientTransaction::Chunk(_) => {
-                                // ignore
+                            ParsedLightClientTransaction::Chunk(chunk) => {
+                                let mut relevant_tx =
+                                    BlobWithSender::new(chunk.body, vec![0], [0; 32], None);
```
kpp: Why `vec![0], [0; 32]`?

rakanalh: Sender & Hash for the chunk are non-existent as the chunk is unsigned

crates/light-client-prover/src/circuit.rs
```diff
@@ -91,92 +111,93 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+                                }
                             } else {
-                                // If not continue to the next blob
-                                continue;
+                                in_memory_chunks.remove(wtxid);
+
+                                let chunk = in_memory_chunks
+                                    .get(wtxid)
+                                    .expect("Chunk with wtxid should exist at this point")
+                                    .to_vec();
```
eyusufatik: why remove and get at the same time?

crates/light-client-prover/src/circuit.rs
```diff
@@ -91,92 +111,93 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+
+                                // The chunk belongs to the current aggregate, append to MMR guest and
+                                // also keep it to reconstruct the complete proof.
+                                let node = MMRNode::new(*wtxid, chunk);
+                                mmr_guest.append(node.clone());
+                                aggregate_chunks.push(node);
```
eyusufatik: don't need to append to mmr here, we only append the mmr with chunks that might be needed later. since this chunk has a an aggregate using itself in the same L1 block, we can discard it as used and not push to the mmr

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -195,12 +210,54 @@ where
...
 
                 assumptions.push(proof);
+            } else if let DaDataLightClient::Aggregate(_txids, wtxids) = batch_proof {
+                // Feed all found chunks which are used by the current aggregate into the MMR
+                // so that a new root can be constructed.
+                for wtxid in &wtxids {
+                    if let Some(body) = unused_chunks.get(wtxid) {
+                        mmr_native.append(MMRNode::new(*wtxid, body.clone()));
+                        // Clear the chunk from the unused chunks.
+                        unused_chunks.remove(wtxid);
```
eyusufatik: unused.remove is correct but don't append to mmr

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -195,12 +210,54 @@ where
...
+                    } else {
+                        return Err(anyhow!(
+                            "Missing chunk with wtxid {} for aggregate",
+                            hex::encode(wtxid)
+                        ));
```
eyusufatik: I guess for MMR hints we should just provide Option<(chunk, proof)> so that when we cant provide hint for an aggregate we can pass on that aggregate in the circuit

crates/light-client-prover/src/da_block_handler.rs
eyusufatik: we will append to MMR only the chunks that weren't used by aggregates in the block we are processing here

eyusufatik (changes requested):

crates/sovereign-sdk/full-node/db/sov-db/src/mmr_db.rs
```diff
@@ -0,0 +1,55 @@
...
+    /// since the DB is unaware of the hash function used by the JMT.
+    pub fn put(&self, mmr_native: MMRNative) -> Result<(), anyhow::Error> {
+        self.db.put::<MMRNodes>(&(), &mmr_native)
+    }
+
+    /// Get an optional value from the database, given a version and a key hash.
+    pub fn get(&self) -> anyhow::Result<Option<MMRNative>> {
+        self.db.get::<MMRNodes>(&())
+    }
+}
```
eyusufatik: are we getting every node all at once?

wouldn't querying by node index whenever we need a specific node be cleverer?

as MMRNodes grow by length, it's going to slow down a lot

eyusufatik (changes requested):

crates/light-client-prover/src/circuit.rs
```diff
@@ -91,92 +111,87 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+                            );
+
+                            if let Err(e) = result {
+                                println!("Error in light client guest: {:?}", e);
+                                continue;
+                            }
+
+                            for wtx_id in &aggregate_wtx_ids {
+                                in_memory_chunks.remove(wtx_id);
+                            }
```
eyusufatik: i guess this is unnecessary now

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -195,12 +210,47 @@ where
...
+                // aggregate in the current block, this tells us that it has been already seen in a previous
+                // L1 block.
+                // Given that we've updated MMR native with existing chunks, we now have a consistent MMR tree
+                // from which we can generate a hint for the guest MMR.
+                for wtxid in wtxids {
+                    if !current_block_wtxids.contains(&wtxid) {
+                        let Some((chunk_from_db, proof)) = mmr_native.generate_proof(wtxid) else {
+                            return Err(anyhow!(
+                                "Failed to generate native MMR proof for blobs at height: {}",
+                                l1_height
```
eyusufatik: we should merge this two for loops

if not in unused_chunks add hint. don't look at all block wtxids

crates/light-client-prover/src/circuit.rs
eyusufatik: i mean we are removing twice

crates/sovereign-sdk/rollup-interface/src/mmr.rs
```diff
@@ -58,24 +65,43 @@ impl MMRNode {
...
+            store: store.clone(),
             nodes: vec![vec![]],
-            leaf_nodes: vec![],
+            leaves: vec![],
+        };
+
+        // Initialize with existing leaves
+        let current_size = store.get_tree_size();
+
+        for i in 0..current_size {
```
eyusufatik: do you need to load everything on MMRNative::new()

I think if you also modified recalculate peaks to pull from and write to storage everything would work great

eyusufatik (changes requested):

crates/sovereign-sdk/rollup-interface/src/mmr.rs
```diff
@@ -0,0 +1,381 @@
...
+    pub fn new(wtxid: [u8; 32], body: Vec<u8>) -> Self {
+        MMRNode { wtxid, body }
+    }
+
+    pub fn hash(&self) -> [u8; 32] {
+        let mut hasher = Sha256::default();
+        hasher.update(self.wtxid);
+        hasher.update(&self.body);
+        hasher.finalize().into()
+    }
```
yaziciahmet: can we just use wtxid as hash as it is already hash of the transaction? I would get it if this was a generic MMR implementation, but since it is tailored specifically for transactions, we can avoid hashing cost here imo.

crates/sovereign-sdk/rollup-interface/src/mmr.rs
rakanalh: Not everything is loaded, we only feed the leaves into the MMR and every append call, invokes `recalculate_peaks`.

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -1,5 +1,6 @@
-use std::collections::{HashMap, VecDeque};
+use std::collections::{BTreeMap, HashMap, VecDeque};
 use std::sync::Arc;
+use std::vec;
```
yaziciahmet: can be removed

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -28,6 +31,8 @@ use tracing::{error, info};
 
 use crate::metrics::LIGHT_CLIENT_METRICS;
 
+type Wtxid = [u8; 32];
```
yaziciahmet: I think we should either make this a global type alias where we use everywhere, or stick with [u8; 32] for consistency sake.

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -143,6 +151,8 @@ where
         let l1_hash = l1_block.header().hash().into();
         let l1_height = l1_block.header().height();
 
+        let mut mmr_native = MMRNative::new(self.mmr_db.clone());
```
yaziciahmet: For every l1 block, we will reload the whole mmr db. Can't we initialize the MMRNative in the L1BlockHandler structure? MMRNative operations already behave as a write-through cache, and update db accordingly.

crates/sovereign-sdk/rollup-interface/src/mmr.rs
```diff
@@ -0,0 +1,381 @@
...
+        let mut mmr = MMRNative {
+            store: store.clone(),
+            nodes: vec![vec![]],
+            leaves: vec![],
+        };
+
+        // Initialize with existing leaves
+        let current_size = store.get_tree_size();
+
+        for i in 0..current_size {
```
yaziciahmet: iter_node on NodeStore trait maybe?

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -166,8 +176,10 @@ where
         );
 
         let mut assumptions = vec![];
+        let mut unused_chunks = BTreeMap::<Wtxid, Vec<u8>>::new();
```
yaziciahmet, line 196: This can be a hashmap as this is native environment and wtxids doesn't have any order which makes insert on btreemap inefficient.

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -166,8 +176,10 @@ where
...
+        let mut mmr_hints = vec![];
 
-        for batch_proof in batch_proofs {
+        for (wtxid, batch_proof) in batch_proofs {
             if let DaDataLightClient::Complete(proof) = batch_proof {
```
yaziciahmet: Wouldn't a match statement be better here?

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -195,12 +207,38 @@ where
...
+                // from which we can generate a hint for the guest MMR.
+                for wtxid in wtxids {
+                    // Cleanup unused_chunks from wtxids which are actually used by the current aggregate
+                    if unused_chunks.contains_key(&wtxid) {
+                        // Clear the chunk from the unused chunks.
+                        unused_chunks.remove(&wtxid);
+                    } else {
+                        let hint = mmr_native.generate_proof(wtxid);
+                        mmr_hints.push(hint);
+                    }
```
yaziciahmet, line 224: are you sure about not putting the current da block's chunks into mmr?

crates/bitcoin-da/src/helpers/parsers.rs
```diff
@@ -127,6 +132,18 @@ impl VerifyParsed for ParsedSequencerCommitment {
     }
 }
 
+impl VerifyParsed for ParsedChunk {
+    fn public_key(&self) -> &[u8] {
+        &[0]
+    }
+    fn signature(&self) -> &[u8] {
+        &[0]
+    }
```
yaziciahmet: unimplemented!("Should not be used with chunks")

crates/bitcoin-da/src/helpers/parsers.rs
```diff
@@ -69,6 +69,11 @@ pub trait VerifyParsed {
     fn signature(&self) -> &[u8];
     fn body(&self) -> &[u8];
 
+    /// Returns the hash of the body
+    fn get_unverified_hash(&self) -> Option<[u8; 32]> {
+        Some(calculate_sha256(self.body()))
+    }
+
     /// Verifies the signature of the inscription and returns the hash of the body
     fn get_sig_verified_hash(&self) -> Option<[u8; 32]> {
```
yaziciahmet, line 78: A note to self irrelevant to this PR: wer are creating Secp256k1 context everytime from scratch in this method which is really really bad practice and quite expensive. Definitely change that.

crates/bitcoin-da/src/service.rs
```diff
@@ -1013,34 +1012,39 @@ impl DaService for BitcoinService {
                             ParsedLightClientTransaction::Complete(complete) => {
                                 if let Some(hash) = complete.get_sig_verified_hash() {
                                     let blob = decompress_blob(&complete.body);
-                                    let relevant_tx =
-                                        BlobWithSender::new(blob, complete.public_key, hash);
-
+                                    let mut relevant_tx =
+                                        BlobWithSender::new(blob, complete.public_key, hash, None);
+                                    relevant_tx.wtxid = Some(wtxid.to_byte_array());
```
yaziciahmet: can be onelined

crates/bitcoin-da/src/service.rs
```diff
@@ -1013,34 +1012,39 @@ impl DaService for BitcoinService {
...
                                 if let Some(hash) = aggregate.get_sig_verified_hash() {
-                                    let relevant_tx = BlobWithSender::new(
+                                    let mut relevant_tx = BlobWithSender::new(
                                         aggregate.body,
                                         aggregate.public_key,
                                         hash,
+                                        None,
                                     );
-
+                                    relevant_tx.wtxid = Some(wtxid.to_byte_array());
```
yaziciahmet: can be one statement'ed xd

crates/bitcoin-da/src/service.rs
```diff
@@ -1013,34 +1012,39 @@ impl DaService for BitcoinService {
...
                                     relevant_txs.push(relevant_tx);
                                 }
                             }
-                            ParsedLightClientTransaction::Chunk(_) => {
-                                // ignore
+                            ParsedLightClientTransaction::Chunk(chunk) => {
+                                let mut relevant_tx =
+                                    BlobWithSender::new(chunk.body, vec![0], [0; 32], None);
+                                relevant_tx.wtxid = Some(wtxid.to_byte_array());
```
yaziciahmet: can be onelined, and pubkey can be just vec![]

crates/bitcoin-da/src/service.rs
```diff
@@ -1013,34 +1012,39 @@ impl DaService for BitcoinService {
...
                                 if let Some(hash) = method_id.get_sig_verified_hash() {
-                                    let relevant_tx = BlobWithSender::new(
+                                    let mut relevant_tx = BlobWithSender::new(
                                         method_id.body,
                                         method_id.public_key,
                                         hash,
+                                        None,
                                     );
-
+                                    relevant_tx.wtxid = Some(wtxid.to_byte_array());
```
yaziciahmet: can be one statement'ed

crates/light-client-prover/src/circuit.rs
```diff
@@ -1,6 +1,11 @@
+use std::collections::{BTreeMap, BTreeSet};
+use std::vec;
```
yaziciahmet: can be removed

crates/light-client-prover/src/circuit.rs
```diff
@@ -91,92 +111,88 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+                        'wtxids_loop: for wtxid in &wtx_ids {
+                            // If the wtxid belongs to a chunk that we've seen in a previous L1 block,
+                            // We use the hints to verify the existence of the chunk.
+                            if in_memory_chunks.contains_key(wtxid) {
+                                let chunk = in_memory_chunks
+                                    .get(wtxid)
+                                    .expect("Chunk with wtxid should exist at this point")
+                                    .to_vec();
+                                aggregate_chunks.push(MMRNode::new(*wtxid, chunk));
+                                in_memory_chunks.remove(wtxid);
```
yaziciahmet: can use BTreeMap::entry api which should lower the map access from 3 to 1.

crates/light-client-prover/src/circuit.rs
```diff
@@ -91,92 +111,88 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+                                    let Some((chunk, proof)) = hint else {
+                                        continue 'wtxids_loop;
+                                    };
+                                    if !mmr_guest.verify_proof(&chunk, &proof) {
+                                        // circuit not provided with enough hints
+                                        continue 'wtxids_loop;
+                                    }
+
+                                    aggregate_chunks.push(chunk);
+                                }
```
yaziciahmet: what if there are 2 aggregates? won't this drain all chunks in the mmr as if all belong to the first aggregate?

yaziciahmet (dismissed):

yaziciahmet:
also I am not sure if this pr encapsulates the case where all chunks except last is in the circuit input mmr, and the "last" chunk of the aggregate is in the da_data with the aggregate tx itself, but aggregate tx appears before the chunk. it won't satisfy the if subset check and will skip processing, and there are no future retry mechanisms.

crates/light-client-prover/src/da_block_handler.rs
rakanalh: Ordering matters since we'd like to maintain the order at which we see a chunk and insert it into the MMR. HashMap loses this order.

crates/light-client-prover/src/circuit.rs
rakanalh: As long as all the nodes in the MMR which are needed to verify the proof are there, this should not be a problem.

crates/light-client-prover/src/da_block_handler.rs
yaziciahmet: You are right about maintaining the order of chunks the same as in the DA, but BTreeMap will literally order them ascending by [u8; 32] (Wtxid), and not in the pushed order.

crates/sovereign-sdk/rollup-interface/src/mmr/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+    fn load_node(&self, level: usize, index: usize) -> Result<Option<MMRNodeHash>>;
+    fn save_chunk(&mut self, wtxid: Wtxid, chunk: MMRChunk) -> Result<()>;
+    fn load_chunk(&self, wtxid: Wtxid) -> Result<Option<MMRChunk>>;
+    fn get_tree_size(&self) -> usize;
+    fn set_tree_size(&mut self, size: usize) -> Result<()>;
+}
+
+#[derive(Serialize, Deserialize, Eq, PartialEq, Clone, Debug, BorshDeserialize, BorshSerialize)]
+pub struct MMRInclusionProof {
+    pub subroot_idx: usize,
```
yaziciahmet: I think this usize could be problematic since MMRInclusionProof is part of the circuit input. usize is 64-bits serialized in native and 32-bits serialized in zk. Didn't confirm this for sure, but their spec https://borsh.io does not specify usize, which is a sign that they are using what compiler returns as the size of this int, which is platform dependent.

This also made me notice that we are using usize for some other fields as well and needs to be fixed asap.

rakanalh: Thank you for sharing this. I noticed some tests failing because of an overflow problem. I'll check if this is related.

crates/light-client-prover/src/circuit.rs
yaziciahmet: All of the mmr_hints are collected into aggregate_chunks variable, and aggregate_chunks is scoped with single iteration. It will successfully verifiy and collect all merkle proofs, but it will be dropped in the next loop. Second aggregate iteration will try to collect hints from mmr_hints into aggregate_chunks one more time, but since mmr_hints is defined outside of the loop, it is already drained.

crates/light-client-prover/src/circuit.rs
```diff
@@ -91,95 +111,83 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+                            if let Some(chunk) = in_memory_chunks.get(wtxid) {
+                                aggregate_chunks.push(MMRChunk::new(*wtxid, chunk.clone()));
+                                in_memory_chunks.remove(wtxid);
                             } else {
-                                // If not continue to the next blob
-                                continue;
+                                while let Some(hint) = mmr_hints.pop_front() {
+                                    // If hint was not provided, which could happen due to the non-existence of the chunk
+                                    // in the same block as aggregate, we skip trying to prove the aggregate.
+                                    let Some((chunk, proof)) = hint else {
```
eyusufatik: if a chunk hint was not provided, we skip the aggreagate as a whole

crates/light-client-prover/src/circuit.rs
```diff
@@ -91,95 +111,83 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
-                            ),
-                        );
+                        let existing_wtx_ids: BTreeSet<Wtxid> =
+                            aggregate_chunks.iter().map(|c| c.wtxid).collect();
+                        let aggregate_wtx_ids: BTreeSet<Wtxid> = wtx_ids.iter().cloned().collect();
+
+                        // Make sure we have all the chunks, perform verification
+                        if aggregate_wtx_ids.is_subset(&existing_wtx_ids) {
+                            // Concatenate complete proof
+                            let complete_proof = aggregate_chunks
```
eyusufatik: we don't need to do the subset check anymore, can just flat_map aggregate_chunks now

crates/light-client-prover/src/circuit.rs
eyusufatik: good call actually, we must define an MMR node as 

wtixd + chunk 

so that we can check here

the native side can still push hints as a Vec and it can't provide a different chunk to the circuit than what the aggregate wants

eyusufatik: I see now that the MMRChunk is already wtxid + chunk, so in the loop we check for the provided wtixd being the same as the gues is looking for if not the same, we stop pulling hints for this aggregate (and we don't consume the last chunk and proof tuple we just queried for)

crates/light-client-prover/src/da_block_handler.rs
eyusufatik: since we need to construct the same MMR tree in native and guest, btree makes sense since both will append with the order of the btree, which will use wtxid

eyusufatik (changes requested):

eyusufatik:
> also I am not sure if this pr encapsulates the case where all chunks except last is in the circuit input mmr, and the "last" chunk of the aggregate is in the da_data with the aggregate tx itself, but aggregate tx appears before the chunk. it won't satisfy the if subset check and will skip processing, and there are no future retry mechanisms.

In that case there is nothing wrong with we can do, except resubmitting the aggregate later (and blobs, if needed)

ercecan (approved):

eyusufatik:
waiting on @ercecan to write tests for the pr, hence the hold merge label

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -302,8 +311,11 @@ where
         // Add unused chunks to MMR native.
         // Up until this point, the proof has been generated by aggregates in the block,
         // so it's okay to update the MMR tree now.
-        for (wtxid, body) in unused_chunks.into_iter() {
-            self.mmr_native.append(MMRChunk::new(wtxid, body))?;
+        if !unused_chunks.is_empty() {
+            println!("Adding {} more chunks to mmr", unused_chunks.len());
```
eyusufatik: no print

crates/light-client-prover/src/tests/mod.rs
```diff
@@ -935,6 +935,483 @@ fn test_mmr_hints() {
...
+        INITIAL_BATCH_PROOF_METHOD_IDS.to_vec(),
+        &batch_prover_da_pub_key.clone(),
+        &method_id_upgrade_authority,
+        Network::Nightly,
+    )
+    .unwrap();
+}
+
+#[test]
+fn test_malformed_mmr_chunk_wtxid() {
```
eyusufatik, line 1224: cant panic because of the wtxid check funny

crates/light-client-prover/src/tests/mod.rs
```diff
@@ -935,6 +935,483 @@ fn test_mmr_hints() {
...
+    let mut mmr_guest = MMRGuest::new();
+    let chunks = vec![
+        ([1; 32], chunk1.clone()),
+        ([2; 32], chunk2.clone()),
+        ([3; 32], chunk3.clone()),
+    ];
+
+    let mut mmr_hints = create_mmr_hints(&mut mmr_guest, chunks);
+
+    // Malform the chunk body
```
eyusufatik: fix comment

eyusufatik (approved):

PR #1624, opened 2024-12-18
CI: Fix validate_and_check_DA_ID CI check 

### LINKED ISSUE #1564: Fix validate_and_check_DA_ID CI check
The step is always failing. Let's fix that.

jfldde:
## Linked Issues
- Fixes #1564


eyusufatik:
so if we only check when ref is release-* we have two problems:

1 - we already do that in release CI, in repo vars we have an expected method id (which also needs to be changed a bit)
2 - we want to know if any pr is changing the commitment, so that we can have releases in an easier way

so if we can have the check method id ci for every pr and push to nightly, that would be awesome. is there a no fail no pass status for CI jobs? like a warning maybe

jfldde:
> so if we only check when ref is release-* we have two problems:
> 
> 1 - we already do that in release CI, in repo vars we have an expected method id (which also needs to be changed a bit) 2 - we want to know if any pr is changing the commitment, so that we can have releases in an easier way
> 
> so if we can have the check method id ci for every pr and push to nightly, that would be awesome. is there a no fail no pass status for CI jobs? like a warning maybe

This would run against any PR opened against release-v* or push to release-v* branches.

eyusufatik (approved):

PR #1629, opened 2024-12-19
Fix debug trace response

### LINKED ISSUE #1616: Fix debug_trace* response type
See newer versions of reth our return type is incorrect.

exeokan:
# Description
Returns `TraceResult` instead of `GethTrace` in `debug_traceBlockByHash` and `debug_traceBlockByNumber` handlers

## Linked Issues
- Fixes #1616


eyusufatik:
Looks ok. I need you to cross reference our response with an ethereum API.

I mean the actual json returned.

crates/evm/src/query.rs
```diff
@@ -1260,7 +1260,7 @@ impl<C: sov_modules_api::Context> Evm<C> {
                 &mut evm_db,
                 l1_fee_rate,
             )?;
-            traces.push(trace);
+            traces.push(TraceResult::new_success(trace, Some(tx.hash())));
```
exeokan, line 1263: We don't save errors to the traces, return Err in case of error, and always push TraceResult::Success. This in line with https://github.com/paradigmxyz/reth/pull/7837

crates/ethereum-rpc/src/trace.rs
```diff
@@ -237,10 +237,17 @@ fn get_traces_with_requested_tracer_and_config(
...
-                                if let GethTrace::CallTracer(call_frame) = trace {
+                                if let TraceResult::Success {
+                                    result: GethTrace::CallTracer(call_frame),
+                                    tx_hash,
+                                } = trace
+                                {
                                     let new_call_frame =
                                         apply_call_config(call_frame.clone(), call_config);
-                                    new_traces.push(GethTrace::CallTracer(new_call_frame));
+                                    new_traces.push(TraceResult::new_success(
```
kpp, line 225: We never push errored traces?

exeokan: In the case of a tracing error, I think it is logical to just the return Err instead of pushing TraceResult::Error, since we need to abort when a tracing error occurs, tx are dependent on each other. This is in line with the discussion here: https://github.com/paradigmxyz/reth/pull/7837#discussion_r1577739078

kpp (approved):

eyusufatik (approved):

eyusufatik:
gonna need to test this against blockscout :D

PR #1637, opened 2024-12-22
Bitcoin header chain verification per network

### LINKED ISSUE #1634: Bitcoin header chain for different networks
Need to adjust the header chain verification code for 

| Citrea Network | Bitcoin Network |
|----------------|-----------------|
| mainnet | mainnet |
| testnet | testnet4 |
| devnet | signet |
| nightly | regtest |

yaziciahmet:
# Description

Implements different header chain verification for each type of bitcoin networks: mainnet, testnet4, signet, regtest.

The only work remaining is to decide on the initial DA blocks for mainnet, testnet4, and signet.

## Linked Issues
- Fixes #1634 

## Testing

## Docs


crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -246,24 +246,24 @@ where
...
         let stored_proof_output = StoredLightClientProofOutput {
             state_root: circuit_output.state_root,
             light_client_proof_method_id: circuit_output.light_client_proof_method_id,
-            da_block_hash: circuit_output.da_block_hash.into(),
-            da_block_height: circuit_output.da_block_height,
-            da_total_work: circuit_output.da_total_work,
-            da_current_target_bits: circuit_output.da_current_target_bits,
-            da_epoch_start_time: circuit_output.da_epoch_start_time,
-            da_prev_11_timestamps: circuit_output.da_prev_11_timestamps,
+            da_block_hash: latest_da_state.block_hash,
```
eyusufatik: update `StoredLightClientProofOutput` also

eyusufatik (changes requested):
well done

yaziciahmet:
One thing I realized is that, we actually don't need total work. We just store it, and add to it, no checks whatsoever. Should we remove it?

crates/bitcoin-da/src/verifier.rs
```diff
@@ -221,64 +222,167 @@ impl DaVerifier for BitcoinVerifier {
...
+        &self,
+        latest_da_state: Option<&LatestDaState>,
+        block_header: &HeaderWrapper,
+    ) -> Result<LatestDaState, ValidationError> {
+        let network_constants = TESTNET4_CONSTANTS;
+        let latest_da_state = latest_da_state.unwrap_or(&INITIAL_TESTNET4_STATE);
+
+        let latest_block_time =
+            latest_da_state.prev_11_timestamps[latest_da_state.block_height as usize % 11];
+        // If more than 20 minutes passed since latest block, reset target
```
ozankaymak: This does not have to be true for all cases. It is true that the difficulty drops to minimum, but a real block with longer time larger than 20 minutes will still pass. See [block 40320](https://mempool.space/testnet4/block/0000000000000011e7cf114aa4980cbd74c2c07fccc8b025480dd4b7e1d96110)

crates/bitcoin-da/src/verifier.rs
```diff
@@ -288,23 +392,115 @@ impl DaVerifier for BitcoinVerifier {
...
+            block_hash: block_header.hash().to_byte_array(),
+            block_height: block_header.height(),
             total_work,
             epoch_start_time,
             prev_11_timestamps,
             current_target_bits,
         })
     }
+
+    fn verify_header_chain_regtest(
```
ozankaymak, line 448: One can get rid of some stuff here, such as the work done per block will always be constant and no need for difficulty change, or calculating the target bytes or compact target for that matter.

ozankaymak (commented):
I implemented all functionalities in a single function, did not seperate according to the network since we can decide it in compile time, see the [pull request](https://github.com/chainwayxyz/risc0-to-bitvm2/pull/19) and the [file](https://github.com/chainwayxyz/risc0-to-bitvm2/blob/844c22be0d4822278ddae3f4119bf4d91ebf8a62/header-chain/src/header_chain.rs). It would also be a good double check for me.

crates/bitcoin-da/src/verifier.rs
yaziciahmet: added epoch block check

crates/bitcoin-da/src/verifier.rs
yaziciahmet: got rid of unnecessary checks

yaziciahmet:
> I implemented all functionalities in a single function, did not seperate according to the network since we can decide it in compile time, see the [pull request](https://github.com/chainwayxyz/risc0-to-bitvm2/pull/19) and the [file](https://github.com/chainwayxyz/risc0-to-bitvm2/blob/844c22be0d4822278ddae3f4119bf4d91ebf8a62/header-chain/src/header_chain.rs). It would also be a good double check for me.

Due to our current code structure, it is hard and ugly for us to decide this on compile-time. Also, this makes it a bit more testable with the unit and integration tests, not limiting us to e2e tests where we compile each target accordingly and test. Hence, I tried to get away with one match statement in runtime and no other overhead.

eyusufatik (approved):

crates/bitcoin-da/src/verifier.rs
```diff
@@ -221,90 +222,299 @@ impl DaVerifier for BitcoinVerifier {
...
+                        // Total work is irrelevant in regtest
+                        total_work: [0; 32],
+                        current_target_bits: REGTEST_CONSTANTS.max_bits,
+                        // Epoch start time is irrelevant in regtest
+                        epoch_start_time: 0,
+                        // Prev 11 timestamps is irrelevant in regtest
+                        prev_11_timestamps: [0; 11],
+                    };
+                    self.verify_header_chain_regtest(&initial_regtest_state, block_header)
+                }
```
eyusufatik: you could do unwrap or here as well

eyusufatik (approved):

ozankaymak (commented):
Lgtm

PR #1639, opened 2024-12-23
batch proof output compatibility

### LINKED ISSUE #1585: StoredBatchProofOutput type migration in LedgerDb

ercecan:
# Description

## Linked Issues
- Fixes #1585
- Enables light client circuit to verify genesis fork proofs.

## Testing

## Docs


eyusufatik (changes requested):
you must also handle it in the light client prover circuit

eyusufatik (changes requested):
just realized we cant use `FORKS` in light client prover as we won't update it. so let's just use a vec of method ids that we start trying from the last one and if proof fails we try the one before... and if we try the first one and it still fails, we ignore that proof.

crates/light-client-prover/src/circuit.rs
```diff
@@ -81,10 +81,20 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+                (l2_genesis_root, 0)
+            },
+            |prev_journal| (prev_journal.state_root, prev_journal.last_l2_height),
+        );
+
     // TODO: Test for multiple assumptions to see if the env::verify function does automatic matching between the journal and the assumption or do we need to verify them in order?
     // https://github.com/chainwayxyz/citrea/issues/1401
     // Parse the batch proof da data
-    for blob in input.da_data {
+    'blobs: for blob in input.da_data {
```
eyusufatik: you don't need 'blobs anymore

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -292,6 +315,11 @@ where
 
         da_data.iter_mut().for_each(|tx| {
             // Check for commitment
+            tracing::warn!("tx sender: {:?}", tx.sender().as_ref());
```
eyusufatik: remove warnings

crates/light-client-prover/src/circuit.rs
```diff
@@ -94,27 +104,65 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
+                                batch_proof_method_ids[0].1
+                            } else {
+                                // If not continue to the next blob
+                                continue 'blobs;
+                            }
+                        } else {
+                            let idx = match batch_proof_method_ids
+                                // Returns err and the index to be inserted, which is the index of the first element greater than the key
+                                // That is why we need to subtract 1 to get the last element smaller than the key
+                                .binary_search_by_key(&last_l2_height, |(height, _)| *height)
```
eyusufatik: you shouldn't be searching by last_l2_height, this is light client proofs last l2 height

extract output first, then use that last_l2 height and use verify.

eyusufatik (changes requested):

eyusufatik (approved):

PR #1650, opened 2024-12-25
Implement EIP-1186 eth_getProof

### LINKED ISSUE #1527: Implement eth_getProof

kpp:
# Description

Implements https://docs.alchemy.com/reference/eth-getproof

Follow-up of https://github.com/chainwayxyz/citrea/pull/1563, because it was easier to create a new PR ontop of nightly.

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/1527


bin/citrea/tests/evm/mod.rs
```diff
@@ -224,6 +229,272 @@ async fn test_genesis_contract_call() -> Result<(), Box<dyn std::error::Error>>
...
+            let storage_proof: jmt::proof::SparseMerkleProof<sha2::Sha256> =
+                borsh::from_slice(&storage_proof.proof[0]).unwrap();
+
+            storage_proof
+                .verify(expected_root_hash, key_hash, proved_value)
+                .expect("Account storage proof must be valid");
+        }
+    }
+
+    let acc_zkproof_1 = seq_test_client
```
eyusufatik: these are not zk proofs. let's rename

bin/citrea/tests/test_client/mod.rs
```diff
@@ -482,6 +482,18 @@ impl TestClient {
         eth_logs
     }
 
+    pub(crate) async fn eth_get_proof(
+        &self,
+        address: Address,
+        keys: Vec<U256>,
+        block_number: Option<BlockNumberOrTag>,
+    ) -> Result<EIP1186AccountProofResponse, Box<dyn std::error::Error>> {
+        self.http_client
```
eyusufatik, line 491: let's use alloy's client here to see if it can parse our proof

bin/citrea/tests/evm/mod.rs
```diff
@@ -224,6 +229,272 @@ async fn test_genesis_contract_call() -> Result<(), Box<dyn std::error::Error>>
...
+                // This check is not actually required, it's for test purposes only
+                assert_eq!(storage_proof.value, storage_value);
+            }
+        }
+    }
+
+    {
+        // Assert historic proof is not the same as the first one queried
+        //  because storage root is different -> all proofs are different too.
+        assert_ne!(acc_zkproof_1, acc_zkproof_2);
```
eyusufatik: we must also assert the first proof queried is exactly the same as the newly queried one with rpc block num = block num at the time of sending the first one.

crates/sovereign-sdk/module-system/sov-state/src/zk_storage.rs
```diff
@@ -149,11 +145,11 @@ where
...
 
     fn open_proof(
-        state_root: Self::Root,
-        state_proof: StorageProof<Self::Proof>,
+        state_root: StorageRootHash,
+        state_proof: StorageProof,
     ) -> Result<(StorageKey, Option<StorageValue>), anyhow::Error> {
         let StorageProof { key, value, proof } = state_proof;
-        let key_hash = KeyHash::with::<H>(key.as_ref());
+        let key_hash = KeyHash::with::<DefaultHasher>(key.as_ref());
```
eyusufatik, line 152: was it the case that we were passing DefaultHasher for previous zk storage? I expect it to be that way but let's confirm

eyusufatik (changes requested):
note to myself: sync with current testnet to see if any storage changes break state root calculation

crates/sovereign-sdk/module-system/sov-state/src/zk_storage.rs
kpp: We passed DefaultHasher everywhere but the type of it was same for mock/not mock stuff. So there is no use in this generic at all.

bin/citrea/tests/test_client/mod.rs
kpp: they don't support block id param. But we use their EIP1186AccountProofResponse so we are good

eyusufatik (approved):

PR #1653, opened 2024-12-26
guest side proof fail

eyusufatik:
# Description
since risc0 guest side proof verification APIs only support proofs that are verifiable, we need to verify failing of proofs with host side verification API.

TODO:
- [x] figure out how to test this in e2e

## Linked Issues
Fixes #1643

PR #1681, opened 2025-01-08
Remove use_latest_elf

jfldde:
# Description

- Remove use_latest_elf config.
- Introduces new Regtest network to be used when needing to test elf switching and backward compatibility
- Use risc0 `sys-getenv` behind testing feature flag to switch fork based on runtime network


eyusufatik:
> Use risc0 sys-getenv behind testing feature flag to switch fork based on runtime network

~can't do that because we are using native-side proof verification functions in the guest to show a zk proof fails since #1653 and that feature is needed for this to function~

Update: misunderstood something

bin/citrea/src/main.rs
```diff
@@ -150,7 +150,11 @@ async fn main() -> Result<(), anyhow::Error> {
 
     let mut network = args.network.into();
     if args.dev {
-        network = Network::Nightly;
+        network = if cfg!(feature = "testing") && network == Network::Regtest {
```
rakanalh: Maybe just use `--network regtest` when needed. I don't think it's going to be convenient to always use regtest when running tests. What do you think?

rakanalh (approved):
LGTM

I can't say i like the `#[cfg(feature = "testing")]` in multiple places for networks but i am guessing that there's no way to pass network cleanly into the guest.

jfldde:
> LGTM
> 
> I can't say i like the `#[cfg(feature = "testing")]` in multiple places for networks but i am guessing that there's no way to pass network cleanly into the guest.

Yes I agree, I don't like it either. The issue I faced is that we want most of the tests to run on default network nightly but a few where we override it and run with regtest. This network has both ELFs and forks. With the current nightly branch behaviour, the network is passed at compile time and there's no way to override it but to re-build with the updated network, makes it very tedious.
This introduces a way to pass env vars at runtime and to override compile time network at runtime. I put everything behind testing flag to make sure we have no overhead/footgun. I might have been overly cautious and we could do without it.

guests/risc0/light-client-proof/bitcoin/src/bin/light_client_proof_bitcoin.rs
```diff
@@ -115,6 +117,16 @@ pub const METHOD_ID_UPGRADE_AUTHORITY_DA_PUBLIC_KEY: [u8; 33] = {
...
 
+fn get_network() -> Network {
+    #[cfg(feature = "testing")]
+    {
+        if std::env::var("CITREA_NETWORK") == Ok(String::from("regtest")) {
+            return Network::Regtest;
+        }
+    }
+    NETWORK
+}
```
eyusufatik: cant say I like this

jfldde: Rationale being no overhead when not under `testing` + able to override only when passing regtest network.
Open to better idea though.

eyusufatik:
still unsure about the pr because of the env var change in the guests 😓

eyusufatik (approved):
Merge after doing two things

- Rename Network::Regtest to something else. maybe "SpecialTestingNetwork" 

- Find another way except testing feature + set env risc0 adapter crate. gotta be solvable with a simpler method

eyusufatik:
enabled sys-getenv in every guest because anyhow errors check env vars. does not change behvaiour of our code.

PR #1682, opened 2025-01-08
Verify aggregate final fixes

yaziciahmet:
# Description

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/light-client-prover/src/circuit.rs
+++ b/crates/light-client-prover/src/circuit.rs
@@ -118,7 +118,7 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
     let mut current_proof_index = 0u32;
     let mut expected_to_fail_hints = input.expected_to_fail_hint.into_iter().peekable();
     // Parse the batch proof da data
-    for blob in input.da_data {
+    'blob_loop: for blob in input.da_data {
         if blob.sender().as_ref() == batch_prover_da_public_key {
             let data = DaDataLightClient::try_from_slice(blob.verified_data());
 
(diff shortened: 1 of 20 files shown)


eyusufatik:
note: even if we close this pr, we should apply changes for Vec<u8> -> &[u8]

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -363,6 +391,11 @@ where
         Ok(())
     }
 
+    /// Verifies complete proof. Returns:
+    ///
+    /// - Ok(true) -> proof is successfully parsed, not a duplicate, and verified
+    /// - Ok(false) -> proof is successfully parsed, not a duplicate, but verification failed
+    /// - Err(_) -> proof is either unparseable or a duplicate
```
eyusufatik, line 364: what is a duplicate

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -265,29 +270,52 @@ where
                         complete_proof.len()
                     );
 
-                    // TODO: Handle error
-                    let complete_proof = self
-                        .da_service
-                        .decompress_chunks(complete_proof.into_iter().flatten().collect())
-                        .unwrap();
+                    let reinsert_used_chunks = || {
```
eyusufatik, line 250: why the functionization

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -265,29 +270,52 @@ where
...
+                            assumptions.push(complete_proof);
+                            proof_index += 1;
+                        }
+                        Ok(false) => {
+                            error!(
+                                "Aggregate batch proof verification failed. wtxid = {}",
+                                hex::encode(wtxid)
+                            );
+
+                            reinsert_used_chunks();
```
eyusufatik: don't reinsert if aggregate can be formed but cannot be verified.

crates/light-client-prover/src/circuit.rs
```diff
@@ -140,59 +140,81 @@ pub fn run_circuit<DaV: DaVerifier, G: ZkvmGuest>(
...
-                            if let Some((wtxid, chunk)) = in_memory_chunks.remove_entry(wtxid) {
-                                // If the wtxid belongs to a chunk that we've seen in the same L1 block,
-                                // We add it to the aggregate.
-                                aggregate_chunks.push(MMRChunk::new(wtxid, chunk));
+                            if in_memory_chunks.contains_key(wtxid) {
+                                continue;
+                            }
+
+                            let hint = mmr_hints_iter.next();
+                            if hint.is_none() || hint.unwrap().0.wtxid != *wtxid {
```
eyusufatik: if we werent able to provide hint for a chunk, now the circuit consumed a hint meant for another aggregate

this was the reason why hints held optional tuples

eyusufatik (changes requested):

crates/light-client-prover/src/circuit.rs
yaziciahmet: it doesnt consume. mmr_hints_iter is created on every iteration once again. this for loop only verifies that we can recreate the aggregate from chunks. the following loop actually takes the values out and recreates the proof. and since the next loop takes the values out from mmr_hints, it is ensured that the newly created mmr_hints_iter on the next loop must match the aggregate.

crates/light-client-prover/src/da_block_handler.rs
yaziciahmet: We should. idk how likely, but there is a chance that (a,b)'s decompressed value somehow is valid serialized structure while true proof is (a,b,c). In such case, we will only see it is not valid when we actually verify.

The same goes for if its not serializable as well, due to the same logic as above.

crates/light-client-prover/src/da_block_handler.rs
yaziciahmet: its used 3 times hence i wanted to just create a closure.

crates/light-client-prover/src/da_block_handler.rs
yaziciahmet: I couldnt find a better way to articulate this condition:
```
        if batch_proof_last_l2_height <= light_client_l2_height && light_client_l2_height != 0 {
            return Err(anyhow::anyhow!(
                "Batch proof l2 height is less than latest light client proof l2 height"
            ));
        }
```

crates/light-client-prover/src/da_block_handler.rs
eyusufatik: don't count for that probability. highly improbable

crates/light-client-prover/src/circuit.rs
eyusufatik: mmr_hints = a, b, c, d, e

assume a,b is for aggr1 
c, d, e is for aggr2 

then for aggr2 we get a and fail

yaziciahmet: we don't get a, we get c. because in this loop we just verify, in the next loop we actually pop a and b.

crates/light-client-prover/src/da_block_handler.rs
yaziciahmet: you are right. this will allow a malicious prover to send a,b,aggregate(a,b) where this aggregate is completely invalid but parseable, and the circuit will hold on to chunks a and b forever.

removed https://github.com/chainwayxyz/citrea/pull/1682/commits/1af16c375c552ce44cc26caeb20847efb3bce456

eyusufatik (approved):

PR #1697, opened 2025-01-13
Implement FromEnv for MonitoringConfig

### LINKED ISSUE #1696: Make MonitoringConfig optional under BitcoinDaServiceConfig
Only the sequencer and the batch prover will make use of the monitoring config, we'll set it to None in full nodes and light client prover.

jfldde:
# Description

- Implement `FromEnv` for `MonitoringConfig`
- Add support for partial parsing with serde defaults

Fixes #1696

### DIFF
--- a/crates/bitcoin-da/src/monitoring.rs
+++ b/crates/bitcoin-da/src/monitoring.rs
@@ -3,6 +3,7 @@ use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
+use anyhow::anyhow;
 use bitcoin::address::NetworkUnchecked;
 use bitcoin::hashes::Hash;
 use bitcoin::{Address, BlockHash, Transaction, Txid};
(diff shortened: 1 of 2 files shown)


crates/bitcoin-da/src/monitoring.rs
```diff
@@ -124,23 +121,59 @@ pub enum MonitorError {
     BitcoinEncodeError(#[from] bitcoin::consensus::encode::Error),
 }
 
+mod defaults {
```
rakanalh:
```suggestion
mod monitoring_defaults {
```

rakanalh (approved):
LGTM

crates/bitcoin-da/src/monitoring.rs
```diff
@@ -124,23 +121,59 @@ pub enum MonitorError {
...
+                |v| v.parse().map_err(Into::<anyhow::Error>::into),
+            )?,
+            history_limit: std::env::var("DA_MONITORING_HISTORY_LIMIT").map_or_else(
+                |_| Ok(defaults::history_limit()),
+                |v| v.parse().map_err(Into::<anyhow::Error>::into),
+            )?,
+            max_history_size: std::env::var("DA_MONITORING_MAX_HISTORY_SIZE").map_or_else(
+                |_| Ok(defaults::max_history_size()),
+                |v| v.parse().map_err(Into::<anyhow::Error>::into),
+            )?,
```
eyusufatik: this always generates a monitoring config. 

for example a full node has a da service but does not need a da monitoring service.

eyusufatik (dismissed):

crates/bitcoin-da/src/monitoring.rs
eyusufatik: i'd do this:

```rust
let check_interval = std::env::var()
let history_limit = std::env::var()
let max_history_size = std::env::var()

if all_err return Err()
else {
  check_interval.parse().expect()
  history_limit.parse().expect()
  max_history_size().expect()
  
  Ok(....)
}
```

on FromEnv impl. of BitcoinDaService you transpose the `MonitoringConfig::from_env()` result.

eyusufatik (approved):

PR #1699, opened 2025-01-13
Support pre fork1 input in generateInput

### LINKED ISSUE #1655: Batch batchProver_generateInput support pre fork 1 input

### LINKED ISSUE #1698: Fix getting current spec in prove_l1
In `prove_l1`, we currently enumerate inputs and check spec according to the sequencer commitment at the same index. Instead, we should check the sequencer commitment range of the input, get the last l2 height of the last commitment and check spec accordingly.

exeokan:
# Description
In `generate_input`, serializes pre-fork1 inputs from `BatchProofCircuitInputV1`

## Linked Issues
- Fixes #1655 
- Fixes #1698

### DIFF
--- a/crates/batch-prover/src/proving.rs
+++ b/crates/batch-prover/src/proving.rs
@@ -240,12 +240,15 @@ where
         .unwrap_or(vec![]);
 
     // Add each non-proven proof's data to ProverService
-    for (i, input) in inputs.into_iter().enumerate() {
+    for input in inputs {
         if !state_transition_already_proven::<StateRoot, Witness, Da, Tx>(&input, &submitted_proofs)
         {
-            let seq_com = sequencer_commitments.get(i).expect("Commitment exists");
-            let last_l2_height = seq_com.l2_end_block_number;
+            let range_end = input.sequencer_commitments_range.1;
 
+            let last_seq_com = sequencer_commitments
+                .get(range_end as usize)
+                .expect("Commitment does not exist");
+            let last_l2_height = last_seq_com.l2_end_block_number;
             let current_spec = fork_from_block_number(last_l2_height).spec_id;
 
             let elf = elfs_by_spec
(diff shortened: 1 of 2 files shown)


crates/batch-prover/src/rpc.rs
```diff
@@ -257,10 +269,6 @@ where
     }
 }
 
-fn serialize_batch_proof_circuit_input<T: BorshSerialize>(item: T) -> Vec<u8> {
```
exeokan, line 260: Removed this function as it needs to have the item's type as `BatchProofCircuitInput` and I wanted to avoid generics.

crates/batch-prover/src/rpc.rs
```diff
@@ -185,7 +186,18 @@ where
         for input in inputs {
             let range_start = input.sequencer_commitments_range.0;
             let range_end = input.sequencer_commitments_range.1;
-            let serialized_circuit_input = serialize_batch_proof_circuit_input(input);
+
+            let seq_com = sequencer_commitments
+                .get(range_end as usize)
```
eyusufatik, line 191: let's also fix your finding in the `prove_l1` function.

crates/batch-prover/src/proving.rs
```diff
@@ -240,12 +240,15 @@ where
...
     // Add each non-proven proof's data to ProverService
-    for (i, input) in inputs.into_iter().enumerate() {
+    for input in inputs {
         if !state_transition_already_proven::<StateRoot, Witness, Da, Tx>(&input, &submitted_proofs)
         {
-            let seq_com = sequencer_commitments.get(i).expect("Commitment exists");
-            let last_l2_height = seq_com.l2_end_block_number;
+            let range_end = input.sequencer_commitments_range.1;
 
+            let last_seq_com = sequencer_commitments
```
exeokan, line 248: #1698

eyusufatik (approved):

PR #1700, opened 2025-01-13
first l2 block copy blocks, transactions, receipts to rlp acessory state

eyusufatik:
we tried to avoid implementing this but here we are.

only affects native execution so don't worry about the cycles.

~still needs DoNot.... structs to convert stuff just like we did with head and head_rlp~


crates/evm/src/evm/primitive_types.rs
```diff
@@ -180,6 +185,42 @@ pub(crate) struct TransactionSignedAndRecovered {
     pub(crate) block_number: u64,
 }
 
+#[cfg(feature = "native")]
+/// This uses the old version of the TransactionSigned launched testnet with with Reth v1.0.4
```
kpp, line 189: This can be moved to a submod to reduce the number of "cfg feature native"

eyusufatik: later

rakanalh (approved):
LGTM, need to think about how to do migrations on the native DB in a better way.

kpp (approved):

PR #1701, opened 2025-01-13
Hotfix

ercecan:


crates/evm/src/lib.rs
```diff
@@ -152,7 +152,7 @@ pub struct Evm<C: sov_modules_api::Context> {
     #[cfg(feature = "native")]
     #[state]
     pub(crate) transactions:
-        sov_modules_api::AccessoryStateVec<TransactionSignedAndRecovered, BcsCodec>,
+        sov_modules_api::AccessoryStateVec<DoNotUseTransactionSignedAndRecovered, BcsCodec>,
```
rakanalh, line 158: It's called `DoNotUse` but it's actually being used.

```suggestion
        sov_modules_api::AccessoryStateVec<CompatibleTransactionSignedAndRecovered, BcsCodec>,
```

rakanalh (approved):
Requesting changes on the PR but i don't want to block it.

crates/evm/src/lib.rs
eyusufatik: that's a convention we have since introducing `DoNotUseSealedHeader`. It does not says it's not used, it instructs the next dev working on the repo to not use that struct

PR #1702, opened 2025-01-14
fix off-by-one error in light client prover

eyusufatik:
# Description
once we deploy the code on a bitcoin network that has built-in constants for starting height, we got an off by one error and everything fails.

we didn't notice this because on e2e, regtest, we don't force the proof to start from a specific DA block and appearently if we set to 200, it would start from 201.

we noticed this while deploying on devnet.

### DIFF
--- a/crates/light-client-prover/src/runner.rs
+++ b/crates/light-client-prover/src/runner.rs
@@ -8,7 +8,6 @@ use jsonrpsee::server::{BatchRequestConfig, ServerBuilder};
 use jsonrpsee::RpcModule;
 use sov_db::ledger_db::{LightClientProverLedgerOps, SharedLedgerOps};
 use sov_db::mmr_db::MmrDB;
-use sov_db::schema::types::SlotNumber;
 use sov_rollup_interface::services::da::DaService;
 use sov_rollup_interface::spec::SpecId;
 use sov_rollup_interface::zk::ZkvmHost;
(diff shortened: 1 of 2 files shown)


kpp (approved):

crates/light-client-prover/src/runner.rs
```diff
@@ -202,3 +202,8 @@ where
         Ok(rpc_methods)
     }
 }
+
+pub(crate) enum StartVariant {
```
rakanalh: Would be nice if we declare types at the top.

crates/light-client-prover/src/runner.rs
```diff
@@ -174,7 +174,7 @@ where
                 mmr_db,
             );
             l1_block_handler
-                .run(last_l1_height_scanned.0, cancellation_token)
+                .run(starting_block, cancellation_token)
```
rakanalh, line 182: I think passing the height + 1 here would be better so that the da block handler doesn't need the variants.

rakanalh (approved):
LGTM with minor comments.

jfldde (approved):

crates/light-client-prover/src/runner.rs
eyusufatik: I think the meaning of the start_block gets lost when passed around. so wanted to handle it where it's needed in case something changes later regarding the loop.

PR #1711, opened 2025-01-15
Database BackupManager

### LINKED ISSUE #1641: RocksDB backups through RPC
Apparently RocksDB has some built-in feature for taking backups. These will come in handy in our infra. Let's see how we can trigger a back up through the RPC.

jfldde:
# Description

- Introduces new `BackupManager` to backup all databases required by a citrea node.
- Introduces two new RPCs: 
    - `backup_create <optional_path>` to create a node backup. Defaults to using path from config if None. Errors if both config backup path and path are not set. 
    - `backup_validate <path>` to validate a node backup at `path`
    - `backup_info <path>` to get info about backup at `path`
 
 - Restore backup via `citrea-cli` backup command
 ```usage:
 citrea-cli backup                                   
error: the following required arguments were not provided:
  --node-kind <NODE_KIND>
  --db-path <DB_PATH>
  --backup-path <BACKUP_PATH>

Usage: citrea-cli backup --node-kind <NODE_KIND> --db-path <DB_PATH> --backup-path <BACKUP_PATH>
 ```
## Linked Issues
- Fixes #1641


crates/sequencer/src/runner.rs
```diff
@@ -761,6 +766,7 @@ where
                         missed_da_blocks_count = 0;
                     }
 
+                    let _l2_lock = backup_manager.start_l2_processing().await;
```
eyusufatik, line 708: does this mean taking backups would lock us here and we won't be able to produce blocks?

eyusufatik: from my understanding coupled with WAL `rocksdb::backup::BackupEngine` doesn't really need you to lock the database.

once concern here maybe would be the state_db and native_db are expected to always be in the same version so we have to guarantee somehow they are backed up at the same point in time -- which is not really possible.

is that the case?

if so, we'll soon be working on fixing this incompatibility problem.

jfldde: This is to err on the safe side and guarantee that all databases are at an initial l2 block state.
The lock is held to access l2 block height and spawn the backup tasks in this safe state and then released. Block production should tick along while backup is on-going

bin/citrea/src/rollup/mod.rs
```diff
@@ -87,25 +100,38 @@ pub trait CitreaRollupBlueprint: RollupBlueprint {
...
         let genesis_config = self.create_genesis_config(runtime_genesis_paths, &rollup_config)?;
 
         let mut storage_manager = self.create_storage_manager(&rollup_config)?;
         let prover_storage = storage_manager.create_finalized_storage()?;
 
+        backup_manager.register_database("ledger".to_string(), ledger_db.clone().db_handle())?;
+        backup_manager
+            .register_database("state".to_string(), storage_manager.get_state_db_handle())?;
+        backup_manager.register_database(
+            "native-db".to_string(),
```
kpp: Why this one is `native-db` whereas ledger and state dont have `-db` prefix?

bin/citrea/src/rollup/mod.rs
```diff
@@ -462,6 +538,23 @@ pub trait CitreaRollupBlueprint: RollupBlueprint {
...
     {
+        let mut backup_manager = BackupManager::new(
+            "light-client-prover",
+            rollup_config.storage.backup_path.clone(),
+            Some(BackupConfig {
+                backup_dirs: vec![
+                    "ledger".to_string(),
+                    "native-db".to_string(),
+                    "state".to_string(),
+                    "mmr".to_string(),
```
kpp, line 549: I can see mmr here but I don't see it's registration

crates/sovereign-sdk/full-node/sov-prover-storage-manager/src/snapshot_manager.rs
```diff
@@ -119,6 +119,10 @@ impl SnapshotManager {
 
         Ok(SnapshotManagerIter::new(db_iter, snapshot_iterators))
     }
+
+    pub fn get_db_handle(&self) -> Arc<sov_schema_db::DB> {
+        self.db.clone()
```
kpp: It would be nice to see the consequences of returning a Weak ptr here. Just curious

crates/common/src/backup/manager.rs
```diff
@@ -0,0 +1,371 @@
...
+        &self,
+        path: Option<PathBuf>,
+        l2_height: u64,
+    ) -> anyhow::Result<CreateBackupInfo> {
+        let backup_path = path
+            .as_ref()
+            .or(self.base_path.as_ref())
+            .context("Missing path and no backup_path found in config.")?;
+
+        let l1_lock = self.l1_processing_lock.lock().await;
```
kpp, line 162: What are these locks for? If we need a FS lock, we can use https://docs.rs/file-lock/latest/file_lock/

jfldde: The idea behind the two locks is to make sure the three backups are created in a consistent L1 and L2 state between them.
We hold both locks, start the backup process in a consistent state, then release locks. The node can continue producing block while the backups are made from the same L2/L1 height

bin/citrea/src/rollup/mod.rs
jfldde: https://github.com/chainwayxyz/citrea/pull/1753 
Thanks for pointing it out, DB_NAME and DB_PATH_SUFFIX were mixed up.

crates/common/src/backup/manager.rs
kpp: Ok, understood. Do we produce l1 blocks in parallel while producing l2 blocks or we are blocked by l1 production? If we are blocked, then there is no need for 2 mutexes?

jfldde: L1 and L2 block productions are independent from each other. The two mutexes are here for the backup manager to hold both locks and have a consistent L1 and L2 height

jfldde:
Holding merge until devops' feedback

crates/common/src/backup/manager.rs
```diff
@@ -0,0 +1,382 @@
...
+        {
+            let dbs = self.databases.read().unwrap();
+            for dir in &self.config.backup_dirs {
+                let path = backup_path.join(dir);
+                let db = dbs.get(dir).context("Missing required db")?.clone();
+                handles.push(tokio::task::spawn_blocking(move || db.create_backup(&path)));
+            }
+        }
+
+        // Wait for all dbs to starting backing up under lock before releasing
```
rakanalh:
```suggestion
        // Wait for all dbs to finish backing up under lock before releasing
```

crates/common/src/backup/manager.rs
```diff
@@ -0,0 +1,382 @@
...
+        drop(l1_lock);
+        drop(l2_lock);
+
+        for handle in handles {
+            handle.await??;
+        }
```
rakanalh, line 197: Given the above comment, i do believe we should await on all backup tasks to finish BEFORE we drop / release the locks. Releasing the locks and then awaiting on the tasks will trigger block production to resume and DB will be updated while it's being copied.

crates/common/src/backup/rpc.rs
```diff
@@ -0,0 +1,152 @@
...
+                ErrorObjectOwned::owned(
+                    INTERNAL_ERROR_CODE,
+                    INTERNAL_ERROR_MSG,
+                    Some(format!("{e}")),
+                )
+            })?
+            .unwrap_or_default();
+
+        self.backup_manager
+            .create_backup(path, l2_height)
```
rakanalh: This is basically a DOS attack vector. Since creating a backup blocks block production, an attacker could choose to make frequent requests to create backups holding the locks for block production and at the same time causing the node storage to fill quickly. What do we do to mitigate this?

If devops can set this up in a way where the RPC is only callable from a specific subnet and make sure this RPC is not publicly available, then we're good.

crates/sovereign-sdk/full-node/db/sov-db/Cargo.toml
```diff
@@ -16,6 +16,7 @@ resolver = "2"
 [dependencies]
 # Maintained by sovereign labs
 jmt = { workspace = true }
+
```
rakanalh: Remove new line since this is considered a new dep group with the new line

rakanalh (dismissed):

bin/citrea/src/cli.rs
```diff
@@ -59,6 +61,12 @@ pub(crate) struct Args {
     /// Logging verbosity
     #[arg(long, short = 'q', action)]
     pub(crate) quiet: bool,
+
+    /// Optionally restore DB state from a backup before starting the node.
+    /// If provided, databases are restored from the specified backup path
+    /// before any services are initialized.
+    #[arg(long)]
+    pub(crate) restore_db: Option<PathBuf>,
```
rakanalh: As part of https://github.com/chainwayxyz/citrea/pull/1796, i am working on `citrea-cli` which provides a bunch of subcommands that we can use to operate the node's storage among potentially other stuff without having to put these things in the node's binary.

I believe this can be a subcommand `restore-backup` or something of the sort which could trigger restoring the DB from backup, rather than having to pass this flag while running the node itself. IMO, not a good design / use case.

crates/common/src/backup/manager.rs
jfldde: Quoting my reply to kpp comment above :
> The idea behind the two locks is to make sure the three backups are created in a consistent L1 and L2 state between them. We hold both locks, start the backup process in a consistent state, then release locks. The node can continue producing block while the backups are made from the same L2/L1 height

RocksDB will take a snapshot at the deterministic L1/L2 block height, then backup from the snapshot. We can thus release the lock as soon as the backup process is started

bin/citrea/src/cli.rs
jfldde: Will we want to restore from backup against a running node ?

crates/common/src/backup/rpc.rs
jfldde: We're not holding the lock but it's still an heavy call that will need to be restricted. Think this can be handled at the devops side though

bin/citrea/src/cli.rs
rakanalh: I don't think so... because there will be some disparity between some in-memory data and what's been written to the database while the node is running. 
Your code restores at the beginning of the binary execution which can be done in a step before running the node which is by invoking `citrea-cli restore-backup ... etc`.

However, i think your PR will get merged before mine so we can address this later.

crates/common/src/backup/rpc.rs
rakanalh: cc: @eyusufatik @Chinacolt

bin/citrea/src/cli.rs
jfldde: Oh you're right, I naively assumed that the `citrea-cli` would be running against the `citrea` node itself

Chinacolt (commented):
# **Backup & Restore Observations and Questions**  

## **Backup Creation (`backup_create`)**  
- Why do we pass the `id` field?  
- I tested whether it blocks ongoing synchronization and RPC operations. It does not. ✅  

## **Backup Information (`backup_info`)**  
- Regardless of which `id` we pass, it always returns all backup information.  
- If `id` is ignored, what is the purpose of passing it?  

## **Backup Validation (`backup_validate`)**  
- If we pass a non-existent path, it fails. Otherwise, it always returns `true`.  
- The `id` field doesn’t seem to have any effect—tested with `id: -73`, and it still returned `true`.  

## **Restore (`restore`)**  
- I tested restore several times with `--restore-db "/mnt/task/backup"`, and it worked consistently. ✅  
- **Question:** Can we specify which `id` to restore? It seems to always restore the latest version.  

---

## **Other Observations (Not necessarily issues, just for awareness)**  

### **1. Version Tracking Behavior**  
- Deleted the latest (6th) backup and then restored. It reverted to the 5th version, which became the latest.  
- After restoring, creating a new backup resulted in version 7.  
- It seems the system tracks previously created versions even after a restore. Is this expected behavior?  

### **2. Partial Backup Deletion & Restore Behavior**  
- Deleted the last 3 versions from `native-db`, last 2 versions from `state`, and the last one from `ledger`.  
- After restore, each database seemed to restore to its latest available backup.  
- However, once the restore was completed, an error occurred:  
```"error": "Timestamp should be greater"```

**Suggestion:** 
- We, ops, can validate if all DBs are at the same version before restoring to avoid inconsistency. 
- This validation doesn’t necessarily have to be handled inside Citrea’s binary, but we need a way to validate individual versions by passing the `id` parameter to `backup_validate`.  

### **3. Post-Restore Backup Behavior**  
After restore completion, a new directory `citrea-db.bak-{EPOCH}` appears.  
**Question:**  
- Does this directory contain a backup of the newly restored version?  
- Or is it the previous `citrea-db` backup just before it was replaced by `citrea-db.tmp-{EPOCH}`?  


**P.S:** All tests are tried on Dev-Net Full-Node EU

**Comment by jfldde** [comment]
> # **Backup & Restore Observations and Questions**
> ## **Backup Creation (`backup_create`)**
> 
>     * Why do we pass the `id` field?
> 
>     * I tested whether it blocks ongoing synchronization and RPC operations. It does not. ✅
> 
> 
> ## **Backup Information (`backup_info`)**
> 
>     * Regardless of which `id` we pass, it always returns all backup information.
> 
>     * If `id` is ignored, what is the purpose of passing it?
> 
> 
> ## **Backup Validation (`backup_validate`)**
> 
>     * If we pass a non-existent path, it fails. Otherwise, it always returns `true`.
> 
>     * The `id` field doesn’t seem to have any effect—tested with `id: -73`, and it still returned `true`.
> 
> 
I think there's a confusion on the API. The three routes expect only a backup path (optional for backup_create if set in config). jsonrpsee typing is a bit loose here and just ignore extra parameters.
id is internal to rocksdb backup mechanism and is used to restore at previous backup iteration. 

> ## **Restore (`restore`)**
> 
>     * I tested restore several times with `--restore-db "/mnt/task/backup"`, and it worked consistently. ✅
> 
>     * **Question:** Can we specify which `id` to restore? It seems to always restore the latest version.
> 
I'll amend so that we can pass an optional id to restore to a previous backup. At the root of the backup dir, you can find a mapping between L2 height and backup id.


> ## **Other Observations (Not necessarily issues, just for awareness)**
> ### **1. Version Tracking Behavior**
> 
>     * Deleted the latest (6th) backup and then restored. It reverted to the 5th version, which became the latest.
> 
>     * After restoring, creating a new backup resulted in version 7.
> 
>     * It seems the system tracks previously created versions even after a restore. Is this expected behavior?
> 

This is probably due to inconsistency between you manually deleting 6th backup and rocksdb backup engine pulling the id from some remaining parts. Just tested with removing the latest on every subdir and it succesfully used the right id

> ### **2. Partial Backup Deletion & Restore Behavior**
> 
>     * Deleted the last 3 versions from `native-db`, last 2 versions from `state`, and the last one from `ledger`.
> 
>     * After restore, each database seemed to restore to its latest available backup.
> 
>     * However, once the restore was completed, an error occurred:
>       `"error": "Timestamp should be greater"`
> 
> 
> **Suggestion:**
> 
>     * We, ops, can validate if all DBs are at the same version before restoring to avoid inconsistency.
> 
>     * This validation doesn’t necessarily have to be handled inside Citrea’s binary, but we need a way to validate individual versions by passing the `id` parameter to `backup_validate`.
> 
The thing is, the backup is made out of 3 different rocksdb backup, one state, one native, one ledger.
If you are to delete - let's say some version of native-db - but don't touch the other ones, they are still valid backups themselves, just not in sync between them anymore. I'll make sure to assert that all backups are at the same expected id on validate. 
[Assert that all subdirs have the same number of incremental backups](https://github.com/chainwayxyz/citrea/pull/1711/commits/7090674b074a407372dee2a476578cd11248fcc8)

> ### **3. Post-Restore Backup Behavior**
> 
> After restore completion, a new directory `citrea-db.bak-{EPOCH}` appears. **Question:**
> 
>     * Does this directory contain a backup of the newly restored version?
> 
>     * Or is it the previous `citrea-db` backup just before it was replaced by `citrea-db.tmp-{EPOCH}`?
> 
> 
This is a backup of the previous citrea-db

**Review by eyusufatik — APPROVED** [review]

**Comment by jfldde** [comment]
Still holding merge, hitting some inconsistency on restore. Flaky backup erroring with :
```
2025-02-13T12:29:03.248201Z ERROR soft_confirmation_rule_enforcer::hooks: error=Timestamp should be greater
2025-02-13T12:29:03.248245Z ERROR citrea_stf::hooks_impl: error=Timestamp should be greater
2025-02-13T12:29:03.248369Z ERROR citrea_fullnode::runner: Could not process L2 block: Timestamp should be greater
```
And 

```
2025-02-13T12:31:02.105776Z ERROR panic: thread 'tokio-runtime-worker' panicked at 'assertion `left == right` failed: Pending head must be set to block 1048, but found block 1050
 ```

PR #1714, opened 2025-01-15
Refactor rollup initialization

rakanalh:
# Description

Problem we have right now is that config along with a lot of other dependencies go through
main -> rollup -> client (sequencer, fullnode ..etc) -> da_block_handler 

So introducing a component in DA block handler that requires a piece of config or a dependency as an example, would require updating the whole flow all the way to that point.

This PR changes this by running RPC & DA block handler as services along side the client (sequencer, fullnode ..etc) not from within. This makes initialization much easier.

# What's been done?
- RPC server is now generalized and is part of `common`
- Main initializes some dependencies
- Each node's crate would be responsible for instantiating the services and returning them (runner, l1 block handler and RPC module (if any))
- Rollup / mod.rs would also help out the designated crate to provide configs / dependencies needed.
- Main spawns services (the runner, L1 block handler, RPC server)

The above makes it easier to pass configs / dependencies towards fullnode's L1 block handler or RPC without having those configs / dependencies go through the runner.

## TODO
- [x] Fix clippy


crates/common/src/da.rs
```diff
@@ -1,20 +1,80 @@
...
-use tokio::sync::Mutex;
-use tracing::warn;
+use tokio::sync::{mpsc, Mutex};
+use tokio::time::sleep;
+use tracing::{error, info, warn};
 
 use crate::cache::L1BlockCache;
 
+pub async fn sync_l1<Da>(
+    start_l1_height: u64,
```
eyusufatik: reaname this to last_scanned_l1_height, since this always starts from height + 1

crates/common/src/da.rs
```diff
@@ -1,20 +1,80 @@
...
+                    Instant::now()
+                        .saturating_duration_since(start)
+                        .as_secs_f64(),
+                );
+                if let Err(e) = sender.send(l1_block).await {
+                    error!("Could not notify about L1 block: {}", e);
+                    continue 'block_sync;
+                }
+            }
+        }
```
eyusufatik: I also want to refactor this part of the code

crates/common/src/rpc/server.rs
eyusufatik, line 1: make sure there was no difference between how sequencer, batch prover, full node, light client prover's rpc servers were launched

bin/citrea/src/cli.rs
```diff
@@ -0,0 +1,132 @@
...
+    #[arg(long, short = 'q', action)]
+    pub(crate) quiet: bool,
+}
+
+pub(crate) enum RollupClient {
+    Sequencer(SequencerConfig),
+    FullNode,
+    BatchProver(BatchProverConfig),
+    LightClientProver(LightClientProverConfig),
+}
```
eyusufatik: I'd call this NodeType.

bin/citrea/src/rollup/mod.rs
```diff
@@ -1,98 +1,159 @@
...
+
+/// Group for storage instances
+pub struct Storage<T: RollupBlueprint> {
+    /// The ledger DB instance
+    pub ledger_db: LedgerDB,
+    /// The prover storage manager instance.
+    pub storage_manager: ProverStorageManager<<T as RollupBlueprint>::DaSpec>,
+    /// The prover storage
+    pub prover_storage: ProverStorage<SnapshotManager>,
+}
```
eyusufatik, line 56: prover_storage = storage_manager.create_finalized_storage()

so maybe we can just have ledger db and storage manager, what do you think?

bin/citrea/src/rollup/mod.rs
```diff
@@ -145,81 +206,40 @@ pub trait CitreaRollupBlueprint: RollupBlueprint {
...
         .unwrap();
 
-        Ok((seq, rpc_methods))
+        Ok(seq)
     }
 
     /// Creates a new rollup.
     #[instrument(level = "trace", skip_all)]
-    async fn create_new_rollup(
+    async fn create_rollup(
```
eyusufatik: while at it let's rename this function to create_new_full_node. because that's what it does.

bin/citrea/src/rollup/mod.rs
```diff
@@ -506,6 +464,67 @@ pub trait CitreaRollupBlueprint: RollupBlueprint {
...
+    fn create_rollup_l1_block_handler(
+        task_manager: &mut TaskManager<()>,
+        ledger_db: LedgerDB,
+        da_service: Arc<<Self as RollupBlueprint>::DaService>,
+        public_keys: RollupPublicKeys,
+        code_commitments_by_spec: HashMap<SpecId, <Self::Vm as Zkvm>::CodeCommitment>,
+        start_l1_height: u64,
+    ) -> FullNodeL1BlockHandler<Self::NativeContext, Self::Vm, Self::DaService, RootHash, LedgerDB>
+    {
+        let l1_block_handler = FullNodeL1BlockHandler::new(
```
eyusufatik: create_full_node_l1_block_handler.

also aren't we going to spawnt da block handlers for sequencer and light client prover here?

eyusufatik (changes requested):
left my initial comments, will need to read rollup/mod.rs a bit more in detail.

bin/citrea/src/rollup/mod.rs
rakanalh: We create this once and share it with both RPC & the node runner

crates/common/src/da.rs
rakanalh: What changes are you looking to have done?

bin/citrea/src/cli.rs
```diff
@@ -0,0 +1,132 @@
...
+    if batch_prover_config.is_some() && light_client_prover_config.is_some() {
+        return Err(anyhow::anyhow!(
+            "Cannot run in both batch prover and light client prover mode at the same time"
+        ));
+    }
+    if light_client_prover_config.is_some() && sequencer_config.is_some() {
+        return Err(anyhow::anyhow!(
+            "Cannot run in both light client prover and sequencer mode at the same time"
+        ));
+    }
```
jfldde: Wouldn't that be already handled by `clap` `conflicts_with_all` above ?

rakanalh: Yes you are correct. I will update.

crates/common/src/rpc/server.rs
rakanalh: They are launched the same way. The difference is the fact that we register additional crate-specific RPC modules / methods.

bin/citrea/src/cli.rs
```diff
@@ -0,0 +1,116 @@
...
+    };
+
+    if let Some(sequencer_config) = sequencer_config {
+        return Ok(NodeType::Sequencer(sequencer_config));
+    } else if let Some(batch_prover_config) = batch_prover_config {
+        return Ok(NodeType::BatchProver(batch_prover_config));
+    } else if let Some(light_client_prover_config) = light_client_prover_config {
+        return Ok(NodeType::LightClientProver(light_client_prover_config));
+    }
+    Ok(NodeType::FullNode)
```
jfldde, line 115: Wish we had it the other way round and passed a `--kind` arg or something. Would let us have a single config field (+ the rollup config) and not be implicit about which node kind is running. 
We would match on the kind here and get the config accordingly.

bin/citrea/tests/test_helpers/mod.rs
```diff
@@ -70,77 +121,147 @@ pub async fn start_rollup(
...
             &mock_demo_rollup,
-            &rt_genesis_paths,
+            genesis_config,
             rollup_config.clone(),
+            da_service,
+            ledger_db,
+            storage_manager,
+            prover_storage,
+            soft_confirmation_channel.0,
+            rpc_module,
```
jfldde: Could remove it here since we return it as is

crates/common/src/rpc/server.rs
```diff
@@ -0,0 +1,75 @@
...
+use jsonrpsee::server::{BatchRequestConfig, RpcServiceBuilder, ServerBuilder};
+use jsonrpsee::RpcModule;
+use tokio::sync::oneshot;
+use tracing::{error, info};
+
+use crate::tasks::manager::TaskManager;
+use crate::RpcConfig;
+
+/// Starts a RPC server with provided rpc methods.
+pub async fn start_rpc_server(
```
jfldde: IIRC the different nodes have subtle differences between them here

bin/citrea/src/cli.rs
rakanalh: I think the way it is right now is:
```
--sequencer
```
which uses the sequencer from env config, or
```
--sequencer /path/to/config
```
which reads the file.

This applies to everything. I can't say i am a fan of this but this is probably something we can improve in the future.

crates/common/src/da.rs
```diff
@@ -1,19 +1,81 @@
...
+        for block_number in last_scanned_l1_height + 1..=new_l1_height {
+            let l1_block =
+                match get_da_block_at_height(&da_service, block_number, l1_block_cache.clone())
+                    .await
+                {
+                    Ok(block) => block,
+                    Err(e) => {
+                        error!("Could not fetch last finalized L1 block: {}", e);
+                        sleep(Duration::from_secs(2)).await;
+                        continue 'block_sync;
```
jfldde: We could drop the outter loop label and just break here and below ?

crates/common/src/da.rs
```diff
@@ -88,3 +150,24 @@ pub async fn get_initial_slot_height(client: &HttpClient) -> u64 {
...
+        Some(height) => height.0,
+        None => {
+            let runner_config = rollup_config
+                .runner
+                .clone()
+                .expect("Runner config should be set");
+            let sequencer_client =
+                HttpClientBuilder::default().build(runner_config.sequencer_client_url)?;
+            get_initial_slot_height(&sequencer_client).await
+        }
```
jfldde, line 175: This seems to be the culprit for batch-prover not becoming responsive as seen here https://github.com/chainwayxyz/citrea/actions/runs/12887840271/job/35931437541?pr=1714.
This can be solved by :
```
pub async fn get_initial_slot_height(client: &HttpClient) -> u64 {
    loop {
        match client.get_soft_confirmation_by_number(U64::from(1)).await {
            Ok(Some(batch)) => return batch.da_slot_height,
            Ok(None) => return 0,
            _ => {
                // sleep 1
                tokio::time::sleep(std::time::Duration::from_secs(1)).await;
                continue;
            }
        }
    }
}
```
But not sure of the correctness of it.

rakanalh: Yea... i figured that the full node is trying to get the initial L1 slot but hangs because the sequencer has not produced any blocks at the point where that is requested. This is handled different than the way you suggested.

bin/citrea/src/rollup/mod.rs
jfldde: I'd expect it to hold `Option<MmrDb>` as well

crates/light-client-prover/src/runner.rs
```diff
@@ -1,209 +1,26 @@
...
-                light_client_proof_commitment,
-                light_client_proof_elfs,
-                mmr_db,
-            );
-            l1_block_handler
-                .run(starting_block, cancellation_token)
-                .await
-        });
-
+    pub async fn run(&mut self, task_manager: TaskManager<()>) -> Result<(), anyhow::Error> {
```
ercecan, line 20: Nit: The name of the function does not make sense now since it does not run anything, but I guess we can still keep it for compatibility

rakanalh: yea it's consistent with the other crate runners. We don't need the runner at all for the light client... but also kept it for consistency

ercecan (approved):

ercecan (commented):
LGTM, I also agree on @jfldde comment for Storage struct to hold optional mmrdb

bin/citrea/src/rollup/mod.rs
rakanalh: MMR is specific to the light client prover which is constructed when we know we need to spawn the light client. However, storage manager and prover storage are needed for every node type.

bin/citrea/src/main.rs
```diff
@@ -100,57 +48,7 @@ async fn main() -> Result<(), anyhow::Error> {
...
-        return Err(anyhow::anyhow!(
-            "Cannot run in both batch prover and light client prover mode at the same time"
-        ));
-    }
-    if light_client_prover_config.is_some() && sequencer_config.is_some() {
-        return Err(anyhow::anyhow!(
-            "Cannot run in both light client prover and sequencer mode at the same time"
-        ));
-    }
+    let client = node_type_from_args(&args)?;
```
eyusufatik: node_type

eyusufatik (approved):
I hope this doesn't break anything. please launch different type of nodes on testnet on your computer. and look out for logs.

rakanalh:
> I hope this doesn't break anything. please launch different type of nodes on testnet on your computer. and look out for logs.

I ran both full node & batch prover, they are both syncing without issues.
Full node:
```
./target/debug/citrea --network testnet --da-layer bitcoin --rollup-config-path resources/configs/testnet/rollup_config.toml --genesis-paths resources/genesis/testnet
```
Batch Prover:
```
PARALLEL_PROOF_LIMIT=1 ./target/debug/citrea --network testnet --da-layer bitcoin --batch-prover resources/configs/testnet/prover_config.toml --rollup-config-path resources/configs/testnet/prover_rollup_config.toml --genesis-paths resources/genesis/testnet
```

image

PR #1715, opened 2025-01-15
Override compute_wtxid() and compute_txid() functions with patched sha256 impl

### LINKED ISSUE #1710: Override compute_wtxid() and compute_txid() functions with patched sha256 impl.
The rust-bitcoin library's `Transaction::compute_wtxid()` or (.wtxid()) functions use custom implementation of sha256.

We can reduce cycle counts with using defining custom functions for the functions that use patched sha256 implementations.

kpp:
# Description

Use our `calculate_double_sha256()` to compute txid/wtxid in zk context.

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/1710


crates/bitcoin-da/src/helpers/mod.rs
```diff
@@ -98,3 +100,36 @@ pub fn calculate_sha256(input: &[u8]) -> [u8; 32] {
     hasher.update(input);
     hasher.finalize().into()
 }
+
+/// Computes the [`Txid`].
+///
+/// Hashes the transaction **excluding** the segwit data (i.e. the marker, flag bytes, and the
+/// witness fields themselves). For non-segwit transactions which do not have any segwit data,
+/// this will be equal to [`Transaction::compute_wtxid()`].
+pub fn citrea_txid(tx: &Transaction) -> [u8; 32] {
```
eyusufatik: lets call this calculate_wtxid. don't want people to think we somehow calculate these differently.

crates/bitcoin-da/src/service.rs
eyusufatik, line 1: don't need to change non-guest side calculations

eyusufatik:
hold-merge label for two reasons:

- @kpp wants to write unit tests for the new functions.
- if we merge this now, the batch proof elfs on testnet and devnet will defer in behaviour -- which might be ok, need to think about it.

eyusufatik (approved):
aproving but don't merge

eyusufatik:
since v0.6.1 is out

we can start merging breaking changes.

let's go.

PR #1729, opened 2025-01-21
Separate ledgerdb tables

### LINKED ISSUE #1619: Separate ledger db tables for nodes
Right now all of our nodes have the same ledger db tables even if they never ever use them, so when implementing migration for a table name change for example for prover 
VerifiedProofs -> VerifiedBatchProofs
We need to implement this migration for every node, which does not make any sense at all. Instead I propose all nodes have their own tables. 
There can still be common tables of course, but other than that the node should have their tables as extra

ercecan:
# Description
Implements separate ledger db tables for nodes so in the future migrations are easied and made only for nodes that involve them

## Linked Issues
- Fixes #1619

## Testing
Tested all nodes manually by first spinning up nodes on 0.6.1 and then on this branch, did not encounter any errors


crates/batch-prover/src/db_migrations/remove_unused_common_tables.rs
```diff
@@ -0,0 +1,33 @@
...
+        ledger_db: Arc<LedgerDB>,
+        tables_to_drop: &mut Vec<String>,
+    ) -> anyhow::Result<()> {
+        // Get difference of LEDGER_TABLES and SEQUENCER_LEDGER_TABLES and drop them
+        let mut diff = LEDGER_TABLES.to_vec();
+        diff.retain(|x| !BATCH_PROVER_LEDGER_TABLES.contains(x));
+        let diff_tables = diff.iter().map(|x| x.to_string()).collect::<Vec<_>>();
+        for table in diff_tables {
+            // Check if table exists in the database
+            if let Ok(_cf_handle) = ledger_db.get_cf_handle(&table) {
```
jfldde:
```suggestion
            if ledger_db.get_cf_handle(&table).is_ok() {
```

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/migrations/mod.rs
```diff
@@ -56,7 +55,11 @@ impl<'a> LedgerDBMigrator<'a> {
     }
 
     /// Run migrations
-    pub fn migrate(&self, max_open_files: Option<i32>) -> anyhow::Result<()> {
+    pub fn migrate(
+        &self,
+        max_open_files: Option<i32>,
+        node_column_families: Vec<String>,
```
kpp, line 61: node_column_families can be `&[&str]`, that will save a lot of code like:

```
            .iter()
            .map(|table| table.to_string())
            .collect::<Vec<_>>();
```

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
```diff
@@ -220,14 +220,6 @@ impl SharedLedgerOps for LedgerDB {
         Ok(())
     }
 
-    /// Gets l1 height of l1 hash
```
kpp: Why this code was touched? Please avoid unnecessary changes <3

kpp (approved):

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/migrations/mod.rs
ercecan: tried it but kind of made it worse inside

jfldde (approved):

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
ercecan: #1735

eyusufatik (approved):

PR #1734, opened 2025-01-21
Get rid of RlpTransaction in Evm call

### LINKED ISSUE #1522: Optimization: Get rid of RlpEvmTransaction
Let's try to get rid of `RlpEvmTransaction` when we try to decode it in zk context. We call either `TransactionSigned::decode_enveloped` or `TransactionSigned::decode_2718` in a zk context whilst we could pass `TransactionSignedNoHash` in evm::Call.

kpp:
# Description

Get rid of Rlp encoding (not really actually =() in Evm::call.

I don't know if this PR is actually worth merging due to `unsafe { core:: mem::transmute }`.

TODO: measure cycle count diff on batch proof with a lot of evm txs.

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/1522

### DIFF
--- a/crates/evm/src/borsh_compat.rs
+++ b/crates/evm/src/borsh_compat.rs
@@ -1,0 +1,531 @@
+use alloy_consensus::{TxEip1559, TxEip2930, TxEip4844, TxEip7702, TxLegacy};
+use alloy_eips::eip7702::{Authorization, SignedAuthorization};
+use alloy_primitives::{Parity, Signature, TxKind, U256};
+use alloy_rpc_types::{AccessList, AccessListItem};
+use borsh::{BorshDeserialize, BorshSerialize};
+use bytes::Bytes;
+use reth_primitives::{Transaction, TransactionSigned};
+
+#[derive(BorshDeserialize, BorshSerialize)]
+enum CompatParity {
+    Eip155(u64),
+    NonEip155(bool),
+    Parity(bool),
+}
+
+impl From<Parity> for CompatParity {
+    fn from(value: Parity) -> Self {
+        match value {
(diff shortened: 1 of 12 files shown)


crates/evm/src/borsh_compat.rs
```diff
@@ -0,0 +1,531 @@
...
+    assert_type::<SignedAuthorization, CompatSignedAuthorization>();
+    assert_type::<Parity, CompatParity>();
+    assert_type::<AccessList, CompatAccessList>();
+    assert_type::<AccessListItem, CompatAccessListItem>();
+    assert_type::<TxKind, CompatTxKind>();
+
+    // Unfortunately `xs.into_iter().map(Into::into).collect()` is not
+    //  optimized due to rustc limitations.
+
+    unsafe { core::mem::transmute(txs) }
```
kpp, line 514: sic! Unfortunately this is a smelly code due to https://github.com/rust-lang/rust/issues/135962.

workingjubilee: As I replied:
> For any pair of pairs `(A0, B0)` and `(A1, B1)`, it is permitted for the physical ordering to be `(A0, B0)` and `(B1, A1)`.

Therefore this is unsound.

workingjubilee (commented):
I cannot advise doing this, as doing it mostly serves as motivation to write the code in our layout algorithm that would tend to break this in actual fact.

kpp:
Unfortunately we have to close this PR due to the lack of optimization in rustc and thus making our casting from `Vec<CompatSignedTransaction> -> Vec<SignedTransaction` a very huge pessimization.

PR #1744, opened 2025-01-24
enable risc0 unstable feature and new precompiles

eyusufatik:
# Description
enables new precompiles.
in bitcoin-da replaces secp256k1 on the guest side with k256 crate, which has a new precompile.

### DIFF
--- a/crates/bitcoin-da/Cargo.toml
+++ b/crates/bitcoin-da/Cargo.toml
@@ -26,11 +26,12 @@ crypto-bigint = { workspace = true }
 hex = { workspace = true, features = ["serde"] }
 itertools = { workspace = true }
 jsonrpsee = { workspace = true, optional = true }
+k256 = { workspace = true }
 metrics = { workspace = true, optional = true }
 pin-project = { workspace = true, optional = true, features = [] }
 rand = { workspace = true }
 reqwest = { workspace = true, optional = true }
-secp256k1 = { workspace = true }
+secp256k1 = { workspace = true, optional = true }
 serde = { workspace = true }
 serde_json = { workspace = true, features = ["raw_value"] }
 thiserror = { workspace = true }
@@ -43,6 +44,7 @@ sha2 = { workspace = true }
 bitcoincore-rpc = { workspace = true, optional = true }
 
 [dev-dependencies]
+bitcoin-da = { path = ".", features = ["native"] }
 citrea-e2e = { workspace = true }
 
 [features]
@@ -59,4 +61,5 @@ native = [
   "dep:bitcoincore-rpc",
   "dep:reqwest",
   "dep:jsonrpsee",
+  "dep:secp256k1",
 ]
--- a/crates/bitcoin-da/src/helpers/parsers.rs
+++ b/crates/bitcoin-da/src/helpers/parsers.rs
@@ -5,10 +5,8 @@ use bitcoin::opcodes::all::OP_CHECKSIGVERIFY;
 use bitcoin::script::Instruction::{Op, PushBytes};
 use bitcoin::script::{Error as ScriptError, PushBytes as StructPushBytes};
 use bitcoin::{Opcode, Script, Transaction};
-use secp256k1::{self, ecdsa, Message, SECP256K1};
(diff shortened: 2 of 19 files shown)


crates/bitcoin-da/src/helpers/parsers.rs
```diff
@@ -71,21 +69,16 @@ pub trait VerifyParsed {
...
-                .is_ok()
-        {
-            Some(hash)
-        } else {
-            None
+        if let Ok(key) = k256::ecdsa::VerifyingKey::from_sec1_bytes(self.public_key()) {
+            use k256::ecdsa::signature::DigestVerifier;
+            let hash = sha2::Sha256::new_with_prefix(self.body());
+            let signature = k256::ecdsa::Signature::from_slice(self.signature());
+            if signature.is_ok() && key.verify_digest(hash.clone(), &signature.unwrap()).is_ok() {
```
eyusufatik, line 76: @kpp will this .clone() clone the hash only or the whole body?

kpp: Hash only. I am surprised it does not implement Copy, but it's okay, its cheap.

guests/risc0/Dockerfile
```diff
@@ -11,6 +11,7 @@ ENV RUSTFLAGS="-C passes=loweratomic -C link-arg=-Ttext=0x00200800 -C link-arg=-
 ENV CARGO_TARGET_DIR="target"
 ENV CC_riscv32im_risc0_zkvm_elf="/root/.local/share/cargo-risczero/cpp/bin/riscv32-unknown-elf-gcc"
 ENV CFLAGS_riscv32im_risc0_zkvm_elf="-march=rv32im -nostdlib"
+ENV RISC0_FEATURE_bigint2=""
```
kpp, line 14: What's the point of this if our .lock files contain r0_bigint2 feature? Can this env be used to disable this feature?

eyusufatik: it tells cargo +risc0 build command to build with unstable enabled I guess. without that the build panics. I've checked risc0-build code and the docker file created there has this env set.

crates/bitcoin-da/Cargo.toml
```diff
@@ -26,11 +26,12 @@ crypto-bigint = { workspace = true }
 hex = { workspace = true, features = ["serde"] }
 itertools = { workspace = true }
 jsonrpsee = { workspace = true, optional = true }
+k256 = { workspace = true }
 metrics = { workspace = true, optional = true }
 pin-project = { workspace = true, optional = true, features = [] }
 rand = { workspace = true }
 reqwest = { workspace = true, optional = true }
-secp256k1 = { workspace = true }
+secp256k1 = { workspace = true, optional = true }
```
kpp, line 34: Where do we use secp256k1 crate after this PR?

guests/risc0/light-client-proof/bitcoin/Cargo.lock
```diff
@@ -2035,6 +2191,16 @@ dependencies = [
  "sov-rollup-interface",
 ]
 
+[[package]]
+name = "spki"
```
kpp, line 2195: I am not sure where this comes from along with the der crate. Maybe we can avoid those deps? If not, it's not a big deal

kpp (approved):

crates/bitcoin-da/Cargo.toml
eyusufatik: on the native side, like tx builders. because I want to make sure the new guest functions are compatible with old signatures and pubkeys etc. 

so the native side still uses old signing, whereas the guest uses new signing

guests/risc0/light-client-proof/bitcoin/Cargo.lock
eyusufatik: not sure

eyusufatik: i think it's k256 related. batch proof already had it through evm. but light client prover is now getting it through bitcoin-da

PR #1746, opened 2025-01-25
NativeDB pruning

### LINKED ISSUE #1198: Pruning: ProverStorageManager
`ProverStorageManager` implements the trait `HierarchicalStorageManager` which provides sequencer / full node / prover with the following methods:

```rust
    fn create_storage_on_l2_height(
        &mut self,
        l2_block_height: u64,
    ) -> anyhow::Result<Self::NativeStorage>;

    /// Finalizes snapshot on given l2 block height
    fn finalize_l2(&mut self, l2_block_height: u64) -> anyhow::Result<()>;

    /// Adds [`Self::NativeChangeSet`] to the storage.
    fn save_change_set_l2(
        &mut self,
        l2_block_height: u64,
        change_set: Self::NativeChangeSet,
    ) -> anyhow::Result<()>;
```

Some of which commit snapshots to DBs. Snapshots as apparent in the methods are tied to L2 block heights. We'd like to prune snapshots while pruning L2 blocks.

rakanalh:
# Description

This PR implements pruning for old native DB records.

When starting at block #1, the version that gets created for all the keys that are stored into the accessory table are assigned 2 which means that the version is ahead of block number by one.

That being said, the current way of pruning is inefficient. Since we have to iterate all entries in the native DB table and compare the version of the key. Unfortunately, there is no way the pruner could figure out which keys specifically changed between two pruning cycles to be able to prune just those.

However, each time the number of entries in this table grows, the pruning process will decrease them significantly which means that the next cycle will need to iterate fewer items.

## Linked Issues
- Fixes #1198


crates/pruning/src/pruners/native_db.rs
```diff
@@ -0,0 +1,32 @@
...
+
+    iter.seek_to_first();
+
+    let mut counter = 1u32;
+    // TODO: Think about a better way to do this as the bigger the table, the slower this will be.
+    while let Some(Ok(entry)) = iter.next() {
+        let version = entry.key.1;
+        // The version value is always ahead of block number by one.
+        if version < up_to_block + 1 {
+            if let Err(e) = native_db.delete::<ModuleAccessoryState>(&entry.key) {
```
eyusufatik: i think we have batch update APIs over rocks db, so we can keep what keys to delete in memory than delete them all in a big go

bin/citrea/tests/e2e/pruning.rs
```diff
@@ -0,0 +1,80 @@
...
+    )
+    .catch_unwind()
+    .await;
+    assert!(panic_2.is_err());
+
+    // Should NOT panic as the data we're requesting here is correct
+    full_node_test_client
+        .eth_get_block_by_number_with_detail(Some(BlockNumberOrTag::Number(21)))
+        .await;
+
```
eyusufatik, line 151: let's do one more round of pruning.

after a round of pruning:

let's call these RPCs also:

- eth_getTransactionByHash
- eth_getBlockReceipts
- eth_getBlockByHash

rakanalh: Done. Added calls to these 3 RPCs and introduced 2 more rounds of pruning

eyusufatik (approved):

PR #1751, opened 2025-01-27
Handle SIGTERM in sequencer

### LINKED ISSUE #1750: Sequencer Service Missing ‘Shutting Down’ Logs: Investigate Graceful Shutdown Behavior
Hey guys

While investigating the incident that occurred on January 26, 2025, I wanted to confirm whether the Citrea services were truly gracefully shutting down. During my research, I ran a query on the logs for each service (sequencer, full-node, batch-prover, and light-prover) from the past 6 months, searching for entries containing the words “shut” or “Shut” (like *shut* or *Shut*).

I encountered an interesting observation: the sequencer has not gracefully shut down even once in the past 6 months. Specifically, it has never logged a message indicating “shutting down.”

Could we make sure if the sequencer can shut down gracefully or not?

jfldde:
# Description

- Handle SIGTERM clean shutdown in sequencer

## Linked Issues
- Fixes #1750

### DIFF
--- a/crates/sequencer/src/runner.rs
+++ b/crates/sequencer/src/runner.rs
@@ -41,6 +41,7 @@ use sov_rollup_interface::stf::StateTransitionFunction;
 use sov_state::ProverStorage;
 use sov_stf_runner::InitParams;
 use tokio::signal;
+use tokio::signal::unix::{signal, SignalKind};
 use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver};
 use tokio::sync::{broadcast, mpsc};
 use tokio::time::sleep;
@@ -626,6 +627,7 @@ where
         let mut block_production_tick = tokio::time::interval(target_block_time);
         block_production_tick.tick().await;
 
+        let mut sigterm = signal(SignalKind::terminate()).expect("failed to setup SIGTERM handler");
         loop {
             tokio::select! {
                 // Receive updates from DA layer worker.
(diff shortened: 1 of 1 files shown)


rakanalh (changes requested):
This is not the idea solution IMO.

Here is what i was thinking.

1. Remove all signal handling from all nodes.
2. Instead of passing the task manager into the `run` method of the runner, pass a cancellation token.
3. In `main`, using task manager, spawn runner just like the other services passing cancellation token.
4. Introduce method `wait` inside the task manager which is called at the end of main which listens for sigterm signal + ctrl_c.

This unifies the way signals are handled across different node types and now it is possible since the services / tasks are spawned outside of each other.

eyusufatik (commented):
if we don't go with rakan's sugesstion, pls use `common::utils::create_shutdown_signal`

eyusufatik:
> This is not the idea solution IMO.
> 
> Here is what i was thinking.
> 
> 1. Remove all signal handling from all nodes.
> 2. Instead of passing the task manager into the `run` method of the runner, pass a cancellation token.
> 3. In `main`, using task manager, spawn runner just like the other services passing cancellation token.
> 4. Introduce method `wait` inside the task manager which is called at the end of main which listens for sigterm signal + ctrl_c.
> 
> This unifies the way signals are handled across different node types and now it is possible since the services / tasks are spawned outside of each other.

agree on this btw. might be a better solution

jfldde:
Closing for https://github.com/chainwayxyz/citrea/pull/1751#pullrequestreview-2575039654

PR #1753, opened 2025-01-27
Fix inconsistent db naming

jfldde:
# Description

- Fix DB_NAME in place of DB_PATH_SUFFIX in NativeDB
- Declare and use LEDGER_DB_NAME in line with the other dbs

## Caveat
Will require a reindex

### DIFF
--- a/crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/migrations/mod.rs
+++ b/crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/migrations/mod.rs
@@ -8,7 +8,7 @@ use tracing::{debug, error, info};
 
 use super::migrations::utils::{drop_column_families, list_column_families};
 use super::LedgerDB;
-use crate::ledger_db::{SharedLedgerOps, LEDGER_DB_PATH_SUFFIX};
+use crate::ledger_db::SharedLedgerOps;
 use crate::rocks_db_config::RocksdbConfig;
 
 /// Utilities for ledger db migrations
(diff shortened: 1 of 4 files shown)


eyusufatik (commented):
not sure if this is worth a reindex on everythin

jfldde:
Re-index could be bypassed by renaming `native-db` to `native`

rakanalh (commented):
Since we're doing this... i have been thinking about separating paths by network.

Example:
Instead of saving to `resources/dbs/native/` or `state` ... etc
we do: `resources/dbs/mainnet/{native,state,ledger} ... etc

Where `make clean-node` cleans `CITREA_NETWORK` and by default `nightly`.

This would make it possible for us to sync our nodes and not have to delete / move the files manually when we're switching different networks.

jfldde:
> Since we're doing this... i have been thinking about separating paths by network.
> 
> Example: Instead of saving to `resources/dbs/native/` or `state` ... etc we do: `resources/dbs/mainnet/{native,state,ledger} ... etc
> 
> Where `make clean-node` cleans `CITREA_NETWORK` and by default `nightly`.
> 
> This would make it possible for us to sync our nodes and not have to delete / move the files manually when we're switching different networks.

Shouldn't that be handled by the config ? We're already needing a config by network since we're setting da.network there

kpp (approved):

kpp:
> Since we're doing this... i have been thinking about separating paths by network.

Let's think how to unify all dbs into one so we can't have inconsistency between them?

rakanalh:
> > Since we're doing this... i have been thinking about separating paths by network.
> 
> Let's think how to unify all dbs into one so we can't have inconsistency between them?

Something like that could work if we introduce the `network` field to ledger DB tables and update all queries to match against current network. But native & state DBs are going to be very HARD to unify.

IMO, separating paths is much much easier.

eyusufatik (commented):
imo no need to merge unless there is a way for nodes to migrate. don't want that hassle for this kind of thing

jfldde:
Closing as too low priority

PR #1755, opened 2025-01-28
Light client prover parallel proof limit defaults to 1

### LINKED ISSUE #1722: Do not enforce `PARALLEL_PROOF_LIMIT` env in light client
Light client will always prove synchronously. Provide the default value of 1.

ercecan:
# Description
Light client prover parallel proof limit defaults to 1

## Linked Issues
- Fixes #1722


jfldde (approved):
Nit but it doesn't really defaults if it's hard set to 1 and we cannot override it

eyusufatik:
> Nit but it doesn't really defaults if it's hard set to 1 and we cannot override it

the reasoning is the light client prover won't need parallel proving at all.

(maybe in the future if we do up to date light client proofs and hanle every branch of the header chain)

eyusufatik (approved):

PR #1756, opened 2025-01-28
Handle shutdown signals

### LINKED ISSUE #1750: Sequencer Service Missing ‘Shutting Down’ Logs: Investigate Graceful Shutdown Behavior

rakanalh:
# Description
Instead of passing `task_manager` into node runners, `spawn` those node runners using the task manager and use the latter to handle shutdown signals for every node type.

## Linked Issues
- Fixes #1750

### DIFF
--- a/bin/citrea/src/main.rs
+++ b/bin/citrea/src/main.rs
@@ -103,6 +103,7 @@ where
     DaC: serde::de::DeserializeOwned + DebugTrait + Clone + FromEnv + Send + Sync + 'static,
     S: CitreaRollupBlueprint<DaConfig = DaC>,
     <<S as RollupBlueprint>::NativeContext as Spec>::Storage: NativeStorage,
+    <S as RollupBlueprint>::NativeRuntime: 'static,
 {
     let rollup_config: FullNodeConfig<DaC> = match rollup_config_path {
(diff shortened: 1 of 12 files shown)


eyusufatik (commented):
snice nothing calls abort now, we don't wait 5 secs before shutting down? do we

rakanalh:
> snice nothing calls abort now, we don't wait 5 secs before shutting down? do we

Am fixing this as we speak.

rakanalh:
- Ctrl-C
- `kill -term PID`
- `kill -int PID` 

All invoke the shutdown signal.

eyusufatik (approved):

jfldde (approved):
Looks great

PR #1761, opened 2025-01-29
Add guest cycle count test with input from file.

### LINKED ISSUE #1467: Add a new test to output the RISC0 cycle count
We should also store nightly cycle count on push, and compare the PR's effect on the cycle count.

We can color based on the change of cycle count.

We should generate inputs and push them to repository and use those inputs for benchmark purposes.

yaziciahmet:
# Description

Adds a test to both generate an input file for the guest circuit, and also the test to build and run the circuit with the input file. Input files are not committed to repo since it is above 500MB. If anyone wants or needs to test cycles, you can download it from our private S3.

## Linked Issues
- Fixes #1467 

## Testing

## Docs


jfldde:
Might want to look into the `perf.yml` and `cycle-diff.sh` script. This can either be used in addition to this PR or removed if not relevant anymore

eyusufatik (approved):

PR #1763, opened 2025-01-30
fix: Safe handling of empty Jellyfish Merkle Tree initialization

crStiv:
This commit improves the handling of empty Jellyfish Merkle Trees (JMT) in the prover storage implementation:

- Adds proper version checking for empty trees
- Prevents unexpected empty roots in non-zero versions
- Improves error handling with context
- Optimizes root hash retrieval
- Adds better error messages

This change ensures safe operation with snapshots by properly validating tree state
and version numbers during initialization.

Resolves TODO: "Fix this before introducing snapshots!"

### DIFF
--- a/crates/sovereign-sdk/module-system/sov-state/src/prover_storage.rs
+++ b/crates/sovereign-sdk/module-system/sov-state/src/prover_storage.rs
@@ -129,23 +129,21 @@ where
         let latest_version = self.version();
         let jmt = JellyfishMerkleTree::<_, DefaultHasher>::new(&self.db);
 
-        // Safe initialization of an empty tree considering snapshots
+        // Handle empty jmt during genesis
         let prev_root = if let Some(root) = jmt.get_root_hash_option(latest_version)? {
             root
         } else {
-            // Initialize an empty tree only if this is truly the first version
-            if latest_version != 0 {
-                anyhow::bail!("Root hash not found for non-zero version {}", latest_version);
-            }
+            // We should only get here during genesis
+            assert_eq!(latest_version, 0, "Root hash missing for non-genesis version");
             
             let empty_batch = Vec::default().into_iter();
             let (root, tree_update) = jmt
                 .put_value_set(empty_batch, latest_version)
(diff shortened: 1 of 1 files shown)


crates/sovereign-sdk/module-system/sov-state/src/prover_storage.rs
```diff
@@ -129,22 +129,27 @@ where
...
+                .context("Failed to initialize empty JMT")?;
 
             self.db
                 .write_node_batch(&tree_update.node_batch)
-                .expect("db write must succeed");
-        }
-        let prev_root = jmt
-            .get_root_hash(latest_version)
-            .expect("Previous root hash was just populated");
+                .context("Failed to write initial empty tree")?;
```
eyusufatik: we only end up running this else block if we are running genesis, or if we forget to run genesis.

so imo, let's revert back to assert and expect in the else blocks and we are good to go.

eyusufatik: the reasoning is that there is no need to keep the node running if it cant even make genesis.

eyusufatik (changes requested):

crates/sovereign-sdk/module-system/sov-state/src/prover_storage.rs
crStiv: > the reasoning is that there is no need to keep the node running if it cant even make genesis.

@eyusufatik Thanks for the feedback! I agree that there's no point in graceful error handling if we can't even create genesis. 

I've switched back to `assert` and `expect` since the node should just crash if genesis fails. Also added some comments to make it clear this is genesis-only code.

Let me know if this looks better now!

eyusufatik (approved):

eyusufatik:
closing PR as bot activity

crStiv:
> closing PR as bot activity

I provided some help to your project, and youclosed my pr, ok that's no problem. But I just wonder why do you consider my activity as bot's?

PR #1800, opened 2025-02-03
use btreemap in db commit method

yaziciahmet:
# Description
Due to HashMap, the order of iteration changes for native and zkvm, I have no idea how it didn't fail so far. @eyusufatik opened a pr to revm to have a version which uses btreemap, until then, we manually convert.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


eyusufatik:
>  have no idea how it didn't fail so far. @eyusufatik opened a pr to revm to have a version which uses btreemap, until then, we manually convert.

just for the records, PR to revm actually is to decrease cycle counts rather than have this fix

eyusufatik (approved):

Issue #1609, opened 2024-12-14
Rollback command line argument


eyusufatik:
We’ve had multiple times, on our own infra, that full nodes or the prover database got corrupted, we had to resync to the chain.

This won’t be acceptable on mainnet, what are we gonna do if the sequencer’s database gets corrupted?  
  
Ideally, there should be two different rollback features.

1. Invoked with a cli arg like so

   `—start-L2-sync-from L2_HEIGHT —start-L1-sync from L1_HEIGHT` 

   This will make sure everything above certain versions for the 3 databases we have will be deleted.

2. On startup, we check the latest versions of the 3 databases, and if we detect an anomaly, we prune the “higher” table, and start syncing form “lower” table.

PR #1808, opened 2025-02-04
Incomplete State DB pruning

### LINKED ISSUE #1193: Pruning: EVM

rakanalh:
# Description

- Add state DB pruning
- Introduce `citrea-cli`
- Add skeleton for rollback

## Linked Issues
- Related #1193


bin/cli/src/commands/prune.rs
```diff
@@ -0,0 +1,38 @@
...
+        distance
+    );
+    let config = PruningConfig { distance };
+
+    let rocksdb_config = RocksdbConfig::new(&db_path, None, None);
+    let ledger_db = LedgerDB::with_config(&rocksdb_config)?;
+    let native_db = NativeDB::<SnapshotManager>::setup_schema_db(&rocksdb_config)?;
+    let state_db = StateDB::<SnapshotManager>::setup_schema_db(&rocksdb_config)?;
+
+    let Some((soft_confirmation_number, _)) = ledger_db.get_head_soft_confirmation()? else {
```
jfldde:
```suggestion
    let Some(soft_confirmation_number) = ledger_db.get_head_soft_confirmation_height()? else {
```

crates/storage-ops/src/pruning/components/state_db.rs
```diff
@@ -0,0 +1,103 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        self.next_result().transpose()
+    }
+}
+
+/// Prune state DB
+pub(crate) fn prune_state_db(state_db: Arc<sov_schema_db::DB>, up_to_block: u64) {
+    debug!("Pruning state DB, up to L2 block {}", up_to_block);
+
+    let Ok(indicies) = StaleNodeIndicesByVersionIterator::new(&state_db, up_to_block + 1) else {
```
jfldde:
```suggestion
    let Ok(indices) = StaleNodeIndicesByVersionIterator::new(&state_db, up_to_block + 1) else {
```
Is that a typo ?

crates/storage-ops/src/rollback/service.rs
```diff
@@ -0,0 +1,35 @@
...
+        loop {
+            select! {
+                biased;
+                _ = cancellation_token.cancelled() => {
+                    return;
+                },
+                Some(num_blocks) = self.receiver.recv() => {
+                    info!("Received signal to rollback {num_blocks} blocks");
+                    if let Err(e) = self.rollback.execute(num_blocks) {
+                        error!("Could not rollback blocks: {:?}", e);
```
jfldde: Shouldn't we abort here ?

jfldde:
Is that supposed to be running against a stopped node ? Wonder what the behaviour would be of rolling back from the cli against a running node. (Pruning feels safe but we might need to be cautious still).
If not, we might want to find a way to prevent running these commands against in-use dbs

eyusufatik:
> Is that supposed to be running against a stopped node ? Wonder what the behaviour would be of rolling back from the cli against a running node. (Pruning feels safe but we might need to be cautious still). If not, we might want to find a way to prevent running these commands against in-use dbs

this is supposed to run against a stopped node.

eyusufatik:
does this close #1609

rakanalh:
> does this close #1609

No, to prevent the PR from getting larger and harder to review... i am only placing the skeleton for rollback here and will implement in a follow up PR

rakanalh:
> Is that supposed to be running against a stopped node ? Wonder what the behaviour would be of rolling back from the cli against a running node. (Pruning feels safe but we might need to be cautious still). If not, we might want to find a way to prevent running these commands against in-use dbs

The design or `Pruner` vs `PrunerService` and `Rollback` vs `RollbackService` will enable us to spawn the service in the future when we use rollback to actually perform rolling back on a running node in case of forks. For now, this is not going to be used.

You cannot use the subcommand on an in-use DB because rocksdb locks the directory of any other process is using the database.

jfldde (approved):

crates/storage-ops/src/pruning/components/state_db.rs
```diff
@@ -62,22 +62,27 @@ impl<'a> Iterator for StaleNodeIndicesByVersionIterator<'a> {
...
         error!("Could not read stale nodes");
         return;
     };
 
-    let indices = indices.into_iter().flatten().flatten().collect::<Vec<_>>();
+    let indices = indices
+        .into_iter()
+        .flatten()
+        .flatten()
+        .collect::<HashSet<_>>();
```
jfldde: Do we expect duplicate indices entries ?

rakanalh:
For the record, this PR no longer closes the state DB pruning issue until we come up with a fix.

crates/storage-ops/src/pruning/components/state_db.rs
```diff
@@ -1,4 +1,155 @@
...
+                error!("Coult not iterate JmtValues: {:?}", e);
+                continue;
+            }
+        };
+
+        if let Err(e) = values_iter.seek(&(key_preimage.clone(), index.node_key.version())) {
+            error!("Could not seek on JmtValues iterator: {:?}", e);
+            continue;
+        }
+
```
eyusufatik, line 94: let's panic here because it's a very big inconsistency to have 

this means a node reported stale is not in the table at all

eyusufatik (approved):

Issue #1795, opened 2025-02-03
EVM new state diff friendly indexed state


eyusufatik:

eyusufatik:
closed by #1816

PR #1816, opened 2025-02-05
Evm accounts repeated account addresses

### LINKED ISSUE #1512: Stateful Statediff: RepeatedWrites (compress keys)
# Introduction

This issue belongs to https://github.com/chainwayxyz/citrea/issues/1237.

Value compression (U256 and code hash) is implemented in https://github.com/chainwayxyz/citrea/issues/1505 and https://github.com/chainwayxyz/citrea/issues/1510.

So now when calculating statediff for evm::Account(address) we store `[address(20) diff(balance) diff(nonce) diff(code_hash)]` instead of `[address(20) balance(32) nonce(8) code_hash(33)]`. And for evm::storage(address, key, value) we store `[address(20) key(32) diff(value)]` instead of `[address(20) key(32) value(32)]`.

However there is a room for optimization how we can store addresses. In zksync(https://docs.zksync.io/build/developer-reference/handling-pubdata-in-boojum) they store u64 as a key for RepeatedWrites (it's a write for a key that was already written before).

zksync uses this struct:

```
pub struct StateDiffRecord {
    /// address state diff occurred at
    pub address: Address,
    /// storage slot key updated
    pub key: U256,
    /// derived_key == Blake2s(bytes32(address), key)
    pub derived_key: [u8; 32],
    /// index in tree of state diff
    pub enumeration_index: u64,
    /// previous value
    pub initial_value: U256,
    /// updated value
    pub final_value: U256,
}
```

And their account::{balance, nonce, code_hash} indeed have a fake keys in evm storage:

```
pub fn get_nonce_key(account: &Address) -> StorageKey {
    let nonce_manager = AccountTreeId::new(NONCE_HOLDER_ADDRESS);

    // The `minNonce` (used as nonce for EOAs) is stored in a mapping inside the `NONCE_HOLDER` system contract
    let key = get_address_mapping_key(account, H256::zero());

    StorageKey::new(nonce_manager, key)
}
```

**Benefits**: Zksync store account data in evm::storage. We don't do that (yet?). And their evm::storage(address, key, value) for InitialWrites is `[derive_key(address(20), key(32)) diff(value)]` instead of `[address(20) key(32) value(32)]` and `[index(u64) diff(value)]` for RepeatedWrites respectfully.

**Conclusion**: We don't need to optimize the other keys for read/write operations except for evm::Account and evm::storage because other operations can be compressed using different techniques.

# Integration

When we deal with storage keys we calculate derived_key:

```
        // Compute the jmt update from the write batch
        let batch = state_accesses
            .ordered_writes
            .into_iter()
            .map(|(key, value)| {
                let key_hash = KeyHash::with::<H>(key.key.as_ref()); // <- This is our derived key

                let key_bytes = Arc::try_unwrap(key.key).unwrap_or_else(|arc| (*arc).clone());
                let value_bytes =
                    value.map(|v| Arc::try_unwrap(v.value).unwrap_or_else(|arc| (*arc).clone()));

                diff.push((key_bytes, value_bytes.clone()));

                (key_hash, value_bytes)
            })
            .collect::<Vec<_>>();
```

However I have no idea what is and `index` for RepeatedWrites.

The issue will be updated.

# TBD

- [ ] Research what is an `index`

kpp:
# Description

Store `address -> idx` and `idx -> AccountInfo` in evm storage instead of `address -> AccountInfo`.
 
- [x] Backwards compatibility with older versions.
- [x] Migration.

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/1512 https://github.com/chainwayxyz/citrea/issues/1795


crates/evm/src/provider_functions.rs
```diff
@@ -1,22 +1,96 @@
...
+use alloy_primitives::{Address, U256};
+use reth_primitives::SealedHeader;
+use sov_modules_api::{StateMapAccessor, StateValueAccessor, StateVecAccessor, WorkingSet};
 
-use crate::Evm;
+use crate::{AccountInfo, Evm};
 
 impl<C: sov_modules_api::Context> Evm<C> {
-    /// Returns the account at the given address.
-    pub fn basic_account(
```
eyusufatik, line 9: the old function `basic_account` shouldn't have had return type Option<Account>, it should be just `Account` as it always returns an account if you see.

if we get a never before used adress it returns this:

```
AccountInfo {
  balance: 0,
  nonce: 0,
  code_hash: EMPTY_KECCAK
}
```

so make the new function return `AccountInfo` and on None return from `self.account_idxs.get()`, return `AccountInfo::default()`

crates/evm/src/provider_functions.rs
```diff
@@ -1,22 +1,96 @@
...
+    pub fn account_delete(&self, address: &Address, working_set: &mut WorkingSet<C::Storage>) {
+        if let Some(idx) = self.account_idxs.get(address, working_set) {
+            self.accounts.delete(&idx, working_set)
+        }
+    }
+
+    /// Get the address of a storage key for the given account
+    fn get_storage_address(account: &Address, key: &U256) -> U256 {
+        let mut hasher: sha2::Sha256 = sha2::Digest::new_with_prefix(account.as_slice());
+        sha2::digest::Update::update(&mut hasher, key.as_le_slice());
```
eyusufatik: hasher.update(key.as_le_slice()) ???

crates/evm/src/provider_functions.rs
```diff
@@ -1,22 +1,96 @@
...
+        let arr = sha2::Digest::finalize(hasher);
```
eyusufatik: hasher.finalize() ?

crates/evm/src/query.rs
```diff
@@ -327,11 +325,8 @@ impl<C: sov_modules_api::Context> Evm<C> {
 
         self.set_state_to_end_of_evm_block_by_block_id(block_id, working_set)?;
 
-        let storage_slot = if self.accounts.get(&address, working_set).is_some() {
-            let db_account = DbAccount::new(address);
-            db_account
-                .storage
-                .get(&index, working_set)
+        let storage_slot = if self.account_exists(&address, working_set) {
```
eyusufatik: no need to check if account_exists()

just self.storage_get().unwrap_or_default() no?

eyusufatik (changes requested):
PR looks really good. thanks 👍 

I want you to create three issues:

1. Backwards compatibility with older versions.
2. Migration.
3. State diff calculation update.

crates/evm/src/provider_functions.rs
kpp: We need this because of reth::AccountReader

rakanalh (approved):

crates/evm/src/lib.rs
```diff
@@ -79,14 +80,30 @@ pub struct Evm<C: sov_modules_api::Context> {
...
+    #[state(rename = "i")]
+    pub(crate) account_idxs: sov_modules_api::StateMap<Address, AccountId, BcsCodec>,
+
     /// Mapping from account address to account state.
     #[state(rename = "a")]
-    pub accounts: sov_modules_api::StateMap<Address, AccountInfo, BcsCodec>,
+    pub accounts_prefork2: sov_modules_api::StateMap<Address, AccountInfo, BcsCodec>,
+
+    /// Mapping from account id to account state.
+    #[state(rename = "an")]
```
eyusufatik: let's rename to something else. save 1 byte :)

crates/evm/src/tests/call_tests.rs
eyusufatik, line 1: these tests under evm should now be running in Fork2 spec. are they?

crates/evm/src/tests/fork_tests.rs
eyusufatik, line 1: can we add a test here which starts from kumquat, then goes fork2 and we see that accounts indeed get migrated to new stroage system?

crates/evm/src/query.rs
```diff
@@ -378,31 +399,18 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
 
         self.set_state_to_end_of_evm_block_by_block_id(block_id, working_set)?;
 
-        let account = self.accounts.get(&address, working_set).unwrap_or_default();
+        let account = self
+            .account_info(&address, citrea_spec, working_set)
+            .unwrap_or_default();
         let code = if let Some(code_hash) = account.code_hash {
-            if current_spec.is_enabled_in(SpecId::CANCUN) {
+            if evm_spec.is_enabled_in(SpecId::CANCUN) {
```
eyusufatik: you can modify this line to be current_spec >= Kumquat btw.

eyusufatik (approved):
looking really good. aproving but have nits

crates/evm/src/lib.rs
kpp: That will not make any sense after statefull stdiffs.

eyusufatik (approved):

PR #1823, opened 2025-02-06
Use sha2 for block hash calculation

### LINKED ISSUE #1819: use precompile for Bitcoin header hash calculation
we use the bitcoin library to calculate the header hash

but just like with txid and wtxid's we can calculate them with accelerated sha2 crate

exeokan:
# Description
Modifies `HeaderWrapper::verify_hash` to use accelerated sha2 crate.

## Linked Issues
- Fixes #1819 

## Testing
Added unit test `calculate_block_hash`

### DIFF
--- a/crates/bitcoin-da/src/spec/header.rs
+++ b/crates/bitcoin-da/src/spec/header.rs
@@ -8,9 +8,8 @@ use borsh::{BorshDeserialize, BorshSerialize};
 use serde::{Deserialize, Serialize};
 use sov_rollup_interface::da::BlockHeaderTrait;
 
-use crate::helpers::calculate_double_sha256;
-
 use super::block_hash::BlockHashWrapper;
+use crate::helpers::calculate_double_sha256;
 
 // HeaderWrapper is a wrapper around BlockHash to implement BlockHeaderTrait
 #[derive(
@@ -37,7 +36,9 @@ impl BlockHeaderTrait for HeaderWrapper {
 
     fn verify_hash(&self) -> bool {
         let mut enc = vec![];
-        self.header.consensus_encode(&mut enc).expect("engines don't error");
+        self.header
+            .consensus_encode(&mut enc)
+            .expect("engines don't error");
         let calculated = calculate_double_sha256(&enc);
 
         self.hash() == BlockHashWrapper::from(calculated)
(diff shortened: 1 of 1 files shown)


exeokan:
Didn't change `HeaderWrapper::block_hash` and `HeaderWrapper::new` as they are not called from guest code, correct me if that's not the case

crates/bitcoin-da/src/spec/header.rs
```diff
@@ -135,3 +143,39 @@ impl From<BitcoinHeader> for BitcoinHeaderWrapper {
...
+            let header_hex = line.unwrap();
+            let header_bytes = hex::decode(&header_hex).unwrap();
+
+            let header = BitcoinHeaderWrapper::deserialize(&mut header_bytes.as_ref()).unwrap();
+
+            let mut enc = vec![];
+            header
+                .consensus_encode(&mut enc)
+                .expect("engines don't error");
+            let calculated = calculate_double_sha256(&enc);
```
kpp: Actually this test tests nothing. You have to provide a fn to calc block hash and then use it in both HeaderWrapper::verify_hash and in this test.

kpp (approved):

crates/bitcoin-da/src/spec/header.rs
exeokan: makes sense

crates/bitcoin-da/src/spec/header.rs
```diff
@@ -33,7 +35,13 @@ impl BlockHeaderTrait for HeaderWrapper {
     }
 
     fn verify_hash(&self) -> bool {
-        self.hash() == BlockHashWrapper::from(self.header.block_hash().to_byte_array())
+        let mut enc = vec![];
+        self.header
```
kpp: use `self.block_hash()` here

crates/bitcoin-da/src/spec/header.rs
```diff
@@ -135,3 +147,45 @@ impl From<BitcoinHeader> for BitcoinHeaderWrapper {
...
+            let inner_header =
+                BitcoinHeaderWrapper::deserialize(&mut header_bytes.as_ref()).unwrap();
+            let header = HeaderWrapper::new(*inner_header.deref(), 0, height, [0; 32]);
+
+            let mut enc = vec![];
+            header
+                .header
+                .consensus_encode(&mut enc)
+                .expect("engines don't error");
+            let calculated = calculate_double_sha256(&enc);
```
kpp:
```suggestion
            let calculated = header.block_hash();
```

crates/bitcoin-da/src/spec/header.rs
exeokan: just as I was doing that :D

kpp:
> Didn't change HeaderWrapper::block_hash and HeaderWrapper::new as they are not called from guest code, correct me if that's not the case

Probably yes but tests now test your code and reflect changes

eyusufatik (approved):

eyusufatik (approved):

PR #1828, opened 2025-02-06
Makefile testing targets improvements

rakanalh:
# Changes

- Github CI: Prevent building the binary for tests twice
  - Added a step for invoking `make build-test` in github checks.yml 
- Github CI: Now invokes `make test-ci` and `make coverage-ci` using the prebuilt binary
- `make test` and `make coverage` are for local use where the binary gets built as a dependency to the make target
- Add two new make targets, to pick and choose which tests to run locally:
  - `make -C bin/citrea mock-e2e`
  - `make -C bin/citrea bitcoin-e2e`  
- Reorganize tests into `mock-e2e` and `bitcoin-e2e`
  - Add `common` stuff in `bin/citrea/tests/common`


jfldde:
I am not sure about this PR `e2e` and `bitcoin` renaming, makes it confusing.
Maybe `e2e/bitcoin` `e2e/mock` would better represent it

jfldde (approved):

eyusufatik (approved):

PR #1830, opened 2025-02-06
make state root concrete

eyusufatik:
no description


yaziciahmet (approved):
One more down ser, lgtm

jfldde (approved):

bin/citrea/tests/bitcoin_e2e/light_client_test.rs
```diff
@@ -1328,9 +1328,9 @@ fn create_serialized_fake_receipt_batch_proof(
         32, 64, 64, 227, 100, 193, 15, 43, 236, 156, 31, 229, 0, 161, 205, 76, 36, 124, 137, 214,
         80, 160, 30, 215, 232, 44, 171, 168, 103, 135, 124, 33,
     ];
-    let batch_proof_output = BatchProofCircuitOutputV2::<BitcoinSpec, [u8; 32]> {
-        initial_state_root,
-        final_state_root,
+    let batch_proof_output = BatchProofCircuitOutputV2::<BitcoinSpec> {
+        initial_state_root: initial_state_root.into(),
```
kpp: Please create an issue to convert all stateroots to [u8; 32] to get rid of .into()

eyusufatik: Well, StorageRootHash is jmt::RootHash 

which is 


pub struct RootHash(pub [u8; 32])

.github/workflows/checks.yml
```diff
@@ -345,31 +345,6 @@ jobs:
           npx mocha test.js
           cd ../../../../..
 
-  check_no_std:
```
rakanalh, line 387: Are we sure about removing this?

I am guessing we keep this to make sure the guest code compiles without native features even if in some sov-sdk crates we don't need the check anymore.

bin/citrea/tests/bitcoin_e2e/light_client_test.rs
kpp: Yes, still we can have [u8; 32] and create jmt::RootHash from it when we need to. Maybe maybe in the future

.github/workflows/checks.yml
kpp, line 387: That's too powerful solution, no?

rakanalh (approved):

crates/batch-prover/src/proving.rs
```diff
@@ -355,11 +331,12 @@ where
             ),
             Err(e) => {
                 info!("Failed to extract post fork 1 output from proof: {:?}. Trying to extract pre fork 1 output", e);
-                let output = Vm::extract_output::<
-                    BatchProofCircuitOutputV1<<Da as DaService>::Spec, StateRoot>,
-                >(&proof)
-                .expect("Should be able to extract either pre or post fork 1 output");
-                let batch_proof_output = BatchProofCircuitOutputV2::<Da::Spec, StateRoot> {
+                let output =
+                    Vm::extract_output::<BatchProofCircuitOutputV1<<Da as DaService>::Spec>>(
```
kpp: Why we can use `BatchProofCircuitOutputV2::<Da::Spec>` but not Da::Spec here?

crates/light-client-prover/src/tests/test_utils.rs
```diff
@@ -16,9 +16,9 @@ pub(crate) fn create_mock_batch_proof(
 ) -> MockBlob {
     let batch_proof_method_id = MockCodeCommitment([2u8; 32]);
 
-    let bp = BatchProofCircuitOutputV2::<MockDaSpec, [u8; 32]> {
-        initial_state_root,
-        final_state_root,
+    let bp = BatchProofCircuitOutputV2::<MockDaSpec> {
+        initial_state_root: initial_state_root.into(),
```
kpp: No really, lots of changes only because its jmt::RootHash not [u8; 32]

kpp (changes requested):
All good except for nostd-check

crates/light-client-prover/src/tests/test_utils.rs
eyusufatik: Ok I'll do it

kpp (approved):

PR #1832, opened 2025-02-08
Citrea CLI

rakanalh:
# Description

Taken from #1808 into it's own PR to unblock @jfldde


crates/storage-ops/src/pruning/components/evm.rs
eyusufatik, line 1: remove evm, since you already did in the other pr?

bin/cli/src/main.rs
```diff
@@ -0,0 +1,58 @@
...
+    },
+    /// Rollback the most recent N blocks
+    Rollback {
+        /// The path of the database to prune
+        #[arg(long)]
+        db_path: PathBuf,
+        /// The number of blocks to rollback
+        #[arg(long)]
+        blocks: u32,
+    },
```
eyusufatik, line 37: make rollback have two fields 

- to_l2_block
- to_l1_block

eyusufatik (changes requested):

bin/cli/src/main.rs
rakanalh: Will do this in the rollback PR

eyusufatik (approved):

jfldde (approved):

PR #1840, opened 2025-02-08
Fix test target

rakanalh:
# Description

Fixes the problem that @kpp had where `make test [TEST-NAME]` did not work as before.

Now we're passing along all additional params we receive for `make test` to make sure that they are available when `test-ci` is invoked.


kpp (approved):

kpp (changes requested):
We actually need to be able to run `make test [TESTNAME] -- --nocapture`. With this PR I need to pass `make test -- -- --nocapture`.

yaziciahmet (approved):

eyusufatik (approved):

kpp (approved):

PR #1843, opened 2025-02-08
refactor dependencies, features, remove prover service trait and sov-stf-runner

eyusufatik:
# Description

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/common/src/lib.rs
```diff
@@ -10,3 +10,13 @@ pub mod tasks;
 pub mod utils;
 
 pub use config::*;
+use sov_rollup_interface::zk::StorageRootHash;
+
+type SoftConfirmationHash = [u8; 32];
```
rakanalh, line 15: Primitives instead of here?

rakanalh (approved):
less code is better code

crates/common/src/lib.rs
eyusufatik: that's a good point. however, did not need to import the common crate anywhere, which means we never used these type aliases except for native context.

eyusufatik: I'm leaving that for another pr as a lot of other places gotta go for a rename

crates/citrea-stf/src/genesis_config.rs
```diff
@@ -78,3 +77,17 @@ fn create_genesis_config<C: Context, Da: DaSpec>(
         soft_confirmation_rule_enforcer_config,
     ))
 }
+
+/// Read genesis file
+pub fn read_json_file<T: serde::de::DeserializeOwned, P: AsRef<Path>>(
+    path: P,
+) -> anyhow::Result<T> {
```
kpp, line 84: anyhow

crates/common/src/lib.rs
kpp: Let's create an issue for it then

crates/citrea-stf/src/genesis_config.rs
eyusufatik: that's native only

kpp (approved):

PR #1849, opened 2025-02-08
Fix make lint-fix fmt

jfldde:
# Description

- Fix an edge case where a `clippy fix` will lead to an out of format code. Runs `cargo fmt` last to be sure the clippy changes end up formatted


eyusufatik (approved):

eyusufatik:
great catch! always made me angry :D

PR #1855, opened 2025-02-09
Remove evm header clone

jfldde:
# Description

- Copy required vars first and prevent cloning header


kpp (changes requested):
Or .seal() can accept &self instead of self

jfldde:
I tend to like seal taking self. Makes sense semantically that after a seal, we cannot access the original AlloyBlock.
I haven't looked deeply yet but I imagine it would trickle down as well

eyusufatik (approved):

kpp (approved):

PR #1859, opened 2025-02-10
Share LedgerDB migrations

jfldde:
# Description

- Share LedgerDB migrations


crates/common/src/migration/mod.rs
rakanalh, line 1: The directory in all nodes is called `db_migrations`. Let's name this one the same for consistency

rakanalh (approved):

eyusufatik (approved):

PR #1862, opened 2025-02-10
Start tests from Fork2

### LINKED ISSUE #1858: Start all tests from Fork2
Now they start from Kumquat. They should start from Fork2 except for fork tests.

kpp:
# Description

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/1858


crates/evm/src/tests/utils.rs
```diff
@@ -323,8 +323,8 @@ pub(crate) fn get_evm_test_config() -> EvmConfig {
     config
 }
 
-pub(crate) fn get_fork_fn_only_fork1() -> impl Fn(u64) -> Fork {
-    |_: u64| Fork::new(SovSpecId::Kumquat, 0)
+pub(crate) fn get_fork_fn_only_fork2() -> impl Fn(u64) -> Fork {
```
rakanalh, line 326: This has to change every time we create a new fork?

How about we implement:

```rust
impl SpecId {
   pub fn latest() -> Self {
        Self::Fork2
   }
}
```

And use this `SpecId::latest()` everywhere?

rakanalh (dismissed):

crates/evm/src/tests/utils.rs
kpp: https://github.com/chainwayxyz/citrea/issues/1869

eyusufatik (approved):

PR #1863, opened 2025-02-10
Add coinbase depth and tx id inclusion proof to `BitcoinLightClient`

### LINKED ISSUE #1777: Update `BitcoinLightClient::setBlockInfo()` with coinbase txid merkle proof depth
The `BitcoinLightClient` contract will store coinbase txid merkle proof depth, which is the number of hashes given in the coinbase tx txid merkle proof (see #1772). This number will be used in `verifyInclusion` function, it will assert given proof has the same number of hashes as stored for that block.

okkothejawa:
# Description
- Adds coinbase depth to `BitcoinLightClient` to solve the merkle tree leaf node weakness issue.
- Adds a `verifyInclusionByTxId` function that acts the same as `verifyInclusion` but uses `txId` instead of `wtxId`, as such, it requires a Bitcoin header parameter that is later validated against stored Bitcoin block hash.

## Linked Issues
- Fixes #1777

## Testing
Added unit tests for coinbase depth fail case and verifying inclusion through tx id.


eyusufatik:
plan is to merge #1935 then update genesises for tests that start with Fork2 in this PR.

we might want to implement fork2 function here so that we can keep backwards compatibility easily. this would also mean we should update the bridge to use this new function. that can be this PR or another one.

only after these we can merge this PR. then we'll need a followup PR for #1913

crates/evm/src/evm/system_contracts/src/BitcoinLightClient.sol
```diff
@@ -32,6 +38,7 @@ contract BitcoinLightClient is IBitcoinLightClient {
         blockNumber = _blockNumber;
     }
 
+    /// @notice Kept for backwards compatibility
```
eyusufatik: add warning that this should not be called and calling it will invalidate L2 block after Fork2 activates

crates/evm/src/evm/system_contracts/src/BitcoinLightClient.sol
```diff
@@ -86,7 +108,24 @@ contract BitcoinLightClient is IBitcoinLightClient {
...
+    function verifyInclusionByTxId(uint256 _blockNumber, bytes32 _txId, bytes calldata _blockHeader, bytes calldata _proof, uint256 _index) external view returns (bool) {
+        bytes32 _blockHash = _blockHeader.hash256View();
+        require(blockHashes[_blockNumber] == _blockHash, "Invalid block header");
+        require(_proof.length == coinbaseDepths[_blockHash] * 32, "Invalid proof length");
+        bytes32 _txRoot = _blockHeader.extractMerkleRootLE();
+        return ValidateSPV.prove(_txId, _txRoot, _proof, _index);
+    }
+
     function _verifyInclusion(bytes32 _blockHash, bytes32 _wtxId, bytes calldata _proof, uint256 _index) internal view returns (bool) {
+        require(_proof.length == coinbaseDepths[_blockHash] * 32, "Invalid proof length");
```
eyusufatik, line 128: what is the actual type of _proof? multiplying by 32 might not be correct

crates/evm/src/evm/system_contracts/src/BitcoinLightClient.sol
```diff
@@ -86,7 +108,24 @@ contract BitcoinLightClient is IBitcoinLightClient {
...
 
+    /// @notice Verifies the inclusion of a transaction ID in the merkle root hash of a block
+    /// @dev Transaction ID and proof elements must be in little endian
+    /// @param _blockNumber Block number of the block
+    /// @param _txId Transaction ID
+    /// @param _blockHeader Block header of the associated Bitcoin block
+    /// @param _proof Merkle proof
+    /// @param _index Index of the transaction
+    /// @return If the transaction ID is included in the merkle root hash of the block
+    function verifyInclusionByTxId(uint256 _blockNumber, bytes32 _txId, bytes calldata _blockHeader, bytes calldata _proof, uint256 _index) external view returns (bool) {
```
eyusufatik, line 119: sneaking in features I see :)

crates/evm/src/hooks.rs
```diff
@@ -381,11 +381,48 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+        .deposit_data
+        .iter()
+        .for_each(|params| {
+            system_events.push(SystemEvent::BridgeDeposit(params.clone()));
+        });
+    system_events
+}
+
 /// Populates system events based on the current soft confirmation info.
 pub fn populate_system_events<'a>(
```
eyusufatik: I might rewrite this

eyusufatik: taking notes

crates/sequencer/src/runner.rs
```diff
@@ -447,6 +450,7 @@ where
 
         let evm_txs = self.get_best_transactions()?;
 
+        let coinbase_depth = short_header_proof.verify().unwrap().coinbase_txid_merkle_proof_height.into();
```
eyusufatik: shouldn't handle it like this

eyusufatik (changes requested):

crates/sequencer/src/runner.rs
okkothejawa: what's wrong with it

crates/evm/src/evm/system_contracts/src/BitcoinLightClient.sol
okkothejawa: concatenated list of 32 bytes of hashes so it works, its later on asserted that `_proof.length % 32 == 0` anyway in the verification lib

guests/risc0/light-client-proof/bitcoin/src/bin/light_client_proof_bitcoin.rs
eyusufatik, line 1: the previous values for mock and bitcoin light client proofs were different, but you set them to the same thing. might be causing problems on tests.

bin/citrea/tests/bitcoin/light_client_test.rs
```diff
@@ -121,12 +121,13 @@ impl TestCase for LightClientProvingTest {
         da.generate(FINALITY_DEPTH).await?;
 
         let batch_proof_l1_height = da.get_finalized_height(None).await?;
-
+        println!("Waiting for light client prover to process batch proofs...");
```
eyusufatik: remove println!

bin/citrea/tests/mock/sequencer_behaviour.rs
```diff
@@ -524,7 +524,9 @@ async fn test_gas_limit_too_high() {
 /// gas limit left from the system transaction(s).
 #[tokio::test(flavor = "multi_thread")]
 async fn test_system_tx_effect_on_block_gas_limit() -> Result<(), anyhow::Error> {
-    // citrea::initialize_logging(tracing::Level::INFO);
+    citrea::initialize_logging(tracing::Level::INFO);
```
eyusufatik: comment out and remove println

crates/common/src/utils.rs
```diff
@@ -122,7 +122,7 @@ async fn update_short_header_proof_from_sys_tx<Da: DaService, DB: SharedLedgerOp
 ) -> anyhow::Result<()> {
     let function_selector: [u8; 4] = tx.input()[0..4].try_into()?;
 
-    if function_selector == BitcoinLightClientContract::setBlockInfoCall::SELECTOR {
+    if function_selector == BitcoinLightClientContract::setBlockInfo_1Call::SELECTOR {
```
eyusufatik: do you think we can define a new constant like 

BitcoinLightClientContract::setBlockInfoPostFork2Call = BitcoinLightClientContract::setBlockInfo_1Call so that we don't see this ugly name?

eyusufatik (changes requested):

eyusufatik (approved):

PR #1867, opened 2025-02-10
Release v0.6.3 changes

eyusufatik:


resources/guests/risc0/testnet/light-0.elf
rakanalh, line 1: Are we going to set new batch & light elfs?

rakanalh (approved):

PR #1871, opened 2025-02-11
Simplify WorkingSet Delta

yaziciahmet:
# Description

Before it was:
WorkingSet -> RevertableWriter -> Delta -> StorageInternalCache -> CacheLog
now:
WorkingSet -> Delta -> CacheLog

Before:
StateCheckpoint -> Delta -> StorageInternalCache -> CacheLog
now:
StateCheckpoint -> Delta -> CacheLog

This PR removes couple of unnecessary layers which makes it harder to follow the path data is taking.

Also it makes both WorkingSet and StateCheckpoint to have Delta instead of WorkingSet having RevertableWriter and StateCheckpoint having Delta, which caused the version to be lost in case of calling `working_set.checkpoint().to_revertable()`. And the loss of version causes problems when trying to replay previous txs on historical blocks.

To make sure there are no changes, and it behaves exactly the same as before, I ran basic_proving_test and test_light_client_proving tests and observed that the ordered reads and writes of each corresponding deltas are equal.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/sovereign-sdk/module-system/sov-modules-core/src/storage/scratchpad.rs
+++ b/crates/sovereign-sdk/module-system/sov-modules-core/src/storage/scratchpad.rs
@@ -253,55 +253,6 @@ pub trait StateReaderAndWriter {
     }
 }
 
-/// A working set accumulates reads and writes on top of the underlying DB,
-/// automating witness creation.
-pub struct Delta<S: Storage> {
-    inner: S,
-    witness: S::Witness,
-    cache: StorageInternalCache,
-}
-
-impl<S: Storage> Delta<S> {
-    fn new(inner: S, version: Option<u64>) -> Self {
-        Self::with_witness(inner, Default::default(), version)
-    }
-
-    fn with_witness(inner: S, witness: S::Witness, version: Option<u64>) -> Self {
-        Self {
-            inner,
-            witness,
-            cache: StorageInternalCache::new(version, CacheMode::State),
-        }
-    }
-
-    fn freeze(&mut self) -> (OrderedReadsAndWrites, S::Witness) {
-        let cache = mem::take(&mut self.cache);
-        let witness = mem::take(&mut self.witness);
-
-        (cache.into(), witness)
-    }
-}
-
-impl<S: Storage> fmt::Debug for Delta<S> {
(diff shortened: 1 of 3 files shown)


crates/sovereign-sdk/module-system/sov-modules-core/src/storage/scratchpad.rs
```diff
@@ -1,18 +1,267 @@
...
+        match self.cache_log.get_value(&cache_key) {
+            ValueExists::Yes(value) => value.map(Into::into),
+            ValueExists::No => {
+                let storage_value = self
+                    .storage
+                    .get_offchain(key, self.version, &mut self.witness);
+                let cache_value = storage_value.as_ref().map(|v| v.clone().into_cache_value());
+
+                self.cache_log
+                    .add_read(cache_key.clone(), cache_value.clone())
```
jfldde: Don't need to clone here

crates/sovereign-sdk/module-system/sov-modules-core/src/storage/scratchpad.rs
```diff
@@ -1,18 +1,267 @@
...
+    }
+
+    fn revert(mut self) -> Self {
+        self.uncommitted_writes.clear();
+        self
+    }
+
+    fn freeze(&mut self) -> (OrderedReadsAndWrites, S::Witness) {
+        let ordered_writes = mem::take(&mut self.cache_log).take_writes();
+        println!("offchain ordered reads: 0");
```
eyusufatik: dont forget logs

eyusufatik (approved):

crates/sovereign-sdk/module-system/sov-modules-core/src/storage/scratchpad.rs
```diff
@@ -480,8 +561,8 @@ impl<S: Storage> WorkingSet<S> {
     where
         S: NativeStorage,
     {
-        // First inner is `RevertableWriter` and second inner is actually a `Storage` instance
-        self.delta.inner.inner.get_with_proof(key, version)
+        // First inner is `R.clone()evertableWriter` and second inner is actually a `Storage` instance
```
jfldde, line 564: Something's gone haywire here

jfldde (approved):

PR #1880, opened 2025-02-12
Rewrite db & storage

yaziciahmet:
# Description

Partial rewrite of the storage system. Here is an overview of changes made:

- No more snapshots. Snapshots were used to handle forks, but we don't deal with that at all at the moment, and give ourselves a finalization margin. The only "snapshot"-like behavior we need is to be able to be query and replay on top of a previous l2 block. To do that, all we need is to use the right version when querying the underlying database.
- Current hierarchy of storage structures goes as following: WorkingSet -> ProverStorage -> StateDB + NativeDB -> DbTransaction -> DB.  DbTransaction is simply a structure that stores local writes in its cache, and queries cache+db whenever get is called. When freezed, returns uncommitted changes to be committed to underlying database.
- StateDB does not have a version in its state anymore, and it completely leaves the version to use to caller. It has an helper method to return the next expected version.
- ProverStorage is the main structure that maintains a version. For every query to database ProverStorage does, it uses its own internal version, and its api no longer accepts version. This simplifies the version flow of storage. ProverStorage also has a flag called is_snapshot. If it is marked as snapshot, it can not be committed to underlying rocksdb, and finalization will panic (to enforce us devs to never do this and catch it on development)
- ProverStorage version indicates which version to query by. But commits will happen to version + 1. So it is like, storage is on version 2 gives the view of version 2 (including 2), and the next commit will be for version 3.
- ProverStorageManager's only job is to create requested ProverStorage on requested version. It has 3 methods to create storage: 1. create_latest_version_storage: creates a prover storage with the latest used version. used when nodes want to run or generate the next block. 2. create_latest_view_storage: creates a prover storage which always returns the latest version values without needing to update the version. this is achieved by specifying version as u64::MAX. this creates a snapshot storage, hence, can not be finalized. this is used by rpc and evm db provider. 3. create_storage_snapshot_on_version: creates a prover storage on the given version. it creates the storage as snapshot, and is useful for querying or replaying blocks on top of older blocks.
- Rpc context has a latest view storage, so it always gets the latest version. When rpcs want to replay transactions, for getting traces, there is a helper method called clone_with_version on ProverStorage, which returns a snapshot storage with the provided version. So any replay can be done without affecting the original ProverStorage.

One side-note that is rather important is that, before, we were committing state db versions as block + 1. so the block 1 would be version 2 in the db. But weirdly, native db was such that block 1 is committed as version 1. This pr changes that and makes native db block 1 to be version 2 (+1). IT IS BREAKING CURRENT NODES. But it is stupid to have such difference, and I think it is something worth to resync native db.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


bin/citrea/src/rollup/mod.rs
```diff
@@ -99,10 +99,10 @@ pub trait CitreaRollupBlueprint: RollupBlueprint {
         &self,
         rollup_config: &FullNodeConfig<Self::DaConfig>,
         rocksdb_config: &RocksdbConfig,
-    ) -> Result<Storage<Self>> {
+    ) -> Result<Storage> {
         let ledger_db = self.create_ledger_db(rocksdb_config);
-        let mut storage_manager = self.create_storage_manager(rollup_config)?;
-        let prover_storage = storage_manager.create_finalized_storage()?;
+        let storage_manager = self.create_storage_manager(rollup_config)?;
+        let prover_storage = storage_manager.create_storage();
```
eyusufatik: we should rename this create_storage function to something like 

create_always_latest storage or maybe 

create_rpc_storage

crates/sovereign-sdk/module-system/sov-modules-api/src/containers/mod.rs
eyusufatik, line 1: are you gonna rewrite these tests?

crates/sovereign-sdk/full-node/db/sov-schema-db/src/lib.rs
```diff
@@ -48,6 +48,31 @@ pub struct DB {
 }
 
 impl DB {
+    /// Opens the DB with a tempdir. Should only be used in tests
+    #[cfg(feature = "test-utils")]
```
eyusufatik, line 52: don't really like having these test workaround functions, let's try to get rid of them

crates/sovereign-sdk/full-node/db/sov-schema-db/src/snapshot.rs
```diff
@@ -1,508 +0,0 @@
...
-pub type SnapshotId = u64;
-
-/// A trait to make nested calls to several [`SchemaBatch`]s and eventually [`crate::DB`]
-pub trait QueryManager {
-    /// Iterator over key-value pairs in reverse lexicographic order in given [`Schema`]
-    type Iter<'a, S: Schema>: Iterator<Item = (SchemaKey, SchemaValue)>
-    where
-        Self: 'a;
-    /// Iterator with given range
-    type RangeIter<'a, S: Schema>: Iterator<Item = (SchemaKey, SchemaValue)>
```
eyusufatik, line 21: try to have both a iter and rangeiter impl

eyusufatik (commented):
not really a detailed review just peeked over the changes

crates/sovereign-sdk/module-system/sov-modules-api/src/containers/mod.rs
yaziciahmet: rewrote them

crates/sovereign-sdk/full-node/db/sov-schema-db/src/transaction.rs
```diff
@@ -0,0 +1,294 @@
...
+            .lock()
+            .expect("Local SchemaBatch lock must not be poisoned");
+        cache.merge(batch);
+        Ok(())
+    }
+}
+
+impl DbTransaction {
+    /// Get a value from current transaction or underlying database
+    pub fn read<S: Schema>(&self, key: &impl KeyCodec<S>) -> anyhow::Result<Option<S::Value>> {
```
eyusufatik, line 56: put these impl block in the other one

crates/sovereign-sdk/full-node/db/sov-schema-db/src/transaction.rs
```diff
@@ -0,0 +1,294 @@
...
+        // Some(Operation) means that key was touched,
+        // but in case of deletion we early return None
+        // Only in case of not finding operation for key,
```
eyusufatik: delete

crates/sovereign-sdk/module-system/sov-state/src/zk_storage.rs
```diff
@@ -155,4 +145,8 @@ where
     fn is_empty(&self) -> bool {
         unimplemented!("Needs simplification in JellyfishMerkleTree: https://github.com/Sovereign-Labs/sovereign-sdk/issues/362")
     }
+
+    fn clone_with_version(&self, _version: jmt::Version) -> Self {
+        Self::new()
```
eyusufatik: unimplemented!()

crates/sovereign-sdk/module-system/sov-state/src/prover_storage.rs
```diff
@@ -1,63 +1,77 @@
...
 /// environment (outside of the zkVM).
-pub struct ProverStorage<Q> {
-    db: StateDB<Q>,
-    native_db: NativeDB<Q>,
+#[derive(Clone)]
+pub struct ProverStorage {
+    db: StateDB,
+    native_db: NativeDB,
+    version: Arc<AtomicU64>,
+    is_snapshot: bool,
```
eyusufatik: commitable

crates/sovereign-sdk/module-system/sov-state/src/prover_storage.rs
```diff
@@ -1,63 +1,77 @@
...
-}
 
-impl<Q> ProverStorage<Q> {
-    /// Creates a new [`ProverStorage`] instance from specified db handles
-    pub fn with_db_handles(db: StateDB<Q>, native_db: NativeDB<Q>) -> Self {
-        Self { db, native_db }
+    /// Creates a new [`ProverStorage`] instace from specified db handles and version.
+    /// When created using this method, storage is marked as snapshot and won't be committed
+    /// to underlying database.
+    pub fn with_version_snapshot(db: StateDB, native_db: NativeDB, version: Version) -> Self {
```
eyusufatik: don't forget commitable name change

crates/sovereign-sdk/module-system/sov-state/src/prover_storage.rs
```diff
@@ -121,31 +120,31 @@ where
         state_accesses: OrderedReadsAndWrites,
         witness: &mut Self::Witness,
     ) -> Result<(StateRootTransition, Self::StateUpdate, StateDiff), anyhow::Error> {
-        let latest_version = self.version();
+        let version = self.version();
         let jmt = JellyfishMerkleTree::<_, DefaultHasher>::new(&self.db);
 
         // Handle empty jmt
         // TODO: Fix this before introducing snapshots!
```
eyusufatik: remove old comments

eyusufatik: look into if there is a empty tree root hash value in jmt

eyusufatik: might not work as state roots might not hold up. let's just try it

crates/sovereign-sdk/full-node/sov-prover-storage-manager/src/lib.rs
```diff
@@ -1,1406 +1,105 @@
...
 
-        let native_db_snapshot = DbSnapshot::new(
-            snapshot_id,
-            ReadOnlyLock::new(self.accessory_snapshot_manager.clone()),
-        );
-
-        let native_db = NativeDB::with_db_snapshot(native_db_snapshot)?;
-        Ok(ProverStorage::with_db_handles(state_db, native_db))
+        storage
     }
```
eyusufatik, line 44: prepare_storage_for_l2_height()

returns storage needed for running block 3

crates/sovereign-sdk/full-node/sov-prover-storage-manager/src/lib.rs
```diff
@@ -1,1406 +1,105 @@
...
-            .into_iter()
-            .filter(|bh| bh != &current_block_hash)
-            .collect();
-
-        while let Some(block_hash) = to_discard.pop() {
-            let child_block_hashes = self.chain_forks.remove(&block_hash).unwrap_or_default();
-            self.blocks_to_parent.remove(&block_hash).unwrap();
+    /// Creates a new [`ProverStorage`] that always has the latest view of the state,
+    /// and can not be committed.
+    pub fn create_latest_view_storage(&self) -> ProverStorage {
```
eyusufatik: create_always_latest_uncommitable_storage()

eyusufatik (changes requested):
small nits. well done.

crates/sovereign-sdk/module-system/sov-state/src/prover_storage.rs
yaziciahmet: removed old comments. empty tree hash issue, we can look as separate issue.

crates/sovereign-sdk/full-node/sov-prover-storage-manager/src/lib.rs
yaziciahmet: renamed in order:

create_storage_for_l2_height(l2_height: u64) -> creates a storage to run l2 height state updates. uncommittable.
create_storage_for_next_l2_height() -> creates a storage to run next l2 height state updates. derives next l2 height from version. committable
create_final_view_storage() -> creates a storage that returns always the final state. uncommittable

lmk what you think

eyusufatik (approved):

PR #1899, opened 2025-02-16
Pruning related RPC error

### LINKED ISSUE #1813: Pruning related error message on RPCs
we might need to return specific RPC error if state or accessory state is pruned.

exeokan:
# Description
Adds `get_last_pruned_l2_height` to `Storage` trait and implements it in `ProverStorage` to read last pruned height from native db.

`WorkingSet` is then used to call `get_last_pruned_l2_height` to determine whether a block number is pruned in rpc handlers. We return `ProviderError::StateAtBlockPruned` if l2 block is pruned. For endpoints taking block hash or transaction hash as parameters, we return `UnknownBlockOrTxIndex` or `None` as before.

## Linked Issues
- Fixes #1813


crates/evm/src/query.rs
```diff
@@ -303,6 +303,12 @@ impl<C: sov_modules_api::Context> Evm<C> {
     ) -> RpcResult<U256> {
         let block_number = self.block_number_from_state(block_id, working_set)?;
```
eyusufatik: is it only eth_getBalance that throws this error?

if not, we can make this check and return err a function and put it on top of the other methods that need this check.

exeokan: There are many others, just wanted to demonstrate with eth_getBalance. will write a check function for and implement for other methods 👍🏻

crates/evm/src/query.rs
```diff
@@ -1689,6 +1699,10 @@ impl<C: sov_modules_api::Context> Evm<C> {
             Some(BlockId::Number(block_num)) => {
                 match block_num {
                     BlockNumberOrTag::Number(num) => {
+                        if num != 0 {
```
eyusufatik, line 1702: don't do num != 0 

invoke the check function regardless

eyusufatik (approved):

eyusufatik:
check if your changes automatically make debug_trace endpoints error out

crates/evm/src/query.rs
exeokan: but then it returns StateAtBlockPruned, and we lose the ability to query the state in genesis block

eyusufatik (approved):

PR #1906, opened 2025-02-17
Remove unused AlignedVec

jfldde:
# Description

- Removes confusing AlignedVec which doesn't offer any guarantee  and replace it with an inner Vec<u8>

### DIFF
--- a/crates/sovereign-sdk/module-system/sov-modules-core/src/common/bytes.rs
+++ b/crates/sovereign-sdk/module-system/sov-modules-core/src/common/bytes.rs
@@ -42,7 +42,7 @@ impl fmt::Display for Prefix {
 
 impl Extend<u8> for Prefix {
     fn extend<T: IntoIterator<Item = u8>>(&mut self, iter: T) {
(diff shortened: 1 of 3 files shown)


crates/sovereign-sdk/module-system/sov-modules-core/src/common/bytes.rs
```diff
@@ -104,24 +42,26 @@ impl fmt::Display for Prefix {
 
 impl Extend<u8> for Prefix {
     fn extend<T: IntoIterator<Item = u8>>(&mut self, iter: T) {
-        self.prefix
-            .extend(&AlignedVec::new(iter.into_iter().collect()))
+        self.prefix.extend(&iter.into_iter().collect::<Vec<u8>>())
```
yaziciahmet: extend accepts iter directly, no need to collect

yaziciahmet (changes requested):
LGTM. one nit

yaziciahmet (approved):

eyusufatik (approved):

PR #1910, opened 2025-02-18
DB rollback implementation

### LINKED ISSUE #1609: Rollback command line argument

rakanalh:
# Description

Rollback all DBs down to a specific block

# RocksDB References
- [Compaction](https://github.com/facebook/rocksdb/wiki/Compaction)
- [Delete a range of keys](https://github.com/facebook/rocksdb/wiki/Delete-A-Range-Of-Keys)
- [DeleteRange](https://github.com/facebook/rocksdb/wiki/DeleteRange)

## Linked Issues
- Fixes #1609


bin/citrea/tests/mock/rollback.rs
```diff
@@ -0,0 +1,147 @@
...
+use crate::common::helpers::{
+    create_default_rollup_config, start_rollup, tempdir_with_children, wait_for_l1_block,
+    wait_for_l2_block, NodeMode,
+};
+use crate::common::TEST_DATA_GENESIS_PATH;
+use crate::mock::evm::init_test_rollup;
+
+/// Trigger rollback native DB data.
+#[tokio::test(flavor = "multi_thread")]
+async fn test_rollback() -> Result<(), anyhow::Error> {
```
jfldde: feel this would be much leaner with citrea-e2e

rakanalh: Why?

jfldde: Well all the config/initialization/stop/restart logic is part of the framework and this would be a few line of test instead

rakanalh: Right... but what difference would it make? these tests are not DA layer related so they could be ran on mock as well as bitcoin.

bin/citrea/tests/mock/rollback.rs
```diff
@@ -0,0 +1,487 @@
...
+        None,
+        None,
+        false,
+    )
+    .await;
+    let full_node_port = full_node_port_rx.await.unwrap();
+    let full_node_test_client = make_test_client(full_node_port).await.unwrap();
+
+    assert_dbs(full_node_test_client, addr).await;
+
```
rakanalh, line 329: Add checks for full node specific tables

bin/citrea/tests/mock/rollback.rs
```diff
@@ -0,0 +1,487 @@
...
+        None,
+        None,
+        false,
+    )
+    .await;
+    let batch_prover_port = batch_prover_port_rx.await.unwrap();
+    let batch_prover_test_client = make_test_client(batch_prover_port).await.unwrap();
+
+    assert_dbs(batch_prover_test_client, addr).await;
+
```
rakanalh, line 483: Add checks for batch node specific tables

crates/storage-ops/src/rollback/mod.rs
```diff
@@ -1,11 +1,67 @@
...
+    ) -> Self {
+        // distance is the only criteria implemented at the moment.
+        Self {
+            ledger_db,
+            state_db,
+            native_db,
+        }
+    }
+
     /// Rollback the provided number of blocks
```
eyusufatik: not number of blocks but to block

crates/storage-ops/src/rollback/components/state_db.rs
```diff
@@ -0,0 +1,82 @@
...
+
+        // Exit loop if we go down below the target block
+        if node_key.version() <= target_version {
+            break;
+        }
+
+        let key_hash = match node {
+            Node::Null => continue,
+            Node::Internal(_) => {
+                if let Err(e) = batch.delete::<JmtNodes>(&node_key) {
```
eyusufatik: don't internal nodes also have keyhash?

crates/storage-ops/src/rollback/components/ledger_db/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+    ledger_db: Arc<sov_schema_db::DB>,
+    target_l2: u64,
+    target_l1: u64,
+) {
+    debug!(
+        "Rolling back Ledger, down to L2 block {}, L1 block {}",
+        target_l2, target_l1
+    );
+
+    match node_type {
```
eyusufatik, line 35: match is unnecessary, no? we do the exact same thing in every node type

crates/storage-ops/src/rollback/components/ledger_db/slots.rs
```diff
@@ -0,0 +1,49 @@
...
+            continue;
+        };
+
+        let slot_height = record.key;
+
+        if slot_height <= SlotNumber(target_l1) {
+            break;
+        }
+
+        if matches!(node_type, StorageNodeType::Sequencer)
```
eyusufatik: for the sequencer case, let's have a cli arg we set ourselves and set the last sequencer commitment sent value to that

bin/citrea/tests/mock/rollback.rs
```diff
@@ -0,0 +1,502 @@
...
+        .await
+        .unwrap();
+
+    drop(rollback);
+    drop(state_db);
+    drop(native_db);
+    drop(ledger_db);
+
+    let (full_node_task_manager, full_node_test_client) =
+        start_full_node(&new_full_node_db_dir, &da_db_dir, seq_port, true).await;
```
eyusufatik, line 561: after rollback let's check if it can keep syncing to the sequencer

bin/citrea/tests/mock/rollback.rs
```diff
@@ -0,0 +1,502 @@
...
+    assert_dbs(full_node_test_client, addr, 40, 40000000000000000000).await;
+
+    full_node_task_manager.abort().await;
+
+    Ok(())
+}
+
+/// Trigger rollback DB data.
+#[tokio::test(flavor = "multi_thread")]
+async fn test_batch_prover_rollback() -> Result<(), anyhow::Error> {
```
eyusufatik, line 593: in this test as well, let's check if batch prover can sync after it's rolledback

eyusufatik (changes requested):

crates/storage-ops/src/rollback/components/state_db.rs
rakanalh: They have a hash but not a key hash. In our case, they give access to the state leafs which for rollback is not that important.

bin/citrea/tests/mock/rollback.rs
rakanalh: Done

bin/citrea/tests/mock/rollback.rs
rakanalh: Done

crates/storage-ops/src/rollback/components/ledger_db/slots.rs
rakanalh: done

crates/storage-ops/src/utils.rs
```diff
@@ -0,0 +1,102 @@
...
+    let mut slots =
+        ledger_db.iter_with_direction::<SlotByHash>(Default::default(), ScanDirection::Backward)?;
+    slots.seek_to_last();
+
+    for record in slots {
+        let Ok(record) = record else {
+            continue;
+        };
+
+        // TODO for pruning this should be less than
```
eyusufatik: todo?

crates/storage-ops/src/rollback/components/ledger_db/slots.rs
```diff
@@ -0,0 +1,35 @@
...
+use crate::pruning::types::StorageNodeType;
+use crate::utils::delete_slots_by_number;
+
+pub(crate) fn rollback_slots(
+    node_type: StorageNodeType,
+    ledger_db: &DB,
+    target_l1: u64,
+) -> anyhow::Result<u64> {
+    let mut slots_to_l2_range = ledger_db
+        .iter_with_direction::<L2RangeByL1Height>(Default::default(), ScanDirection::Backward)?;
```
eyusufatik: we figured out L2RangeByL1Height was not used by anything. we'll be deleting that soon. let's remove everything related to pruning it

crates/storage-ops/src/rollback/components/ledger_db/soft_confirmations.rs
```diff
@@ -0,0 +1,52 @@
...
+            break;
+        }
+
+        if matches!(node_type, StorageNodeType::Sequencer)
+            || matches!(node_type, StorageNodeType::FullNode)
+        {
+            ledger_db.put::<LastSequencerCommitmentSent>(
+                &(),
+                &SoftConfirmationNumber(last_sequencer_commitment_l2_height),
+            )?;
```
eyusufatik: should this be in the for loop now?

bin/citrea/tests/mock/rollback.rs
```diff
@@ -0,0 +1,705 @@
...
+        rollback_l1_height,
+        rollback_l2_height,
+    )
+    .await
+    .unwrap();
+
+    let (seq_task_manager, seq_test_client, _) =
+        start_sequencer(&new_sequencer_db_dir, &da_db_dir, true).await;
+
+    assert_dbs(&seq_test_client, addr, 30, 30000000000000000000).await;
```
eyusufatik: pls assert that sequencer can keep publishing blocks after rollback

bin/citrea/tests/mock/rollback.rs
```diff
@@ -0,0 +1,705 @@
...
+        &new_full_node_db_dir,
+        rollback_l2_height,
+        rollback_l1_height,
+        rollback_l2_height,
+    )
+    .await
+    .unwrap();
+
+    //------------------
+    // Make sure nodes are able to sync after rollback
```
eyusufatik, line 441: can we also try syncing a rolledback full node can sync against a sequencer which was not rolledback?

also since we rollback L1 related stuff. can you check for L1 related data as well? like sequencer commitments.

bin/citrea/tests/mock/rollback.rs
rakanalh: This is done in the other tests since we can make sure that blocks are being published and the other nodes are syncing.

crates/storage-ops/src/utils.rs
rakanalh: Let me handle this in a different PR.

crates/storage-ops/src/rollback/components/ledger_db/slots.rs
rakanalh: I don't think this is a good idea,,, there's no other way to loop through recorded L1s

bin/citrea/tests/mock/rollback.rs
rakanalh: Done

crates/storage-ops/src/rollback/components/ledger_db/slots.rs
rakanalh: Done

bin/citrea/tests/mock/rollback.rs
```diff
@@ -452,6 +452,99 @@ async fn test_fullnode_rollback() -> Result<(), anyhow::Error> {
...
+    )
+    .unwrap();
+    let (full_node_task_manager, full_node_test_client) =
+        start_full_node(&new_full_node_db_dir, &da_db_dir, seq_port, true).await;
+
+    for _ in 0..10 {
+        seq_test_client.spam_publish_batch_request().await.unwrap();
+    }
+    wait_for_l2_block(&seq_test_client, 40, None).await;
+    wait_for_l2_block(&full_node_test_client, 40, None).await;
```
eyusufatik, line 567: can you get state roots with an rpc and compare?

eyusufatik (approved):

PR #1911, opened 2025-02-18
witness over blocks

### LINKED ISSUE #1769: Batch Proof witnesses works over range
Witness is a widely used term through out our repository. For this part of the documentation we’ll define 

**Witness**: Hints for batch proof circuit for state data.

We have two witnesses in the batch proof circuit.

**State witness:** Vector of (value, JMT proof) pairs.

**JMT proof**: Jellyfish-Merkle Tree proof with respect to the root.

**Offchain witness**: Vector of values. This is introduced in the Kumquat fork. The verification of the data provided by the offchain witness is left for `sov-modules`. This is only used for EVM smart contracts right now. The hash of the code of smart contracts in EVM is already committed to the state. So instead of putting the smart contract code into the state tree (JMT), thus the code ends up in DA through our state diffs, we commit the hash to the state, and check if the provided code by the offchain witness has the hash in the state.

The witness generation is done by the prover. When the prover runs an L2 block, it keeps track of state reads (and writes) through the `WorkingSet` passed to the STF. At the end of the run of the L2 block, the prover saves the witnesses generated to LedgerDB. When the prover creates input for the batch proof, these witnesses are passed to the circuit as input. The working set is also used in the circuit side, and the circuit storage implementation `ZkStorage` pops witness elements from the witnesses given as input.   
  
The `WorkingSet` also acts as a cache on top of the storage, so once a value is read from the storage, it is then served from the `WorkingSet.` 

The ProverStorage-Witness-ZkStorage relationship is only implemented per L2 block. However, we run multiple L2 blocks in a single run of the batch proof circuit. With the current implementation, we reset the working set after every L2 block, meaning that keys read from the state or the offchain witness are re-read (and re-verified) if they are used again in another L2 block in the same run of the circuit.

If we could generate the witnesses for a range of L2 blocks, we would do way less witness verifications resulting in dramatic decrease in cycle counts.

Needed changes:

1. On the circuit side, `WorkingSet` ‘s are not resetted after a run of the L2 block so that we can keep using it as a cache over multiple L2 blocks. There is some problems regarding state root calculation and JMT update proofs that we'll need to figure out here.
2. The batch prover can create witnesses over an L2 block range.
3. The batch prover is aware of the memory limitations of the circuit, so it make sure for a given range of L2 blocks, the size of the cache (WorkingSet) will not go over 20 MBs. Meaning that for a range found in sequencer commitments, the batch prover prepares it Vec<(Witness, OffchainWitness)> for subranges that won’t cause the circuit to go over its memory limits.

yaziciahmet:
# Description

Before, on every block hook cycle, we were caching the values we read from the storage and to-be committed writes, and discarding it after cycle is completed, starting an empty cache on a new block hook cycle. The problem with this was that even if block 2 reads the same key from the storage as block 1, we have to generate witness per-block basis, which made the input big, and caused unnecessary merkle proof verification.

With this PR, I introduced a reuseable cache log which is returned at the end of finalize hook, and can be re-inputted back into to apply soft confirmation hook, which will allow us to avoid passing and verifying witnesses for the same keys to zk guest.

When applying blocks in the runners of the any node type, this is irrelevant, and blocks are applied one-by-one with no cache reuse. This is only used when replaying transactions to generate witness input for the batch proof circuit.

This cache is called `ReadWriteLog`, and its lifecycle looks like this:
1. Initialize it empty for block N
2. Block N applied and log is populated with reads and writes
3. Block N finished and log is returned
4. Initialize the log of Block N+1 with the previously returned cache, and ensure that all writes are marked as reads while doing so (previous block's writes are current blocks reads)
5. Same loop continues until all blocks are applied

One problem with this approach is that, cache could grow quite big and that would cause OOM issues in the guest environment. Second problem is that, sha2 is already quite cheap in zk, and memory (paging) overhead of this approach in zk might not give as much benefit as we hope it would.

The solution to OOM issue is resolved by pruning half of the cache when a single cache size limit has exceeded (64 MB). Keep in mind there are 2 caches we cumulate, state and offchain. This size calculation excludes the pointer and BTreeMap node overhead since that is relatively hard to consider, but more importantly, these sizes actually change in 64-bit vs 32-bit environments, causing mismatch in between native and zk. So the size calculation is just an approximation. 

Since it is an unwanted overhead to keep track of the cache size in zk, I instead calculated the size on the native side, and passed the l2 heights to prune the cache to the zk as input.

## Linked Issues
- Fixes #1769 

## Testing

## Docs


yaziciahmet:
Ran 50k transaction test

**Nightly:**
2025-02-19T00:28:20.472983Z INFO risc0_zkvm::host::server::exec::executor: execution time: 8280.911531875s
2025-02-19T00:28:20.473242Z INFO risc0_zkvm::host::server::session: number of segments: 105422
2025-02-19T00:28:20.473244Z INFO risc0_zkvm::host::server::session: 110542979072 total cycles
2025-02-19T00:28:20.473246Z INFO risc0_zkvm::host::server::session: 60711324018 user cycles (54.92%)
2025-02-19T00:28:20.473248Z INFO risc0_zkvm::host::server::session: 49377295738 paging cycles (44.67%)
2025-02-19T00:28:20.473250Z INFO risc0_zkvm::host::server::session: 454359316 reserved cycles (0.41%)
2025-02-19T00:28:20.473251Z INFO risc0_zkvm::host::server::session: ecalls
2025-02-19T00:28:20.473253Z INFO risc0_zkvm::host::server::session:  86443169 BigInt2 calls, 11439295685 cycles, (10.35%)
2025-02-19T00:28:20.473254Z INFO risc0_zkvm::host::server::session:  51965433 Sha2 calls, 4272667138 cycles, (3.87%)
2025-02-19T00:28:20.473361Z INFO risc0_zkvm::host::server::session:  140651271 BigInt calls, 1406512710 cycles, (1.27%)
2025-02-19T00:28:20.473362Z INFO risc0_zkvm::host::server::session:  6149837 Software calls, 114072443 cycles, (0.10%)
2025-02-19T00:28:20.473363Z INFO risc0_zkvm::host::server::session:  0 Input calls, 0 cycles, (0.00%)
2025-02-19T00:28:20.473365Z INFO risc0_zkvm::host::server::session: syscalls
2025-02-19T00:28:20.473366Z INFO risc0_zkvm::host::server::session:  5818953 Keccak calls
2025-02-19T00:28:20.473367Z INFO risc0_zkvm::host::server::session:  313073 Read calls
2025-02-19T00:28:20.473368Z INFO risc0_zkvm::host::server::session:  8884 VerifyIntegrity calls
2025-02-19T00:28:20.473370Z INFO risc0_zkvm::host::server::session:  8884 ProveKeccak calls
2025-02-19T00:28:20.473378Z INFO risc0_zkvm::host::server::session:  40 Write calls

**Witness-over-blocks**:
2025-02-18T21:51:06.403376Z INFO risc0_zkvm::host::server::exec::executor: execution time: 8191.903570875s
2025-02-18T21:51:06.403792Z INFO risc0_zkvm::host::server::session: number of segments: 105675
2025-02-18T21:51:06.403794Z INFO risc0_zkvm::host::server::session: 110808268800 total cycles
2025-02-18T21:51:06.403796Z INFO risc0_zkvm::host::server::session: 58297458026 user cycles (52.61%)
2025-02-18T21:51:06.403798Z INFO risc0_zkvm::host::server::session: 52055645948 paging cycles (46.98%)
2025-02-18T21:51:06.403799Z INFO risc0_zkvm::host::server::session: 455164826 reserved cycles (0.41%)
2025-02-18T21:51:06.403800Z INFO risc0_zkvm::host::server::session: ecalls
2025-02-18T21:51:06.403802Z INFO risc0_zkvm::host::server::session:  86442890 BigInt2 calls, 11439256554 cycles, (10.32%)
2025-02-18T21:51:06.403816Z INFO risc0_zkvm::host::server::session:  49429975 Sha2 calls, 4084805722 cycles, (3.69%)
2025-02-18T21:51:06.403817Z INFO risc0_zkvm::host::server::session:  140651271 BigInt calls, 1406512710 cycles, (1.27%)
2025-02-18T21:51:06.403819Z INFO risc0_zkvm::host::server::session:  5952898 Software calls, 110954726 cycles, (0.10%)
2025-02-18T21:51:06.403820Z INFO risc0_zkvm::host::server::session:  0 Input calls, 0 cycles, (0.00%)
2025-02-18T21:51:06.403821Z INFO risc0_zkvm::host::server::session: syscalls
2025-02-18T21:51:06.403822Z INFO risc0_zkvm::host::server::session:  5818593 Keccak calls
2025-02-18T21:51:06.403824Z INFO risc0_zkvm::host::server::session:  116494 Read calls
2025-02-18T21:51:06.403825Z INFO risc0_zkvm::host::server::session:  8884 VerifyIntegrity calls
2025-02-18T21:51:06.403900Z INFO risc0_zkvm::host::server::session:  8884 ProveKeccak calls
2025-02-18T21:51:06.403901Z INFO risc0_zkvm::host::server::session:  40 Write calls

Total cycles did not have any meaningful change, but the user cycles dropped and paging cycles increased, expectedly since we now rely heavily on the cache. And based on the discussions with risc0 team, this seems to be something we would want due to the fact that paging cycles will be drastically improved with risc0 version 1.4.0, and our bottleneck will be user cycles.

crates/sovereign-sdk/module-system/sov-state/src/zk_storage.rs
```diff
@@ -79,17 +79,16 @@ where
             }
         }
 
-        let mut diff = Vec::with_capacity(state_accesses.ordered_writes.len());
```
eyusufatik, line 82: why?

crates/sovereign-sdk/module-system/sov-state/src/zk_storage.rs
eyusufatik, line 1: also I've realized we could make state diff calc take less cycles now due to how we use caches now

eyusufatik: can bring down cycles a lot

eyusufatik:
for the same blocks, we should check if nightly and this branch yield same size of state diffs btw

crates/batch-prover/src/da_block_handler.rs
```diff
@@ -414,29 +398,132 @@ pub(crate) async fn get_batch_proof_circuit_input_from_commitments<
...
     Ok((
         short_header_proofs,
         state_transition_witnesses,
+        cache_prune_l2_heights,
         committed_l2_blocks,
         da_block_headers_of_l2_blocks,
     ))
 }
 
+async fn generate_cumulative_witness<'txs, Da: DaService, DB: BatchProverLedgerOps>(
```
eyusufatik: better if in proving.rs

crates/batch-prover/src/da_block_handler.rs
```diff
@@ -37,50 +38,52 @@ use crate::errors::L1ProcessingError;
...
     DB: BatchProverLedgerOps,
-    Witness: Default + BorshSerialize + BorshDeserialize + Serialize + DeserializeOwned,
 {
     prover_config: BatchProverConfig,
     prover_service: Arc<ParallelProverService<Da, Vm>>,
     ledger_db: DB,
     da_service: Arc<Da>,
+    storage_manager: ProverStorageManager,
     sequencer_pub_key: Vec<u8>,
+    sequencer_k256_pub_key: Vec<u8>,
```
eyusufatik, line 46: how the hell did everything work before when we didn't supply the prover with k256 pubkey before @ercecan?

crates/batch-prover/src/da_block_handler.rs
```diff
@@ -307,45 +312,32 @@ pub(crate) async fn get_batch_proof_circuit_input_from_commitments<
...
-            Err(e) => {
-                return Err(anyhow!(
-                    "Failed to get soft confirmations from the ledger db: {}",
-                    e
-                ));
-            }
-        };
+
+        let soft_confirmations_in_commitment = ledger_db
+            .get_soft_confirmation_range(
```
eyusufatik: get range might return less than what we ask for, in that case return err so that prover can try later (when it syncs to those blocks)

crates/batch-prover/src/da_block_handler.rs
```diff
@@ -414,29 +398,132 @@ pub(crate) async fn get_batch_proof_circuit_input_from_commitments<
...
+                    offchain_log.prune_half();
+                    cache_prune_l2_heights.push(l2_height);
+                }
+
+                cumulative_state_log = Some(state_log);
+                cumulative_offchain_log = Some(offchain_log);
+            }
+
+            witnesses.push((
+                soft_confirmation_result.witness,
```
eyusufatik: here we are pushing some roots and update proof when we read everything from cache right?

crates/batch-prover/src/da_block_handler.rs
```diff
@@ -414,29 +398,132 @@ pub(crate) async fn get_batch_proof_circuit_input_from_commitments<
...
+        let should_use_cache = fork_from_block_number(
+            l2_blocks_in_commitment
+                .last()
+                .expect("must have at least one")
+                .l2_height(),
+        )
+        .spec_id
+            >= SpecId::Fork2;
+
+        for l2_block in l2_blocks_in_commitment {
```
eyusufatik: btw we are not resetting caches when we move over to another sequencer commitment right?

crates/sovereign-sdk/module-system/sov-modules-core/src/storage/cache.rs
```diff
@@ -94,6 +106,18 @@ impl Access {
         }
     }
 
+    pub fn size(&self) -> usize {
```
eyusufatik, line 110: here shouldn't you include the size of the enum selector?

btw we can and should make the Access enum repr() with u8 -- or u32 (zkvm memsize)

crates/sovereign-sdk/module-system/sov-modules-core/src/storage/cache.rs
```diff
@@ -292,6 +313,43 @@ impl CacheLog {
...
+        }
+
+        let mid_idx = self.log.len() / 2;
+        let mid_key = self.log.keys().nth(mid_idx).expect("must exist").clone();
+        self.log.split_off(&mid_key);
+    }
+
+    /// Returns the estimated heap size of the `CacheLog`. This doesn't account for
+    /// pointer and node overhead coming from BTreeMap.
+    pub fn estimated_size(&self) -> usize {
```
eyusufatik, line 349: this not called in zk now right?

crates/batch-prover/src/runner.rs
eyusufatik, line 1: remove saving witness on normal L2 sync

eyusufatik (changes requested):

crates/sovereign-sdk/module-system/sov-state/src/zk_storage.rs
yaziciahmet: ordered_writes are not known beforehand anymore, because we don't convert the CacheLog into writes and we keep it as CacheLog, which is mixed with read and write accesses. I will try to improve the cache logic in general, and this is something I will try to address when I am doing that. And this is not taking that much of a space anymore, because the diff is wrapped with Rc, what this Vec actually stores is just pointers.

crates/batch-prover/src/da_block_handler.rs
yaziciahmet: yes, since we re-feed the cache back into the apply_soft_confirmation method, the witness it generates is with respect to the already existing cache. and we collect them here.

crates/batch-prover/src/da_block_handler.rs
yaziciahmet: no, cache is only resetted when it exceeds a certain memory limit.

crates/sovereign-sdk/module-system/sov-modules-core/src/storage/cache.rs
yaziciahmet: we sure can, there are lots of things we don't take into account (lots as number of things, but quite small relatively in bytes), but this should relatively easy to do so.

crates/sovereign-sdk/module-system/sov-modules-core/src/storage/cache.rs
yaziciahmet: nope, only prune_half is called in zk. this is called in native, if size is greater than limit, the block number is pushed as an input, and in zk when we encounter this block number hint, we just call prune_half.

crates/batch-prover/src/da_block_handler.rs
yaziciahmet: it is already checked in the data_to_prove method before calling this function. so maybe panic here is more appropriate?

crates/batch-prover/src/da_block_handler.rs
yaziciahmet: done

crates/batch-prover/src/da_block_handler.rs
yaziciahmet: done

crates/sovereign-sdk/module-system/sov-modules-core/src/storage/cache.rs
yaziciahmet: done

crates/batch-prover/src/runner.rs
yaziciahmet: I think lets keep it temporarily, in case something goes wrong it might be good to have witness stored. After testing it a bit on testnet, we can remove it right away no problem.

crates/batch-prover/src/da_block_handler.rs
yaziciahmet: or at the end of the proving

crates/batch-prover/src/da_block_handler.rs
ercecan: In runner.rs

eyusufatik (approved):

Issue #1931, opened 2025-02-19
Consider borsh::DEFAULT_SERIALIZER_CAPACITY


kpp:
This is how it's used:

```
pub(super) const DEFAULT_SERIALIZER_CAPACITY: usize = 1024;

pub fn to_vec<T>(value: &T) -> Result<Vec<u8>>
where
    T: BorshSerialize + ?Sized,
{
    let mut result = Vec::with_capacity(DEFAULT_SERIALIZER_CAPACITY);
    value.serialize(&mut result)?;
    Ok(result)
}
```

Probably we may want to reduce that number and test cycles.

jfldde:
Was just having a look at it this morning. I'll open a PR

kpp:
@jfldde  Please also take a look at https://github.com/chainwayxyz/citrea/pull/1925

jfldde:
Fixed in #1934

PR #1934, opened 2025-02-20
Preallocate exact size for borsh serialization

kpp:
# Description

- [x] Optimize AlloyU256, AlloyAddress, ModuleAddress, AccountInfo
- [x] Replace `impl<T> StateValueCodec<Vec<T>> for BorshCodec` with a concrete impls for missing T

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/1931

### DIFF
--- a/crates/evm/src/evm/mod.rs
+++ b/crates/evm/src/evm/mod.rs
@@ -89,7 +89,9 @@ impl StateValueCodec<AccountInfo> for BorshCodec {
     type Error = std::io::Error;
 
     fn encode_value(&self, value: &AccountInfo) -> Vec<u8> {
-        borsh::to_vec(&value).unwrap()
+        let mut buf = Vec::with_capacity(32 + 8 + 32 + 1);
+        BorshSerialize::serialize(value, &mut buf).unwrap();
+        buf
     }
 
     fn try_decode_value(&self, bytes: &[u8]) -> Result<AccountInfo, Self::Error> {
--- a/crates/sovereign-sdk/module-system/sov-state/src/codec/borsh_codec.rs
+++ b/crates/sovereign-sdk/module-system/sov-state/src/codec/borsh_codec.rs
@@ -76,7 +76,9 @@ macro_rules! impl_borsh_codec {
     ($t:tt) => {
         impl StateKeyCodec<$t> for BorshCodec {
             fn encode_key(&self, value: &$t) -> Vec<u8> {
-                borsh::to_vec(value).expect("Failed to serialize key")
+                let mut buf = Vec::with_capacity(8);
+                BorshSerialize::serialize(value, &mut buf).unwrap();
+                buf
             }
         }
 
@@ -84,7 +86,9 @@ macro_rules! impl_borsh_codec {
             type Error = std::io::Error;
 
             fn encode_value(&self, value: &$t) -> Vec<u8> {
-                borsh::to_vec(value).expect("Failed to serialize value")
+                let mut buf = Vec::with_capacity(8);
+                BorshSerialize::serialize(value, &mut buf).unwrap();
+                buf
             }
 
             fn try_decode_value(&self, bytes: &[u8]) -> Result<$t, Self::Error> {
@@ -121,6 +125,8 @@ where
     T: BorshSerialize,
 {
     fn encode_key_like(&self, borrowed: &[T]) -> Vec<u8> {
-        borsh::to_vec(borrowed).unwrap()
+        let mut buf = Vec::with_capacity(64);
+        BorshSerialize::serialize(borrowed, &mut buf).unwrap();
+        buf
     }
 }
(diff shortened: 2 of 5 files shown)


crates/evm/src/evm/mod.rs
```diff
@@ -89,7 +89,9 @@ impl StateValueCodec<AccountInfo> for BorshCodec {
     type Error = std::io::Error;
 
     fn encode_value(&self, value: &AccountInfo) -> Vec<u8> {
-        borsh::to_vec(&value).unwrap()
+        let mut buf = Vec::with_capacity(32 + 8 + 32 + 1);
```
yaziciahmet, line 92: do you think it would be better if we check code_hash.is_some and add the last +32 accordingly? It is also possible that compiler might make better optimizations if the capacity is constant, and extra +32 bytes almost never gonna mean much.

crates/sovereign-sdk/module-system/sov-state/src/codec/borsh_codec.rs
```diff
@@ -41,43 +48,43 @@ impl StateValueCodec<AlloyU256> for BorshCodec {
...
 
         impl StateValueCodec<$t> for BorshCodec {
             type Error = std::io::Error;
 
             fn encode_value(&self, value: &$t) -> Vec<u8> {
-                borsh::to_vec(value).expect("Failed to serialize value")
+                let mut buf = Vec::with_capacity(8);
+                BorshSerialize::serialize(value, &mut buf).unwrap();
+                buf
             }
```
yaziciahmet, line 101: maybe 8 is too small? not sure where this is used exactly, but on the top off my head 64 or 128 seems a bit more appropriate

crates/evm/src/evm/mod.rs
kpp: Unfortunately we go into Allocator which doesn't care much about that difference. We need to bench this case but I doubt it will matter at all

crates/sovereign-sdk/module-system/sov-state/src/codec/borsh_codec.rs
kpp: So for int primitives we may allocate exact size. For String it should be 4 + str.len()

crates/evm/src/evm/mod.rs
yaziciahmet: then lets keep it this way, its better than 1024 in any case

crates/sovereign-sdk/module-system/sov-state/src/codec/borsh_codec.rs
yaziciahmet: you can take the capacity as 2nd arg into the macro for easily doing it for uint and int types. for string, you can manually implement it

crates/sovereign-sdk/module-system/sov-state/src/codec/mod.rs
```diff
@@ -19,6 +19,18 @@ mod tests {
 
     use super::*;
 
+    impl StateValueCodec<Vec<i32>> for BorshCodec {
+        type Error = std::io::Error;
+
+        fn encode_value(&self, value: &Vec<i32>) -> Vec<u8> {
+            borsh::to_vec(value).unwrap()
```
yaziciahmet, line 26: size = 4 + (4* value.len)

jfldde: Just used locally in this test

crates/sovereign-sdk/module-system/sov-state/src/codec/borsh_codec.rs
jfldde: We only use String in tests, rest should be fine

jfldde: AFAICT only `usize` and `u64` are used out of tests

yaziciahmet (approved):

crates/sovereign-sdk/module-system/sov-state/src/codec/borsh_codec.rs
```diff
@@ -114,6 +125,8 @@ where
     T: BorshSerialize,
 {
     fn encode_key_like(&self, borrowed: &[T]) -> Vec<u8> {
-        borsh::to_vec(borrowed).unwrap()
+        let mut buf = Vec::with_capacity(64);
```
kpp: Why?

jfldde: Prevent default allocating default 1024 with borsh::to_vec().
Chose the lowest power of 2 above the common (only?) value of 37. We can try fine tuning it

jfldde (approved):

PR #1939, opened 2025-02-20
Lift L1 fields out of L2Header

### LINKED ISSUE #1791: Remove L1 related fields from L2BlockHeader
May be done before or after #1766, does not matter.

The `l1_hash` and `l1_height` should be removed from soft confirmation and signed soft confirmation structures, because #1788 will now handle system transactions inside normal txs field of the soft confirmation.

jfldde:
# Description

- Lift L1 related fields out of header and keep at L2Block level for backward compatibility reasons.
Further versioning work should take care of removing this fields from this L2Block variant as well.

## Linked Issues
- Fixes #1791

### DIFF
--- a/crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
+++ b/crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
@@ -11,8 +11,6 @@ use serde::{Deserialize, Serialize};
 #[derive(PartialEq, Eq, BorshDeserialize, BorshSerialize, Serialize, Deserialize, Clone, Debug)]
 pub struct L2Header {
     l2_height: u64,
-    da_slot_height: u64,
-    da_slot_hash: [u8; 32],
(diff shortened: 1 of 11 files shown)


crates/sovereign-sdk/rollup-interface/src/state_machine/soft_confirmation.rs
```diff
@@ -100,6 +92,10 @@ pub struct L2Block<'txs, Tx: Clone + BorshSerialize> {
     pub blobs: Cow<'txs, [Vec<u8>]>,
     /// Deposit data
     pub deposit_data: Vec<Vec<u8>>,
+    /// L1 height
```
eyusufatik, line 99: can you put a comment describing everythin except txs and header is here for compatability reasons?

kpp, line 99: So these should be `// TODO: Remove before mainnet`?

eyusufatik (approved):

PR #1941, opened 2025-02-20
Use tinyvec in Prefix and StorageKey

### LINKED ISSUE #1933: Optimize StorageKey::new
Each statemap get/set creates a StorageKey with `StorageKey::new` which internally allocates a Vec. We may optimize out allocation by using https://docs.rs/smallvec/latest/smallvec/struct.SmallVec.html

kpp:
# Description

Use TinyVec in Prefix and StorageKey for optimizations.

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/1933


yaziciahmet:
I benched cycle counts of both nightly and this pr 3 times.

**Nightly:**
lowest: 13.60 billion
highest: 13.63 billion

**This branch:**
lowest: 13.40 billion
highest: 13.43 billion

This PR seems to have a stable 1.5% cycle reduction. Great job!

Cargo.toml
```diff
@@ -104,6 +104,7 @@ sha2 = { version = "0.10.8", default-features = false, features = ["asm"] }
 schemars = { version = "0.8.16", features = ["derive"] }
 secp256k1 = { version = "0.29.0", default-features = false, features = ["global-context", "recovery"] }
 thiserror = { version = "1.0.50", default-features = false }
+tinyvec = { version = "1", default-features = false, features = ["alloc", "rustc_1_61", "rustc_1_57", "rustc_1_55"] }
```
yaziciahmet, line 107: do all these features needed?

crates/sovereign-sdk/module-system/sov-modules-core/Cargo.toml
```diff
@@ -57,5 +58,5 @@ std = [
     "thiserror",
 ]
 serde = []
-sync = ["borsh/rc", "serde/rc"]
+sync = ["borsh/rc", "serde/rc", "tinyvec/serde"]
```
yaziciahmet, line 61: why enabled with sync feat?

crates/sovereign-sdk/module-system/sov-modules-core/src/common/bytes.rs
```diff
@@ -13,22 +16,32 @@ use crate::module::Context;
...
-    )
-)]
+#[cfg_attr(feature = "sync", derive(serde::Serialize, serde::Deserialize,))]
 pub struct Prefix {
-    prefix: Vec<u8>,
+    pub(crate) data: SmallData,
+}
+
+/// Mostly always inlined data
+pub type SmallData = TinyVec<[u8; 64]>;
```
yaziciahmet, line 25: SmallData is a bit confusing. TinyVec64 or smh. feels more appropriate to me

yaziciahmet:
I get the StorageKey being 64 bytes. But arent all our prefixes even smaller than that? Max 16 bytes (maybe even 8 bytes didnt check exactly) should be enough for prefixes. Maybe we can have 16 sized TinyVec for Prefix?

yaziciahmet:
> I get the StorageKey being 64 bytes. But arent all our prefixes even smaller than that? Max 16 bytes (maybe even 8 bytes didnt check exactly) should be enough for prefixes. Maybe we can have 16 sized TinyVec for Prefix?

I see what you did now. you use same size for key and prefix so when prefix is extended with key, the same type can be used without allocating or creating a new TinyVec. Good thinking.

yaziciahmet:
Forgive me guys, I actually forgot to pull the last nightly merge commit while benching... Here are the actual cycle counts:

**Nightly (same because I had it correctly when benching)**:
lowest: 13.60 billion
highest: 13.63 billion

This branch:
lowest: 13.55 billion
highest: 13.58 billion

It still seems that this pr still brings improvement, but much less visible.

crates/sovereign-sdk/module-system/sov-modules-core/Cargo.toml
kpp: It used to be that way. I didn't change anything.

crates/sovereign-sdk/module-system/sov-modules-core/src/common/bytes.rs
kpp: Well, the underlying type can be smallvec::SmallVec too. Or any other crate that implements this functionality. That's why it's SmallData

Cargo.toml
kpp: Yes...

eyusufatik (approved):

PR #1947, opened 2025-02-21
ReadWriteLog::iter_ordered_writes returns DoubleEndedIterator

kpp:
# Description

Because DoubleEndedIterator is more powerful and I will need it in ststdiff


yaziciahmet (approved):
Looks good, keep in mind that the order is not the write order of the keys, but literally their byte-ordering.

kpp:
Well this PR was implemented because I needed to iter ordered writes in reverse order to be able to calc diff from [last -> first] write. Now it makes no sense actually.

PR #1950, opened 2025-02-21
Versionned Transaction

jfldde:
# Description

- Keep transactions variant behind main Transaction enum
- Remove blobs from blocks. This can now be derived by correct transaction variant.
- Uses magic marker to keep backward compatibility with previous serialization format


crates/sovereign-sdk/module-system/sov-modules-api/src/transaction.rs
```diff
@@ -26,192 +32,205 @@ pub struct Transaction {
...
-            chain_id: value.chain_id,
-            nonce: value.nonce,
+/// The versioned transaction type
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Transaction {
+    /// Fork 2 version using k256 signatures
+    V2(TransactionV2),
+    /// Pre Fork 2 version using ed25519 signatures
+    V1(Box<TransactionV1>),
+}
```
yaziciahmet: two things:
1. please put V1 before V2 :D 
2. why V1 is boxed?

jfldde: 2 - Fixes `large size difference between variants` clippy lint
1 - Dubious but why not

eyusufatik (approved):
aproving, @ercecan should take a look

I hope this does not break bacwards compatibility. not too sure. however, we can fix those concerns later, there is already that problem on nightly as well

jfldde:
It does sync against current testnet sequencer (with switching tx_merkle_root to optional). I'll add some test fixtures for testing backward ser/deser compatibility in another PR

ercecan (approved):

PR #1954, opened 2025-02-21
New Witness type

yaziciahmet:
# Description

There is no reason for us to track index in Witness. We can just use VecDequeue. It seems to slightly lower cycle counts ~1% most likely due to less Read Calls when deserializing. The main aim of this PR is not even lowering cycle counts, but versioning the Witness so that I can play around with new Witness types easily without thinking about anything else.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/sovereign-sdk/rollup-interface/src/state_machine/witness.rs
+++ b/crates/sovereign-sdk/rollup-interface/src/state_machine/witness.rs
@@ -1,3 +1,5 @@
+//! Defines versioned witness types to be used.
+
 use std::collections::VecDeque;
 
 use borsh::{BorshDeserialize, BorshSerialize};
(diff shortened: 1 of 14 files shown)


crates/sovereign-sdk/rollup-interface/src/state_machine/witness.rs
```diff
@@ -1,5 +1,16 @@
...
 use borsh::{BorshDeserialize, BorshSerialize};
 
+/// PreFork2Witness should only be used for forks < Fork2
+#[derive(BorshDeserialize, BorshSerialize)]
+pub struct PreFork2Witness {
+    next_idx: usize,
+    hints: Vec<Vec<u8>>,
+}
+
 /// A [`Vec`]-based implementation of [`Witness`] with no special logic.
```
eyusufatik: VecDequeu based

eyusufatik (approved):
I think the %1 is error  the benchmark but this is still better

PR #1959, opened 2025-02-22
Don't remove Evm.latest_block_hashes

### LINKED ISSUE #1957: On Fork2 don't delete block hashes
Don't delete block hashes if after Fork2, but keep limiting access through the revm trait

kpp:
# Description

There is no reason to remove them from the state at all.
We remove them only before Fork2 for backwards compatibility.

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/1957


eyusufatik (changes requested):
requested changes irl

eyusufatik (approved):

PR #1962, opened 2025-02-22
batch proof always outputs latest l1 hash on light client contract

eyusufatik:
# Description
new findings new implementations, this way is more secure

### DIFF
--- a/bin/citrea/tests/bitcoin/batch_prover_test.rs
+++ b/bin/citrea/tests/bitcoin/batch_prover_test.rs
@@ -790,6 +790,13 @@ impl TestCase for L1HashOutputTest {
         }
     }
 
+    /// The test consists of two parts.
+    /// Part 1 we make lots of DA blocks so the Bitcoin Light Client Contract gets updated many times
+    /// in a single proof.
+    /// Then we assert that the proof output has the latest L1 hash.
+    ///
+    /// Then we don't make any new DA blocks, make a single proof and assert that we still have
+    /// the same L1 hash outputted.
     async fn run_test(&mut self, f: &mut TestFramework) -> Result<()> {
         let da = f.bitcoin_nodes.get(0).unwrap();
         let sequencer = f.sequencer.as_ref().unwrap();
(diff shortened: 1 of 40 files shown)


crates/citrea-stf/src/verifier.rs
```diff
@@ -77,17 +83,108 @@ where
...
+    last_hash_witness: &mut Witness,
+    final_state_root: StorageRootHash,
+) -> [u8; 32] {
+    let prefix = {
+        let temp_evm = Evm::<ZkDefaultContext>::default();
+        temp_evm.storage.prefix().clone()
+    };
+
+    // key for light client contract next l1 height
+    let inner_evm_key = Evm::<ZkDefaultContext>::get_storage_address(
```
yaziciahmet: this function is used in both native and zk. is it ok to use it with ZkDefaultContext?

crates/short-header-proof-provider/src/native.rs
```diff
@@ -45,20 +45,24 @@ impl<Da: DaSpec> ShortHeaderProofProvider for NativeShortHeaderProofProviderServ
...
+                let return_cond = txs_commitment == l1_update_info.tx_commitment
                     && block_hash == l1_update_info.header_hash
                     && prev_hash_cond
-                    && l1_height == l1_update_info.block_height);
+                    && l1_height == l1_update_info.block_height;
+
+                if return_cond {
+                    self.quried_and_verified_hashes
+                        .lock()
+                        .expect("Should lock quried and verified hashes")
```
yaziciahmet: typo: queried

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/batch_proof/input/mod.rs
```diff
@@ -158,6 +160,7 @@ where
                 da_block_headers_of_soft_confirmations: self.da_block_headers_of_l2_blocks,
                 sequencer_commitments: self.sequencer_commitments,
                 cache_prune_l2_heights: self.cache_prune_l2_heights,
+                last_hash_witness: self.last_hash_witness,
```
yaziciahmet: add l1 somewhere in this variable for clarity.

yaziciahmet (changes requested):
looks good. not so detailed review. will do review one more time.

crates/citrea-stf/src/verifier.rs
eyusufatik: yeah forgot to put a generic for that

eyusufatik:
check usage of l1_hash.to_be_bytes

bin/citrea/tests/bitcoin/batch_prover_test.rs
```diff
@@ -790,6 +790,13 @@ impl TestCase for L1HashOutputTest {
...
     }
 
+    /// The test consists of two parts.
+    /// Part 1 we make lots of DA blocks so the Bitcoin Light Client Contract gets updated many times
+    /// in a single proof.
+    /// Then we assert that the proof output has the latest L1 hash.
+    ///
+    /// Then we don't make any new DA blocks, make a single proof and assert that we still have
+    /// the same L1 hash outputted.
     async fn run_test(&mut self, f: &mut TestFramework) -> Result<()> {
```
yaziciahmet, line 800: on line 817, wait mempool len should be 4 with its current state, as 100 blocks will invoke 2 commitments

bin/citrea/tests/bitcoin/batch_prover_test.rs
```diff
@@ -829,21 +836,90 @@ impl TestCase for L1HashOutputTest {
...
+        assert_eq!(hash_from_rpc.as_raw_hash().to_byte_array(), l1_hash);
+
+        // part 2
+        for _ in 0..110 {
+            sequencer.client.send_publish_batch_request().await?;
+        }
 
-        for (height, hash) in (start_l1_height..=start_l1_height + 100).zip(l1_hashes) {
-            let hash_from_rpc = sequencer.da.get_block_hash(height).await?;
+        da.wait_mempool_len(6, None).await?;
```
yaziciahmet, line 855: This should now be 4?

bin/citrea/tests/bitcoin/batch_prover_test.rs
```diff
@@ -829,21 +836,90 @@ impl TestCase for L1HashOutputTest {
...
 
-            assert_eq!(hash_from_rpc.as_raw_hash().to_byte_array(), hash);
+        for _ in 0..51 {
+            sequencer.client.send_publish_batch_request().await?;
         }
 
+        da.wait_mempool_len(8, None).await?;
```
yaziciahmet, line 861: 6?

crates/citrea-stf/src/verifier.rs
```diff
@@ -77,17 +84,157 @@ where
...
+                None => {
+                    // If this is the first proof in Fork2 and we haven't changed the height yet
+                    // we must get from pre fork2 storage.
+                    //
+                    // We don't even check storage cache with pre fork2 keys here because
+                    // if pre fork2 storage is read in Fork2,
+                    // it would be written also be written to the cache with fork2 keys.
+                    // And we wouldn't be here.
+
+                    let pre_fork2_key = Evm::<C>::get_storage_key_pre_fork2(
```
yaziciahmet, line 159: Don't we know the spec here? If so use it here to avoid trying to understand via key existence.

yaziciahmet (dismissed):

crates/citrea-stf/src/verifier.rs
eyusufatik: Yes but migration of storage happens by invocation basis on fork2. So on the first fork2 proof, there is a chance this happens.

bin/citrea/tests/bitcoin/batch_prover_test.rs
eyusufatik: 6 because 2 from zk proof before

4 from 2 new sequencer commitments

bin/citrea/tests/bitcoin/batch_prover_test.rs
eyusufatik: well no, catchup blocks does not cause a commitment like that. I think we are going to be rewriting that part.

bin/citrea/tests/bitcoin/batch_prover_test.rs
eyusufatik: > 6 because 2 from zk proof before

we did not mine it

bin/citrea/tests/bitcoin/batch_prover_test.rs
eyusufatik: 8 because same as above

yaziciahmet (approved):

PR #1966, opened 2025-02-24
Port e2e syncing tests

jfldde:
# Description

- Port e2e syncing tests
- This catches an issue where nonce is not properly increased on `update_sequencer_authority` and block fails with `dry_run_transactions should have already checked this: HookError(SovTxBadNonce)`
- This means that any block with multiple sov TX would fail since we don't increase account nonce between the two txs.
- No tests were previously checking for this. This is now properly checked by `test_sync_status`.
- MockDA tests are not catching this, I assume because they run with `test_mode: false` and don't deterministically produce blocks

- Adding HOLD-MERGE tag since this new test will fail until we fix nonce increase in sequencer runner


crates/sequencer/src/runner.rs
```diff
@@ -510,18 +516,23 @@ where
...
             let raw_message = <CitreaRuntime<DefaultContext, Da::Spec> as EncodeCall<
                 citrea_evm::Evm<DefaultContext>,
             >>::encode_call(call_txs);
 
-            let signed_tx = self.sign_tx(
-                raw_message,
-                &mut working_set,
-                soft_confirmation_info.current_spec(),
-            )?;
+            println!("signing tx in produce l2 block got evm tx");
```
eyusufatik: remove print

crates/sequencer/src/runner.rs
```diff
@@ -534,10 +545,11 @@ where
             if next_fork.spec_id == SpecId::Fork2
                 && soft_confirmation_info.l2_height() + 1 == next_fork.activation_height
             {
-                let (signed_blob, signed_tx) = self.update_sequencer_authority(&mut working_set, soft_confirmation_info.current_spec()).expect("Should create and sign soft confirmation rule enforcer authority change call messages");
+                let (signed_blob, signed_tx) = self.update_sequencer_authority( soft_confirmation_info.current_spec(), nonce).expect("Should create and sign soft confirmation rule enforcer authority change call messages");
                 blobs.push(signed_blob);
                 txs.push(signed_tx);
             }
+            println!("updating seq auth");
```
eyusufatik: remove print

crates/sequencer/src/runner.rs
```diff
@@ -1121,7 +1130,8 @@ where
             >,
         >>::encode_call(rule_enforcer_call_tx);
 
-        let signed_tx = self.sign_tx(raw_message, working_set, current_spec)?;
+        println!("singing tx and blob in update seq auth");
```
eyusufatik: remove print

eyusufatik (approved):

PR #1972, opened 2025-02-24
fix authority change sov nonce increase

### LINKED ISSUE #1967: Nonce not increased when multiple sov txs are part of the same block

ercecan:
# Description

## Linked Issues
- Fixes #1967

### DIFF
--- a/crates/sequencer/src/runner.rs
+++ b/crates/sequencer/src/runner.rs
@@ -239,6 +239,7 @@ where
                     raw_message,
                     &mut working_set_to_discard,
                     soft_confirmation_info.current_spec(),
+                    false,
                 )?;
 
                 let txs = vec![signed_tx];
@@ -297,6 +298,7 @@ where
                             raw_message,
                             &mut working_set_to_discard,
                             soft_confirmation_info.current_spec(),
+                            false,
                         )?;
 
                         let txs = vec![signed_tx];
@@ -499,6 +501,7 @@ where
                 raw_message,
                 &mut working_set,
                 soft_confirmation_info.current_spec(),
+                false,
             )?;
             blobs.push(signed_tx.to_blob()?);
             txs.push(signed_tx);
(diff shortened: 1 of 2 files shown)


crates/sequencer/src/runner.rs
```diff
@@ -832,10 +835,14 @@ where
...
         spec_id: SpecId,
+        increment_nonce: bool,
     ) -> anyhow::Result<Transaction> {
         // if a batch failed need to refetch nonce
         // so sticking to fetching from state makes sense
-        let nonce = self.get_nonce(working_set, spec_id)?;
+        let mut nonce = self.get_nonce(working_set, spec_id)?;
+        if increment_nonce {
+            nonce += 1;
+        }
```
yaziciahmet: are we sure about this? get_nonce seems to return a 0 on nonexistent, and we will create the first tx with  1, but nonce should start from 0

crates/sequencer/src/runner.rs
```diff
@@ -512,7 +515,7 @@ where
             if next_fork.spec_id == SpecId::Fork2
                 && soft_confirmation_info.l2_height + 1 == next_fork.activation_height
             {
-                let (signed_blob, signed_tx) = self.update_sequencer_authority(&mut working_set, soft_confirmation_info.current_spec()).expect("Should create and sign soft confirmation rule enforcer authority change call messages");
+                let (signed_blob, signed_tx) = self.update_sequencer_authority(&mut working_set, soft_confirmation_info.current_spec(), !txs.is_empty()).expect("Should create and sign soft confirmation rule enforcer authority change call messages");
```
yaziciahmet: why only increment nonce when txs not empty

yaziciahmet (changes requested):

crates/sequencer/src/runner.rs
eyusufatik, line 1: i think the better solution and long term better api would be to have sign_tx accept nonce as param and we call get_nonce outside and pass it to the function. 

that way we have better control of the nonce

eyusufatik (changes requested):

crates/sequencer/src/runner.rs
ercecan: will do

crates/sequencer/src/runner.rs
ercecan: That'll only work for a specific case, but it's a bad api rn so I'll just pass nonce to the fn as Esad said

PR #1976, opened 2025-02-25
update risc0 to 1.3.0

eyusufatik:
# Description
Got some errors in compilation and also bn precompile does not have tag yet hence the label.

Just opening PR to see something that fails on my end in CI.


eyusufatik:
closing this as risc0 will update to v2 soon

PR #1993, opened 2025-02-26
Decrease logging

rakanalh:
# Description
DEBUG is getting verbose with 3rd party deps & some `debug!` calls on our end which add little value IMO.

### DIFF
--- a/crates/common/src/rpc/mod.rs
+++ b/crates/common/src/rpc/mod.rs
@@ -97,7 +97,7 @@ where
         let req_id = req.id();
         let req_method = req.method_name().to_string();
 
-        tracing::trace!(id = ?req_id, method = ?req_method, params = ?req.params().as_str(), "rpc_request");
+        tracing::debug!(id = ?req_id, method = ?req_method, params = ?req.params().as_str(), "rpc_request");
 
         let service = self.0.clone();
         async move {
--- a/crates/common/src/rpc/mod.rs
+++ b/crates/common/src/rpc/mod.rs
@@ -97,13 +97,13 @@ where
         let req_id = req.id();
         let req_method = req.method_name().to_string();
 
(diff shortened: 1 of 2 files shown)


crates/common/src/rpc/mod.rs
```diff
@@ -97,13 +97,13 @@ where
...
 
-        tracing::debug!(id = ?req_id, method = ?req_method, params = ?req.params().as_str(), "rpc_request");
+        tracing::trace!(id = ?req_id, method = ?req_method, params = ?req.params().as_str(), "rpc_request");
 
         let service = self.0.clone();
         async move {
             let resp = service.call(req).await;
             if resp.is_success() {
-                tracing::debug!(id = ?req_id, method = ?req_method, result = ?resp.as_result(), "rpc_success");
+                tracing::trace!(id = ?req_id, method = ?req_method, result = ?resp.as_result(), "rpc_success");
```
yaziciahmet: not sure but cloud team specifically asked for this AFAIK, if we make this trace it won't be so useful anymore. @eyusufatik can you check this?

eyusufatik: yeah that's true.

yaziciahmet: maybe what we can do here is, instead of debug logging `rpc_request` `rpc_success` twice, we can make one trace and the other debug, this way at least we see the incoming requests in debug level, but its not duplicated so less noise.

I suggest keeping the first `rpc_request` debug level so we can see the request params if we wanna further debug it with the parameters upon a problem reported by a user, and lower the `rpc_success` log to trace.

WDYT @rakanalh @eyusufatik ?

rakanalh: Updated as suggested

eyusufatik (approved):

yaziciahmet (approved):

PR #2017, opened 2025-03-03
Unify DA namespace

### LINKED ISSUE #1774: Merge sequencer commitment and ZK proof namespaces
All the DA inclusion & completeness checks will be handled by the light client proof so it makes sense to just use a single namespace.

kpp:
# Description

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/1774


crates/sovereign-sdk/adapters/mock-da/src/service.rs
```diff
@@ -450,7 +450,6 @@ impl DaService for MockDaService {
     fn extract_relevant_blobs_with_proof(
         &self,
         block: &Self::FilteredBlock,
```
eyusufatik, line 450: I thought we were going to keep the da namespace here so that we can still provide inputs for the kumquat elf for example

eyusufatik: Maybe the Namespace type can stay as is

but we introduce a 3rd variant called ALL_NAMESPACE or smth than start using that on all fork2 stuff?

no?

eyusufatik (changes requested):
when we first talked about this issue last week, you suggested doing this the way I just wrote in my comments, remember?

if we don't do something like that, our transition to fork2 will be a little hard.

crates/bitcoin-da/src/helpers/mod.rs
```diff
@@ -10,17 +8,16 @@ pub mod merkle_tree;
 pub mod parsers;
 
 /// Type represents a typed enum for LightClient kind
-#[repr(u16)]
 enum TransactionKindLightClient {
```
kpp: TODO: unify TransactionKind and parser/builder functions

crates/sovereign-sdk/adapters/mock-da/src/service.rs
eyusufatik: decided to go another way

crates/bitcoin-da/src/helpers/mod.rs
```diff
@@ -9,9 +9,9 @@ pub mod builders;
 pub mod merkle_tree;
 pub mod parsers;
 
-/// Type represents a typed enum for LightClient kind
+/// Type represents a typed enum for transaction kind
 #[repr(u16)]
```
eyusufatik, line 13: should we go for u8?

crates/bitcoin-da/tests/service.rs
```diff
@@ -65,11 +63,11 @@ impl TestCase for BitcoinServiceTest {
         // Extracts relevant batch proof blobs with proof correctly
         {
             let (mut txs, inclusion_proof, completeness_proof) =
-                service.extract_relevant_blobs_with_proof(&block, DaNamespace::ToBatchProver);
+                service.extract_relevant_blobs_with_proof(&block);
             assert_eq!(inclusion_proof.wtxids.len(), 33);
             assert_eq!(inclusion_proof.wtxids[1..], block_wtxids[1..]);
             // 3 valid commitments, and 1 invalid commitment with wrong public key
```
eyusufatik: can you fix comment?

crates/bitcoin-da/tests/service.rs
```diff
@@ -154,24 +142,11 @@ impl TestCase for BitcoinServiceTest {
         // Batch proof tx blob signed with different private key should still be
         // returned as blob with sender recovered correctly.
         {
```
eyusufatik: wrong pubkey here shouldn't have been removed, this points to something is probably wrong.

crates/bitcoin-da/tests/verifier.rs
```diff
@@ -47,9 +47,9 @@ impl TestCase for BitcoinVerifierTest {
         let (block, _, _, _) = generate_mock_txs(&service, da_node, &mut task_manager).await;
 
         let (mut b_txs, b_inclusion_proof, b_completeness_proof) =
-            service.extract_relevant_blobs_with_proof(&block, DaNamespace::ToBatchProver);
+            service.extract_relevant_blobs_with_proof(&block);
```
eyusufatik, line 50: we don't need both b_txs and l_txs.

should have single txs, inclusion_proof and completeness_proof now

crates/bitcoin-da/tests/verifier.rs
```diff
@@ -288,9 +248,8 @@ impl TestCase for BitcoinVerifierTest {
                     &block.header,
                     inclusion_proof,
                     b_completeness_proof.clone(),
-                    DaNamespace::ToBatchProver,
                 ),
-                Err(ValidationError::RelevantTxNotInProof),
+                Err(ValidationError::IncorrectInclusionProof),
```
eyusufatik: I should double check this change here

eyusufatik (changes requested):

crates/bitcoin-da/src/helpers/mod.rs
eyusufatik: @kpp said this can be done however, a lot of serialized txs should be recreted, which is a big pain in the ass so we're not doing it

crates/bitcoin-da/tests/service.rs
eyusufatik: this is not going to work. need to regenerate false sig mock block.
but below test shows it works so removing test

eyusufatik (approved):

PR #2023, opened 2025-03-03
L1 block hashes' JMT

### LINKED ISSUE #1986: Light client circuit builds JMT of L1 block hashes
Since we need to verify if a batch proof's last_l1_hash output exists in the header chain (or does not exist so we can ignore the proof), we need a merklized structure that has inclusion and exclusion proofs.

On our benchmarks we've found a JMT with keyhashes as Bitcoin blockhashes and values as empty vec's perform on par with regular merkle trees.

So we'll be creating a JMT of blockhashes starting from the first L1 block the light client proof processes.

jfldde:
# Description

- Light client circuit builds JMT of L1 block hashes

## Linked Issues
- Fixes #1986

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2053,6 +2053,7 @@ dependencies = [
  "citrea-common",
  "citrea-primitives",
  "hex",
+ "jmt",
  "jsonrpsee",
  "metrics",
  "metrics-derive",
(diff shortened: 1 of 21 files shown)


eyusufatik:
we decided to go another way with the light client circuit.

PR #2025, opened 2025-03-03
Fix cargo deny

rakanalh:
# Description
Fails on nightly

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -184,7 +184,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Run cargo-deny
-        uses: EmbarkStudios/cargo-deny-action@v1
+        uses: EmbarkStudios/cargo-deny-action@v2
         with:
           command: check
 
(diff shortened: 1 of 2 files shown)


.github/workflows/checks.yml
```diff
@@ -184,7 +184,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - name: Run cargo-deny
-        uses: EmbarkStudios/cargo-deny-action@v1
+        uses: EmbarkStudios/cargo-deny-action@v2
```
rakanalh, line 187: Seems like rust 1.81.1 is somehow installable via this new version

deny.toml
```diff
@@ -1,19 +1,17 @@
...
-
 allow = [
   "Apache-2.0",
   # "Apache-2.0 WITH LLVM-exception",
   "BSD-2-Clause",
   "BSD-3-Clause",
   "BSL-1.0",
   "CC0-1.0",
-  # "GPL-3.0",
+  "GPL-3.0",
```
eyusufatik, line 9: I am not sure if we should start allowing this

eyusufatik: both are okay

eyusufatik (approved):

PR #2029, opened 2025-03-04
Support js tracer

### LINKED ISSUE #1524: Support JS tracer
for debug_trace* calls add support for JsTracer.

You can look at code in reth.

yaziciahmet:
# Description

Brings support for js tracer in eth_debug* rpcs.

One thing I am concerned about is wrapping &mut EvmDb with a RefCell to be able to have interior mutability in EvmDbRef and implement DatabaseRef trait. We have to be actively careful about our future EvmDbRef usage and ensure  it is only used in sync code.

Other way I can do this was to add `get_ref` type of methods the whole storage chain from EvmDb to StateAccessors to WorkingSet which just takes WorkingSet as immutable ref, and WorkingSet does not commit anything to its cache when these functions are called.

TODO:
- [x] Add mock e2e test.
- [x] Check if I can somehow convert the cache traces into js traces so js tracer can benefit from cache. **ANSWER:** Not possible as far as I see. Even if possible, should be almost duplicate of revm JsInspector, so not worth it.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


eyusufatik (commented):
looks like this would work.

one question though

have you looked into wrapping EvmDb with something new that has interior mutability?

like revm's CacheDB?

yaziciahmet:
> have you looked into wrapping EvmDb with something new that has interior mutability?
> 
> like revm's CacheDB?

No, I will, but it might actually makes sense to have a wrapper, this way EvmDb wouldn't have the RefCell, but CacheDb wrapper would contain RefCell<EvmDb>, and we would be just creating the CacheDb when we need DatabaseRef trait, which is only used with JsTracer currently.

eyusufatik:
> > have you looked into wrapping EvmDb with something new that has interior mutability?
> > like revm's CacheDB?
> 
> No, I will, but it might actually makes sense to have a wrapper, this way EvmDb wouldn't have the RefCell, but CacheDb wrapper would contain RefCell, and we would be just creating the CacheDb when we need DatabaseRef trait, which is only used with JsTracer currently.

sounds like it makes sense

crates/evm/src/evm/db.rs
```diff
@@ -152,6 +152,74 @@ impl<'a, C: sov_modules_api::Context> Database for EvmDb<'a, C> {
     }
 }
 
+#[cfg(feature = "native")]
+pub mod immutable {
```
eyusufatik, line 156: that's what I'm talking about baby.

eyusufatik (approved):

PR #2039, opened 2025-03-06
Update genesis files

kpp:
# Description

To please CI check


kpp:
Fixed by https://github.com/chainwayxyz/citrea/pull/2038

PR #2058, opened 2025-03-09
Build guests concurrently

### LINKED ISSUE #2053: Compile guest elfs in parallel
Now mock/bitcoin guests are compiled sequentially. We can improve our compilation speed if we compile them in parallel.

rakanalh:
# Description

Not so much of an improvement on the CI level... but when executing `make build`, it does feel a bit faster than when compiling sequentially.

However, one caveat is `Blocking waiting for file lock on package cache` which can be observed when compiling the guest code. Meaning that even when build operations are concurrent, some sequential locking is still there based on what Cargo is downloading / building.

## Linked Issues
- Fixes #2053


jfldde:
Is it really worth it ? Quite complex for dubious gains

rakanalh:
Discussed with @kpp and we don't think this is effective given the Cargo locks which prevent concurrent guest builds.

PR #2063, opened 2025-03-10
Remove old `setBlockInfo` from `BitcoinLightClient`

okkothejawa:
Removes the `setBlockInfo` function that was kept for backwards compatability and runs `make genesis` afterwards.


resources/genesis/mock/evm.json
```diff
@@ -141,6 +141,9 @@
   ],
   "chain_id": 5655,
   "limit_contract_code_size": null,
+  "spec": {
```
eyusufatik: spec here is now removed pls update your genesis script and run make genesis again. 

don't forget to pull first because I merged my branch into yours a minute ago

okkothejawa: done

PR #2068, opened 2025-03-11
Zk_storage: Remove redundant code

rakanalh:
# Description

Follow up to: #2052 

Both branches have the exact same code. So it seems like in zk_storage, the parameter is not needed compared to prover storage.

### DIFF
--- a/crates/sovereign-sdk/module-system/sov-state/src/zk_storage.rs
+++ b/crates/sovereign-sdk/module-system/sov-state/src/zk_storage.rs
@@ -54,7 +54,7 @@ impl Storage for ZkStorage {
         &self,
         state_log: &ReadWriteLog,
         witness: &mut Witness,
-        accumulate_diff: bool,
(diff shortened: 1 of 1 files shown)


crates/sovereign-sdk/module-system/sov-state/src/zk_storage.rs
```diff
@@ -70,32 +70,19 @@ impl Storage for ZkStorage {
...
-        let batch = if accumulate_diff {
-            state_log
-                .iter_ordered_writes()
-                .map(|(key, value)| {
-                    let key_hash = KeyHash::with::<DefaultHasher>(key.key.as_ref());
-
-                    let key_bytes = key.key.clone();
-                    let value_bytes = value.as_ref().map(|v| v.value.clone());
-
-                    diff.push((key_bytes, value_bytes.clone()));
```
eyusufatik, line 82: one builds the diff, the other does not.

PR #2070, opened 2025-03-11
Remove brotli unwraps

### LINKED ISSUE #2059: Remove .unwrap() on brotli decompress
If a relevant complete proof cannot be decompressed with brotli, BitcoinVerifier panics right now.

If an error occurs during decompression, we can just skip that blob.

jfldde:
# Description

- Propagate brotli errors
- Skip blob in DaVerifier when decompression fails
- Make extract_* methods infallible

## Linked Issues
- Fixes #2059

### DIFF
--- a/crates/bitcoin-da/src/service.rs
+++ b/crates/bitcoin-da/src/service.rs
@@ -760,7 +760,7 @@ impl DaService for BitcoinService {
         &self,
         block: &Self::FilteredBlock,
         prover_da_pub_key: &[u8],
-    ) -> Result<Vec<Proof>> {
+    ) -> Vec<Proof> {
         let mut completes = Vec::new();
         let mut aggregate_idxs = Vec::new();
 
@@ -791,7 +791,8 @@ impl DaService for BitcoinService {
                             };
 
                             let DataOnDa::Complete(zk_proof) = data else {
-                                bail!("{}: Complete: unexpected kind", tx_id);
+                                warn!("{}: Complete: unexpected kind", tx_id);
+                                continue;
                             };
                             completes.push((i, zk_proof));
                         }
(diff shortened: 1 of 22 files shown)


crates/bitcoin-da/src/service.rs
```diff
@@ -994,7 +994,7 @@ impl DaService for BitcoinService {
                 match tx {
                     ParsedTransaction::Complete(complete) => {
                         if let Some(hash) = complete.get_sig_verified_hash() {
-                            let blob = decompress_blob(&complete.body);
+                            let blob = decompress_blob(&complete.body)?;
```
eyusufatik: we cant return err on extract_* function 

we have to ignore that certain tx and go on.

eyusufatik (changes requested):

crates/bitcoin-da/src/service.rs
jfldde: It's already returning err here though. You mean we should discard decompression error altogether in extract_* ?

crates/bitcoin-da/src/service.rs
```diff
@@ -936,11 +946,11 @@ impl DaService for BitcoinService {
     fn extract_relevant_blobs_with_proof(
         &self,
         block: &Self::FilteredBlock,
-    ) -> (
+    ) -> anyhow::Result<(
```
eyusufatik: can you revert the return type change?

crates/primitives/Cargo.toml
```diff
@@ -12,6 +12,7 @@ sov-rollup-interface = { path = "../sovereign-sdk/rollup-interface" }
 
 # 3rd-party deps
 alloy-eips = { workspace = true }
+anyhow = { workspace = true }
```
eyusufatik: pls don't use anyhow in primitives crate. we'd like to deactivate sys-getenv feature of risc0 at one point.

eyusufatik (changes requested):

eyusufatik (approved):

crates/bitcoin-da/src/service.rs
```diff
@@ -782,9 +782,14 @@ impl DaService for BitcoinService {
...
                             // push only when signature is correct
-                            let body = decompress_blob(&complete.body);
-                            let data = DataOnDa::borsh_parse_complete(&body)
-                                .map_err(|e| anyhow!("{}: Failed to parse complete: {e}", tx_id))?;
+                            let Ok(body) = decompress_blob(&complete.body) else {
+                                continue;
+                            };
+
+                            let Ok(data) = DataOnDa::borsh_parse_complete(&body) else {
+                                continue;
```
eyusufatik, line 792: can you change other bail!'s in this function to continue?

btw bail on chunk parse should continue on 'aggregate

eyusufatik (approved):

PR #2071, opened 2025-03-11
Tone down sequencer RPC Logger logging level

jfldde:
# Description

- `eth_sendRawTransaction` RPC response has lots of valid, non-warn level reason to fail, such as Tx `already known` or `insufficient funds for gas`, etc.
This reduces sequencer logs warn level clutter


eyusufatik (approved):
good idea!

Issue #1905, opened 2025-02-17
Rewrite full node L2 block status tracking


eyusufatik:
When a full node finds a sequencer commitment on DA, it verifies the commitment and marks it as finalized (should be committed maybe?).

However, if it sees a sequencer commitment it can't verify, it should be aware that there might be another fork of the chain which is not aware of.

Another possibility is, it finds two or more sequencer commitment for the same range and only one of these it will be able to verify.

They both mean, from the point of view of the full node, there is another fork the full node is not aware of which will only be resolved once a batch proof hits the DA layer which verifies one of these commitments. (For the case of two different proof attesting to two different sequencer commitments, the first proof to land on DA will mark the canonical chain because that's the proof the light client proof accepts as Citrea's next state update).

~~So the full node should mark these L2 ranges (and the following L2 heights) as conflicting and should wait on a batch proof to resolve the dispute.~~

I propose the following change to the state status from the full node perspective as such:

- Hold two heights for "committed" and "proven".
- The committed height is ~~decreased or~~ increased according to sequencer commitments.
- The proven height is  increased according to batch proofs ~~also resolving disputes among sequencer commitments~~.

It is up to the actor that uses the Citrea full node to ask regularly these heights (or querying by height) to act on the Citrea state changes even if the full node keeps syncing.

eyusufatik:
The full nodes should apply the same logic for state status update as the light client prover.

ercecan:
after #2014 

- [ ]	Create two new ledger database tables to track the full node’s committed and proven L2 heights. Each table should use a key-value structure where the key is an enum (Proven or Committed), and the value is a tuple (u64, u64), representing the last L2 end height and the index of the last sequencer commitment.
- [ ]	When detecting a sequencer commitment on DA:
  - If the node already knows a sequencer commitment with the same index, discard the new sequencer commitment.
  - Retrieve the committed height from the 
  - Compare the index and L2 height with the stored values:
  - ~~If the new index is greater than the one in the database, and the L2 height is also greater, compute the Merkle root from db_height + 1 to l2_end_height.~~
  -  If the new index is **one more**  than the last index in the db and the l2 height is also greater (`new index == old_index + 1 && new_end_l2 > old_end_l2`), compute the Merkle root from db_height + 1 to l2_end_height (`MerkleTree([old_end_l2 + 1..new_end_l2]).root()`).
[...]

PR #2073, opened 2025-03-12
Batch proof input output changes

### LINKED ISSUE #2019: Batch proof outputs sequencer commitment index range and hashes
For the range of sequencer commitments the batch proof "verifies" it will make two new outputs.

1. seq_comm_range: (u32, u32) -> The index-range of the sequencer commitments that was fed into the batch proof.
2. seq_comm_hashes: Vec<[u8; 32]> -> Batch proof will hash serialized sequencer commitments, accumulate them in a vector and output that.

### LINKED ISSUE #2069: Put last sequencer commitment to the input and output of the batch proof
Talked with @eyusufatik and decided this is safer than #2067 
This is needed so that sequencer and batch prover cannot get together and fool light client prover. With this the batch prover will have to start from the last sequencer commitment's l2 end height and will not be able to deceive light client prover.

We also discussed on keeping the start l2 height for commitments however we realized that sequencer and batch prover can still deceive lcp by giving false ranges with correct index

All in all we need to put the previous sequencer commitment to the next batch proof in put and put that sequencer commitment's index and hash to the output

ercecan:
# Description
Adds
-  Previous sequencer commitment to the input of the batch proof (the sequencer commitment with the previous index of the current batch proofs first commitment)
- Sequencer Commitment index range (range of indexes of sequencer commitments proven in the batch proof)
- Index of the previous sequencer commitment that was inputted
- Hash of the previous sequencer commitment that was inputted
Updates
- Instead of outputting sequencer commitment merkle roots we now output sequencer commitment hashes
Removes
- L2 start height from sequencer commitment

With these changes, sequencer and batch prover cannot trick light client prover to prover false state transitions

## Linked Issues
- Fixes #2069 #2019

### DIFF
--- a/bin/citrea/tests/bitcoin/batch_prover_test.rs
+++ b/bin/citrea/tests/bitcoin/batch_prover_test.rs
@@ -315,7 +315,6 @@ impl TestCase for SkipPreprovenCommitmentsTest {
             .map(|response| SequencerCommitment {
                 merkle_root: response.merkle_root,
                 index: response.index.to(),
-                l2_start_block_number: response.l2_start_block_number.to(),
                 l2_end_block_number: response.l2_end_block_number.to(),
             })
             .collect();
--- a/bin/citrea/tests/bitcoin/batch_prover_test.rs
+++ b/bin/citrea/tests/bitcoin/batch_prover_test.rs
@@ -177,219 +177,220 @@ async fn basic_prover_test() -> Result<()> {
         .await
 }
 
-#[derive(Default)]
-struct SkipPreprovenCommitmentsTest {
-    task_manager: TaskManager<()>,
-}
+// TODO: Should I remove this?
+// #[derive(Default)]
+// struct SkipPreprovenCommitmentsTest {
+//     task_manager: TaskManager<()>,
+// }
 
-#[async_trait]
-impl TestCase for SkipPreprovenCommitmentsTest {
-    fn test_config() -> TestCaseConfig {
-        TestCaseConfig {
-            with_batch_prover: true,
-            with_full_node: true,
-            ..Default::default()
-        }
-    }
+// #[async_trait]
+// impl TestCase for SkipPreprovenCommitmentsTest {
+//     fn test_config() -> TestCaseConfig {
+//         TestCaseConfig {
+//             with_batch_prover: true,
+//             with_full_node: true,
+//             ..Default::default()
+//         }
+//     }
 
-    fn sequencer_config() -> SequencerConfig {
-        SequencerConfig {
-            min_soft_confirmations_per_commitment: 1,
-            ..Default::default()
-        }
-    }
+//     fn sequencer_config() -> SequencerConfig {
+//         SequencerConfig {
+//             min_soft_confirmations_per_commitment: 1,
+//             ..Default::default()
+//         }
+//     }
 
-    fn scan_l1_start_height() -> Option<u64> {
-        Some(170)
-    }
+//     fn scan_l1_start_height() -> Option<u64> {
+//         Some(170)
+//     }
 
-    async fn run_test(&mut self, f: &mut TestFramework) -> Result<()> {
-        let da = f.bitcoin_nodes.get(0).unwrap();
-        let sequencer = f.sequencer.as_ref().unwrap();
-        let batch_prover = f.batch_prover.as_ref().unwrap();
-        let full_node = f.full_node.as_ref().unwrap();
+//     async fn run_test(&mut self, f: &mut TestFramework) -> Result<()> {
+//         let da = f.bitcoin_nodes.get(0).unwrap();
+//         let sequencer = f.sequencer.as_ref().unwrap();
+//         let batch_prover = f.batch_prover.as_ref().unwrap();
+//         let full_node = f.full_node.as_ref().unwrap();
 
-        let da_config = &f.bitcoin_nodes.get(0).unwrap().config;
-        let bitcoin_da_service_config = BitcoinServiceConfig {
-            node_url: format!(
-                "http://127.0.0.1:{}/wallet/{}",
-                da_config.rpc_port,
-                NodeKind::Bitcoin
-            ),
-            node_username: da_config.rpc_user.clone(),
-            node_password: da_config.rpc_password.clone(),
-            network: bitcoin::Network::Regtest,
-            da_private_key: Some(
-                // This is because the prover has a check to make sure that the commitment was
-                // submitted by the sequencer and NOT any other key. Which means that arbitrary keys
-                // CANNOT submit preproven commitments.
-                // Using the sequencer DA private key means that we simulate the fact that the sequencer
-                // somehow resubmitted the same commitment.
-                sequencer
-                    .config()
-                    .rollup
-                    .da
-                    .da_private_key
-                    .as_ref()
-                    .unwrap()
-                    .clone(),
-            ),
-            tx_backup_dir: Self::test_config()
-                .dir
-                .join("tx_backup_dir")
-                .display()
-                .to_string(),
-            monitoring: Default::default(),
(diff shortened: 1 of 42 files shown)


crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
```diff
@@ -519,29 +525,74 @@ where
...
+                // should be the one after the last commitment
+                previous_batch_proof_l2_end_height =
+                    previous_sequencer_commitment.l2_end_block_number;
+                (
+                    Some(previous_sequencer_commitment.index),
+                    Some(previous_sequencer_commitment.serialize_and_calculate_sha_256()),
+                )
+            } else {
+                // If this is the first batch proof, then the first commitment idx should be 0
+                // TODO: First commitment index is not in 0 parallel_proving_test why?
```
eyusufatik: can you eloborate?

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
eyusufatik, line 1: can you modify these two tables:
- PendingSequencerCommitment
- LastSequencerCommitmentSent

so that they map to a u64, which we can use as key for your table SequencerCommitmentByIndex

to get the sequencer commitment.

you can leave the existing ledger db functions as is so that I can jsut call get_last_sequencer_commitment() without an argument and get the last sequencer commitment

crates/sequencer/src/commitment/mod.rs
```diff
@@ -272,7 +273,20 @@ where
                     .delete_pending_commitment(pending_db_comm.index)?;
             } else {
                 // Submit commitment
-                let range = SoftConfirmationNumber(pending_db_comm.l2_start_block_number)
+                let l2_start_block_number = if pending_db_comm.index == 0 {
+                    if get_fork2_activation_height() == 0 {
+                        1
+                    } else {
+                        get_fork2_activation_height()
+                    }
```
eyusufatik: you don't need to call get_fork2_activation_height twice here

crates/fullnode/src/da_block_handler.rs
```diff
@@ -135,6 +136,13 @@ where
             l1_block,
             &self.sequencer_da_pub_key,
         );
+
+        for commitment in sequencer_commitments.iter() {
+            self.ledger_db
+                .put_commitment_by_index(commitment)
+                .expect("Should save commitment to ledger db");
+        }
```
eyusufatik: shouldn't we do this after we verified the sequencer commitments?

crates/batch-prover/src/da_block_handler.rs
```diff
@@ -316,16 +316,32 @@ pub(crate) fn break_sequencer_commitments_into_groups<DB: BatchProverLedgerOps>(
 
     let mut range = 0usize..=0usize;
     let mut cumulative_state_diff = StateDiff::new();
+    let first_l2_block_number = if sequencer_commitments[0].index == 0 {
+        // TODO: Handle this better
+        if get_fork2_activation_height() == 0 {
```
eyusufatik: same

crates/batch-prover/src/proving.rs
```diff
@@ -715,12 +767,26 @@ pub(crate) fn save_commitments<DB>(
     DB: BatchProverLedgerOps,
 {
     for sequencer_commitment in sequencer_commitments.iter() {
+        let l2_start_block_number = if sequencer_commitment.index == 0 {
+            if get_fork2_activation_height() == 0 {
+                1
```
eyusufatik: same

bin/citrea/tests/bitcoin/tx_chain.rs
```diff
@@ -595,6 +595,7 @@ impl TestCase for TestProverTransactionChaining {
             .await?;
         let txs = &block.txdata;
 
+        // TODO: Why is this very flaky now?
```
eyusufatik, line 598: should we adress this in this pr?

bin/citrea/tests/bitcoin/batch_prover_test.rs
```diff
@@ -177,220 +172,220 @@ async fn basic_prover_test() -> Result<()> {
...
-}
-
-#[tokio::test]
-async fn prover_skips_preproven_commitments_test() -> Result<()> {
-    TestCaseRunner::new(SkipPreprovenCommitmentsTest::default())
-        .set_citrea_path(get_citrea_path())
-        .run()
-        .await
-}
+// TODO: Should I remove this?
```
eyusufatik, line 175: well we can easily fix batch prover behaviour by making it not prove any sequencer commitment it had proven before

and if we remove the assert_eq!s on batch proof outputs in this test, it would pass.

eyusufatik (changes requested):

crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
ercecan: It was a bug and I fixed it but forgot to remove comment

bin/citrea/tests/bitcoin/tx_chain.rs
ercecan: Maybe I can open an issue?

bin/citrea/tests/bitcoin/batch_prover_test.rs
eyusufatik: can open issue about this btw or we can handle this in #1905

eyusufatik: so maybe we can extend 1905 to include this also

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
ercecan: With this cahnge we will need migrations for PendingSequencerCommitment and LastSequencerCommitmentSent
LastSequencerCommitmentSent will be just () => () initially because prefork2 index does not exist, will open issue

ercecan: #2077

bin/citrea/tests/bitcoin/batch_prover_test.rs
ercecan: #2078

bin/cli/src/main.rs
```diff
@@ -65,9 +65,6 @@ enum Commands {
         /// The target sequencer commitment index to rollback to
         #[arg(long)]
         sequencer_commitment_index: u32,
-        /// The L2 block number at which there was a sequencer commitment that was sent.
```
eyusufatik, line 68: why did you remove this?

we still need to set something like this somehow.

crates/fullnode/src/da_block_handler.rs
```diff
@@ -324,110 +311,77 @@ where
...
-        }
-        for root in merkle_roots {
-            // make sure sequencer commitment soft confirmation merkle root match
-            // since we wouldn't have the sequencer commitment in the ledger db
-            // this makes sure the sequencer commitment exists
-            let seq_comm_range = self
-                .ledger_db
-                .get_l2_range_by_commitment_merkle_root(root)?
-                .ok_or(SyncError::SequencerCommitmentNotFound(root))?;
+            assert_eq!(
```
eyusufatik: instead of assert we should compare and return err - or whatever is needed to  ignore the proof

eyusufatik (changes requested):

bin/cli/src/main.rs
ercecan: because it was not used

eyusufatik (approved):

PR #2074, opened 2025-03-12
Do not check `humantime` crate in cargo deny

ercecan:
# Description
`humantime` library is not maintained for over 4 years now so cargo deny is rightfully erroring out, however since it comes from some reth crate we cannot get rid of it just yet so we can exclude it for now


eyusufatik (changes requested):
you can fix this by adding "RUSTSEC-2025-0014" to deny.toml ignore list. that way on our machines deny works the same as ci 👍

eyusufatik (approved):

PR #2081, opened 2025-03-13
Add WCBTC as a predeploy

okkothejawa:
Adds WCBTC to `0x3100000000000000000000000000000000000006`. Updates the genesis generation script, and runs `make genesis`. Adds a system contract test for deposit / withdraw flow.


eyusufatik (approved):

kpp (approved):

eyusufatik (commented):
do you need to add somethign to the the prod genesis script?

If so please fix geensis prod script to use numbers (not strings) for block gas limit and remove spec field from the genesis generated.

PR #2082, opened 2025-03-13
Remove generic Tx from L2Block

### LINKED ISSUE #2075: Remove L2Block generic transaction
- Let's keep Transaction type concrete. What we used it for can be better serve with a versionned Transaction enum

jfldde:
# Description

- Remove generic Tx from `L2Block`. We won't be supporting any other Tx type than sov-tx. Versionning is now done at the Transaction level since #1950 
- Extract keys related code to dedicated `sov-keys` crate. This was required to fix deep cyclical dependency between `sov-modules-api` and `rollup-interface`.
- Remove complex and redundant `test_rpc.rs` file which is better and already tested via bitcoin_e2e tests

## Linked Issues
- Fixes #2075


rakanalh (approved):
LGTM, though following the sov naming convention, the new crate should be called `sov-modules-keys` but it doesn't matter that much.

crates/sovereign-sdk/module-system/sov-keys/src/serde_pub_key.rs
eyusufatik, line 1: I think you can now remove DefaultPublicKey and dalek dependencies

eyusufatik (changes requested):
pls base this on `fork2-no-backwards-compat`

eyusufatik (approved):

PR #2084, opened 2025-03-13
Fix eth_getproof

kpp:
# Description

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/2080


crates/ethereum-rpc/src/lib.rs
```diff
@@ -296,7 +291,7 @@ where
         let nonce = account.nonce;
         let code_hash = account.code_hash.unwrap_or(KECCAK_EMPTY);
 
-        fn generate_account_proof_postfork2<C>(
+        fn generate_account_proof<C>(
```
eyusufatik, line 294: we should somehow return account proof as concatanation of two proofs:
1 - address => index proof
2 - index => account info proof

eyusufatik (changes requested):

crates/ethereum-rpc/src/lib.rs
kpp: How?

PR #2102, opened 2025-03-16
Correct verbosity handling

rakanalh:
# Description

The default verbosity for our nodes is `INFO` which has a value of 2. However, the verbosity flag counts the `v`s and we decide the verbosity based on the number of `v`s which starts from 0 for ERROR all the way up to TRACE.
However, our initial verbosity is 2, which means that when we pass `-v` or `--verbose`, that should indicate that we'd like to increase verbosity STARTING from `INFO` which takes us to `DEBUG`.

This PR means that:
- Passing `-v` enables debug logging
- Passing `-vv` enables trace logging

### DIFF
--- a/bin/citrea/src/cli.rs
+++ b/bin/citrea/src/cli.rs
@@ -54,7 +54,7 @@ pub(crate) struct Args {
     pub(crate) light_client_prover: Option<Option<String>>,
 
     /// Logging verbosity
-    #[arg(long, short = 'v', action = clap::ArgAction::Count, default_value = "0")]
+    #[arg(long, short = 'v', action = clap::ArgAction::Count, default_value = "0", env = "CITREA_VERBOSITY")]
     pub(crate) verbose: u8,
     /// Logging verbosity
     #[arg(long, short = 'q', action)]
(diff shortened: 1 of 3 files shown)


jfldde (approved):

bin/citrea/src/cli.rs
```diff
@@ -54,7 +54,7 @@ pub(crate) struct Args {
     pub(crate) light_client_prover: Option<Option<String>>,
 
     /// Logging verbosity
-    #[arg(long, short = 'v', action = clap::ArgAction::Count, default_value = "2")]
+    #[arg(long, short = 'v', action = clap::ArgAction::Count, default_value = "0", env = "CITREA_VERBOSITY")]
```
eyusufatik, line 57: we also have RUST_LOG for log level env var. are we sure about doing this?

rakanalh: RUST_LOG overrides our configuration which silence 3rd party logs. This is why we added the verbosity which basically lets us control how verbose citrea is while not increasing the verbosity of 3rd party libs that add no value to the logs

eyusufatik (approved):

PR #2106, opened 2025-03-17
DA RPC enhancements

jfldde:
# Description

- Add full transaction to MonitoredTxResponse
- Revert U64 serialization changes
- Remove alloy from bitcoin-da crate altogether

### DIFF
--- a/crates/bitcoin-da/src/monitoring.rs
+++ b/crates/bitcoin-da/src/monitoring.rs
@@ -3,7 +3,6 @@ use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
 use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
-use alloy_primitives::U64;
 use anyhow::anyhow;
 use bitcoin::address::NetworkUnchecked;
 use bitcoin::hashes::Hash;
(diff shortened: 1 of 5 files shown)


crates/bitcoin-da/src/rpc.rs
```diff
@@ -35,13 +34,14 @@ impl From<(Txid, MonitoredTx)> for MonitoredTxResponse {
...
-            initial_broadcast: U64::from(tx.initial_broadcast),
-            initial_height: U64::from(tx.initial_height),
+            base_fee,
+            vsize: tx.tx.vsize(),
+            initial_broadcast: tx.initial_broadcast,
+            initial_height: tx.initial_height,
             prev_txid: tx.prev_txid,
             next_txid: tx.next_txid,
             status: tx.status,
+            transaction: tx.tx,
```
eyusufatik: make sure this has hex serialized tx inside.

crates/bitcoin-da/src/monitoring.rs
```diff
@@ -33,20 +32,20 @@ pub enum TxStatus {
...
         in_mempool: bool,
-        base_fee: U64,
-        timestamp: U64,
+        base_fee: u64,
+        timestamp: u64,
     },
     #[serde(rename_all = "camelCase")]
     Confirmed {
         block_hash: BlockHash,
-        block_height: U64,
```
eyusufatik, line 42: we wanted to unify all RPCs to have hex numbers. that's why we converted these to U64 before, why change it no?

jfldde: Well it doesn't make sense in bitcoin context since it doesn't hex serialize u64 nor expect hex serialized u64 as input.
It's also remove alloy dependency from bitcoin-da which feels more consistent

eyusufatik (approved):

PR #2111, opened 2025-03-18
For non segwit blocks the witness root is txs merkle root

### LINKED ISSUE #2110: Non segwit block witness root should be txs merkle root for bitcoin blocks

ercecan:
# Description
When implementing reorg syncer @eyusufatik and I realized there is a bug in verifying short header proofs, turns out we made a mistake at non segwit block witness roots, it should be txs merkle root.

## Linked Issues
- Fixes #2110


eyusufatik:
discussed on slack to go a different way on this and base pr on nightly

PR #2114, opened 2025-03-18
Fix try_from slice for K256PublicKey

yaziciahmet:
# Description

- fixes try_from slice for k256 pubkey. even though provided slice was invalid, filling it with 0 during copy process might make it look like it is valid and might cause some unwanted trouble

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


kpp (approved):
Very good

PR #2115, opened 2025-03-18
Full node L2 block status tracking

### LINKED ISSUE #1905: Rewrite full node L2 block status tracking

jfldde:
# Description

- Tracks L2 block status

## Linked Issues
- Fixes #1905

# Missing :

- [x] Handle conflicting commitments ranges
- [x] Handle out of order commitments
- [x] Test for conflicting commitment ranges
- [x] Test for out of order commitments
- [x] Rollback handling
- [x] new RPC for Clementine https://github.com/chainwayxyz/citrea/pull/2115#issuecomment-2736556635


eyusufatik:
clementine team will also need an rpc that gives you the highest L2 height that would have proven status once an L1 block is processed.

so we need an RPC that takes L1 height and maps it into highest committed and highest proven up to that point

eyusufatik (changes requested):
one issue I can clearly see is this:

imagine this scenario

### L1 block 1
seq comm index 1 end 10

### L1 block 2 
seq comm index 3 end 30


now we return highest committed 30 even though we never saw seq comm index 2

eyusufatik:
one problem I'd like to address in this PR is also that the light client prover can keep a batch proof "pending" meaning that if it does not show that a sequencer commitment mismatch hasn't happened etc. it'll keep it for future processing so that once it is sure it's verified or not, it can remove it from the "pending" (or so called "unstitched" in craft) output.

We should also do a similar thing for the full node.

Think about the full node L2 status tracking like it's mimicking light client prover.

crates/fullnode/src/da_block_handler.rs
```diff
@@ -125,6 +125,14 @@ where
...
             .unwrap();
 
+        if let Err(e) = self.process_pending_proofs(l1_block).await {
+            error!("Error processing pending proofs: {e:?}");
+        }
+
+        if let Err(e) = self.process_pending_commitments(l1_block).await {
+            error!("Error processing pending commitments: {e:?}");
+        }
+
```
ercecan: I believe that processing pending commitments and proofs at the end of the l1 block, after you processed all of the commitments and proofs, makes more sense because of scenarios like this:

block num: 1, commitments in block: [1,2,3]
block num: 2, commitments in block: [5,6]
in block#2 the commitments 5 and 6 are in pending
block num: 3 commitments in block: [4]
pendin commitments are processed but 4 is still not processes so, 5 and 6 are still pending
4 is now the committed index but 6 should have been the committed index
now we have to wait for another l1 block to process them, instead if we process pendings at the end of block#3 we will not need to wait for the next l1

crates/fullnode/src/da_block_handler.rs
```diff
@@ -350,6 +462,24 @@ where
             None => get_fork2_activation_height_non_zero(),
         };
 
+        for index in sequencer_commitment_index_range.0..=sequencer_commitment_index_range.1 {
+            if self.ledger_db.get_commitment_by_index(index)?.is_none() {
```
ercecan: maybe a get_indexes_by_range query could be useful

ercecan (changes requested):

crates/fullnode/src/da_block_handler.rs
jfldde: Makes total sense, I'll switch it to the end of `process_l1_block`

crates/fullnode/src/da_block_handler.rs
eyusufatik, line 1: in process_l1_block right now we are processing proofs first and sequencer commitments second.

ideally, it's better to process these in the order they are on da.

can you do that please?

crates/fullnode/src/da_block_handler.rs
```diff
@@ -207,24 +226,76 @@ where
...
             l1_block.header().height(),
         );
 
+        let committed_height = self
+            .ledger_db
+            .get_highest_l2_height_for_status(L2HeightStatus::Committed, None)?
+            .unwrap_or_default();
+
+        // Only proceed if the commitment height and index are higher than the stored one
+        if end_l2_height <= committed_height.height {
```
eyusufatik: this is a good check actually. are we doing this in LCP @ercecan?

@yaziciahmet we should also check for this at the batch prover. just discard if a commitment with index n does not have higher end height than commitment index n - 1

crates/fullnode/src/da_block_handler.rs
```diff
@@ -207,24 +226,76 @@ where
...
+                );
+            return Ok(());
+        }
+
+        if sequencer_commitment.index != committed_height.commitment_index + 1 {
+            info!(
+                "Skipping sequencer commitment with index {} as it is not increasing by one",
+                sequencer_commitment.index,
+            );
+            return Ok(());
```
eyusufatik: isn't this unreachable?

crates/fullnode/src/da_block_handler.rs
```diff
@@ -322,6 +401,39 @@ where
...
+            .ledger_db
+            .get_highest_l2_height_for_status(L2HeightStatus::Committed, None)?
+            .unwrap_or_default();
+
+        if proven_height > committed_height {
+            return Err(SyncError::ProvenHeightExceedsCommittedHeight(
+                proven_height,
+                committed_height,
+            ));
+        }
```
eyusufatik: these checks might be irrelevant.

for a proof, if all the sequencer commitments in the output exists and the hashes match, we can just push it to something like "pending proofs" even if proof that should precede this one haven't been seen yet.

crates/fullnode/src/da_block_handler.rs
```diff
@@ -350,6 +462,24 @@ where
...
+                    "Commitment with index {} is missing for proof. Storing proof as pending.",
+                    index
+                );
+
+                // Store as pending for later processing
+                self.ledger_db.store_pending_proof(
+                    sequencer_commitment_index_range.0,
+                    sequencer_commitment_index_range.1,
+                    raw_proof,
+                )?;
```
eyusufatik: don't push as pending before making sure all known sequencer commitment indexes hashes match.

so verify as many indices as you can, if there are missing, push to pending. if any mismatch just discard proof

eyusufatik (changes requested):
didn't review tests for now, but the implementation is pretty solid.

eyusufatik:
there's rumors that @ercecan is about to open the new LCP PR with the final logic. you can cheat from there :D

crates/fullnode/src/da_block_handler.rs
ercecan: This is also not the case anymore since we discard proofs with missing commitments
Resolving

crates/fullnode/src/da_block_handler.rs
```diff
@@ -207,24 +241,76 @@ where
...
+                warn!(
+                    "Conflicting sequnecer commitments with different merkle roots at index: {}.
+                    Already processed merkle root: 0x{}, conflicting merkle root: 0x{}",
+                    sequencer_commitment.index,
+                    hex::encode(existing_commitment.merkle_root),
+                    hex::encode(sequencer_commitment.merkle_root)
+                );
+            }
+            return Ok(());
+        }
```
ercecan: please also add a warning even if the merkle roots are matching, because it is still undesired behaviour

crates/fullnode/src/da_block_handler.rs
```diff
@@ -207,24 +241,76 @@ where
...
+                None => {
+                    // Store the out of order commitment as pending
+                    info!(
+                            "Commitment with index {} is missing its predecessor (index {}). Storing as pending.",
+                            sequencer_commitment.index,
+                            sequencer_commitment.index - 1
+                        );
+                    self.ledger_db
+                        .store_pending_commitment(sequencer_commitment.clone())?;
+                    return Ok(());
```
ercecan: please check the highest committed l2 height before storing a commitment as pending, if highest committed height >= sequencer_commitment.l2_end_block_number then discard and log with a warning

crates/fullnode/src/da_block_handler.rs
```diff
@@ -207,24 +241,76 @@ where
...
+            }
+
+            if sequencer_commitment.index != committed_height.commitment_index + 1 {
+                info!(
+                    "Skipping sequencer commitment with index {} as it is not increasing by one",
+                    sequencer_commitment.index,
+                );
+                return Ok(());
+            }
+        }
```
ercecan: Correct me if I'm wrong but these two ifs should be unreachable so maybe panic makes more sense here because if we are here we made a big mistake

crates/fullnode/src/da_block_handler.rs
```diff
@@ -320,8 +414,51 @@ where
...
+
+        let proven_height = self
+            .ledger_db
+            .get_highest_l2_height_for_status(L2HeightStatus::Proven, None)?
+            .unwrap_or_default();
+
+        let end_l2_height = batch_proof_output.last_l2_height();
+
+        if end_l2_height <= proven_height.height
+            && sequencer_commitment_index_range.1 <= proven_height.commitment_index
```
ercecan:
```suggestion
        if end_l2_height <= proven_height.height
            || sequencer_commitment_index_range.1 <= proven_height.commitment_index
```

ercecan: We can skip it either way because if end_l2_height is smaller or equal then the index has to be smaller or equal as well, as they are strictly correlated

crates/fullnode/src/da_block_handler.rs
ercecan: I agree we can just add them to pending proofs and process them at the end of the block

crates/fullnode/src/da_block_handler.rs
```diff
@@ -369,25 +506,53 @@ where
...
+            info!(
+                "Commitments {missing_commitments:?} were missing for proof. Storing as pending."
+            );
+            self.ledger_db.store_pending_proof(
+                sequencer_commitment_index_range.0,
+                sequencer_commitment_index_range.1,
+                raw_proof,
+            )?;
+            return Ok(());
+        }
```
ercecan: If any of the commitments are missing we need to discard that proof, if we have all the commitments in the db but let's say the first commitment of the batch proof is not the next commitment index we expected we store it in pending proofs

example scenario:
l1 block #1 data: 
sequencer commitments with indices : [1,2,3,4,5,6,7]
last committed index = 5 
l1 block #2 data: batch proof with commitment index range: [1,2,3]
last proven index = 3
l1 block #3 data: batch proof with commitment index range: [6,7]
we would expect the bathc proof indexes to start with 4 but it starts with 6
however we did store commitments 6 and 7, what we do is we check the hashes of commitments with index:
5 because it is the previous commitment and we have a field called previous commitment index and hash in the bp output
6,7 because they are in the index rnage of bp output, if hashes match, we add this proof to the pending proofs
l1 block #4 data: batch proof with commitment index range: [4,5]
last proven index = 5
also calling process_pending_proofs() at the end of every l1 block
process_pending_proofs() will see that we have the proof with indices [6,7]
sets last proven index = 7
l1 block #5 data ....... goes on like this

ercecan (changes requested):

crates/fullnode/src/da_block_handler.rs
```diff
@@ -125,66 +131,100 @@ where
...
-                        unreachable!("Error irrelevant!")
-                    }
-                }
-            }
+        if let Err(e) = self.process_pending_proofs(l1_block).await {
+            error!("Error processing pending proofs: {e:?}");
+        }
+
+        if let Err(e) = self.process_pending_commitments(l1_block).await {
+            error!("Error processing pending commitments: {e:?}");
```
ercecan, line 219: I  feel that processing pending commitments before processing pending proofs makes more sense, because we might update the committed height and since there is a proven height < committed height check this should keep things in line I believe

crates/fullnode/src/da_block_handler.rs
ercecan: This is fixed now after @jfldde push to pending if proof index is not sequential but has all commitments with correct hashes

ercecan (changes requested):
one small ask

ercecan (approved):

crates/fullnode/src/rpc.rs
```diff
@@ -0,0 +1,134 @@
...
+pub fn register_rpc_methods<DB: NodeLedgerOps + Clone + 'static>(
+    mut rpc_methods: jsonrpsee::RpcModule<()>,
+    rpc_context: RpcContext<DB>,
+) -> Result<jsonrpsee::RpcModule<()>, jsonrpsee::core::RegisterMethodError> {
+    let rpc = create_rpc_module(rpc_context);
+    rpc_methods.merge(rpc)?;
+    Ok(rpc_methods)
+}
+
+#[rpc(client, server, namespace = "fullNode")]
```
eyusufatik: let's just call this namespace citrea if that's not a huge change

crates/fullnode/src/da_block_handler.rs
```diff
@@ -125,66 +131,100 @@ where
...
+            .iter()
+            .filter_map(|item| match item {
+                ProofOrCommitment::Commitment(commitment) => Some(commitment.l2_end_block_number),
+                _ => None,
+            })
+            .max()
+            .unwrap_or(0);
+
+        if l2_end_block_number > 0 {
             // If the L2 range does not exist, we break off the current process call
```
eyusufatik, line 151: cool trick but isn't this a problem if we have something like this:

10 20 u64::max

10 and 20 could be processed

crates/fullnode/src/da_block_handler.rs
```diff
@@ -125,66 +131,100 @@ where
...
+                ProofOrCommitment::Commitment(commitment) => {
+                    if let Err(e) = self
+                        .process_sequencer_commitment(l1_block, &commitment)
+                        .await
+                    {
+                        match e {
+                            SyncError::MissingL2(msg, start_l2_height, end_l2_height) => {
+                                warn!("Could not completely process sequencer commitments. Missing L2 blocks {:?} - {:?}, msg = {}", start_l2_height, end_l2_height, msg);
+                                return;
+                            }
```
eyusufatik, line 170: can't we put these into pending? instead of discarding everything on that L1 blcok we can process the rest tbh

crates/fullnode/src/da_block_handler.rs
```diff
@@ -125,66 +131,100 @@ where
...
+                                unreachable!("Error irrelevant!")
+                            }
+                            SyncError::SequencerCommitmentWithIndexNotFound(_) => {
+                                unreachable!("Error irrelevant!")
+                            }
+                            SyncError::ProvenHeightExceedsCommittedHeight(
+                                proven_height,
+                                committed_height,
+                            ) => {
+                                error!("Could not process ZK proofs: L2 status proven height {proven_height:?} above committed height {committed_height:?}")
```
eyusufatik: If this happens when we get a commitment for an index that we already know that is proven, message can be simpler: "Discarding already proven commitment."

eyusufatik: can put index end height etc. into the log

crates/fullnode/src/da_block_handler.rs
```diff
@@ -206,19 +246,77 @@ where
...
+            .get_highest_l2_height_for_status(L2HeightStatus::Committed, None)?
+        {
+            // Only proceed if the commitment height and index are higher than the stored one
+            if end_l2_height <= committed_height.height {
+                info!(
+                    "Skipping sequencer commitment with height {end_l2_height} as it is not strictly superior to existing commitment with height {}",
+                    committed_height.height,
+                );
+                return Ok(ProcessingResult::Discarded);
+            }
```
eyusufatik, line 282: @ercecan check if we do this in LCP

crates/fullnode/src/da_block_handler.rs
```diff
@@ -319,9 +425,52 @@ where
...
+            .get_highest_l2_height_for_status(L2HeightStatus::Committed, None)?
+            .unwrap_or_default();
+
+        if proven_height > committed_height {
+            return Err(SyncError::ProvenHeightExceedsCommittedHeight(
+                proven_height,
+                committed_height,
+            ));
+        }
+
```
eyusufatik: I guess we can panic on this should never happen. if it does, obvious bug.

crates/fullnode/src/da_block_handler.rs
```diff
@@ -369,30 +518,48 @@ where
...
+                l2_start_height = sequencer_commitment.l2_end_block_number + 1;
+            } else {
                 return Err(anyhow!(
-                    "Proof verification: For a known and verified sequencer commitment. Hash mismatch - expected 0x{} but got 0x{}. Skipping proof.",
-                    hex::encode(sequencer_commitment.serialize_and_calculate_sha_256()),
-                    hex::encode(expected_hash)
-                ).into());
+                    "Commitment index {index} is missing for proof, discarding proof."
+                )
+                .into());
```
eyusufatik: want to make a SyncError for this?

crates/fullnode/src/da_block_handler.rs
```diff
@@ -369,30 +518,48 @@ where
...
+            .zip(commitments_hashes)
         {
             // Check if hash matches
-            let sequencer_commitment = self
-                .ledger_db
-                .get_commitment_by_index(index)?
-                .ok_or(SyncError::SequencerCommitmentWithIndexNotFound(index))?;
-
-            if sequencer_commitment.serialize_and_calculate_sha_256() != expected_hash {
+            if let Some(sequencer_commitment) = self.ledger_db.get_commitment_by_index(index)? {
```
eyusufatik, line 515: there is a put_l2_block_status in this in this if call. should we get rid of that now we have a different tracking API?

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
```diff
@@ -69,6 +70,9 @@ pub const SEQUENCER_LEDGER_TABLES: &[&str] = &[
     ShortHeaderProofBySlotHash::table_name(),
     CommitmentMerkleRoots::table_name(),
     SequencerCommitmentByIndex::table_name(),
+    L2StatusHeights::table_name(),
```
eyusufatik, line 72: let's get rid of L2BlockStatus and old status RPCs that use this table.

bin/citrea/tests/bitcoin/full_node.rs
```diff
@@ -80,3 +100,1508 @@ async fn test_full_node_restart() -> Result<()> {
...
+        da.invalidate_block(&initial_height_hash).await?;
+        let block_count = da.get_block_count().await?;
+        assert_eq!(block_count, f.initial_da_height);
+        // Restart and remove rolledback txs from mempool
+        da.restart(None, None).await?;
+        assert_eq!(
+            da.get_raw_mempool().await?.len(),
+            0,
+            "Mempool should be empty"
+        );
```
eyusufatik, line 866: instead of resending out of order

you can set a very high number for max l2 blocks per commitment so that the sequencer does not commit, than calculate the commitment yourself in the tests and send out of order. 

you did a pretty similar thing below actually, no?

bin/citrea/tests/bitcoin/full_node.rs
```diff
@@ -80,3 +100,1508 @@ async fn test_full_node_restart() -> Result<()> {
...
+            merkle_root: merkle_root4,
+            l2_end_block_number: max_l2_blocks_per_commitment * 4,
+            index: 4,
+        };
+
+        let proof3 = wait_for_zkproofs(full_node, proof_l1_height, None, 1)
+            .await
+            .unwrap()[0]
+            .clone()
+            .proof;
```
eyusufatik, line 849: maybe making asserts here about what commitments were proven in the proofs would be a good idea.

bin/citrea/tests/bitcoin/full_node.rs
```diff
@@ -80,3 +100,1508 @@ async fn test_full_node_restart() -> Result<()> {
...
+        let committed_height = full_node
+            .client
+            .http_client()
+            .get_last_committed_l2_height()
+            .await?
+            .unwrap();
+        assert_eq!(committed_height.height, max_l2_blocks_per_commitment * 2);
+        assert_eq!(committed_height.commitment_index, 2);
+
+        // Make sure proof was discarded even after processing its commitment range
```
eyusufatik, line 969: why?

eyusufatik: bc you havent sent the proof ok.

eyusufatik (changes requested):

bin/citrea/tests/bitcoin/full_node.rs
jfldde: Because at the time of proof processing, the commitment range didn't exist. I'll explicit the comment a bit more

crates/fullnode/src/da_block_handler.rs
jfldde: They would be processed. `l2_end_block_number` would resolve to u64::max and be above > 0.
It would be 0 in the case where we don't have any commitment.

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
jfldde: This and `put_l2_block_status` above is also used by batch-prover. Shall we replace it with this new status tracking or just remove it ?

eyusufatik:
another thing I forgot to mention is we can throw in a full node in one of the light client tests (that run with batch prover and sequencer) and see if we have the same block statuses with the LCP

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
eyusufatik: batch prover does not need to track any L2 block status

eyusufatik: CC @yaziciahmet right?

yaziciahmet: no need to track L2 block status in batch prover.

jfldde: We don't use it for any batch prover logic but we do expose it via `get_l2_block_status` RPC. Shall we remove all of this from batch-prover ?

crates/fullnode/src/da_block_handler.rs
jfldde: With the early return, we'll re-try processing this L1 block as we're not popping it out of the l1 queue. It's kind of pending but we stay on this block. It's the behaviour wanted since we want to give time for the l2 block processing to catch up.

crates/fullnode/src/da_block_handler.rs
eyusufatik: no what I mean is we can process the commitments with 10 and 20 end block heights because of the u64::MAX commitment and stuck on that l1 block forever

jfldde: That was the existing behaviour. This change is just to have access to proofs and commitments in order. 
Shall we put a cap on the max `max_l2_blocks_per_commitment` value and use this as a max delta here ?

eyusufatik: we decided on making an issue for stuff that are not in the happy path

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
eyusufatik: yes we can remove that

eyusufatik (approved):

PR #2133, opened 2025-03-20
Fullnode sync: handle response size limit with retry

### LINKED ISSUE #2119: Full node try smaller range on ledger_getL2BlockRange response too big error
If the L2 blocks are full, the default value of 10 might go over the sequencer's response body size limit. We should retry pulling with a lower length range.

rakanalh:
# Description

## Linked Issues
- Fixes #2119

### DIFF
--- a/crates/common/src/l2.rs
+++ b/crates/common/src/l2.rs
@@ -367,7 +367,7 @@ async fn sync_l2(
 
         if l2_blocks.is_empty() {
             debug!(
-                "Soft Confirmation: no batch at starting height {}, retrying...",
+                "L2 block: no batch at starting height {}, retrying...",
                 start_l2_height
             );
 
@@ -377,7 +377,7 @@ async fn sync_l2(
 
         start_l2_height += l2_blocks.len() as u64;
 
-        // Make sure soft confirmations are sorted for us to make sure they are processed
+        // Make sure L2 blocks are sorted for us to make sure they are processed
         // in the correct order.
         l2_blocks.sort_by_key(|l2_block| l2_block.header.height);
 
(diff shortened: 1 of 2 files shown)


crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/rpc.rs
```diff
@@ -71,7 +71,7 @@ impl LedgerRpcProvider for LedgerDB {
             l2_block_ids.len() <= MAX_L2_BLOCK_PER_REQUEST as usize,
             "requested too many l2 blocks. Requested: {}. Max: {}",
             l2_block_ids.len(),
-            MAX_BATCHES_PER_REQUEST
+            MAX_L2_BLOCK_PER_REQUEST
```
rakanalh, line 74: I noticed this while debugging this PR... we are comparing against `MAX_L2_BLOCK_PER_REQUEST` so that is what we should have in the error message.

crates/common/src/l2.rs
```diff
@@ -314,68 +321,63 @@ where
...
+                    current_sync_blocks_count /= 2;
+                    if current_sync_blocks_count == 1 {
+                        warn!("Very slow sync at 1 block/s");
+                    } else if current_sync_blocks_count == 0 {
+                        error!("L2 blocks are getting too big. It is recommended to increase response size");
+                        // Stop the sync since we cannot fetch new soft confirmations.
+                        return;
+                    }
+                    continue;
+                }
```
eyusufatik, line 359: we also need to increase the sync blocks count back up imo (up to the original).

rakanalh: Done

eyusufatik (approved):

PR #2137, opened 2025-03-20
Better error handling & logging around system transactions

### LINKED ISSUE #2124: Better logging for system txs

eyusufatik:
# Description

## Linked Issues
Closes #2124

### DIFF
--- a/crates/evm/src/hooks.rs
+++ b/crates/evm/src/hooks.rs
@@ -258,6 +258,7 @@ impl<C: sov_modules_api::Context> Evm<C> {
 }
 
 /// Initializes system contracts
+#[cfg(feature = "native")]
 pub fn create_initial_system_events(
     current_slot_hash: [u8; 32],
     current_da_txs_commitment: [u8; 32],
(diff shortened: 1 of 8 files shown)


jfldde (approved):

crates/evm/src/hooks.rs
```diff
@@ -254,6 +259,7 @@ impl<C: sov_modules_api::Context> Evm<C> {
 }
 
 /// Initializes system contracts
+#[cfg(feature = "native")]
 pub fn create_initial_system_events(
     current_slot_hash: [u8; 32],
     current_da_txs_commitment: [u8; 32],
```
ercecan, line 265: I believe call populate_set_block_info_event here instead of creating the event in place, so we have the logs,+ maybe logging bridge initialize params could be useful as well? I thinj it would be nice since it is a one time thing

ercecan: nvm just saw you already did it in sequencer

crates/sovereign-sdk/rollup-interface/src/state_machine/stf.rs
```diff
@@ -143,6 +143,8 @@ pub enum L2BlockModuleCallError {
     EvmSystemTransactionPlacedAfterUserTx,
     /// System tx failed to parse
     EvmSystemTxParseError,
+    /// System tx not run successfully
+    EvmSystemTransactionFailed,
```
ercecan: Because you check if result and state is not success, this means the transaction reverted rather than failed, sos maybe EvmSystemTransactionReverted makes more sense for that specific if
but also keeping EvmSystemTransactionFailed and putting on the error hand of match arm of evm transact can also be useful

ercecan (approved):

PR #2144, opened 2025-03-21
Increase L1 sync speed

rakanalh:
# Description

I noticed while running the nodes including full node and batch prover that the L1 sync speed was quite slow especially on the batch prover. This is because `pending_l1_blocks` was behind a Mutex where DA block handler and the L1 syncer had a  read/write lock contention on the Mutex.

The reason we wanted to have this is because we did not want pending l1 blocks to grow beyond limits when trying to process those blocks. The better lock-free solution was to use a bounded channel whose limit is 10 blocks similar to how the we previously manually did with the lock + check on the size.

This lock-free approach led to a much faster L1 sync.

On my other PRs, i had to wait for a while (> 5 minutes) until the batch prover found block ~ 1000 at which there was a commitment to prove. However, with this new version, the sync took seconds to reach to the desired block.

### DIFF
--- a/crates/common/src/da.rs
+++ b/crates/common/src/da.rs
@@ -1,4 +1,3 @@
-use std::collections::VecDeque;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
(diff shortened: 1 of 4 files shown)


crates/common/src/da.rs
```diff
@@ -56,13 +55,12 @@ pub async fn sync_l1<Da>(
                     }
                 };
 
-            let mut queue = block_queue.lock().await;
-
-            if queue.len() < 10 {
-                queue.push_back(l1_block.clone());
-            } else {
-                debug!("Block queue is full, will try later...");
-                break;
```
rakanalh, line 65: This check was replaced by having a channel which has a size of 10 instead of checking the queue length WHILE holding the lock.

eyusufatik:
We've talked about this before.

The reason we need a mutex is simple. We don't want to keep pushing new L1 blocks all the time (growing the queue indefinitely) because when an L1 block cannot be processed, let's say due to missing L2 blocks, we have an always growing list of blocks. This is not acceptable.

PR #2147, opened 2025-03-21
fix sequencer resubmission logic

### LINKED ISSUE #1998: Update sequencer commitment resubmission logic
Right now the sequencer commitment resubmission logic works like this:
- We get our pendind commitments from ledger dn
- We get the submitted pending commitments from mempool
- We get the last commitment l1 height (which is the da slot height of latest l2 block of latest commitment that was submitted)
- We scan starting from that l1 height to the head of l1 and get mined commitments as well
- We iterate over our pending commitments from db if we see a commitment which is already mined or in mempool we remove it from db else we resubmit

This won't work now because after fork2 there is no da data such as da slot height in l2 block

Thus we need to change the logic

eyusufatik:
# Description

## Linked Issues
Closes #1998

## Testing

## Docs

### DIFF
--- a/crates/sequencer/src/commitment/mod.rs
+++ b/crates/sequencer/src/commitment/mod.rs
@@ -10,7 +10,7 @@ use parking_lot::RwLock;
 use rs_merkle::algorithms::Sha256;
 use rs_merkle::MerkleTree;
 use sov_db::ledger_db::SequencerLedgerOps;
-use sov_db::schema::types::{L2BlockNumber, SlotNumber};
+use sov_db::schema::types::L2BlockNumber;
 use sov_modules_api::{StateDiff, WorkingSet};
 use sov_rollup_interface::da::{BlockHeaderTrait, DaTxRequest, SequencerCommitment};
 use sov_rollup_interface::services::da::{DaService, TxRequestWithNotifier};
--- a/crates/sequencer/src/commitment/mod.rs
+++ b/crates/sequencer/src/commitment/mod.rs
@@ -178,9 +178,12 @@ where
         let start = Instant::now();
         let ledger_db = self.ledger_db.clone();
 
-        ledger_db
-            .put_commitment_by_index(&commitment)
-            .map_err(|_| anyhow!("Sequencer: Failed to store sequencer commitment by index"))?;
+        // Even though the commitment service does not shutdown before we get a response from the DA service,
+        // we still need to store the commitment in the pending commitments, so that if anything happens
+        // from the time we send the tx to the DA service and until we get a response (so the tx wasn't even submitted yet),
+        // e.g. server shutdown, we can resubmit the pending commitments.
+        //
+        // So pending status for a commitment spans from da service submission to entrance to mempool.
         ledger_db.put_pending_commitment(&commitment).map_err(|_| {
             anyhow!("Sequencer: Failed to store sequencer commitment in pending commitments")
         })?;
@@ -195,6 +198,10 @@ where
                 .saturating_duration_since(start)
                 .as_secs_f64(),
         );
+
+        ledger_db
+            .put_commitment_by_index(&commitment)
+            .map_err(|_| anyhow!("Sequencer: Failed to store sequencer commitment by index"))?;
 
         ledger_db.delete_pending_commitment(commitment.index)?;
         ledger_db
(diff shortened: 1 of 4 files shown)


crates/sequencer/src/commitment/mod.rs
```diff
@@ -196,6 +202,10 @@ where
                 .as_secs_f64(),
         );
 
+        ledger_db
+            .put_commitment_by_index(&commitment)
+            .map_err(|_| anyhow!("Sequencer: Failed to store sequencer commitment by index"))?;
+
```
yaziciahmet, line 208: is there any specific reason you carried saving the commitment after da submission?

crates/sequencer/src/commitment/mod.rs
```diff
@@ -207,26 +217,53 @@ where
...
             pending_mempool_commitments
         );
 
-        let last_commitment_l1_height = self
-            .ledger_db
-            .get_l1_height_of_last_commitment()?
-            .unwrap_or(SlotNumber(1));
-        let mined_commitments = self
-            .get_mined_commitments_from(last_commitment_l1_height)
+        // to determine which L1 block to start scannign from
```
yaziciahmet: scanning

crates/sequencer/src/commitment/mod.rs
```diff
@@ -207,26 +217,53 @@ where
...
+            let l2_height = {
+                let last_commitment = self.ledger_db.get_last_commitment()?;
+
+                if let Some(last_commitment) = last_commitment {
+                    last_commitment.l2_end_block_number
+                } else {
+                    // if no commitment yet, just use block number 1
+                    1
+                }
+            };
```
yaziciahmet: let l2_height = ledger_db.get_last_commitment()?.map(...).unwrap_or(1);

crates/sequencer/src/commitment/mod.rs
```diff
@@ -207,26 +217,53 @@ where
...
+            // set state to end of evm block l2_height
+            working_set.set_archival_version(l2_height + 1);
+
+            let l1_height = get_last_l1_height_in_light_client(
+                &Evm::<DefaultContext>::default(),
+                &mut working_set,
+            )
+            .expect("There must be a last l1 height");
+
+            l1_height.to::<u64>() + 1u64
```
yaziciahmet: + 1 (no need for u64 postfix)

yaziciahmet: + 1 (no need for u64 postfix)

crates/sequencer/src/commitment/mod.rs
```diff
@@ -236,21 +273,33 @@ where
...
+
         for pending_db_comm in pending_db_commitments {
             if pending_commitments_to_remove
                 .iter()
                 .any(|commitment| commitment.index == pending_db_comm.index)
             {
+                // thins pending commiment is either mined or in the L1 mempool
+                // so we can delete it from the pending db
+                // and also update last sequencer commitment l2 height
+
```
yaziciahmet, line 286: I think this is easy to understand from the variable names and fns already. comment can be removed.

crates/sequencer/src/commitment/mod.rs
```diff
@@ -306,18 +355,22 @@ where
 
     async fn get_mined_commitments_from(
         &self,
-        _da_height: SlotNumber,
+        start_height: u64,
     ) -> anyhow::Result<Vec<SequencerCommitment>> {
+        info!(
```
yaziciahmet, line 355: not info, debug

yaziciahmet (changes requested):

crates/sequencer/src/commitment/mod.rs
yaziciahmet: dont know how it got duplicated

crates/sequencer/src/commitment/mod.rs
eyusufatik: it makes more sense that way because if somehow sequencer dies right before commitment is sent to mempool, we can use pending to restruct but the commitment by index table being filled might be a little problematic.


It used to be the case that we were not waiting on DA send, but the commitment service runs in a different task now so we don't need to care about waiting on DA tx mining.

crates/sequencer/src/commitment/mod.rs
eyusufatik: everytime I read this code again I have to spend time understanding the underlying logic. so comment is fine.

crates/sequencer/src/commitment/mod.rs
eyusufatik: imo info, esp.given that we had problems with this before, it helps to have these kind of logs (and also this runs only once)

yaziciahmet (approved):

PR #2151, opened 2025-03-24
Cleanup ledger db & remove last commitment table

yaziciahmet:
# Description

- Some put operations was being done as SchemaBatch even though it was a single write. There is no need for ledger db to do that, even though under the hood sov DB is already doing a SchemaBatch write for single puts.
- Carried put_l2_block trait fn to be an inner fn as its only used by ledger db itself.
- Removed LastSequencerCommitmentSent table to avoid unnecessary double writes. The name first suggested to me that this is a table where we track the last commitment that is sent to DA, but the usage of this table rather suggests that it is the commitment that has the highest index. Hence, I completely removed this table and implemented get_last_commitment fn by reading the last element in the SequencerCommitmentsByIndex table.
 
One bug that existed most likely due to confusion of having 2 such tables and set_last_commitment fn taking full commitment as input even though just index would have sufficed, on resubmitting pending commitments in sequencer, if submission was complete, we only saved the index to LastSequencerCommitmentSent table, but didnt save the actual commitment to SequencerCommitmentsByIndex table. Fixed that as well.

Also on rollback, we weren't pruning the SequencerCommitmentsByIndex table. Fixed that.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
+++ b/crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
@@ -243,12 +243,7 @@ impl SharedLedgerOps for LedgerDB {
             None => (l2_height, l2_height),
         };
 
-        let mut schema_batch = SchemaBatch::new();
-
-        schema_batch.put::<L2RangeByL1Height>(&l1_height, &new_range)?;
-        self.db.write_schemas(schema_batch)?;
-
-        Ok(())
+        self.db.put::<L2RangeByL1Height>(&l1_height, &new_range)
     }
 
     #[instrument(level = "trace", skip(self), err, ret)]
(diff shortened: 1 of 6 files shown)


crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
```diff
@@ -141,6 +140,16 @@ impl LedgerDB {
         }
     }
 
+    fn put_l2_block(
+        &self,
+        l2_block: &StoredL2Block,
+        l2_block_number: &L2BlockNumber,
```
eyusufatik: can't you get number from block?

eyusufatik (approved):

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
yaziciahmet: sure can do

PR #2157, opened 2025-03-24
Update l1 fee overhead calculations

exeokan:
# Description

Updates l1 fee overhead calculations after no-backwards-compatibility changes


l1_fee_overhead.md
```diff
@@ -1,36 +1,42 @@
...
+```452f482f0000000000000000``` = 12 bytes
 
+```67483f22e31f1ca26cb5475507b3473e9a0fa9913d38364dc81eac79f161fddc``` = 32 bytes
 
-## sc rule enforcer
+12 + 32 = 44 bytes 
+
+44 * 256 = 11264 bytes in a batch
+
+### account amount
```
eyusufatik, line 20: well actually, if we update calc_diff size, we can just charge the address => idx mapping?

I think we should do that

eyusufatik: we are actually charging it. so no need to include it in the overhead imo

ercecan (approved):

PR #2166, opened 2025-03-26
Refactor commitment service

### LINKED ISSUE #2079: Rewrite sequencer commitment service

rakanalh:
# Description

## Linked Issues
- Fixes #2079


crates/sequencer/src/commitment/helpers.rs
```diff
@@ -0,0 +1,21 @@
...
+        .max();
+    // max index from last commitment:
+    let max_last = db.get_last_commitment().unwrap().map(|s| s.index);
+    // maximum of pending and last:
+    let max_db = max_pending.max(max_last);
+    if let Some(max_db) = max_db {
+        max_db + 1
+    } else {
+        // if comms are empty, then index is 1
+        1
```
eyusufatik, line 19: should be 1 after latest LCP pr by @ercecan , be careful while merging nightly

crates/sequencer/src/commitment/service.rs
```diff
@@ -0,0 +1,377 @@
...
+            .map_err(|_| anyhow!("Sequencer: Failed to store sequencer commitment by index"))?;
+
+        ledger_db.delete_pending_commitment(commitment.index)?;
+
+        info!("New commitment. L2 range: #{}-{}", l2_start.0, l2_end.0);
+
+        Ok(())
+    }
+
+    // TODO Re-write since da_slot_height is not indexed as part of L2Block. Ref https://github.com/chainwayxyz/citrea/issues/1998
```
eyusufatik: todo should be unnecessary now.

crates/sequencer/src/commitment/controller.rs
```diff
@@ -21,147 +19,97 @@ where
...
+                debug!("Enough state diff size to submit commitment");
+                return Ok(Some(L2BlockNumber(l2_start)..=L2BlockNumber(l2_height)));
+            }
+
+            let compressed_state_diff = compress_blob(&uncompressed_state_diff).unwrap();
+            if compressed_state_diff.len() > MAX_TXBODY_SIZE {
+                debug!("Enough state diff size to submit commitment");
+                return Ok(Some(L2BlockNumber(l2_start)..=L2BlockNumber(l2_height)));
+            }
+        }
```
eyusufatik: I mean this will work ofc, but why not keep a latest state diff as we did before?

the sequencer can still save by L2 height, but we can keep a in-memory merged state diff from last committed height up to the number we are checking. (and this could be constructed on commitment service start up.)

+ once we commit, we should delete state diffs for given height range. as they increase storage usage a lot.

eyusufatik (changes requested):

crates/sequencer/src/commitment/helpers.rs
rakanalh: It is 1... am i understanding this incorrectly?

rakanalh: https://github.com/chainwayxyz/citrea/commit/af9fa450c95aa4a55561a913b0416e798e861658#diff-989a25c8b80179292326c46f154a068f32320dae3f54a4fe1a1741885aa728df

crates/sequencer/src/commitment/controller.rs
rakanalh: The problem is that, before this change, we used to send the L2 height + state diff to the commitment service which in turn calculated the overall state diff.

Now since the service is independent of the sequencer, it wouldn't be able to know the current L2 state diff unless we store that.

However, you have a good point about removing the state diffs if we decide to commit. I'll implement that.

crates/sequencer/src/commitment/helpers.rs
eyusufatik: weird, I saw 0 when I was reviewing :D

bin/citrea/tests/bitcoin/batch_prover_test.rs
```diff
@@ -111,7 +111,7 @@ impl TestCase for BasicProverTest {
         }
 
         // Wait for blob inscribe tx to be in mempool
-        da.wait_mempool_len(4, None).await?;
+        da.wait_mempool_len(2, None).await?;
```
eyusufatik: should be 4. making 2 sequencer commitments above

bin/citrea/tests/bitcoin/batch_prover_test.rs
```diff
@@ -177,7 +177,7 @@ impl TestCase for BasicProverTest {
...
```
eyusufatik: same

bin/citrea/tests/bitcoin/batch_prover_test.rs
```diff
@@ -573,7 +573,7 @@ impl TestCase for ParallelProvingTest {
         }
 
         // Wait for 2 commitments (4 txs) to hit DA mempool
-        da.wait_mempool_len(4, Some(Duration::from_secs(420)))
+        da.wait_mempool_len(2, Some(Duration::from_secs(420)))
```
eyusufatik: again.

bin/citrea/tests/bitcoin/batch_prover_test.rs
```diff
@@ -586,18 +586,18 @@ impl TestCase for ParallelProvingTest {
             .await?;
 
         // Wait for batch proof txs to hit mempool
-        da.wait_mempool_len(4, Some(Duration::from_secs(420)))
+        da.wait_mempool_len(2, Some(Duration::from_secs(420)))
```
eyusufatik: iirc this test expects two batch proofs here.

eyusufatik: seeing below changes to proof counts then I believe the sequencer started to commit only one proof now?

I think this is because the commitment service  compares last committed to head and decides to commit to the whole range. 

let's renmae min l2 blocks to commit max l2 blocks to commit (don't forget the env var names also) and commit to last_l2_end + 1, last_l2_end + 1 + max l2 blocks to commit

bin/citrea/tests/bitcoin/light_client_test.rs
```diff
@@ -222,7 +222,7 @@ impl TestCase for LightClientProvingTestMultipleProofs {
 
         let min_l2_blocks_per_commitment = sequencer.min_l2_blocks_per_commitment();
 
-        let n_commitments = 2;
+        let n_commitments = 1;
```
eyusufatik: this will also change due to my proposed change on the previous comment

crates/sequencer/src/commitment/controller.rs
rakanalh: Implemented keeping track of threshold internally

crates/sequencer/src/commitment/service.rs
eyusufatik, line 1: just realised that if we change the controlling loop here from always checking from last l2 committed to head block num to an incremental loop of last l2 committed, n (where n in [last committed + 1, head]) we would have better committing strategy in terms of state diff checks (and all the comments about the changes in tests would be resolved.)

eyusufatik (changes requested):

Cargo.toml
```diff
@@ -160,7 +160,7 @@ alloy-eips = { version = "0.4.2", default-features = false }
 alloy-consensus = { version = "0.4.2", default-features = false, features = ["serde", "serde-bincode-compat"] }
 alloy-network = { version = "0.4.2", default-features = false }
 
-citrea-e2e = { git = "https://github.com/chainwayxyz/citrea-e2e", rev = "7b392a2" }
+citrea-e2e = { git = "https://github.com/chainwayxyz/citrea-e2e", branch = "rakanalh/max-l2-blocks" }
```
rakanalh: We have to set the revision here once https://github.com/chainwayxyz/citrea-e2e/pull/84 is merged

crates/sequencer/src/commitment/service.rs
```diff
@@ -0,0 +1,385 @@
...
+                                error!("Error while checking commitment criteria: {:?}", e);
+                                continue;
+                            },
+                            _ => {
+                                continue;
+                            }
+                        };
+
+                        let index = self.next_commitment_index;
+                        self.next_commitment_index += 1;
```
eyusufatik: should we increment this in self.commit() ?

bin/citrea/tests/bitcoin/sequencer_commitments.rs
```diff
@@ -215,19 +215,16 @@ impl TestCase for SequencerSendCommitmentsToDaTest {
...
+        // publish max_l2_blocks_per_commitment - 1 confirmations, no commitments should be sent
+        for _ in 0..max_l2_blocks_per_commitment - 1 {
             sequencer.client.send_publish_batch_request().await?;
         }
         sequencer
-            .wait_for_l2_height(min_l2_blocks_per_commitment - 1, None)
+            .wait_for_l2_height(max_l2_blocks_per_commitment - 1, None)
             .await?;
 
-        da.generate(FINALITY_DEPTH).await?;
```
eyusufatik, line 228: you can also remove the for loop below, i guess originally it was testing that a commitment was not sent but it's not really working tbh :D

eyusufatik (approved):

PR #2169, opened 2025-03-26
Rework Batch Prover

### LINKED ISSUE #2042: Batch prover can generate proof over range of sequencer commitments over multiple da blocks
Can be implemented after #2014 with the new sequencer commitment system
This will be like some type of a rewrite for batch prover da block handler:

- L1 scanner will read for sequencer commitments from da blocks, this time instead of generating a proof input for every l1 block, it will only verify and save them to a ledger db maybe.
- After that in a separate thread the prover will get as many non-proven sequencer commitments as possible to generate a proof input where:
- Sequencer commitment indexes are sequential, meaning no batch proof should commit to commitments with indexes like 1,2,4
- The total state diff rule is the same
- If for example batch prover sees not sequential commitments like 1,2 and 4 it should generate a proof for commitments 1,2 and wait until the commitment with index 3 to arrive to generate the next proof
- All in all batch prover generates proofs for multiple sequencer commitments over multiple da blocks as long as the commitments are sequential and the last proven commitment index is sequential to the first commitment index of the next proof

### LINKED ISSUE #2078: Fix batch prover behaviour by making it not prove any sequencer commitment it had proven before
context: https://github.com/chainwayxyz/citrea/pull/2073#discussion_r1991141415

With this change we can also uncomment prover_skips_preproven_commitments_test

yaziciahmet:
# Description

Rework of Batch Prover. Now, batch prover has 3 core services: L1Syncer, L2Syncer, Prover. All these services run in their own tokio task.

**L1Syncer**
Watches L1 blocks from DA, extracts sequencer commitments, and saves them to 2 tables, commitment by index and pending commitments. Pending commitments table is where we store the commitments that have been scanned by L1Syncer, but is not yet processed by the Prover. Anytime an L1Block is processed, sends an empty to signal to prover for a potential proving job invocation. I say potential because based on the rules that will be defined in Prover, it might just decide to skip proving temporarily, or prove partially.

**L2Syncer**
Watches L2 blocks from Sequencer, and processes and saves them to database. It already has a broadcast channel for pinging new L2Block arrival, which will be used to ping the Prover for a potential proving job invocation. This signal is useful for the special case where L1Syncer has seen a commitment in which the L2Syncer has not synced up to yet. If Prover is in a state where it expects to sync up to a certain L2 block, this signal is useful for immediately starting the job.

**Prover**
This is the core and biggest component of batch prover. It listens for signals from L1Syncer and L2Syncer as described in their section. With the new prover, we no longer drive our proving process on L1 basis. L1Syncer is just there to scan commitments, and Prover works solely on these pending commitments. It reads pending commitments (that are populated by L1Scanner), filters out the ones that have l2 blocks that are not yet synced, partitions commitments, starts each partitions as its own proving job and saves the jobs to db. Job ids are uuid v7, which ensure global uniqueness, avoid id state, and also contain timestamp in them which could be useful for debugging purposes.

This PR is not finished yet, but the overall skeleton and flow should be pretty obvious and reviewable.

Remaining TODOs:
- [x] Figure out what would be the best way to save proofs to db (by l1, by job id, by l2 block range etc.)
- [x] Bonsai session recovery, and proof DA submission recovery
- [x] Implement prove rpc to invoke proving by skipping sampling
- [x] Implement set commitment rpc to either override or add future commitments to database as pending
- [x] Implement job rpc for debugging currently running proving jobs
- [ ] General cleanup of TODOs
- [x] Fix tests 

Future issues which I will open and implement:
- [x] Add tests for the job rpc endpoints

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


eyusufatik:
> Figure out what would be the best way to save proofs to db (by l1, by job id, by l2 block range etc.)

I think the clearest way would be to map job id => proof

and then map other stuff to job id

this could be something like 

sequencer comm index => job id
and maybe
l1 => vec<job ids>

bin/citrea/src/main.rs
```diff
@@ -284,26 +284,16 @@ where
...
-            };
+            task_manager.spawn(TaskType::Secondary, |cancellation_token| async move {
+                l1_syncer.run(cancellation_token).await
+            });
 
             task_manager.spawn(TaskType::Secondary, |cancellation_token| async move {
-                l1_block_handler
-                    .run(l1_start_height, cancellation_token)
-                    .await
+                prover.run(cancellation_token).await
```
eyusufatik: prover job probably primary? cc @rakanalh

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
```diff
@@ -106,19 +107,17 @@ pub const BATCH_PROVER_LEDGER_TABLES: &[&str] = &[
     L2BlockByNumber::table_name(),
     L2BlockByHash::table_name(),
     ShortHeaderProofBySlotHash::table_name(),
-    L2RangeByL1Height::table_name(),
```
eyusufatik, line 114: can delete this table from everywhere, nothing uses this anymore

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
```diff
@@ -106,19 +107,17 @@ pub const BATCH_PROVER_LEDGER_TABLES: &[&str] = &[
...
     L2GenesisStateRoot::table_name(),
     ProverLastScannedSlot::table_name(),
-    L2BlockStatus::table_name(),
-    CommitmentsByNumber::table_name(),
-    ProofsBySlotNumber::table_name(),
-    ProofsBySlotNumberV2::table_name(),
-    PendingProvingSessions::table_name(),
     ProverStateDiffs::table_name(),
     LastPrunedBlock::table_name(),
-    CommitmentMerkleRoots::table_name(),
```
eyusufatik, line 124: this one also remove from everywhere

#SequencerCommitmentByIndex
eyusufatik, line 1: don't forget to delete

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
```diff
@@ -558,6 +555,52 @@ impl BatchProverLedgerOps for LedgerDB {
...
+        iter.seek_to_first();
+
+        let mut commitment_indices = vec![];
+        for el in iter {
+            let (index, _) = el?.into_tuple();
+            commitment_indices.push(index);
+        }
+
+        Ok(commitment_indices)
+    }
```
eyusufatik: this is gonna get messy. sequencer also has one of these pending commitment set and gets

full node about to have one also with jeremy's pr

give distinctive names pls

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
```diff
@@ -558,6 +555,52 @@ impl BatchProverLedgerOps for LedgerDB {
...
+        self.db.write_schemas(schema_batch)
+    }
+
+    #[instrument(level = "trace", skip(self), err)]
+    fn set_proving_job_finished(&self, id: Uuid) -> anyhow::Result<()> {
+        self.db.delete::<ProverRunningJobs>(&id)
+    }
 }
 
 impl ProvingServiceLedgerOps for LedgerDB {
```
eyusufatik: does this need a clean up now? afaik this is from our old proving "recovery" which we don't use anymore

crates/batch-prover/src/l1_syncer.rs
```diff
@@ -0,0 +1,184 @@
...
+                        warn!("Got commitment index {} that was already in db", index);
+                        // Sanity checks
+                        assert_eq!(
+                            commitment.l2_end_block_number, db_commitment.l2_end_block_number,
+                            "Found duplicate commitment with different l2 block numbers"
+                        );
+                        assert_eq!(
+                            commitment.merkle_root, db_commitment.merkle_root,
+                            "Found duplicate commitment with different merkle roots"
+                        );
```
eyusufatik: no don't make an assertion like this. just ignore the new commitment with known index.

crates/batch-prover/src/l1_syncer.rs
```diff
@@ -0,0 +1,184 @@
...
+                    }
+                    None => {
+                        self.ledger_db
+                            .put_commitment_by_index(commitment)
+                            .expect("Should store commitment");
+                        self.ledger_db
+                            .put_pending_commitment(index)
+                            .expect("Should set commitment status to pending");
+
+                        commitment_indices.push(index);
```
eyusufatik: we should either make merkle root checks here, or when we are about to prove. if merkle root mismatch, just don't continue proving. there is something very wrong.

crates/batch-prover/src/prover.rs
```diff
@@ -0,0 +1,764 @@
...
+        let pending_commitment_indices = self.ledger_db.get_pending_commitments()?;
+
+        let mut commitments = Vec::with_capacity(pending_commitment_indices.len());
+        for index in pending_commitment_indices {
+            let commitment = self
+                .ledger_db
+                .get_commitment_by_index(index)?
+                .expect("Unproven commitment must exist by index");
+            commitments.push(commitment);
+        }
```
eyusufatik: you can make a ledger db function that does this. afaik there is one in sequencer for example

crates/batch-prover/src/prover.rs
```diff
@@ -0,0 +1,764 @@
...
+    /// Partition the commitments into provable chunks. Here are the rules when partitioning in Normal mode:
+    /// 1. If there is an index gap in between commitments, partition is formed
+    /// 2. If ƒork has changed, partition is formed
+    /// 3. If max state diff limit is surpassed, partition is formed
+    fn partition_commitments<'a>(
+        &self,
+        commitments: &'a [SequencerCommitment],
+        mode: PartitionMode,
+    ) -> anyhow::Result<Vec<Partition<'a>>> {
+        let start_l2_height = if commitments[0].index == 0 {
```
eyusufatik: put todo here to fix first commitment index to be 1 after #2180.

crates/batch-prover/src/prover.rs
```diff
@@ -0,0 +1,764 @@
...
+            // If this is the first commitment ever, start from 1
+            1
+        } else {
+            let previous_commitment_index = commitments[0].index - 1;
+            // If this is not the first commitment, start l2 height will be end block number + 1 of the previous commitment
+            self.ledger_db
+                .get_commitment_by_index(previous_commitment_index)?
+                .expect("Previous commitment must exist")
```
eyusufatik: no don't panic. if this happens just don't prove that commitment. also maybe the previous index is in commitments sent so this is not ok.

crates/batch-prover/src/prover.rs
```diff
@@ -0,0 +1,764 @@
...
+}
+
+#[inline(always)]
+fn assert_state_diff_threshold(state_diff: &StateDiff) {
+    let serialized_diff = borsh::to_vec(state_diff).expect("Diff serialization cannot fail");
+    let compressed_diff = compress_blob(&serialized_diff).expect("Diff compression cannot fail");
+    assert!(
+        compressed_diff.len() > MAX_TXBODY_SIZE,
+        "Got single commitment bigger than txbody limit"
+    );
```
eyusufatik: don't assert, we should still prover that commitmnet as a single item in the partition.

crates/batch-prover/src/prover.rs
eyusufatik: think about this scenario. you get commitment indices

3 4 5

but you don't know 2.

you could still prove (4, 5)

crates/batch-prover/src/prover.rs
```diff
@@ -0,0 +1,764 @@
...
+                .into_iter()
+                .map(|comm| comm.index)
+                .collect::<Vec<_>>();
+
+            self.ledger_db
+                .insert_new_proving_job(id, &commitment_indices)
+                .context("Failed to insert prover job")?;
+            self.ledger_db
+                .delete_pending_commitments(commitment_indices)
+                .context("Failed to delete pending commitments")?;
```
eyusufatik: were these inserted as normal non-pending commitments before?

crates/batch-prover/src/prover.rs
```diff
@@ -0,0 +1,764 @@
...
+                job_rxs = remaining_rxs;
+                // TODO: this is very sketchy. swap_remove is deterministic, and that is what select_all is using to remove the completed job,
+                // but still relying on this to keep the correct order is nasty. try to find another way
+                let job_id = job_ids.swap_remove(idx);
+
+                ledger_db
+                    .set_proving_job_finished(job_id)
+                    .expect("Should delete job id");
+
+                // TODO: how to save proof? by l1? by job id? by commitment? by l2 block range?
```
eyusufatik: job id => proof

you already have job id => vec<seq com index> am i right? 

I think this would suffice

crates/batch-prover/src/prover.rs
```diff
@@ -0,0 +1,764 @@
...
+        };
+        let (id, rx) = self
+            .prover_service
+            .start_proving(proof_data, ReceiptType::Groth16)
+            .await;
+
+        (id, rx)
+    }
+
+    fn watch_proving_jobs(&self, proving_jobs: Vec<(Uuid, oneshot::Receiver<Proof>)>) {
```
eyusufatik, line 452: aren't we recording proving jobs somewhere? or is that handled in proverservice?

eyusufatik (changes requested):

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
yaziciahmet: yes will do cleanup of ledger

crates/batch-prover/src/l1_syncer.rs
yaziciahmet: but isn't it problematic if we receive same index commitment with different block numbers? it means the world is broken

crates/batch-prover/src/prover.rs
yaziciahmet: but if too big to fit in a tx, even if you prove, how do you submit that?

eyusufatik: that's why we have chunk / aggregators

crates/batch-prover/src/prover.rs
eyusufatik: there is a case we should watch out btw.

if we can't verify 4, that does not mean 4 itself must be wrong, but it might mean 3 is wrong etc.

now I realize it's safer to wait here.

eyusufatik: we can work on such cases later

crates/batch-prover/src/prover.rs
yaziciahmet: that's what I am saying, this assertion is for the case where single commitment causes state diff bigger than the txbody limit. if there is a single commitment that big, that means sequencer messed up and we have crucial problems with our chunking and commitment creation strategy.

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
yaziciahmet: I will cleanup the whole ledger db messiness in another pr.

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
yaziciahmet: another pr

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
yaziciahmet: I did, but to tackle this issue from the core, we should implement ledger db individually for each node type.

crates/batch-prover/src/prover.rs
yaziciahmet: in try_proving we start and record the proving job, and here we spawn a background task to watch the finalization of the proof, and update db accordingly

crates/batch-prover/src/prover.rs
yaziciahmet: in l1_syncer we insert both commidx => comm (CommitmentByIndex), and commidx => () (ProverPendingCommitments). and when list of commitments is partitioned into a proving job, they are removed from ProverPendingCommitments, and a new entry in ProverPendingJobs is created.

crates/batch-prover/src/l1_syncer.rs
eyusufatik: yes. but better to keep on proving on what you know.

yaziciahmet: done

crates/batch-prover/src/l1_syncer.rs
yaziciahmet: we can't do it here because here it is not guaranteed that batch prover is synced up to commitment's l2 block, so I did it in the prover.

crates/batch-prover/src/prover.rs
yaziciahmet: done https://github.com/chainwayxyz/citrea/pull/2169/commits/3cb6cf235fc8848f42f14d0af6b874d368832fdf

bin/citrea/src/main.rs
rakanalh: Yes, should be primary

rakanalh: The runner is primary though which is correct

bin/citrea/src/main.rs
```diff
@@ -284,26 +284,16 @@ where
...
             });
 
             task_manager.spawn(TaskType::Primary, |cancellation_token| async move {
-                if let Err(e) = prover.run(cancellation_token).await {
+                if let Err(e) = runner.run(cancellation_token).await {
```
rakanalh: IMO, as @eyusufatik pointed out. This prover job should be a primary. 
Also, we don't need the runner anymore since the all the runner does is make sure the L2 syncer is running. So i suggest that we just spawn the l2 syncer here without having 1 additional layer of abstraction.

crates/batch-prover/src/prover.rs
```diff
@@ -0,0 +1,824 @@
...
+pub enum PartitionMode {
+    /// Groups commitments the normal way
+    /// Generates proof(s) given l1 height using the same strategy of batch prover
+    Normal,
+    /// Every commitment is a group on their own
+    /// Generates a proof for every commitment
+    OneByOne,
+}
+
+struct PartitionState<'a> {
```
rakanalh: Maybe move these to `types.rs` or `state.rs`

crates/batch-prover/src/prover.rs
```diff
@@ -0,0 +1,824 @@
...
+
+type CommitmentStateTransitionData = (
+    VecDeque<Vec<u8>>,
+    VecDeque<Vec<(Witness, Witness)>>,
+    Vec<u64>,
+    VecDeque<Vec<L2Block>>,
+    Witness,
+);
+
+#[allow(clippy::too_many_arguments)]
```
rakanalh: The code here and below can be moved to `helpers.rs` or `utils.rs` for the sake of keeping this prover module a bit cleaner

rakanalh (commented):
Structural comments

guests/risc0/light-client-proof/mock/Cargo.lock
```diff
@@ -2266,6 +2267,7 @@ dependencies = [
  "serde",
  "sha2",
  "sov-keys",
+ "uuid",
```
eyusufatik: if we make uuid optional and activated on native it wouldn't end up here

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
```diff
@@ -21,6 +24,16 @@ pub mod light_client_proof;
 /// The ZK proof generated by the [`ZkvmHost::run`] method.
 pub type Proof = Vec<u8>;
 
+#[derive(Debug, Clone)]
+/// Wrapper around `Proof` to associate it with a job
+pub struct ProofWithJob {
```
eyusufatik, line 29: maybe make native?

crates/batch-prover/src/prover.rs
```diff
@@ -0,0 +1,792 @@
...
+            .collect()
+    }
+
+    /// Partition the commitments into provable chunks. Here are the rules when partitioning in Normal mode:
+    /// 1. If there is an index gap in between commitments, partition is formed
+    /// 2. If ƒork has changed, partition is formed
+    /// 3. If max state diff limit is surpassed, partition is formed
+    /// This function expects each commitment to have previous commitment, so, ensure filtering commitments with
+    /// `filter_prev_missing_commitments` before calling this function.
+    fn partition_commitments<'a>(
```
eyusufatik, line 297: can we write unit tests for this function? should be easy

crates/risc0/src/host/bonsai.rs
```diff
@@ -0,0 +1,354 @@
...
+        };
+        self.ledger_db
+            .upsert_pending_bonsai_session(job_id, db_session)
+            .context("Failed to upsert bonsai snark session")?;
+
+        let groth16_receipt = self.wait_snark_receipt(&snark_session).await?;
+        groth16_receipt
+            .verify_integrity_with_context(&VerifierContext::default())
+            .context("Failed to verify bonsai groth16 proof integrity")?;
+
```
eyusufatik, line 186: no need to invoke snark proving if the proof type is stark

eyusufatik (changes requested):

crates/risc0/src/host/bonsai.rs
yaziciahmet: I check for that in between line 167-169. if succinct, we just return the receipt immediately.

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
yaziciahmet: done

guests/risc0/light-client-proof/mock/Cargo.lock
yaziciahmet: done

crates/batch-prover/src/prover.rs
```diff
@@ -0,0 +1,830 @@
...
+            let state_root = self
+                .ledger_db
+                .get_l2_state_root(commitment.l2_end_block_number)?
+                .expect("State root must exist");
+            assert_eq!(
+                commitment.merkle_root, state_root,
+                "Invalid state root in commitment {}",
+                commitment.index
+            );
+        }
```
ercecan: merkle root in commitment is the merkle root of state roots from the commitments start l2 height to end l2 height

ercecan (dismissed):

crates/batch-prover/src/prover.rs
ercecan: sorry not the root of state roots the root of the tree of the commitment ranges l2 block hashes
root(commitment l2 start height block hash .. commitment l2 end height block hash)

yaziciahmet: ty very much, missed that commitment merkle root has a different meaning now. fixed https://github.com/chainwayxyz/citrea/pull/2169/commits/64ab8bc914465443405fab1ea9d06c30e51f4690

crates/batch-prover/src/partition.rs
```diff
@@ -0,0 +1,128 @@
...
+    partition_start_height: u64,
+    partition_start_idx: usize,
+    ledger_db: DB,
+}
+
+impl<'a, DB: BatchProverLedgerOps> PartitionState<'a, DB> {
+    pub fn new(commitments: &'a [SequencerCommitment], ledger_db: DB) -> anyhow::Result<Self> {
+        let start_l2_height = if commitments[0].index == 1 {
+            // If this is the first commitment ever, start from 1
+            1
```
ercecan: Fork2_start_height_non_zero() makes more sense here

eyusufatik: yes, indexed commitments will only be fore fork2

crates/sovereign-sdk/rollup-interface/src/node/rpc/mod.rs
```diff
@@ -168,13 +169,39 @@ pub struct LightClientProofResponse {
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct JobRpcResponse {
+    /// Job id
+    pub id: Uuid,
+    /// Commitments being proven in the job
+    pub commitments: Vec<SequencerCommitmentResponse>,
+    /// Proof result of the job. If proof is None, job still continues,
+    /// and if it is Some but l1_tx_id is 0-value, it is being submitted to L1.
+    pub proof: Option<BatchProofResponse>,
+}
```
ercecan, line 182: What is the result if the job had failed?

crates/sovereign-sdk/rollup-interface/src/node/rpc/mod.rs
```diff
@@ -168,13 +169,39 @@ pub struct LightClientProofResponse {
...
+    /// Merkle root of the commitment
+    #[serde(with = "hex::serde")]
+    pub merkle_root: [u8; 32],
+    /// Index of the commitment
+    pub index: u32,
+    /// L2 end block number of the commitment
+    pub l2_end_block_number: u64,
+    /// L1 height of the commitment
+    pub l1_height: u64,
+}
```
ercecan, line 196: Does this request also overwrite existing commitments?

crates/batch-prover/src/prover.rs
```diff
@@ -0,0 +1,844 @@
...
+            l1_signal_rx,
+            l2_block_rx,
+            request_rx,
+            sync_target_l2_height: None,
+            proving_paused: false,
+        }
+    }
+
+    pub async fn run(mut self, mut shutdown_signal: GracefulShutdown) {
+        self.recover_proving_sessions().await;
```
ercecan, line 102: In the old implementation there used to be an if check here with a config to either recover or not recover pending sessions, should we still do that just in case of a buggy proving session in bonsai or sth?
@eyusufatik

crates/sovereign-sdk/rollup-interface/src/node/rpc/mod.rs
yaziciahmet: yes, based on request of @eyusufatik

crates/sovereign-sdk/rollup-interface/src/node/rpc/mod.rs
yaziciahmet: error is not stored or returned in this case, its just logged. the exact rpc response in case of failure depends. if it failed during proving, then the proof will be null, but if proving was successful and something went wrong with da submission, then proof will be there, but the inside l1_tx_id will be null.

crates/batch-prover/src/prover.rs
```diff
@@ -0,0 +1,844 @@
...
+
+                    debug!("Got L2 signal to try proving");
+                    if let Err(e) = self.try_proving(PartitionMode::Normal, true).await {
+                        error!("Failed to start proving: {}", e);
+                    }
+                }
+                request = self.request_rx.recv() => {
+                    let Some(request) = request else {
+                        // no need to panic if for some reason request channel is closed
+                        error!("Prove request sender channel closed abruptly");
```
ercecan: ~~If the prove request channel is closed why not shut down the prover as well? I believe it would be easier for us to notice that there is a problem, not so easy with just a log I think~~

ercecan: nvm, this is only for rpc proving

yaziciahmet: I thought about that as well, and concluded to not do that just because we are not going to be able to send rpc requests. It can happen if the rpc server crashes for some reason. I am still in between... What do you say?

crates/batch-prover/src/prover.rs
eyusufatik: no worries. we can do that if we need to

crates/batch-prover/src/prover.rs
ercecan: Since this is not just some full node rpc maybe we can crash? Because we will only use this when we need to, so instead of wasting time trying to figure out if rpc is crashed maybe just bring the whole thing down?
@eyusufatik we can't decide what do you say?

eyusufatik: not a bad thing to do tbh. lets shut everything down in that case

crates/batch-prover/src/prover.rs
```diff
@@ -0,0 +1,844 @@
...
+        let prover_service = self.prover_service.clone();
+        let code_commitments_by_spec = self.code_commitments_by_spec.clone();
+
+        let mut proving_jobs = proving_jobs
+            .into_iter()
+            .map(|(job_id, rx)| async move {
+                let proof = rx.await.expect("Proof channel should never close");
+                (job_id, proof)
+            })
+            .collect::<FuturesUnordered<_>>();
```
ercecan, line 465: cool

bin/citrea/src/main.rs
```diff
@@ -282,29 +282,17 @@ where
...
-                }
-            };
+            task_executor.spawn_with_graceful_shutdown_signal(|shutdown_signal| async move {
+                l1_syncer.run(shutdown_signal).await
+            });
 
             task_executor.spawn_with_graceful_shutdown_signal(|shutdown_signal| async move {
-                l1_block_handler.run(l1_start_height, shutdown_signal).await
+                l2_syncer.run(shutdown_signal).await
             });
```
eyusufatik, line 291: maybe critical?

crates/batch-prover/src/prover.rs
```diff
@@ -0,0 +1,840 @@
...
+            // check state diff threshold
+            if compressed_diff.len() > MAX_TXBODY_SIZE {
+                cumulative_state_diff = commitment_state_diff;
+                state.add_partition(i - 1, PartitionReason::StateDiff)?;
+                continue;
+            }
+        }
+
+        // Add all remaining commitments as last partition
+        state.add_partition(commitments.len() - 1, PartitionReason::Finish)?;
```
eyusufatik, line 363: index gaps here?

crates/batch-prover/src/prover.rs
```diff
@@ -0,0 +1,840 @@
...
+        let proving_jobs = self
+            .prover_service
+            .start_session_recovery()
+            .expect("Failed to start proving session recovery");
+        let mut proving_jobs = proving_jobs
+            .into_iter()
+            .map(|rx| async move { rx.await.expect("Proof recovery channel closed abruptly") })
+            .collect::<FuturesUnordered<_>>();
+
+        info!("Recovering {} proving sessions", proving_jobs.len());
```
eyusufatik, line 505: log which jobs will be recovered

crates/batch-prover/src/rpc.rs
```diff
@@ -32,253 +29,231 @@ pub struct ProverInputResponse {
...
 #[rpc(client, server, namespace = "batchProver")]
 pub trait BatchProverRpc {
-    /// Generate state transition data for the given L1 block height, and return the data as a borsh serialized hex string.
-    #[method(name = "generateInput")]
-    async fn generate_input(
-        &self,
-        l1_height: u64,
-        group_commitments: Option<GroupCommitments>,
-    ) -> RpcResult<Vec<ProverInputResponse>>;
+    /// Manually set commitments. It overrides the commitment already if exists, so use with caution.
```
eyusufatik, line 70: let's add send to da rpc

eyusufatik: input is serialized receipt in BatchProofOutput (hex)

should verify the proof by method id of spec of last_l2_height before sending.

eyusufatik: create issue and let's do this later.

bin/citrea/tests/bitcoin/batch_prover_test.rs
```diff
@@ -885,22 +923,41 @@ impl TestCase for L1HashOutputTest {
...
+            .await
+            .unwrap();
 
-        let zkp = wait_for_proving_finish(batch_prover, finalized_height, None).await?;
+        // Wait for proving job to start
+        let job_ids = wait_for_prover_job_count(batch_prover, 1, None)
+            .await
+            .unwrap();
+        assert_eq!(job_ids.len(), 1);
+        println!("JOB IDS {:?}", job_ids);
```
eyusufatik: prints

eyusufatik (changes requested):

eyusufatik (approved):

PR #2150, opened 2025-03-21
Implement new batch proof verification logic in light client

ercecan:
# Description
At the current state of the pr, if there are missing commitments of a batch proof, info about those batch proofs are added to light client output for later verification as well
## Linked Issues
- Fixes #2020

PR #2180, opened 2025-03-27
New batch proof verification and chaining

### LINKED ISSUE #2020: Update batch proof verification logic in light client proof
here's my rusty-python pseudo code.

```python
verify_seq_comm_relation(proof):
    if jmt[proof.last_l1_hash] is null:
        return ProofProcess::Discard

    ret = ProofProcess::Correct
	for i in proof.seq_comm_range:
          
          seq_comm_hash = jmt[i] if jmt[i] is not null else:
            if ret == ProofProcess::Correct
                  ret = ProofProcess::MissingSeqComm([])
            ret.inner().push((i, proof.seq_comm_hashes[i]))
          if seq_comm_hash == proof.seq_comm_hashes[i]:
            continue
          else:
            return ProofProcess::Discard              
    return ret



for data in da:
    match data:
        SeqComm(seq_comm) => {
            if jmt[seq_comm.index] is null:
                jmt[seq_comm.index] = hash(seq_comm)

              for p in lcp.semi_unstitched:
                    for tup in p.tup:
                        if seq_comm.index == tup.0 && hash(seq_comm) == tup.1:
                            p.tup.remove(tup)
                            if p.tup.is_empty():
                                lcp.semi_unstitched.remove(p)
                                lcp.full_unstitched.push((
                                    p.1,
                                    p.2,
                                    p.3
                                ))
        }
        Proof(proof) => {
             if verify_zk(proof):
                match verify_seq_comm_relation(proof):
                    ProofProcess::Correct => {
                        if proof.seq_comm_range.0 == lcp.last_seq_comm_index + 1  && proof.prev_root == lcp.last_root:
                            lcp.last_seq_comm_index = proof.seq_comm_range.1
                            lcp.last_root = proof.final_root
                        elif proof.seq_comm_range.0 == lcp.last_seq_comm_index + 1:
                            # processes next seq comm, seq comms are known, but prev root is not same????
                            DISCARD PROOF
                        elif proof.prev_root == lcp.last_root:
                            # prev roots match, seq comms are known, but not next index????
                            DISCARD PROOF
                        else:
                            # both roots and indices do not match, missing proof
                            lcp.full_unstitched.push((
                                proof.prev_root,
                                proof.next_root,
                                proof.seq_comm_range
                            ))
                    }
                    ProofProcess::MissingSeqComm(tup) => {
                        lcp.semi_unstitched.push((
                            tup,
                            proof.prev_root,
                            proof.next_root,
                            proof.seq_comm_range
                        ))
                    }
                    ProofProcess::Discard => _
        }
   
m
try_stitch_unstitched(lcp, full_unstitched)

# prune should prune by looking at seq_comm_ranges
# and lcp.last_proven_seq_comm_index

# remove entry if lcp.last_proven_seq_comm_index >= item.2
prune(lcp.full_unstitched)

# remove entry if lcp.last_proven_seq_comm_index >= item.3
prune(lcp.semi_unstitched)
```

A batch proof verification is seperated into two parts.

1. ZK verification
   - You parse output, figure out what method id to use.
   - Verify the SNARK
2. Seqeuncer commitment relationship
   - From the batch proof output you take sequencer commimtment index range
   - For the range outputted for sequencer commitment indices, LCP qeuriest its state
     - If there is even one hash mismatch, we discard proof
     - If there is even one unknown sequencer commitment index (LCP got batch proof before sequencer commitment), we push batch proof to not "fully verified unchained batch proofs"

ercecan:
# Description
With this pr the **output size of the light client proof** will be **constant**
supersedes #2150 
Issue updated couple times, rewrite made more sense
**Caveat**
There are some extra stuff done int this pr
- The sequencer [commitment](https://github.com/chainwayxyz/citrea/pull/2180) index starts from 1 instead of 0
- The batch proof output has a new filed called `state_roots` and initial and final state roots are removed, this is a vector with the first element being the initial state root, meaning the previous commitments state root to the state root of the last commitment it has in its commitment range.
So If a batch proof has a commitment range of `[2..4]` the state roots vec will be sth like: 
```rust
[ [1u8; 32], [2u8; 32], [3u8;32], [4u8; 32] ]
```
With this we can now chain proofs with index range like the following two:
```[2..5] + [4..7] => 7``` (7 is the last commitment index of the lcp)
before this the output would be just 5 would be the last commitment index of lcp and the  other batch proof would have been ignored
## Linked Issues
- Fixes #2020

### CODE REVIEW

File: crates/sovereign-sdk/rollup-interface/src/state_machine/zk/light_client_proof/output.rs [hunk 1]
```diff
@@ -17,19 +17,21 @@ pub struct LightClientCircuitOutput {
     /// Latest da state output of the previous light client proof
     /// If None, initial hardcoded da block will be used for verification
     pub latest_da_state: LatestDaState,
-    /// Batch proof info from current or previous light client proofs that were not changed and unable to update the state root yet
-    pub unchained_batch_proofs_info: Vec<BatchProofInfo>,
     /// Last l2 height the light client proof verifies
     pub last_l2_height: u64,
     /// L2 activation height of the fork and the Method ids of the batch proofs that were verified in the light client proof
+    // TODO: Move to jmt
```
eyusufatik: very important to do this in this pr

crates/sovereign-sdk/rollup-interface/src/node/rpc/mod.rs
```diff
@@ -130,8 +130,8 @@ impl From<MMRGuest> for MMRGuestRpcResponse {
     }
 }
 
-impl From<BatchProofInfo> for BatchProofInfoRpcResponse {
-    fn from(info: BatchProofInfo) -> Self {
+impl From<SequencerCommitmentInfo> for BatchProofInfoRpcResponse {
```
eyusufatik: I'm not sure if we should call this `SequencerCommitmentInfo`

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/batch_proof/output/v3.rs
```diff
@@ -10,10 +10,10 @@ use crate::zk::StorageRootHash;
...
 #[derive(Debug, Serialize, Deserialize, BorshSerialize, BorshDeserialize)]
 pub struct BatchProofCircuitOutputV3 {
-    /// The state of the rollup before the transition
-    pub initial_state_root: StorageRootHash,
-    /// The state of the rollup after the transition
-    pub final_state_root: StorageRootHash,
+    /// All of the state roots of commitments from initial state (previous commitments state root) to the last sequencer commitment
+    /// First is initial state root, last is the final state root of batch proof
+    /// If there is no previous commitment the first element of the vector will be genesis state root
+    pub state_roots: Vec<StorageRootHash>,
```
eyusufatik, line 16: pls verify in some test that these actually match the l2 state roots

crates/light-client-prover/src/circuit/accessors.rs
```diff
@@ -121,6 +122,48 @@ impl<S: Storage> SequencerCommitmentAccessor<S> {
     }
 }
 
+pub struct SequencerCommitmentInfoAccessor<S: Storage> {
+    phantom: core::marker::PhantomData<S>,
+}
+
+impl<S: Storage> SequencerCommitmentInfoAccessor<S> {
```
eyusufatik: implement unit tests.

crates/light-client-prover/src/circuit/mod.rs
```diff
@@ -298,21 +428,22 @@ impl<S: Storage, DS: DaSpec, Z: Zkvm> LightClientProofCircuit<S, DS, Z> {
...
+        while let Some(sequencer_commitment_info) = SequencerCommitmentInfoAccessor::<S>::get(
+            last_sequencer_commitment_index + 1,
+            &mut working_set,
+        ) {
+            if sequencer_commitment_info.initial_state_root == last_l2_state_root {
+                last_l2_state_root = sequencer_commitment_info.final_state_root;
+                last_l2_height = sequencer_commitment_info.last_l2_height;
+                last_sequencer_commitment_index += 1;
+            } else {
+                break;
```
eyusufatik: if this happens maybe we should panic, as this implies a bug in our implementation?

eyusufatik: not sure though

crates/light-client-prover/src/circuit/mod.rs
```diff
@@ -58,12 +55,132 @@ impl<S: Storage, DS: DaSpec, Z: Zkvm> LightClientProofCircuit<S, DS, Z> {
...
+            _ => {
+                // If there are no previous commitments then this should be the first batch proof
+                // The first batch proof's first commitment index should be 1
+                if batch_proof_output.sequencer_commitment_index_range().0 != 1 {
+                    println!(
+                        "Previous commitment index is not set, but sequencer commitment index range start is not 1: {}",
+                        batch_proof_output.sequencer_commitment_index_range().0
+                    );
+                    return false;
+                }
```
eyusufatik, line 114: this is unreachable right? as we assert this in batch proof.

if we don't assert this in batch proof, let's do it.

also this can stay the way it is in case we miss something this is extra protection.

crates/light-client-prover/src/circuit/mod.rs
```diff
@@ -58,12 +55,132 @@ impl<S: Storage, DS: DaSpec, Z: Zkvm> LightClientProofCircuit<S, DS, Z> {
...
+            if i as u32 == last_index - first_index
+                && jmt_commitment.l2_end_block_number != batch_proof_output.last_l2_height()
+            {
+                println!(
+                    "Last sequencer commitment l2 height mismatch, expected: {}, got: {}",
+                    jmt_commitment.l2_end_block_number,
+                    batch_proof_output.last_l2_height()
+                );
+                return false;
+            }
```
eyusufatik: this is also unreachble, as hashes wouldn't match otherwise, add comment explaning pls. code can leave as is in case there are any bugs in batch proof

crates/light-client-prover/src/circuit/mod.rs
```diff
@@ -107,14 +227,36 @@ impl<S: Storage, DS: DaSpec, Z: Zkvm> LightClientProofCircuit<S, DS, Z> {
...
+        }
+
+        for (idx, seq_comm_index) in (batch_proof_output.sequencer_commitment_index_range().0
+            ..=batch_proof_output.sequencer_commitment_index_range().1)
+            .enumerate()
+        {
+            if seq_comm_index <= last_sequencer_commitment_index
+                || SequencerCommitmentInfoAccessor::<S>::get(seq_comm_index, working_set).is_some()
+            {
+                continue;
```
eyusufatik, line 255: add comment explaining this decision

crates/light-client-prover/src/circuit/mod.rs
eyusufatik: maybe we could even remove this "sequencer commitment info" at this point, we should think about that

eyusufatik (changes requested):
leaving initial review. now going to review tests seperately.

crates/light-client-prover/src/tests/test_utils.rs
```diff
@@ -63,27 +105,43 @@ pub(crate) fn create_mock_batch_proof(
...
+    let prev_index = if sequencer_commitments[0].index == 1 {
+        None
+    } else {
+        Some(sequencer_commitments[0].index - 1)
+    };
+
+    let mut state_roots = vec![initial_state_root];
+
+    // For the sake of easiness of impl tests, we can use merkle root as state root
+    state_roots.extend(sequencer_commitments.iter().map(|c| c.merkle_root));
```
eyusufatik, line 130: should we???? are we sure this is not going to cause some false tests somehow?

bin/citrea/tests/bitcoin/light_client_test.rs
```diff
@@ -811,14 +796,93 @@ impl TestCase for LightClientUnverifiableBatchProofTest {
...
+
+        let _ = bitcoin_da_service
+            .send_transaction_with_fee_rate(
+                DaTxRequest::SequencerCommitment(fake_sequencer_commitment_2.clone()),
+                1,
+            )
+            .await
+            .unwrap();
+
+        da.wait_mempool_len(10, None).await?;
```
jfldde: Any reason for the multiple `wait_mempool_len` ?

crates/light-client-prover/src/tests/mod.rs
```diff
@@ -218,23 +521,37 @@ fn create_unchainable_outputs_then_chain_them_on_next_block() {
...
-    // Fin state root
+    let unchained_info3 =
+        SequencerCommitmentInfoAccessor::<ProverStorage>::get(3, &mut working_set).unwrap();
+    assert_eq!(unchained_info3.initial_state_root, seq_comm_2.merkle_root);
     assert_eq!(
-        output_1.unchained_batch_proofs_info[0].final_state_root,
-        [4; 32]
+        unchained_info3.last_l2_height,
+        seq_comm_3.l2_end_block_number
     );
```
eyusufatik: assert last_commitment_index here

crates/light-client-prover/src/tests/mod.rs
```diff
@@ -269,9 +586,7 @@ fn create_unchainable_outputs_then_chain_them_on_next_block() {
         .unwrap();
 
     // Check that the state transition actually happened from 1-4 now
-
     assert_eq!(output_2.l2_state_root, [4; 32]);
-    assert!(output_2.unchained_batch_proofs_info.is_empty());
     assert_eq!(output_2.last_l2_height, 4);
```
eyusufatik, line 603: assert last_seq_comm_index here

crates/light-client-prover/src/tests/mod.rs
```diff
@@ -417,9 +791,7 @@ fn test_unverifiable_batch_proofs() {
     // Check that the state transition actually happened but only for verified batch proof
     // and assert the unverified is ignored, so it is not even in the unchained outputs
     assert_eq!(output_1.l2_state_root, [2; 32]);
-    assert!(output_1.unchained_batch_proofs_info.is_empty());
     assert_eq!(output_1.last_l2_height, 2);
-    assert_eq!(output_1.unchained_batch_proofs_info.len(), 0);
 }
```
eyusufatik: assert we didn't save any `SequencerCommitmentInfo` for unverified batch proof

crates/light-client-prover/src/tests/mod.rs
```diff
@@ -710,13 +1130,18 @@ fn test_light_client_circuit_verify_chunks() {
...
+    let seq_comm_1_blob = create_mock_sequencer_commitment_blob(seq_comm_1.clone());
+
     let serialized_mock_proof = create_serialized_mock_proof(
         l2_genesis_state_root,
-        [2u8; 32],
         101,
         true,
         Some(state_diff),
         block_header_1.hash.0,
+        vec![seq_comm_1],
```
eyusufatik: can you use multiple seq comms for this aggregate proof?

crates/light-client-prover/src/circuit/mod.rs
ercecan: yes we assert that in batch proof, leaving it as is

bin/citrea/tests/bitcoin/light_client_test.rs
```diff
@@ -811,14 +796,93 @@ impl TestCase for LightClientUnverifiableBatchProofTest {
...
+        let _ = bitcoin_da_service
+            .send_transaction_with_fee_rate(
+                DaTxRequest::SequencerCommitment(fake_sequencer_commitment.clone()),
+                1,
+            )
+            .await
+            .unwrap();
+        da.wait_mempool_len(2, None).await?;
+
+        let fake_sequencer_commitment_5 = SequencerCommitment {
```
eyusufatik: fake_sequencer_commitment_2?

bin/citrea/tests/bitcoin/light_client_test.rs
```diff
@@ -811,14 +796,93 @@ impl TestCase for LightClientUnverifiableBatchProofTest {
...
+        let _ = bitcoin_da_service
+            .send_transaction_with_fee_rate(
+                DaTxRequest::SequencerCommitment(fake_sequencer_commitment_3.clone()),
+                1,
+            )
+            .await
+            .unwrap();
+        da.wait_mempool_len(6, None).await?;
+
+        let fake_sequencer_commitment_4 = SequencerCommitment {
```
eyusufatik, line 792: why are all namings out of place?

bin/citrea/tests/bitcoin/light_client_test.rs
```diff
@@ -1661,68 +1831,480 @@ async fn test_unknown_l1_hash_batch_proof_in_light_client() -> Result<()> {
...
+            .send_transaction_with_fee_rate(DaTxRequest::ZKProof(bp), 1)
+            .await
+            .unwrap();
+
+        da.wait_mempool_len(4, None).await?;
+
+        da.generate(FINALITY_DEPTH).await?;
+
+        // Make sure all of them are in the block
+        da.wait_mempool_len(0, Some(TEN_MINS)).await?;
```
jfldde:
```suggestion
      let mempool = da.get_raw_mempool().await?;
      assert!(mempool.is_empty());
```

bin/citrea/tests/bitcoin/light_client_test.rs
ercecan: one of them is for commitments and the other is for batch proofs, they are in different blocks

bin/citrea/tests/bitcoin/light_client_test.rs
```diff
@@ -1661,68 +1831,480 @@ async fn test_unknown_l1_hash_batch_proof_in_light_client() -> Result<()> {
...
+            .send_transaction_with_fee_rate(DaTxRequest::ZKProof(bp), 1)
+            .await
+            .unwrap();
+
+        da.wait_mempool_len(2, None).await?;
+
+        da.generate(FINALITY_DEPTH).await?;
+
+        // Make sure all of them are in the block
+        da.wait_mempool_len(0, Some(TEN_MINS)).await?;
```
jfldde: Same as above. Issue is that `wait_mempool_len` wait for the mempool to be at least this size, so it will always pass the condition on 0 even if it still has pending txs

eyusufatik (changes requested):
reviewed tests

crates/light-client-prover/src/circuit/mod.rs
ercecan: I believe panic is safer

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/light_client_proof/output.rs
eyusufatik: well we can do it in a seperate pr but still open an issue if you wanna go that way.

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/batch_proof/output/v3.rs
ercecan: I already do that in a bunch of tests
I just call functions like final_state_root() or initial_state_root() which basically gets the data from the vec

crates/sovereign-sdk/rollup-interface/src/node/rpc/mod.rs
ercecan: What do u have in mind?

eyusufatik:
In general we are asserting very few lcp_output.last_sequencer_commitment_index  's so we should do that

crates/light-client-prover/src/tests/test_utils.rs
ercecan: Not sure how is that possible, that hack is only used in unit tests and fake commitments

crates/light-client-prover/src/circuit/mod.rs
eyusufatik: let's not do it right away and discuss in the office tomorrow morning

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/batch_proof/output/v3.rs
eyusufatik: no verify intermediary state roots also match. 

for instance, in basic prover test, we use two seq comms in a single batch proof, we can assert all 3 of the state roots match the actual values.

crates/sovereign-sdk/rollup-interface/src/node/rpc/mod.rs
eyusufatik: VerifiedStateTransitionForSequencerCommitmentIndex

bin/citrea/tests/bitcoin/light_client_test.rs
jfldde: I wasn't clear. We wait for 2, then 4, then 6, etc. Wonder if there's a reason for us not to do a single wait at the end ?

eyusufatik:
also we haven't thought about if a batch proof is allowed to prove for a sequencer commitment which does not include any blocks from it's fork. let's talk about that tomorrow.

crates/light-client-prover/src/circuit/accessors.rs
```diff
@@ -121,16 +126,115 @@ impl<S: Storage> SequencerCommitmentAccessor<S> {
...
+    phantom: core::marker::PhantomData<S>,
+}
+
+impl<S: Storage> BatchProofMethodIdAccessor<S> {
+    const PREFIX: u8 = b'm';
+    const KEY: u8 = b'm';
+
+    fn key() -> StorageKey {
+        // use `StorageKey::singleton_owned` as a hack to create no serialization key
+        let mut key = [0u8; 2]; // 1 prefix + 4 bytes
```
jfldde:
```suggestion
        let mut key = [0u8; 2]; // 1 prefix + 1 byte
```

crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
```diff
@@ -384,8 +384,8 @@ where
             } else {
                 // If this is the first batch proof, then the first commitment idx should be 0
```
jfldde:
```suggestion
                // If this is the first batch proof, then the first commitment idx should be 1
```

crates/light-client-prover/src/circuit/mod.rs
ercecan: Discussed and decided to panic also added explanation

bin/citrea/tests/bitcoin/light_client_test.rs
ercecan: you're right will remove

crates/light-client-prover/src/circuit/accessors.rs
```diff
@@ -121,16 +126,119 @@ impl<S: Storage> SequencerCommitmentAccessor<S> {
...
+    }
+}
+
+pub struct BatchProofMethodIdAccessor<S: Storage> {
+    phantom: core::marker::PhantomData<S>,
+}
+
+impl<S: Storage> BatchProofMethodIdAccessor<S> {
+    const PREFIX: u8 = b'm';
+    const KEY: u8 = b'm';
```
eyusufatik: can just use "prefix" as key don't need two bytes

crates/light-client-prover/src/circuit/accessors.rs
```diff
@@ -121,16 +126,119 @@ impl<S: Storage> SequencerCommitmentAccessor<S> {
...
+        working_set: &mut WorkingSet<S>,
+    ) {
+        let key = Self::key();
+        let mut method_ids = Self::get(working_set).unwrap_or_default();
+        method_ids.extend(initial_batch_proof_method_ids);
+
+        let value: StorageValue = borsh::to_vec(&method_ids)
+            .expect("Batch proof method ids serialization should not fail")
+            .into();
+        working_set.set(&key, value);
```
eyusufatik: just set why get extend set?

bin/citrea/tests/bitcoin/light_client_test.rs
```diff
@@ -609,10 +631,15 @@ impl TestCase for LightClientBatchProofMethodIdUpdateTest {
...
             .get_light_client_proof_by_l1_height(method_id_l1_height)
             .await?;
-        let lcp_output = lcp.unwrap().light_client_proof_output;
+        let _lcp_output = lcp.unwrap().light_client_proof_output;
+        let batch_proof_method_ids = light_client_prover
+            .client
+            .http_client()
+            .get_batch_proof_method_ids()
+            .await?;
         // Verify the current batch proof method ids
```
eyusufatik: Need to double check here

bin/citrea/tests/bitcoin/light_client_test.rs
```diff
@@ -790,30 +823,30 @@ impl TestCase for LightClientUnverifiableBatchProofTest {
...
-        };
 
-        let _ = bitcoin_da_service
-            .send_transaction_with_fee_rate(
-                DaTxRequest::SequencerCommitment(fake_sequencer_commitment_2.clone()),
-                1,
-            )
-            .await
-            .unwrap();
+        // let fake_sequencer_commitment_5 = SequencerCommitment {
```
eyusufatik: here as well

eyusufatik (changes requested):

eyusufatik (approved):

PR #2185, opened 2025-03-28
[Sub PR] More revm/reth updates

### LINKED ISSUE #2174: Implement EIP-2537
I will benchmark according to example inputs in the EIP. 

Then we can decide if we are going to include the eip or not.

kpp:
# Description

Another incremental update of revm and reth

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -144,8 +144,7 @@ reth-rpc = { git = "https://github.com/paradigmxyz/reth", tag = "v1.3.7", defaul
 reth-stages = { git = "https://github.com/paradigmxyz/reth", tag = "v1.3.7", default-features = false }
 
 revm = { version = "21.0.0", features = ["serde"], default-features = false }
-# TODO: figure out how to use substrate-bn instead of arkbn254
-revm-precompile = { version = "18.0.0", default-features = false, features = ['secp256r1', 'blst'] }
+revm-precompile = { version = "18.0.0", default-features = false, features = ['secp256r1', 'blst', 'bn'] }
 revm-primitives = { version = "17.0.0", default-features = false }
 revm-bytecode = { version = "2.0.0", default-features = false }
 alloy-genesis = { version = "0.13", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -103,7 +103,7 @@ serde = { version = "1", default-features = false, features = ["alloc", "derive"
 serde_json = { version = "1", default-features = false, features = ["alloc"] }
 sha2 = { version = "0.10.8", default-features = false, features = ["asm"] }
 schemars = { version = "0.8.16", features = ["derive"] }
-secp256k1 = { version = "0.29.0", default-features = false, features = ["global-context", "recovery"] }
+secp256k1 = { version = "0.30.0", default-features = false, features = ["global-context", "recovery"] }
 thiserror = { version = "2", default-features = false }
 tinyvec = { version = "1", default-features = false, features = ["alloc", "rustc_1_61", "rustc_1_57", "rustc_1_55"] }
 tracing = { version = "0.1.41", default-features = false, features = ["attributes"] }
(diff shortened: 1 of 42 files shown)


eyusufatik:
https://github.com/chainwayxyz/citrea/issues/2205

Cargo.toml
```diff
@@ -103,7 +103,7 @@ serde = { version = "1", default-features = false, features = ["alloc", "derive"
 serde_json = { version = "1", default-features = false, features = ["alloc"] }
 sha2 = { version = "0.10.8", default-features = false, features = ["asm"] }
 schemars = { version = "0.8.16", features = ["derive"] }
-secp256k1 = { version = "0.29.0", default-features = false, features = ["global-context", "recovery"] }
+secp256k1 = { version = "0.30.0", default-features = false, features = ["global-context", "recovery"] }
```
eyusufatik, line 106: why

crates/evm/src/evm/handler.rs
```diff
@@ -150,328 +161,454 @@ impl CitreaExternalExt for CitreaExternal {
...
+    fn contains(&self, address: &Address) -> bool {
+        self.inner.contains(address)
+    }
 }
 
-pub(crate) struct CitreaHandler<SPEC, EXT, DB> {
-    _phantom: std::marker::PhantomData<(SPEC, EXT, DB)>,
+impl Default for CitreaPrecompiles {
+    fn default() -> Self {
+        Self::new_with_spec(SpecId::CANCUN)
```
eyusufatik: no prague

eyusufatik (changes requested):

crates/evm/src/evm/call.rs
```diff
@@ -2,19 +2,33 @@
...
-use revm::context::{BlockEnv, CfgEnv, TransactionType, TxEnv};
+use revm::context::{BlockEnv, CfgEnv, TxEnv};
 
 use crate::caller_gas_allowance;
 
 pub(crate) fn create_txn_env(
     block_env: &BlockEnv,
     request: TransactionRequest,
     cap_to_balance: Option<U256>,
+    nonce_if_req_has_no_nonce: Option<u64>,
```
kpp, line 18: Why make it optional instead of just u64?

Cargo.toml
kpp: Because alloy PK requires v0.30

eyusufatik: that's actually a big problem as it probably means our ecrecovers on ethereum txs might be using secp256k1 instead of the accelerated k256 crate. we need to get rid of that.

eyusufatik: looks like we do ecrecover through reth-primitives-traits and it's secp256k1 feature is not enabled so it's good.

guests/risc0/batch-proof/bitcoin/Cargo.lock
```diff
@@ -293,12 +378,48 @@ dependencies = [
...
 ]
 
+[[package]]
+name = "android-tzdata"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
+
+[[package]]
+name = "android_system_properties"
```
eyusufatik, line 388: wth

guests/risc0/batch-proof/bitcoin/Cargo.lock
```diff
@@ -791,7 +918,7 @@ dependencies = [
  "bitcoin_hashes",
  "hex-conservative",
  "hex_lit",
- "secp256k1",
+ "secp256k1 0.29.1",
```
eyusufatik, line 921: let's also upgrade bitcoin crate?

crates/primitives/src/basefee.rs
```diff
@@ -7,9 +7,9 @@ pub fn calculate_next_block_base_fee(
     gas_limit: u64,
     base_fee: u64,
     base_fee_params: BaseFeeParams,
-) -> u128 {
+) -> u64 {
     std::cmp::max(
-        MIN_BASE_FEE_PER_GAS,
-        calc_next_block_base_fee(gas_used, gas_limit, base_fee, base_fee_params) as u128,
+        MIN_BASE_FEE_PER_GAS as u64,
```
eyusufatik: update const type

crates/evm/src/lib.rs
```diff
@@ -100,7 +101,7 @@ pub struct Evm<C: sov_modules_api::Context> {
     /// Mapping from code hash to code. Used for lazy-loading code into a contract account.
     #[state(rename = "c")]
     pub(crate) offchain_code:
-        sov_modules_api::OffchainStateMap<B256, revm::primitives::Bytecode, BcsCodec>,
+        sov_modules_api::OffchainStateMap<B256, revm::state::Bytecode, BcsCodec>,
```
eyusufatik, line 104: when hoevered over state it says optimism specific constants types and helpers. need to check whats going on.

crates/evm/src/evm/executor.rs
```diff
@@ -3,37 +3,43 @@ use alloy_eips::Typed2718;
...
     /// Creates a new Citrea EVM with the given parameters.
-    pub fn new(db: DB, block_env: BlockEnv, config_env: CfgEnvWithHandlerCfg, ext: EXT) -> Self {
-        let evm_env = Env::boxed(config_env.cfg_env, block_env, Default::default());
-        let evm_context = EvmContext::new_with_env(db, evm_env);
-        let context = Context::new(evm_context, ext);
-        let handler = citrea_handler(config_env.handler_cfg);
-        let evm = revm::Evm::new(context, handler);
+    pub fn new(db: DB, block_env: BlockEnv, config_env: CfgEnv, ext: &'a mut CitreaChain) -> Self {
+        let mut journal = Journal::<DB>::new(db);
+        journal.set_spec_id(config_env.spec().into());
```
eyusufatik: why do we need to set explicitly here?

crates/evm/src/evm/executor.rs
```diff
@@ -3,37 +3,43 @@ use alloy_eips::Typed2718;
...
+        let evm = Context {
+            block: block_env,
+            cfg: config_env,
+            chain: ext,
+            tx: Default::default(),
```
eyusufatik, line 38: huh?

crates/evm/src/evm/handler.rs
```diff
@@ -150,328 +161,454 @@ impl CitreaExternalExt for CitreaExternal {
...
+
+/// Type alias for the error type of the CitreaEvm.
+type CitreaError<CTX> =
+    EVMError<<<CTX as ContextTr>::Db as Database>::Error /*CitreaTransactionError*/>;
+
+impl<CTX, INSP> ExecuteEvm for CitreaEvm<CTX, INSP>
+where
+    CTX: CitreaContextTr + ContextSetters,
+{
+    type Output = Result<ResultAndState<HaltReason /*TODO CitreaHaltReason */>, CitreaError<CTX>>;
```
eyusufatik, line 262: why do you think we need CitreaHaltReason?

crates/evm/src/evm/handler.rs
```diff
@@ -150,328 +161,454 @@ impl CitreaExternalExt for CitreaExternal {
...
+/// Returns precompiles.
+pub fn citrea_precompiles() -> &'static Precompiles {
+    static INSTANCE: OnceBox<Precompiles> = OnceBox::new();
+    INSTANCE.get_or_init(|| {
+        // Berlin because:
+        // 1. POINT_EVALUATION precompile(0x0A) is enabled in Cancun
+        // 2. BLS12_381 precompiles (0x0b..0x11) are enabled in Prague
+        // TODO: add bls_12381 precompiles https://github.com/chainwayxyz/citrea/issues/2174
+
+        let mut precompiles = Precompiles::berlin().clone();
```
eyusufatik, line 365: I think using what spec citrea is on then removing is less error prone

crates/evm/src/evm/handler.rs
```diff
@@ -150,328 +161,454 @@ impl CitreaExternalExt for CitreaExternal {
...
+#[cfg(feature = "native")]
+impl<EVM, ERROR, FRAME> InspectorHandler for CitreaHandler<EVM, ERROR, FRAME>
+where
+    EVM: InspectorEvmTr<
+        Context: CitreaContextTr,
+        Inspector: Inspector<<<Self as Handler>::Evm as EvmTr>::Context, EthInterpreter>,
+    >,
+    ERROR: EvmTrError<EVM> + FromStringError,
+    // TODO `FrameResult` should be a generic trait.
+    // TODO `FrameInit` should be a generic.
```
eyusufatik, line 595: how important are these todos

crates/evm/src/rpc_helpers/tracing_utils.rs
```diff
@@ -266,122 +243,92 @@ pub(crate) fn trace_transaction<C: sov_modules_api::Context>(
...
 
-pub(crate) fn inspect_with_citrea_handle<'a, C, I>(
+pub(crate) fn inspect_with_citrea_handler<'a, C, I>(
     db: EvmDb<'a, C>,
-    config_env: CfgEnvWithHandlerCfg,
+    config_env: CfgEnv,
     block_env: BlockEnv,
     tx_env: TxEnv,
-    ext: &mut I,
+    l1_fee_rate: u128,
```
eyusufatik, line 305: for this function l1_fee_rate should always be zero

crates/evm/Cargo.toml
```diff
@@ -22,6 +22,7 @@ citrea-primitives = { path = "../primitives" }
 borsh = { workspace = true, features = ["rc"] }
 hex = { workspace = true, optional = true }
 jsonrpsee = { workspace = true, features = ["macros", "client-core", "server"], optional = true }
+once_cell = { workspace = true }
```
eyusufatik, line 25: not used

eyusufatik (changes requested):

crates/evm/src/evm/executor.rs
eyusufatik: I mean the tx part

crates/evm/src/rpc_helpers/tracing_utils.rs
eyusufatik: it is the case already

crates/evm/src/evm/handler.rs
eyusufatik: I mean it should be

crates/evm/Cargo.toml
eyusufatik: it's used

guests/risc0/batch-proof/bitcoin/Cargo.lock
kpp: Yes we can do that :ok:

crates/evm/src/evm/executor.rs
kpp: Because it's basically this code:

```
    pub fn with_db<ODB: Database>(
        self,
        db: ODB,
    ) -> Context<BLOCK, TX, CFG, ODB, Journal<ODB>, CHAIN> {
        let spec = self.cfg.spec().into();
        let mut journaled_state = Journal::new(db);
        journaled_state.set_spec_id(spec);
        Context {
            tx: self.tx,
            block: self.block,
            cfg: self.cfg,
            journaled_state,
            chain: self.chain,
            error: Ok(()),
        }
    }
```

crates/evm/src/evm/handler.rs
kpp: Well we can do that instead of

```
                return Err(ERROR::from_string(format!(
                    "Not enough funds for L1 fee: {}",
                    l1_fee
                )));
```

But that's optional.

crates/evm/src/evm/executor.rs
kpp: tx is of `TxEnv` type. It is changed each time we execute tx in 

```
    pub(crate) fn transact(
        &mut self,
        tx: &Recovered<TransactionSigned>,
    ) -> Result<ResultAndState, EVMError<DB::Error>> {
        self.evm.ctx().chain().set_current_tx_hash(tx.hash());
        self.evm.transact(create_tx_env(tx))
    }
```

crates/evm/src/evm/handler.rs
kpp: They are not. That's a ^C^V from revm

guests/risc0/batch-proof/bitcoin/Cargo.lock
kpp: That's ok. it exists in cargo.lock only, compiles only on android.

eyusufatik (approved):

PR #2192, opened 2025-04-02
Fix lack of precompiles in simulation queries

### LINKED ISSUE #2183: `eth_call` does not have same custom precompiles
## Summary
As we don't append revm citrea handler to revm inside eth_call's we don't have the same precompile set as normal execution.

eyusufatik:
# Description
On endpoints such as 

- eth_call
- eth_createAccessList
- eth_estimateGas

Uses citrea evm handler so that everything works the same with normal transaction execution.

## Linked Issues
Fixes #2183 

## Testing

## Docs

### DIFF
--- a/crates/evm/src/evm/handler.rs
+++ b/crates/evm/src/evm/handler.rs
@@ -240,8 +240,7 @@ where
         self.inspector.create_end(context, inputs, outcome)
     }
     fn selfdestruct(&mut self, contract: Address, target: Address, value: U256) {
-        // why not  self.inspector.selfdestruct(contract, target, value) ?
-        (&mut self.inspector as &mut dyn Inspector<DB>).selfdestruct(contract, target, value)
+        self.inspector.selfdestruct(contract, target, value)
     }
 }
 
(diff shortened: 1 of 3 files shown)


crates/evm/src/query.rs
```diff
@@ -709,11 +712,14 @@ impl<C: sov_modules_api::Context> Evm<C> {
         // can consume the list since we're not using the request anymore
         let initial = request.access_list.take().unwrap_or_default();
 
+        // TODO: update here.
```
eyusufatik: todo

crates/evm/src/evm/handler.rs
```diff
@@ -240,6 +240,7 @@ where
         self.inspector.create_end(context, inputs, outcome)
     }
     fn selfdestruct(&mut self, contract: Address, target: Address, value: U256) {
+        // why not  self.inspector.selfdestruct(contract, target, value) ?
```
eyusufatik: @kpp

kpp: Do it if you can do that. It wasnt available on prev rust versions

kpp (approved):

PR #2197, opened 2025-04-03
Fix ungraceful shutdown issues during l2 block commits

### LINKED ISSUE #2179: Start L2 sync from min(state_db.version, native_db.version)
Instead of using ledger db for source as start_l2_height, we should save L2 block to ledger db first and then commit sate db and native db, but use these db's versions as start sync height. this would solve our ungraceful shut down problems.

exeokan:
# Description
Our db write order is:
StateDB -> NativeDB -> LedgerDB
Since the `l2_start_height` is based on the ledger db, an interrupt between stateDB write and ledgerDB write would cause the l2 block to be applied again, causing the panic "Prover storage version is corrupted"

With this PR:
For full nodes and batch provers, interrupts between db writes are handled in the `CitreaRollupBlueprint::sync_ledger_and_state_db`, if statedb.version == ledgerdb.version+1, rollbacks dbs to ledgerdb.version

## Linked Issues
- Fixes #2179


bin/citrea/src/rollup/mod.rs
```diff
@@ -142,6 +145,57 @@ pub trait CitreaRollupBlueprint: RollupBlueprint {
         )
     }
 
+    /// In case of an interrupt between l2 block commits of LedgerDB and StateDB,
+    /// this function will reapply the last l2 block to the stateDB.
+    /// Returns Err if the LedgerDB version > StateDB version + 1
```
rakanalh: The case described in the comment should NOT happen to begin with if the node was shutdown gracefully. I am not sure why we're doing this

eyusufatik: sometimes the node does not shutdown gracefully. this is still a problem we face rarely.

rakanalh: IMO, rollback might be a perfect use case when ledger DB and state are different where we rollback to `min(state_version - 1, last_l2_height_from_ledger)`

crates/common/src/l2.rs
```diff
@@ -120,12 +120,13 @@ pub async fn process_l2_block<Da: DaService, DB: SharedLedgerOps>(
...
     }
 
-    storage_manager.finalize_storage(l2_block_result.change_set);
-
     let tx_hashes = compute_tx_hashes::<DefaultContext>(&l2_block.txs, current_spec);
     let tx_bodies = if include_tx_body { tx_bodies } else { None };
 
     ledger_db.commit_l2_block(l2_block, tx_hashes, tx_bodies)?;
+    // Write to state/native db after the ledger db commit, otherwise an interrupt 
+    // after the finalize_storage might cause ledger db to miss the block
```
eyusufatik: if we change this here we should also change it in the sequencer btw. but I'm not sure if we should commit to ledger db before state / acc. state

bin/citrea/src/rollup/mod.rs
exeokan: I just pushed this, we need to perform a rollback for full nodes as the new `sync_ledger_and_state_db` can only be used if `include_tx_body` was `true`. I think we can get rid of `sync_ledger_and_state_db` and only execute rollback as it is a simpler solution. 

Btw I added a `pop_head_l2_block` method to ledger db but I can also use `citrea_storage_ops::rollback`

eyusufatik (approved):

bin/citrea/src/rollup/mod.rs
exeokan: reverted changes on the db write order, now using `citrea_storage_ops::rollback` to rollback state/native db to ledger db version

PR #2201, opened 2025-04-04
Call resubmit commitments when the service starts

### LINKED ISSUE #2200: Call resubmit_pending_commitments inside commitment service
It blocks the sequencer main loop. should happen inside the service.

rakanalh:
# Description

## Linked Issues
- Fixes #2200

### DIFF
--- a/crates/sequencer/src/commitment/service.rs
+++ b/crates/sequencer/src/commitment/service.rs
@@ -6,12 +6,14 @@ use std::time::{Duration, Instant};
 use anyhow::anyhow;
 use citrea_evm::{get_last_l1_height_in_light_client, Evm};
 use citrea_primitives::forks::get_fork2_activation_height_non_zero;
+use citrea_primitives::types::L2BlockHash;
 use citrea_stf::runtime::DefaultContext;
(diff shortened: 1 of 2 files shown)


crates/sequencer/src/commitment/service.rs
```diff
@@ -62,7 +64,22 @@ where
...
     }
 
-    pub async fn run(mut self, cancellation_token: CancellationToken) {
+    pub async fn run(
+        mut self,
+        storage_manager: ProverStorageManager,
+        l2_block_hash: L2BlockHash,
+        cancellation_token: CancellationToken,
+    ) {
+        if l2_block_hash != [0; 32] {
```
eyusufatik, line 73: btw why are we not calling this function on initial startup? cant remember the reason

rakanalh: The fact that it requires storage manager and `init_params` which is generated from `genesis` makes it a mess to have this work on startup.

eyusufatik: the way we init storage now, it shouldn't be a problem. are you sure?

eyusufatik (approved):

PR #2208, opened 2025-04-06
fix how we handle diff size calc. for JournaledState

eyusufatik:
# Description

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


eyusufatik:
no tests are affected as we don't have a test for diff size calculation on a tx like this:

eoa --> smart contract, set storage --> smart contract, set storage

kpp (approved):

PR #2216, opened 2025-04-07
lock backup manager before processing missed blocks

eyusufatik:
# Description

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/sequencer/src/runner.rs
+++ b/crates/sequencer/src/runner.rs
@@ -654,6 +654,8 @@ where
                 // The RPC from which the sender can be called is only registered for test mode. This means
                 // that evey though we check the receiver here, it'll never be "ready" to be consumed unless in test mode.
                 _ = self.l2_force_block_rx.recv(), if self.config.test_mode => {
+                    let _l2_lock = backup_manager.start_l2_processing().await;
+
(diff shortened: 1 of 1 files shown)


crates/sequencer/src/runner.rs
```diff
@@ -654,6 +654,8 @@ where
                 // The RPC from which the sender can be called is only registered for test mode. This means
                 // that evey though we check the receiver here, it'll never be "ready" to be consumed unless in test mode.
                 _ = self.l2_force_block_rx.recv(), if self.config.test_mode => {
+                    let _l2_lock = backup_manager.start_l2_processing().await;
```
jfldde, line 657: Might be better to lock only inside the `process_missed_da_blocks` after the `get_block_at` RPC resolves so that we're not stuck in the exponential_backoff

eyusufatik: if we get stuck in that backoff, block production already stopped. so should it matter?

jfldde: Not really, just mean that during this odd case where we get stuck here, we won't be able to backup.

jfldde (approved):

PR #2218, opened 2025-04-07
Fork2 evm compat

ercecan:
# Description

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/evm/src/evm/handler.rs
+++ b/crates/evm/src/evm/handler.rs
@@ -356,23 +356,25 @@ pub struct CitreaPrecompiles {
     inner: EthPrecompiles,
 }
 
-/// Returns precompiles.
-pub fn citrea_precompiles(spec: SpecId) -> &'static Precompiles {
+// Returns precompiles for Citrea's Cancun spec.
+pub fn cancun() -> &'static Precompiles {
     static INSTANCE: OnceBox<Precompiles> = OnceBox::new();
     INSTANCE.get_or_init(|| {
         // Berlin because POINT_EVALUATION precompile(0x0A) is enabled in Cancun
-        // then we add prague precompiles
-        // and then we add the rest of the precompiles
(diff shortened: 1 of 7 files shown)


crates/evm/src/evm/handler.rs
```diff
@@ -356,19 +357,21 @@ pub struct CitreaPrecompiles {
...
     INSTANCE.get_or_init(|| {
         // Berlin because POINT_EVALUATION precompile(0x0A) is enabled in Cancun
         // then we add prague precompiles
         // and then we add the rest of the precompiles
         let mut precompiles = Precompiles::berlin().clone();
 
-        // Add prague precompiles
-        // Effectively skipping kzg precompiles in Cancun
-        precompiles.extend(bls12_381::precompiles());
+        if let SpecId::PRAGUE = spec {
```
kpp: `if spec >= CANCUN { enable P256VERIFY, SCHNORRVERIFY }`
`if spec >= PRAGUE { enable bls12_381 }`

kpp (approved):

crates/evm/src/evm/handler.rs
eyusufatik: no this is correct

eyusufatik (changes requested):
as INSTANCES is static once cell, the precompiles are not going to be activated based on spec.

crates/evm/src/genesis.rs
```diff
@@ -202,6 +203,8 @@ impl<C: sov_modules_api::Context> Evm<C> {
 
         self.cfg.set(&chain_cfg, working_set);
 
+        let evm_spec = citrea_spec_id_to_evm_spec_id(sov_modules_api::SpecId::Genesis);
```
eyusufatik: if we can get what spec the block number 0 correlates to that's better

eyusufatik (approved):

PR #2219, opened 2025-04-08
Final calc_diff_size touches

### LINKED ISSUE #2215: add l1 fee overhead directly to diff size

eyusufatik:
# Description
Charges account indices correctly
Applies discount at the end saving us from losing 1 byte each time

## Linked Issues
Fixes #2215 

## Testing

## Docs


crates/evm/src/evm/handler.rs
```diff
@@ -728,21 +724,12 @@ where
                     continue;
                 }
 
+                // transferred cbtc causes account diff change on target
```
kpp:
```suggestion
                // transferred balance causes account diff change on target
```

crates/evm/src/evm/handler.rs
```diff
@@ -754,13 +741,14 @@ where
 
     let mut diff_size = 0usize;
 
+    let mut addresses_to_check = Vec::with_capacity(account_changes.len());
+
     for (addr, account) in account_changes {
-        if account.created {
-            diff_size += ACCOUNT_IDX_KEY_SIZE + ACCOUNT_IDX_SIZE;
-        }
+        // cloning addresses to avoid borrowing issues
```
kpp, line 749:
```suggestion
```

crates/evm/src/evm/handler.rs
```diff
@@ -754,13 +741,14 @@ where
 
     let mut diff_size = 0usize;
 
+    let mut addresses_to_check = Vec::with_capacity(account_changes.len());
```
kpp, line 743:
```suggestion
    // Check if it's a new address to charge for new index
    let mut addresses_to_check = Vec::with_capacity(account_changes.len());
```

kpp (approved):

crates/evm/src/evm/db.rs
```diff
@@ -121,6 +121,32 @@ impl<C: sov_modules_api::Context> Database for EvmDb<'_, C> {
     }
 }
 
+/// A trait to check if an account is newly created.
+/// This is useful when calculating diff size for a transactions
+pub trait IsAccountNewlyCreatedProvider {
```
kpp:
```
pub trait AccountExistsProvider: AccountReader {
    fn account_exists(&self, address: &Address) -> ProviderResult<bool>;
}
```

crates/evm/src/evm/db.rs
```diff
@@ -121,6 +121,32 @@ impl<C: sov_modules_api::Context> Database for EvmDb<'_, C> {
...
+    /// Check if an account is newly created
+    /// By querying `Evm::account_exists`
+    fn is_account_new(&mut self, address: &Address) -> bool;
+}
+
+impl<C: sov_modules_api::Context> IsAccountNewlyCreatedProvider for EvmDb<'_, C> {
+    fn is_account_new(&mut self, address: &Address) -> bool {
```
kpp: How it can be new if it does not exist? This semantics `new/old` is "artificial" and exists only in state diff calc.

crates/evm/src/evm/db.rs
eyusufatik: I renamed but not extending the trait

eyusufatik:
accumulating different type of diff changes then applying discounts would also be a good touch

eyusufatik:
also let's change those diff constants to be u64

crates/evm/src/evm/handler.rs
```diff
@@ -704,11 +700,11 @@ where
             }
             JournalEntry::CodeChange { address } => {
                 let account = account_changes.entry(address).or_default();
-                account.code_changed = true;
+                account.account_info_changed = true;
             }
+            // Only addded to the journal on smart contract creation
```
kpp:
```suggestion
            // Only added to the journal on smart contract creation
```

bin/citrea/tests/mock/evm/diff_sizes.rs
```diff
@@ -0,0 +1,187 @@
...
+        false,
+    )
+    .await;
+
+    let seq_port = seq_port_rx.await.unwrap();
+
+    let seq_test_client = init_test_rollup(seq_port).await;
+
+    // compare sending cBTC to the same address
+    // resulting in different sizes because the second address was just created
```
jfldde: Not clear. Was already created ?

bin/citrea/tests/mock/evm/diff_sizes.rs
```diff
@@ -0,0 +1,187 @@
...
+            )
+            .unwrap(),
+            U64::from(30)
+        );
+    }
+
+    // make a transaction to a contract that calls another contract
+    // resulting in storage changes in both
+    // all storage changes should be charged
+    {}
```
jfldde: TODO ?

crates/evm/src/evm/handler.rs
```diff
@@ -752,15 +739,18 @@ where
         "Total accounts for diff size"
     );
 
-    let mut diff_size = 0usize;
+    // Check if it's a new address to charge for new index
+    let mut addresses_to_check = Vec::with_capacity(account_changes.len());
```
jfldde, line 743: ~Do we need this since we push all addr form account_changes ? Can't we just iter over account_changes a second time ?~ Guess we'll face the same mutable borrow issue

bin/citrea/tests/mock/evm/diff_sizes.rs
eyusufatik: updated comment will push in a bit

crates/evm/src/evm/handler.rs
```diff
@@ -771,20 +761,27 @@ where
...
-                (db_account_size + DB_ACCOUNT_KEY_SIZE) * ACCOUNT_DISCOUNTED_PERCENTAGE / 100;
+            account_based_diff += db_account_size + DB_ACCOUNT_KEY_SIZE;
         }
 
         // Apply size of changed slots
         let slot_size = STORAGE_KEY_SIZE + STORAGE_VALUE_SIZE; // key + value;
 
-        diff_size +=
-            slot_size * account.storage_changes.len() * STORAGE_DISCOUNTED_PERCENTAGE / 100;
+        storage_based_diff += slot_size * account.storage_changes.len();
```
eyusufatik, line 770: checking storage_changes.len() > 0 saves us a few cycles maybe?

crates/evm/src/evm/handler.rs
```diff
@@ -752,15 +739,18 @@ where
...
+
+    let mut account_based_diff = 0usize;
+    let mut storage_based_diff = 0usize;
 
     for (addr, account) in account_changes {
-        if account.created {
-            diff_size += ACCOUNT_IDX_KEY_SIZE + ACCOUNT_IDX_SIZE;
-        }
+        // cloning addresses to avoid borrowing issues
+        addresses_to_check.push(*addr);
```
jfldde, line 750: Can't we infer from entry variants in journal above a subset that could be `is_first_time_committing_address`. Maybe only `AccountCreated` is relevant here ?

ercecan (approved):

crates/evm/src/evm/handler.rs
eyusufatik: unfortunately only entry about an account first time appearing is for contract creations.

crates/evm/src/evm/handler.rs
kpp: No, because r0 says each instruction takes exact same resource. Except for special ones

jfldde (approved):

PR #2221, opened 2025-04-08
Recover from MinRelayFeeNotMet error

### LINKED ISSUE #2203: Infinite loop on min relay fee not met
Since we retry recreating and sending bitcoin TXs are there is a chance we end up in an infinite loop.

This happened before.

See my "fix" on release-v0.6.11

But I think we can have a batter aproach to this.

jfldde:
# Description

- Add a `BitcoinServiceError` to be able to recover from
- Recover form `MinRelayFeeNotMet` variant and increase base fee rate multiplier
- Cap fee rate multiplier at 2.0f64
- Handle base fee and calculation in FeeService

## Linked Issues
- Fixes #2203 

## Testing
- Add `test_min_relay_fee_handling` test. Drains sequencer UTXOs just enough to hit a min relay fee not met then let da service recover.


eyusufatik (approved):
looking good

PR #2226, opened 2025-04-09
Rename fork2 to Tangerine

rakanalh:
# Description


crates/common/src/l2.rs
```diff
@@ -95,7 +95,7 @@ pub async fn process_l2_block<Da: DaService, DB: SharedLedgerOps>(
         .context("Failed to parse transactions")?;
 
     let l2_block_result = {
-        // Since Post fork2 we do not have the slot hash in l2 blocks we inspect the txs and get the slot hashes from set block infos
+        // Since Post tangerine we do not have the slot hash in l2 blocks we inspect the txs and get the slot hashes from set block infos
```
jfldde:
```suggestion
        // Post tangerine, we do not have the slot hash in l2 blocks we inspect the txs and get the slot hashes from set block infos
```
Or something, looks a bit weird with `Since Post`

jfldde (approved):
Tangerine should always be capitalized in doc and comments imo

eyusufatik (approved):

PR #2254, opened 2025-04-17
Fork activation e2e testing

### LINKED ISSUE #2242: Fork testing

jfldde:
# Description

- Add tests with assertion that precompile and `eip7702` are not available before Tangerine fork
- Assert that first commitment happens only after Tangerine with index 1
- Assert that first proof happens only after Tangerine 
- Restore `ALL_FORKS` to include Genesis and Kumquat

## Missing 

- Rollback and assert that everything is desactivated as it should

Fixes #2242


bin/citrea/tests/bitcoin/fork.rs
```diff
@@ -0,0 +1,455 @@
...
+                p256_result.is_err(),
+                "P256_VERIFY precompile should not be available before Tangerine fork"
+            );
+        }
+
+        let eip7702_result = self.send_eip7702_transaction(client).await;
+        assert!(
+            eip7702_result.is_err(),
+            "eip7702 tx should fail before Tangerine fork"
+        );
```
eyusufatik, line 267: i don't understand how this is failing tbh, the sequencer only sets eip7702 allowed on mempool on start up, and that does not depend on forks at all.

bin/citrea/tests/bitcoin/fork.rs
```diff
@@ -0,0 +1,455 @@
...
+                g1_add_result.is_ok(),
+                "G1_ADD should be available after Tangerine fork"
+            );
+        }
+
+        let eip7702_result = self.send_eip7702_transaction(client).await;
+        assert!(
+            eip7702_result.is_ok(),
+            "eip7702 tx should succeed after Tangerine"
+        );
```
eyusufatik: not just ok, assert result is 0x00000...01 on p256 and schnorr calls, look up the expected result of g1 add call

bin/citrea/tests/bitcoin/fork.rs
```diff
@@ -0,0 +1,455 @@
...
+                )
+                .await;
+
+            assert!(
+                g1_add_result.is_ok(),
+                "G1_ADD should be available after Tangerine fork"
+            );
+        }
+
+        let eip7702_result = self.send_eip7702_transaction(client).await;
```
eyusufatik: eth_sendRawTransaction returning ok is not enough, wait for the transactions inclusion to a block by calling eth_getTransactionReceipt or the relevant method of the alloy return type

eyusufatik (changes requested):

bin/citrea/tests/bitcoin/fork.rs
jfldde: Fails on `estimate_gas` with
```
Err(server returned an error response: error code -32003: transaction type not supported)
```

eyusufatik: ah I see, that's because those endpoints are ofc fork aware.

bin/citrea/tests/bitcoin/fork.rs
```diff
@@ -344,21 +346,23 @@ impl ForkActivationTest {
                 )
                 .await;
 
-            assert!(
-                g1_add_result.is_ok(),
+            assert_eq!(
+                g1_add_result.unwrap(),
+                "0x0000000000000000000000000000000000000000000000000000000000000001",
```
eyusufatik: 01? very suspicious

eyusufatik (changes requested):

eyusufatik (approved):

PR #2255, opened 2025-04-17
Simplify create_txn_env nonce handling

### LINKED ISSUE #2207: Quality of Code on create_txn_env
Why make it optional instead of just u64?

_Originally posted by @kpp in https://github.com/chainwayxyz/citrea/pull/2185#discussion_r2030194193_

jfldde:
# Description

- Pass down non-optional nonce to `create_txn_env`

## Linked Issues
- Fixes #2207

### DIFF
--- a/crates/evm/src/evm/call.rs
+++ b/crates/evm/src/evm/call.rs
@@ -15,7 +15,7 @@ pub(crate) fn create_txn_env(
     block_env: &BlockEnv,
     request: TransactionRequest,
     cap_to_balance: Option<U256>,
-    nonce_if_req_has_no_nonce: Option<u64>,
+    nonce: u64,
     chain_id_to_set: u64,
 ) -> EthResult<TxEnv> {
     let tx_type = if request.authorization_list.is_some() {
@@ -39,7 +39,6 @@ pub(crate) fn create_txn_env(
         gas,
         value,
         input,
-        nonce,
         access_list,
         chain_id,
         authorization_list,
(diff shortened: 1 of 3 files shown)


crates/evm/src/evm/call.rs
```diff
@@ -39,7 +39,6 @@ pub(crate) fn create_txn_env(
         gas,
         value,
         input,
-        nonce,
```
kpp, line 42: We have to fail tx if user sends a tx with invalid nonce.

jfldde: This is handled by `revm` validation

kpp: No, you always overwrite request.Some(nonce) with correct nonce from DB. So if a user sends tx.nonce=3 when in DB nonce=4, tx is expected to fail but it wont.

jfldde: Where does the behaviour changes ?
We used to set request.nonce = None then pass Some(nonce). Not we just pass nonce and use it

jfldde: We don't always overwrite, we do `let nonce = request.nonce.unwrap_or(account.nonce);`

eyusufatik (approved):

crates/evm/src/evm/call.rs
eyusufatik: right now behaviour is equivalent to the version before

PR #2258, opened 2025-04-17
ZkVM FakeReceipt Generation RPC

yaziciahmet:
# Description

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/batch-prover/src/rpc.rs
```diff
@@ -196,6 +230,121 @@ where
...
+        let commitments = ledger_db
+            .get_commitment_by_range(index_start..=index_end)
+            .map_err(|e| internal_rpc_error(e.to_string()))?;
+        if commitments.len() as u32 != index_end - index_start + 1 {
+            return Err(internal_rpc_error(
+                "Missing some commitment indices from the range",
+            ));
+        }
+
+        let last_commitment = commitments.last().expect("Must have at least 1");
```
rakanalh:
```suggestion
        let last_commitment = commitments.last().expect("Must have at least 1");
```
Use ok_or and return an error instead of panicing here

rakanalh (approved):

crates/batch-prover/src/rpc.rs
yaziciahmet: but with this:
```rs
if index_start > index_end {
       return Err(internal_rpc_error("Invalid index range"));
}
```
and this checks:
```rs
if commitments.len() as u32 != index_end - index_start + 1 {
             return Err(internal_rpc_error(
                 "Missing some commitment indices from the range",
             ));
}
```
it is already ensured that there is definitely at least one commitment. because the range is inclusive

jfldde (approved):

crates/batch-prover/src/rpc.rs
```diff
@@ -196,6 +230,121 @@ where
...
+        let last_l2_block = ledger_db
+            .get_l2_block_by_number(&L2BlockNumber(last_commitment.l2_end_block_number))
+            .map_err(|e| internal_rpc_error(e.to_string()))?
+            .ok_or_else(|| internal_rpc_error("Not synced up to latest L2 block yet"))?;
+
+        let mut start_l2_height = previous_commitment.l2_end_block_number + 1;
+        let initial_state_root = ledger_db
```
eyusufatik, line 280: no this should be previous commitment l2 end state root

bin/citrea/tests/bitcoin/batch_prover_test.rs
```diff
@@ -1045,3 +1045,64 @@ async fn test_batch_proof_l1_hashes_added_output() -> Result<()> {
...
+        TestCaseConfig {
+            with_batch_prover: true,
+            with_light_client_prover: true,
+            ..Default::default()
+        }
+    }
+
+    fn sequencer_config() -> SequencerConfig {
+        SequencerConfig {
+            max_l2_blocks_per_commitment: 5,
```
jfldde, line 1065: Why not default value of 4 ?

crates/batch-prover/src/rpc.rs
```diff
@@ -196,6 +230,121 @@ where
...
+                .expect("L2 state root must exist");
+            state_roots.push(end_state_root);
+
+            start_l2_height = end_l2_height + 1;
+        }
+
+        let storage = self
+            .context
+            .storage_manager
+            .create_storage_for_l2_height(last_l2_block.height + 1);
```
eyusufatik, line 316: double check this

btw we can also use eth_call here. might be easier not to introduce bugs

eyusufatik (changes requested):

crates/batch-prover/src/rpc.rs
eyusufatik: it's like last state root befoer blocks in the proof

crates/batch-prover/src/rpc.rs
yaziciahmet: double checked, create_storage_for_l2_height takes in the height for the l2 block to run, so to get state of block N, we pass in N+1

crates/batch-prover/src/rpc.rs
yaziciahmet: good catch, fixed

bin/citrea/tests/bitcoin/batch_prover_test.rs
yaziciahmet: tbh, following multiples of 5 feels more natural, that is the only reason.

bin/citrea/tests/bitcoin/batch_prover_test.rs
```diff
@@ -1045,3 +1047,180 @@ async fn test_batch_proof_l1_hashes_added_output() -> Result<()> {
...
+        // verify lcp output
+        let lcp_output = light_client
+            .client
+            .http_client()
+            .get_light_client_proof_by_l1_height(finalized_height)
+            .await?
+            .unwrap()
+            .light_client_proof_output;
+        assert_eq!(lcp_output.last_sequencer_commitment_index.to::<u32>(), 4);
+        assert_eq!(lcp_output.last_l2_height.to::<u32>(), 20);
```
eyusufatik, line 1214: you should assert all other fields too

ercecan (approved):

eyusufatik (approved):

PR #2263, opened 2025-04-18
Deposit Event Bug Fix in Deposit Id

erdkocak:
# Description

- Emit `depositTxIds.length - 1` in the `Deposit` event so each deposit gets a unique ID.
- Add `testSecondDepositId` to verify that the second deposit’s `depositId` is `1`.


crates/evm/src/evm/system_contracts/test/Bridge.t.sol
```diff
@@ -360,4 +360,42 @@ contract BridgeTest is Test {
         bridge.deposit(depositParams);
         vm.stopPrank();
     }
+
+    function testSecondDepositId() public {
+    doDeposit();
```
eyusufatik: formatting

crates/evm/src/evm/system_contracts/out/ProxyAdmin.sol/ProxyAdmin.json
eyusufatik, line 1: revert

eyusufatik (changes requested):
revert other contract changes

okkothejawa (approved):

eyusufatik (approved):

PR #2267, opened 2025-04-21
CLI: Clear fullnode pending proofs and commitments

jfldde:
# Description

- Drop both `PendingSequencerCommitments` and `PendingProofs` column families that will be re-created on next fullnode startup


eyusufatik (commented):
are we sure dropping the column family won't cause issues on the next start up?

jfldde:
Yes, since we're opening with RocksDbConfig `create_if_missing(true);` option. Tested it on devnet and worked as expected on restart.

eyusufatik (approved):

PR #2268, opened 2025-04-22
Fix possible attack vector on full node for commitment with very high unsynced l2 end block number

### LINKED ISSUE #2214: Possible L1 process halting attack vector on full node
- [ ] https://github.com/chainwayxyz/citrea/pull/2115#discussion_r2028830900

https://github.com/chainwayxyz/citrea/pull/2115/files/e0df646e1250543569054faf014ff290386dc92a#diff-9bddebe2983b37c2a6f39fbd521183021f7d118e96a128487161d7b56abcfb0bR142-R151

When processing the commitments we get the max l2 height of all the commitments in the L1 block, then we check if we have all the L2 blocks necessary to calculate and verify the given merkle roots in the commitments, however what if in an l1 block we have the following commitments:
index | L2 end height
1 | 10
2 | 20
3 | u64:MAX

We go through the commitments and find that the l2 block we should have synced up to is u64::MAX, however that is not really possible, which causes the commitments to be never processed and l1 to practically halt.

Same problem happens some where else as well:
- [ ] https://github.com/chainwayxyz/citrea/pull/2115#discussion_r2028832499

https://github.com/chainwayxyz/citrea/pull/2115/files/e0df646e1250543569054faf014ff290386dc92a#diff-9bddebe2983b37c2a6f39fbd521183021f7d118e96a128487161d7b56abcfb0bR160-R171

In here the full node returns early if the necessary l2 hieght of commitment is not synced, but if there is again a malicious commitment with u64::MAX l2 end height this will halt there as well.

ercecan:
# Description
When processing the commitments we get the max l2 height of all the commitments in the L1 block, then we check if we have all the L2 blocks necessary to calculate and verify the given merkle roots in the commitments, however what if in an l1 block we have a commitment with a very high l2 end block number, then the full node l1 processing will halt.

This pr fixes that issue by storing unsynced commitments to pending commitments storage, also storing unsynced proofs to pending proofs storage

## Linked Issues
- Fixes #2214


crates/fullnode/src/da_block_handler.rs
```diff
@@ -329,11 +302,25 @@ where
...
         if stored_l2_blocks.len() < ((end_l2_height - start_l2_height) as usize) {
-            return Err(SyncError::MissingL2(
-                "L2 range not synced yet",
-                start_l2_height,
-                end_l2_height,
-            ));
+            if self
+                .ledger_db
+                .get_pending_commitment_by_index(sequencer_commitment.index)?
+                .is_none()
```
eyusufatik: instead of loading all first and checking len, 

check if start and end exist in db, only then load all blocks.

crates/fullnode/src/da_block_handler.rs
```diff
@@ -521,11 +503,57 @@ where
...
+                            "Proof verification: For a pending sequencer commitment. Hash mismatch - expected 0x{} but got 0x{}. Skipping proof.",
+                            hex::encode(pending_commitment.serialize_and_calculate_sha_256()),
+                            hex::encode(expected_hash)
+                        ).into());
+                }
+                info!("Proof has a pending commitment with index: {}.", index);
+                proof_is_pending = true;
             } else {
                 return Err(SyncError::SequencerCommitmentMissingForProof(index));
             }
```
eyusufatik: for commitment and previous commitment the verification can be a common function

bin/citrea/tests/bitcoin/full_node.rs
```diff
@@ -1675,3 +1677,562 @@ async fn test_overlapping_proof_ranges() -> Result<()> {
...
+    async fn cleanup(self) -> Result<()> {
+        self.task_manager
+            .graceful_shutdown_with_timeout(Duration::from_secs(1));
+        Ok(())
+    }
+
+    async fn run_test(&mut self, f: &mut TestFramework) -> Result<()> {
+        /*
+        Sequencer max l2 blocks is 10000 so it does not x
+        Sequencer publish 1-10 x
```
eyusufatik: remove x's

bin/citrea/tests/bitcoin/full_node.rs
```diff
@@ -1675,3 +1677,562 @@ async fn test_overlapping_proof_ranges() -> Result<()> {
...
+#[async_trait]
+impl TestCase for UnsyncedCommitmentL2RangeTest {
+    fn test_config() -> TestCaseConfig {
+        TestCaseConfig {
+            with_full_node: true,
+            with_sequencer: true,
+            with_batch_prover: true,
+            with_citrea_cli: true,
+            ..Default::default()
+        }
```
eyusufatik: - no need for citrea-cli
- remove unused code comments
- set up light client prover and compare results (proven height and commitment index) with full node

eyusufatik (changes requested):

eyusufatik (approved):

PR #2269, opened 2025-04-22
Failed Deposits Now Go To Vault

erdkocak:
## Description
Forward failed deposits to the designated vault address instead of the recipient.

## Testing
Added a unit test covering the fallback path.


crates/evm/src/evm/system_contracts/out/BitcoinLightClient.sol/BitcoinLightClient.json
eyusufatik, line 1: why is this file changed?

eyusufatik (changes requested):
looks ok except out.json changes

eyusufatik (approved):

PR #2271, opened 2025-04-23
Log missing environment variable

### LINKED ISSUE #2261: Log what env was missing

yaziciahmet:
# Description

## Linked Issues
- Fixes #2261 

## Testing

## Docs

### DIFF
--- a/crates/bitcoin-da/src/monitoring.rs
+++ b/crates/bitcoin-da/src/monitoring.rs
@@ -171,7 +171,7 @@ impl FromEnv for MonitoringConfig {
             read_env("DA_MONITORING_HISTORY_LIMIT"),
             read_env("DA_MONITORING_MAX_HISTORY_SIZE"),
         ) {
-            (Err(_), Err(_), Err(_)) => Err(anyhow!("Missing monitoring config")),
+            (Err(_), Err(_), Err(_)) => Err(anyhow!("At least one of the monitoring envs must exist: DA_MONITORING_CHECK_INTERVAL, DA_MONITORING_HISTORY_LIMIT, DA_MONITORING_MAX_HISTORY_SIZE")),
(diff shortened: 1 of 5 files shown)


jfldde (approved):

crates/bitcoin-da/src/monitoring.rs
```diff
@@ -166,9 +167,9 @@ impl Default for MonitoringConfig {
...
     fn from_env() -> anyhow::Result<Self> {
         match (
-            std::env::var("DA_MONITORING_CHECK_INTERVAL"),
-            std::env::var("DA_MONITORING_HISTORY_LIMIT"),
-            std::env::var("DA_MONITORING_MAX_HISTORY_SIZE"),
+            read_env("DA_MONITORING_CHECK_INTERVAL"),
+            read_env("DA_MONITORING_HISTORY_LIMIT"),
+            read_env("DA_MONITORING_MAX_HISTORY_SIZE"),
         ) {
             (Err(_), Err(_), Err(_)) => Err(anyhow!("Missing monitoring config")),
```
eyusufatik: should update this line

eyusufatik (changes requested):
are you sure if we call println!(err) on every from_env() err case?

crates/bitcoin-da/src/monitoring.rs
yaziciahmet: didn't update it initially because monitoring config error was unique here

updated it here to be more explicit: https://github.com/chainwayxyz/citrea/pull/2271/commits/f45cd0211f2c214d0bf624ba6cc4348b6b7640e4

yaziciahmet:
Yes, we handle all env call when starting the rollup and if there is an error we log it. This won't show ALL the envs that are missing, but the most recent checked one only, so resolving of env variables will show up one by one.

eyusufatik (approved):

PR #2284, opened 2025-04-28
Better logging in Light Client Prover

### LINKED ISSUE #2245: Light client circuit macro for logging::info! on native println! on zk
On our logging system println!'s don't show up as they are not json (my guess)

Just like or native_info! and native_error! macros, we can have a macro that resolves to logging::info etc. or println!.

eyusufatik:
# Description
Uses `tracing::info!` in native context so that we get to capture logs created by the circuit better. ZK logs still won't show up in our logging system but still can be reached if needed, so this change is enough for now.

## Linked Issues
Closes #2245 

~Adding HOLD-MERGE label as I can't verify this works as intended due to bitcoin v29 docker not working correctly on my system.~


rakanalh (approved):
The macro is a good idea :+1:

PR #2288, opened 2025-04-28
Discard 0 index commitments

### LINKED ISSUE #2274: Make sure sequencer commitment with index 0 is discarded everywhere
Because the first commitment index is 1, If we somehow get commitment with 0 index we should discard it.

ercecan:
# Description
Makes sure 0 indexed sequencer commitments is never used

## Linked Issues
- Fixes #2274


crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
```diff
@@ -374,6 +374,13 @@ where
         // and this should start from proving the first l2 block
         let (previous_commitment_index, previous_commitment_hash) =
             if let Some(previous_sequencer_commitment) = previous_sequencer_commitment {
+                // The only way there would be a 0 indexed commitment is if the previous commitment somehow has index 0
+                // This assertion will block that
+                assert!(
+                    previous_sequencer_commitment.index != 0,
+                    "Previous sequencer commitment index must be non-zero"
+                );
```
eyusufatik, line 382: among all changes in this pr this is actually enough.

crates/fullnode/src/da_block_handler.rs
```diff
@@ -140,6 +140,14 @@ where
...
             match commitment_or_proof {
                 ProofOrCommitment::Commitment(commitment) => {
+                    if commitment.index == 0 {
+                        // Skip the first commitment as it is not a sequencer commitment
+                        error!(
+                            "Detected sequencer commitment with index 0 at L1 height {}, skipping...",
+                            l1_block.header().height()
+                        );
+                        continue;
+                    }
```
eyusufatik: maybe this can stay yes.

crates/light-client-prover/src/circuit/mod.rs
eyusufatik, line 1: as we enforce no 0 indexed sequencer commitment is processed in the batch proof, we can just revert all lcp changes.

eyusufatik (changes requested):
On the light client proof, we start checking state roots from proofs by initally looking at index 1, even though we process a commitment with index 0, it's just added to the state tree and nothing is done, so we don't need an extra if.

crates/fullnode/src/da_block_handler.rs
```diff
@@ -140,6 +140,14 @@ where
         for commitment_or_proof in commitments_and_proofs {
             match commitment_or_proof {
                 ProofOrCommitment::Commitment(commitment) => {
+                    if commitment.index == 0 {
+                        // Skip the first commitment as it is not a sequencer commitment
```
eyusufatik: Change the comment to reflect more on why we ignore 0 index commitments and start from 1

eyusufatik (approved):

PR #2293, opened 2025-04-29
Update full node chunk proof processing

### LINKED ISSUE #2277: Full node and light client circuit chunk proof processing difference
In light client prover circuit when we get chunk proofs we store them to the jmt with it's wtxid, and when we get aggregate data we iterate through the wtxids of that aggregate and for each wtxid check if the chunk exists in the jmt, if it does not we discard that aggregate proof.

However in full nodes, when we get an aggregate proof, using the txids from the aggregate we query the DA, if we can get all the chunks we combine them into one complete proof and continue processing, if we cannot we skip that proof

The difference is in full nodes the chunks can come after the aggregate data.

In order to have the same behavior with light client prover fix is simple on the full node side:
if any chunk is after the aggregate data ignore that proof

We can also make light client prover to work like full node as well, we can store the aggregate if we do not have the chunks, once we have all the missing chunks we can continue processing

I would suggest the first approach tho

ercecan:
# Description
In light client circuit we process aggregate proofs only if we have all the chunks of it before the aggregate, this was not the case in full node which may create some difference in chunk proof processing
This pr aims to fix that difference by making sure full nodes also only accept aggregates with chunks that come before the aggregate

## Linked Issues
- Fixes #2277


crates/bitcoin-da/src/spec/transaction.rs
```diff
@@ -11,7 +11,7 @@ use serde::{Deserialize, Serialize};
 #[derive(Clone, PartialEq, Eq, Debug, Hash, Deserialize, Serialize)]
 #[repr(transparent)]
 #[serde(transparent)]
-pub struct TransactionWrapper(Transaction);
+pub struct TransactionWrapper(pub(crate) Transaction);
```
eyusufatik: use `TransactionWrapper::inner()` don't make inner pub

crates/bitcoin-da/src/service.rs
```diff
@@ -944,7 +954,57 @@ impl DaService for BitcoinService {
...
+                    continue 'aggregate;
+                }
+                if tx_block_height == block.header.height as usize {
+                    if let Some(chunk_idx) = chunks.get(&chunk_id) {
+                        if *chunk_idx >= i {
+                            // This means the chunk comes after the aggregate in the same block
+                            // This is not a valid case because lcp expects all chunks to come before their aggregate
+                            error!("{}:{}: Chunk comes after aggregate", tx_id, chunk_id);
+                            continue 'aggregate;
+                        }
```
eyusufatik: these should be warn

crates/bitcoin-da/src/service.rs
```diff
@@ -1397,3 +1457,61 @@ fn calculate_witness_root(txdata: &[TransactionWrapper], tx_count: usize) -> [u8
         .collect();
     BitcoinMerkleTree::new(hashes).root()
 }
+
+impl TryFrom<GetRawTransactionResult> for TransactionWrapper {
```
eyusufatik: if i'm not mistaken, there should be From<GetRawTransactionResult> for Transaction. use that and From<Transaction> for TransactionWrapper

crates/bitcoin-da/src/service.rs
```diff
@@ -944,7 +954,57 @@ impl DaService for BitcoinService {
...
+                let tx_block_height = if let Some(tx_block_hash) = tx_raw.blockhash {
+                    let res = retry_backoff(exponential_backoff, || async move {
+                        self.client
+                            .get_block_info(&tx_block_hash)
+                            .await
+                            .map_err(|e| match e {
+                                BitcoinError::Io(_) => backoff::Error::transient(e),
+                                _ => backoff::Error::permanent(e),
+                            })
+                    })
```
eyusufatik: I wonder if we can leverage L1BlockCache here.

crates/bitcoin-da/src/service.rs
```diff
@@ -944,7 +954,57 @@ impl DaService for BitcoinService {
...
+                            // This means the chunk comes after the aggregate in the same block
+                            // This is not a valid case because lcp expects all chunks to come before their aggregate
+                            error!("{}:{}: Chunk comes after aggregate", tx_id, chunk_id);
+                            continue 'aggregate;
+                        }
+                    } else {
+                        // This means the block hashes are somehow different
+                        // Should be infallible
+                        unreachable!("{}:{}: Chunk not found in chunks map", tx_id, chunk_id);
+                    }
```
eyusufatik: you could make a similar if check before sending all these getblock requests so that we send less get block requests.

if a chunk_id is in chunks hashmap, you can compare the indices right away.

crates/bitcoin-da/src/service.rs
```diff
@@ -944,7 +954,57 @@ impl DaService for BitcoinService {
                     }
                 };
 
-                let wrapped: TransactionWrapper = tx_raw.into();
+                let tx_block_height = if let Some(tx_block_hash) = tx_raw.blockhash {
```
eyusufatik: this function is already hard to follow before these changes. any way we can move the newly added code to somewhere else?

crates/bitcoin-da/src/error.rs
```diff
@@ -5,6 +5,16 @@ use tokio::task::JoinError;
...
+#[derive(Debug, Error)]
+pub enum BitcoinTxConversionError {
+    #[error("Input Txid not found")]
+    InputTxidNotFound,
+    #[error("Input script sig not found")]
+    ScriptSigNotFound,
+    #[error("Witness not found")]
+    WitnessNotFound,
+}
+
```
eyusufatik: If my from + from comment works changes in the file is unnecessary.

eyusufatik (changes requested):

crates/bitcoin-da/src/service.rs
ercecan: Unfortunately there is not

crates/bitcoin-da/src/service.rs
ercecan: I can but I would need to use traits because I cannot put the cache inside da service trait
Should I do it?

crates/bitcoin-da/src/spec/transaction.rs
ercecan: Not sure how that helps, I can move the conversion of rpc transaction under spec/transaction.rs and feature gate it, that way I would not need to change the visibility, would you prefer that?

crates/bitcoin-da/src/service.rs
eyusufatik: then you can use `GetRawTransactionResult::hex` and consensus decode to convert the transaction.

I don't really want this custom impl here.

crates/bitcoin-da/src/spec/transaction.rs
eyusufatik: https://github.com/chainwayxyz/citrea/pull/2293#discussion_r2070206388 this would remove the need for this change.

crates/bitcoin-da/src/service.rs
ercecan: just saw there is a function called 'transaction' that already does that :D Sorry

crates/bitcoin-da/src/lib.rs
```diff
@@ -15,6 +15,9 @@ pub mod fee;
 #[cfg(feature = "native")]
 pub mod rpc;
 
+#[cfg(feature = "native")]
```
eyusufatik: make it test only

bin/citrea/tests/bitcoin/full_node.rs
```diff
@@ -2104,6 +2108,342 @@ async fn test_unsynced_commitment_l2_range_test() -> Result<()> {
...
+        // Check that the proof was not processed and the last proven height is still the same
+        let last_proven_l2_height = full_node
+            .client
+            .http_client()
+            .get_last_proven_l2_height()
+            .await?
+            .unwrap();
+        assert_eq!(last_proven_l2_height.height, 20);
+        assert_eq!(last_proven_l2_height.commitment_index, 2);
+
```
eyusufatik, line 2493: after we asserted the chunks + aggregate combo was not processed, you can try sending the same data using the same function and correct order to make sure your newly written function is working correctly

eyusufatik (approved):

PR #2297, opened 2025-04-30
compare risc0-zkvm and r0vm versions on LocalProver startup

eyusufatik:
# Description
Will be useful next time we forgot to update r0vm on infra

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/risc0/src/host/local.rs
+++ b/crates/risc0/src/host/local.rs
@@ -30,6 +30,14 @@ impl LocalProver {
         let dev_mode = env::var("RISC0_DEV_MODE").is_ok();
         let r0vm_path = get_r0vm_path().expect("Could not get r0vm path");
 
+        // Check if the version of the r0vm matches the version of the risc0 crate
+        // the `get_r0vm_path` function will return the path to the r0vm binary with the correct version
+        // IF the version is installed
+        // else the function will just return "r0vm" as path
+        // then we'd be defaulting to the system default
+        // so we need to check if the version is correct
+        compare_risc0_versions(&r0vm_path).expect("Something is wrong with system r0vm");
+
         Self {
             dev_mode,
             r0vm_path,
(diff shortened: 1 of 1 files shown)


crates/risc0/src/host/local.rs
```diff
@@ -30,6 +30,14 @@ impl LocalProver {
         let dev_mode = env::var("RISC0_DEV_MODE").is_ok();
         let r0vm_path = get_r0vm_path().expect("Could not get r0vm path");
 
+        // Check if the version of the r0vm matches the version of the risc0 crate
+        // the `get_r0vm_path` function will return the path to the r0vm binary with the correct version
+        // IF the version is installed
+        // else the function will just return "r0vm" as path
+        // then we'd be defaulting to the system default
+        // so we need to check if the version is correct
+        compare_risc0_versions(&r0vm_path).expect("Something is wrong with system r0vm");
```
rakanalh, line 39:
```suggestion
        if let Err(e) = compare_risc0_versions(&r0vm_path) {
            panic!("Found an error comparing risc0 version: {}", e);
        }
```

This is better IMO since multiple errors are emitted from that function. We need to see which one.

rakanalh (approved):
Looks reasonable.

crates/risc0/src/host/local.rs
eyusufatik: why? expect will already panic

crates/risc0/src/host/local.rs
```diff
@@ -159,3 +167,37 @@ fn get_r0vm_path() -> anyhow::Result<PathBuf> {
...
+        .strip_prefix("risc0-r0vm ")
+        .expect("r0vm must return version string in this format");
+
+    if version != r0vm_version {
+        return Err(anyhow!(
+            "RISC0 version {} does not match r0vm version {}",
+            version,
+            r0vm_version
+        ));
+    }
```
yaziciahmet, line 200: maybe just a `contains` check in case risc0 for any reason changes the format of the version output?

crates/risc0/src/host/local.rs
rakanalh: This is better IMO since multiple errors are emitted from that function. We need to see which one.

yaziciahmet (approved):

crates/risc0/src/host/local.rs
eyusufatik: but doesn't expect also show you the unwrapped err?

crates/risc0/src/host/local.rs
eyusufatik: tbf, does not matter because our tests would also fail when we update risc0 version

crates/risc0/src/host/local.rs
rakanalh: You're right! https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=2cf8acb9fcafc96c471e508ddbd2635a

PR #2302, opened 2025-05-01
Propagate coverage env to citrea-e2e

jfldde:
# Description

- Propagate coverage env to citrea-e2e with companion PR https://github.com/chainwayxyz/citrea-e2e/pull/96
- Remove build-test dependency. We need to use same binary shared between nextest and spawned via citrea-e2e so name mangling matches and we hit coverage. This has the additional benefits of reducing CI dependencies, start running coverage straight away and remove the need for additional build step.
- Add tests folder with placeholder to `citrea-cli` so it gets compiled by nextest alongside main `citrea` binary
- Remove setting `CITREA_E2E_TEST_BINARY` as this would override CI env and is handled by test framework `.set_citrea_path(get_citrea_path())`

### DIFF
--- a/bin/cli/tests/mod.rs
+++ b/bin/cli/tests/mod.rs
@@ -1,0 +1,7 @@
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn test_placeholder() {
+        assert_eq!(40 + 2, 42);
+    }
+}
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -194,7 +194,6 @@ jobs:
   coverage:
     needs:
       - docker-setup
-      - build-test
     runs-on: ubicloud-standard-30
     if: github.event.pull_request.draft == false
     steps:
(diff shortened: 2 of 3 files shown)


bin/cli/tests/mod.rs
```diff
@@ -0,0 +1,7 @@
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn test_placeholder() {
+        assert_eq!(40 + 2, 42);
```
eyusufatik, line 5: 😂

eyusufatik (approved):
Looks like I don't need to pass CITREA_E2E_BINARY= env var when running tests, right?

jfldde:
> Looks like I don't need to pass CITREA_E2E_BINARY= env var when running tests, right?

Yes, this shouldn't be need anymore. It will default to `target/debug/citrea` via
```
pub(super) fn get_citrea_path() -> PathBuf {
    std::env::var("CITREA_E2E_TEST_BINARY").map_or_else(
        |_| {
            let manifest_dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
            manifest_dir
                .ancestors()
                .nth(2)
                .expect("Failed to find workspace root")
                .join("target")
                .join("debug")
                .join("citrea")
        },
        PathBuf::from,
    )
}
```

PR #2303, opened 2025-05-01
Fix always pending first commitment

exeokan:
# Description
Since our criteria for a pending proof was that its predecessor must have been processed, if l2 sync wasn't complete for the commitment 1, it will never be processed again. This PR solves it by handling commitment index 1 separately

## Testing
Added e2e test `test_unsynced_first_commitment`

### DIFF
--- a/bin/citrea/tests/bitcoin/full_node.rs
+++ b/bin/citrea/tests/bitcoin/full_node.rs
@@ -2137,7 +2137,7 @@ impl TestCase for UnsyncedFirstCommitmentTest {
 
         Start full node
         Full node is unable to process commitments, assert last committed l2 is None
-    
+
         Start sequencer
         Full node sync 1-10
         Generate one l1 block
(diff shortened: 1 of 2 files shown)


ercecan (approved):
Good catch

eyusufatik:
in `process_sequencer_commitment` we keep a commitment as pending if we are not synced up to that point. are you sure this change is needed?

exeokan:
> in `process_sequencer_commitment` we keep a commitment as pending if we are not synced up to that point. are you sure this change is needed?

A change is certainly needed since when commitment 1 is pending, we are querying whether commitment 0 is present.
But as you pointed out, we have the check in `process_sequencer_commitment`, so instead of` end_l2_height <= head_l2_height`, we can just put `true` for commitment 1 and always re-process "pending commitment 1" per l1 block.

eyusufatik (approved):

PR #2307, opened 2025-05-02
Add job_id to submit_proof future

### LINKED ISSUE #2299: Log job-id of proof sent
I'd like to see what proof is sent to da, currently I filter by "blob inscribe sent" but there could be a better log for this that includes job id, ideally the partition of the proof.

kpp:
# Description

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/2299

### DIFF
--- a/crates/batch-prover/src/prover.rs
+++ b/crates/batch-prover/src/prover.rs
@@ -419,7 +419,7 @@ where
         Ok(state.into_inner())
     }
 
-    #[instrument(skip_all, fields(_uuid = _uuid.to_string()))]
+    #[instrument(skip_all, fields(uuid = _uuid.to_string()))]
     fn create_circuit_input(
         &self,
         partition: &Partition<'_>,
--- a/crates/batch-prover/src/prover.rs
+++ b/crates/batch-prover/src/prover.rs
@@ -33,7 +33,7 @@ use tokio::sync::{broadcast, mpsc, oneshot};
 use tracing::level_filters::LevelFilter;
 use tracing::{debug, error, info, instrument, warn};
 use tracing_subscriber::layer::SubscriberExt;
-use uuid::{uuid, Uuid};
+use uuid::Uuid;
 
 use crate::partition::{Partition, PartitionMode, PartitionReason, PartitionState};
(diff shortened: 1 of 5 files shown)


eyusufatik:
merge nightly to rakan's pr and rebase this one again pls

crates/batch-prover/src/prover.rs
```diff
@@ -469,10 +470,12 @@ where
         })
     }
 
+    #[instrument(skip_all, fields(uuid))]
```
rakanalh: Isn't this part of the batch prover span? which means that this is going to mention that the method has been called inside a span creating yet a new one... and we're also creating one more in the embedded `start_proving` call.

rakanalh: I am just wondering, but if we want this level of verbosity, that's fine by me.

kpp: It adds a span with uuid. I am not sure how verbose that is. I can fix it if needed though

rakanalh (approved):

crates/prover-services/src/parallel.rs
```diff
@@ -90,19 +90,21 @@ where
 
     /// Runs proving in a blocking manner. This just calls `start_proving` and waits for the result.
     pub async fn prove(&self, data: ProofData, receipt_type: ReceiptType) -> anyhow::Result<Proof> {
-        let (_, rx) = self.start_proving(data, receipt_type).await?;
+        let id = uuid!("00000000-0000-0000-0000-000000000000");
```
yaziciahmet: Uuid::nil()

yaziciahmet (approved):

crates/prover-services/src/parallel.rs
eyusufatik: does this even work?

why pass 000 here? start_proving used to create legit uuid?

kpp: Because it's a temporary uuid which needs to be passed but actually it's called from rpc only. There is no such job, there is only input to prove.

PR #2313, opened 2025-05-02
Fix wrong l1 height map for pending da data

### LINKED ISSUE #2283: Wrong L1 height mapping for unordered batch proofs on DA
For full node, consider this case:
proof of commitment with index 3 on DA:200
proof of commitment with index 2 on DA:205

We write the proof of index 3 to the ledger db table `VerifiedBatchProofsBySlotNumber` after processing the proof of index 2, and we use the `height=205` while doing so

so the proof of index 3 becomes available through the rpc `get_verified_batch_proofs_by_slot_height` if we call it with height=205.

ercecan:
# Description
Pending commitments and proofs are mapped to wrong l1 height, this pr fixes it.
**Migrations Tested locally** ✅ 

## Linked Issues
- Fixes #2283


crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
```diff
@@ -516,12 +516,12 @@ define_table_with_seek_key_codec!(
 
 define_table_with_default_codec!(
     /// Out of order sequencer commitments
-    (PendingSequencerCommitments) u32 => SequencerCommitment
+    (PendingSequencerCommitments) u32 => (SequencerCommitment, u64)
```
jfldde: Makes more sense to change the index to (u64, u32) instead and will let us replace the clear-pending cli command by properly handling it via rollback

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
```diff
@@ -516,12 +516,12 @@ define_table_with_seek_key_codec!(
...
 );
 
 define_table_with_default_codec!(
     /// Out of order proofs
-    (PendingProofs) (u32, u32) => Proof
+    (PendingProofs) (u32, u32) => (Proof, u64)
```
jfldde: Same as above

crates/fullnode/src/da_block_handler.rs
```diff
@@ -382,12 +402,13 @@ where
 
     async fn process_zk_proof(
         &self,
-        l1_block: &Da::FilteredBlock,
+        current_l1_block_height: &u64,
+        found_in_l1_block_height: &u64,
```
jfldde:
```suggestion
        current_l1_block_height: u64,
        found_in_l1_block_height: u64,
```

jfldde: Since we end up copying down the line

jfldde (changes requested):
Good catch. It makes more sense to add this l1 height to the key though. We also need to handle rollback now that we can and remove clear-pending from citrea-cli if not needed anymore

ercecan:
I understand your point, but there’s an issue with that approach. If you look at line 670 of da_block_handler.rs, we’re using the index to retrieve a single pending commitment and verify its existence. When I need to check for a pending commitment, I only have its index—not the L1 block height—so I can’t use height-based filtering at that stage.

While it’s possible to fetch all pending commitments and search through them, doing this on every L1 block processing step would be inefficient. Instead, I suggest performing this once during the rollback phase: retrieve all pending commitments and proofs, filter by the relevant L1 height, and roll those back accordingly.

crates/fullnode/src/da_block_handler.rs
```diff
@@ -632,8 +665,10 @@ where
         let sequencer_commitment =
             if let Some(sequencer_commitment) = self.ledger_db.get_commitment_by_index(idx)? {
                 sequencer_commitment
-            } else if let Some(sequencer_commitment) =
-                self.ledger_db.get_pending_commitment_by_index(idx)?
+            } else if let Some(sequencer_commitment) = self
+                .ledger_db
+                .get_pending_commitment_by_index(idx)?
+                .map(|x| x.0)
```
jfldde:
```suggestion
      } else if let Some((sequencer_commitment, _)) =
           self.ledger_db.get_pending_commitment_by_index(idx)?
```

crates/fullnode/src/db_migrations/pending_commitments.rs
```diff
@@ -0,0 +1,56 @@
...
+impl LedgerMigration for PendingCommitmentsL1HeightFieldAdd {
+    fn identifier(&self) -> (MigrationName, MigrationVersion) {
+        ("PendingCommitmentsL1HeightFieldAdd".to_owned(), 1)
+    }
+
+    /// Pending commitments need l1 heights that they were found in.
+    /// This migration adds the l1 height field to the pending commitments
+    /// and updates the existing pending commitments with the l1 height,
+    /// but because the ones that are already in the db cannot be known,
+    /// the l1 height is set to 0.
```
jfldde: Shouldn't we force a re-index of L1 instead of setting bogus values ? We can do another round of rollback/clear-pending and use the new table format. This would even remove the need for migration ?

ercecan: I also prefer forcing a re-index instead of migrations however the ledger db won't throw errors when opening after the value type is changed, so we do not really have a valid way to force re-indexing. Thus I suggest this can stay for now and people can choose to either go with the migrated version or just re index l1

jfldde (approved):

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
```diff
@@ -516,12 +516,12 @@ define_table_with_seek_key_codec!(
...
 );
```
eyusufatik: pls comment the u64's

eyusufatik (changes requested):

eyusufatik (commented):
let's do a reindex instead of db migration like @jfldde suggested.

eyusufatik (approved):

Issue #2318, opened 2025-05-05
Replace citrea-cli clear-pending command with proper rollback handling


jfldde:
Now that we index the l1 height alongside the pending commitments/proofs, we can cleanly roll it back against target l1 height

eyusufatik:
choosing milestone v0.7.3 as we'll still need this in v0.7.2

jfldde:
Needs #2321

PR #2321, opened 2025-05-06
Refactor rollback

### LINKED ISSUE #2294: rollback_slot_by_hash is wrongly implemented
## Summary

depends on an iterator of SlotByHash by these are hash => number mappings, which cant be iterated over yielding increasing / decreasing block heights

### LINKED ISSUE #2301: Refactor rollback by separating logic based on node type

rakanalh:
# Description

## Linked Issues
- Fixes #2301 
- Fixes #2294


crates/storage-ops/src/rollback/node.rs
```diff
@@ -0,0 +1,178 @@
...
+    target_l1: u64,
+) {
+    log_result_or_error!(
+        "l2_blocks",
+        rollback_l2_blocks(
+            &ledger_db,
+            target_l2,
+            move |_ledger_db, _l2_block_number, _l2_block_hash| { Ok(()) }
+        )
+    );
```
eyusufatik: this can be removed directly

crates/storage-ops/src/rollback/node.rs
```diff
@@ -0,0 +1,178 @@
...
+    log_result_or_error!(
+        "slots",
+        rollback_batch_prover_slots(StorageNodeType::BatchProver, &ledger_db, target_l1)
+    );
+    let _ = ledger_db.put::<ProverLastScannedSlot>(&(), &SlotNumber(target_l1));
+    let _ = ledger_db.flush();
+}
+
+pub(crate) fn rollback_light_client(
```
eyusufatik: need to roll back proof by slot

crates/storage-ops/src/rollback/node.rs
```diff
@@ -0,0 +1,178 @@
...
+                l1_height, e
+            );
+        }
+        if let Err(e) = ledger_db.delete::<L2StatusHeights>(&(L2HeightStatus::Proven, l1_height)) {
+            error!(
+                "Could not delete proven L2StatusHeight at {}: {:?}",
+                l1_height, e
+            );
+        }
+    }
```
eyusufatik: I'm sure there is better way to do this @jfldde wdyt?

eyusufatik (changes requested):

eyusufatik:
also I thought we were going the rollback defined for every table seperately way as per our discussion btw?

eyusufatik (commented):
also instead of returning deleted counts from functions, can we just write how many deleted for *every table* in the function they are deleted?

crates/storage-ops/src/rollback/node.rs
jfldde: Yes that's the naive implementation I added at the time but we could only iter backwards over the `L2StatusHeights` table and break once we hit target_l1 height

crates/storage-ops/src/rollback/node/batch_prover.rs
```diff
@@ -0,0 +1,175 @@
...
+}
+
+impl LedgerNodeRollback for BatchProverLedgerRollback {
+    fn ignored_tables(&self) -> Vec<&'static str> {
+        vec![
+            "ExecutedMigrations",
+            "L2GenesisStateRoot",
+            "ProverLastScannedSlot",
+            "LastPrunedBlock",
+            "",
```
eyusufatik: ?

crates/storage-ops/src/rollback/node/light_client.rs
```diff
@@ -0,0 +1,92 @@
...
+
+            if l2_block_number <= L2BlockNumber(l2_target) {
+                break;
+            }
+
+            self.ledger_db.delete::<L2BlockByNumber>(&l2_block_number)?;
+            increment_table_counter!("L2BlockByNumber", rollback_result);
+        }
+
+        Ok(rollback_result)
```
eyusufatik: again light client prover does not need this. just ignore this table in light client prover.

crates/storage-ops/src/rollback/node/sequencer.rs
```diff
@@ -0,0 +1,153 @@
...
+
+            self.ledger_db
+                .delete::<SequencerCommitmentByIndex>(&comm_idx)?;
+            increment_table_counter!("SequencerCommitmentByIndex", rollback_result);
+        }
+
+        Ok(rollback_result)
+    }
+
+    fn rollback_slots(&self, l1_target: u64, mut rollback_result: RollbackResult) -> Result {
```
eyusufatik, line 79: there is last scanned L1 block tables for nodes. can you set the value for that table for nodes that this applies to?

crates/storage-ops/src/rollback/node/sequencer.rs
```diff
@@ -0,0 +1,153 @@
...
+    fn ignored_tables(&self) -> Vec<&'static str> {
+        vec![
+            "ExecutedMigrations",
+            "L2GenesisStateRoot",
+            "PendingSequencerCommitment",
+            "MempoolTxs",
+            "LastPrunedBlock",
+            "ProverLastScannedSlot",
+            // Unprunable
+            "ShortHeaderProofBySlotHash",
```
eyusufatik: shouldn't we rollback this?

eyusufatik (commented):
looks ok. didn't review pruning though.

crates/storage-ops/src/rollback/node/sequencer.rs
rakanalh: I am already doing it in this file line 146

crates/storage-ops/src/rollback/node/sequencer.rs
rakanalh: Since we don't have access SlotsByHash, there is no way to map a slot number to a slot hash to be able to delete short header by slot hash

Therefore, i ignored this table.

jfldde: Can't we use the l1 cache ?

rakanalh: We don't have `SlotsByHash` table for sequencer to construct the cache. The table is defined in the sequencer for the sake of table being there when we use the sequencer backup as a fullnode database.

crates/storage-ops/src/pruning/ledger/l2_blocks.rs
```diff
@@ -0,0 +1,48 @@
...
+
+        ledger_db.delete::<L2BlockByHash>(&record.value.hash)?;
+
+        if matches!(node_type, StorageNodeType::BatchProver) {
+            ledger_db.delete::<ProverStateDiffs>(&l2_block_number)?;
+        }
+
+        if matches!(node_type, StorageNodeType::FullNode) {
+            ledger_db.delete::<L2StatusHeights>(&(L2HeightStatus::Committed, l2_block_number.0))?;
+        }
```
jfldde, line 42: Should be pruned against slot

crates/storage-ops/src/rollback/node/fullnode.rs
```diff
@@ -0,0 +1,196 @@
...
+            }
+
+            self.ledger_db.delete::<L2BlockByNumber>(&l2_block_number)?;
+            increment_table_counter!("L2BlockByNumber", rollback_result);
+
+            self.ledger_db.delete::<L2BlockByHash>(&l2_block_hash)?;
+            increment_table_counter!("L2BlockByHash", rollback_result);
+
+            self.ledger_db
+                .delete::<L2StatusHeights>(&(L2HeightStatus::Committed, l2_block_number.0))?;
```
jfldde: Same as above

crates/storage-ops/src/rollback/node/fullnode.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+impl LedgerNodeRollback for FullNodeLedgerRollback {
+    fn ignored_tables(&self) -> Vec<&'static str> {
+        vec![
+            "Table ExecutedMigrations",
+            "L2GenesisStateRoot",
+            "ProverLastScannedSlot",
+            "LastPrunedBlock",
+            "PendingSequencerCommitments",
```
jfldde: This shouldn't be ignored since we now index by commitment index

jfldde: Can be taken care of in dedicated issue https://github.com/chainwayxyz/citrea/issues/2318

eyusufatik:
putting hold-merge label until v0.7.2 is released

rakanalh:
This is also pending a fix by @jfldde that we discussed earlier 

> The issue is that we were creating proof data based on pre-rolled back bitcoin da.
Now that we properly rollback blockbyhash, the L1 hash isn't valid anymore (part of the pre-rollback chain), and we are correctly hitting the Batch proof output last_l1_hash_on_bitcoin_light_client_contract isn't known

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
```diff
@@ -133,7 +133,6 @@ pub const BATCH_PROVER_LEDGER_TABLES: &[&str] = &[
 pub const LIGHT_CLIENT_PROVER_LEDGER_TABLES: &[&str] = &[
     ExecutedMigrations::table_name(),
     SlotByHash::table_name(),
-    L2BlockByNumber::table_name(),
```
jfldde, line 136: We still have calls to this table in light client prover that crashes the node with 
```
2025-06-05T07:37:01.577321Z ERROR sov_db::ledger_db: error=DB::cf_handle not found for column family name: L2BlockByNumber
2025-06-05T07:37:02.601874Z ERROR panic: thread 'main' panicked at 'Could not start light client prover: Failed to get head l2 block: DB::cf_handle not found for column family name: L2BlockByNumber': bin/citrea/src/main.rs:321
```

ercecan:
Should we also rollback LastPrunedL2Height and LastPrunedBlock iff they have a value at the time of rollback?

ercecan:
Is CommitmentIndicesByJobId rolledback? If not we can do it right after rollback of JobIdOfCommitment

ercecan (approved):
I have two questions and after https://github.com/chainwayxyz/citrea/pull/2321#discussion_r2128196635 is resolverd
LGTM

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
rakanalh: Removed query for L2 block from light client code since light client does not record L2 block numbers

rakanalh: On a second thought, it got too messy so i just restored the table.

rakanalh:
> Is CommitmentIndicesByJobId rolledback? If not we can do it right after rollback of JobIdOfCommitment

Problem with this table is that using the commitment index, we need to loop the entire table to find the commitment index and then delete the records where the match happens. We can skip this one.

eyusufatik (approved):

eyusufatik:
> > Is CommitmentIndicesByJobId rolledback? If not we can do it right after rollback of JobIdOfCommitment
> 
> Problem with this table is that using the commitment index, we need to loop the entire table to find the commitment index and then delete the records where the match happens. We can skip this one.

table should be small. we can do this

eyusufatik (changes requested):
my last comment

eyusufatik (approved):

PR #2323, opened 2025-05-06
Clean up build requirements for e2e tests

jfldde:
# Description

- Follow up of https://github.com/chainwayxyz/citrea/pull/2302
- Remove the need for `build-test` altogether. Use binary outputted by `nextest` and `coverage` and pass its path down to `citrea-e2e`
- Remove the dependency on `build` job for `coverage` and `nextest` CI jobs
- Remove local duplicate compilation on `make test` 
- Remove `test-ci` and `coverage-ci` makefile targets since they were in place to bypass this redundant compilation and are not needed anymore. Simply restored to `test` and `coverage`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ build: ## Build the project
 	@cargo build
 
 .PHONY: build-test
-build-test: $(EF_TESTS_DIR) ## Build the project
+build-test: ## Build the project
 	@cargo build --locked $(TEST_FEATURES)
 
 build-reproducible: build-sp1 ## Build the project in release mode with reproducible guest builds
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -50,16 +46,15 @@ clean-all: clean clean-node clean-txs
 test-legacy: ## Runs test suite with output from tests printed
 	@cargo test -- --nocapture -Zunstable-options --report-time
```
rakanalh, line 47: No need for `test-legacy` target since we can now use `--no-capture` with nextest

rakanalh, line 47:
```suggestion

```
No need for `test-legacy` target since we can now use `--no-capture` with nextest

rakanalh (approved):

Makefile
jfldde: How so ? 
I tried  `make test -- --nocapture` but ended up with :
```
RISC0_DEV_MODE=1 PARALLEL_PROOF_LIMIT=1 cargo nextest run -j15 --locked --workspace --all-features --no-fail-fast --nocapture
error: the argument '--test-threads <N>' cannot be used with '--no-capture'

Usage: cargo nextest run --test-threads <N> --locked --workspace --all-features --no-fail-fast --retries <N> [FILTERS]... [-- <FILTERS_AND_ARGS>...]
```

rakanalh: I forgot about the `-j`... but in general, `--no-capture` is provided by nextest if we decide to remove the `-j`.

Can we experiment without the `-j` with this PR?

eyusufatik (approved):

PR #2339, opened 2025-05-08
feat: Use Cow::Borrowed for BorshCodec::encode_key

### LINKED ISSUE #1926: StateKeyCodec::encode_key return Cow instread of Vec
For primitive keys like u64, Address, U256 we know how to serialize keys for borsh/bsc codecs: it's plain LE or BE slice of bytes of the key. We don't need to bcs::to_vec(key) or borsh::to_vec(key) to get LE/BE slice of the key (it's literally a NOP for primitives).

Given that, we may reduce 1 alloc for every .get/.set when constructing keys for Storage.

Also may fix borsh::to_vec for structs with a known size to allocate a required amount of memory.

kpp:
# Description

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/1926


eyusufatik:
I need zk benches to merge this + we are kinda in a code freeze for things that are inside stf

kpp:
Thank you @yaziciahmet!

nightly:

```
Execution stats: SessionStats { segments: 6328, total_cycles: 6634864640, user_cycles: 5385506879, paging_cycles: 1065239067, reserved_cycles: 184118694 }
```

kpp's branch:

```
Execution stats: SessionStats { segments: 6254, total_cycles: 6557270016, user_cycles: 5336957485, paging_cycles: 1038315423, reserved_cycles: 181997108 }
```

1.3% drop in cycles

eyusufatik:
%1 is not a joke. but I'm not entirely sure about merging this 🤔 

will this require any storage rewrites etc.? or just pure functional changes which we can sneak into an upgrade?

if so, I'd like to put this in mainnet's first upgrade.

kpp:
> will this require any storage rewrites etc.?

No. This PR must construct same key/value pairs as before on the storage. The problem can be in the future only with endianess. As long as we run it on the same endianess we are good.

Here is where the problem can be:

- `key.as_bytes())` for i32, u32, u64, usize
- alloy or ruint U256 - see https://github.com/recmo/uint/blob/9bc4c717fbe126dabaa722489a284021f404652f/src/support/borsh.rs for explanation how it can affect LE/BE.  

UPD:

Fixed in https://github.com/chainwayxyz/citrea/pull/2339/commits/51109a95d90662156df101a89c709bab5b618be0

guests/risc0/batch-proof/bitcoin/Cargo.lock
```diff
@@ -2595,7 +2595,7 @@ version = "0.2.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "77957b295656769bb8ad2b6a6b09d897d94f05c41b069aede1fcdaa675eaea04"
 dependencies = [
- "zerocopy",
+ "zerocopy 0.7.35",
```
eyusufatik: maybe we should just use this version of zerocopy so that we don't have two?

jfldde: Let me look into it

jfldde (approved):

eyusufatik (approved):

PR #2342, opened 2025-05-09
Rename L2HeightStatus to L2CommitmentStatus

rakanalh:
# Description

We have a table called `L2StatusHeight` which is confused with `L2HeightStatus` enum.

The better name for `L2HeightStatus` enum is `L2CommitmentStatus`


jfldde (approved):

crates/sovereign-sdk/full-node/db/sov-db/src/schema/types/mod.rs
```diff
@@ -66,7 +66,7 @@ pub struct L2HeightAndIndex {
 )]
 #[repr(u8)]
 #[borsh(use_discriminant = true)]
-pub enum L2HeightStatus {
+pub enum L2CommitmentStatus {
```
eyusufatik, line 71: this is L2 height's status though

rakanalh: We use this Commitment status and L2 height to get the committed height.
```
        if let Some(committed_height) = self
            .ledger_db
            .get_highest_l2_height_for_status(L2CommitmentStatus::Committed, None)?
        {
```

rakanalh: Also defined:
```
pub enum L2CommitmentStatus {
    /// Commmited sequencer commitment
    Committed = 0,
    /// Proven sequencer commitment
    Proven = 1,
}
```

jfldde: True... but I always thought that the naming from this PR was confusing as well. Maybe `L2Status` so that it just takes care of representing the various status a l2 can be ?

rakanalh: `Committed` and `Proven` represent a commitment rather than an L2 block or are we referring to L2 being committed or proven?

jfldde: image

Main usage for now is by batch explorer landing page :
https://citrea.xyz/batch-explorer?page=1&limit=10
Which gives `Last Commited Height` and `Last Proven Height`, so we're interested in the L2 height itself. 

I'd say the rename is correct since we index both the commitment height and index + we recently changed to return the index alongside the height in the RPC response

kpp (approved):

PR #2344, opened 2025-05-12
Feature gate imports & methods for native

rakanalh:
# Description

These warnings were always bugging me.


rakanalh:
```
batch-proof-bitcoin:   = note: the `riscv32im-risc0-zkvm-elf` target may not be installed
batch-proof-bitcoin:   = help: consider downloading the target with `rustup target add riscv32im-risc0-zkvm-elf`
```

```
error: component 'rust-std' for target 'riscv32im-risc0-zkvm-elf' is unavailable for download for channel '1.85'
```

PR #2347, opened 2025-05-12
Short header proof unit tests

### LINKED ISSUE #2312: Add short proof provider fail tests

yaziciahmet:
# Description

Implemented both successful and fail test cases for native and zk short header proof providers.

## Linked Issues
- Fixes #2312 

## Testing

## Docs


eyusufatik (commented):
as you are running the same tests for native and zk, cant you just make a tests.rs and add those tests here which would run the test case against both impls?

crates/short-header-proof-provider/tests/shp_integration.rs
```diff
@@ -0,0 +1,312 @@
...
+    let proof_bytes = borsh::to_vec(&mock_proof).unwrap();
+    ledger_db
+        .put_short_header_proof_by_l1_hash(&block_hash, proof_bytes.clone())
+        .unwrap();
+
+    let ok = native_service
+        .get_and_verify_short_header_proof_by_l1_hash(
+            block_hash, [5u8; 32], // different prev_block_hash
+            101,       // different height
+            [6u8; 32], // different txs_commitment
```
eyusufatik: also make instances where you only passs single incorrect value and other arguments are correct.

eyusufatik (changes requested):

crates/short-header-proof-provider/tests/shp_integration.rs
yaziciahmet: https://github.com/chainwayxyz/citrea/pull/2347/commits/afd5854542dbe49e0ee163571ed580d1ede0ff25

eyusufatik (approved):

PR #2348, opened 2025-05-12
Test: Proof using circuit input directly

### LINKED ISSUE #2337: Call `batchProver_createInput` in a test and then invoke a batch proof proving session inside the test
From the input returned from RPC, theoretically, you can successfully prove. Let's see that it works.

rakanalh:
## Linked Issues
- Fixes #2337


jfldde (commented):
I don't understand the test flow. We added a RPC to get circuit inputs but there's no way for us to use it without going through internal methods ? What will be the use case for this RPC ?

eyusufatik:
> I don't understand the test flow. We added a RPC to get circuit inputs but there's no way for us to use it without going through internal methods ? What will be the use case for this RPC ?

this test is to make sure createInput rpc works ok.

after we get the input on our end, we can invoke r0vm on local, call bonsai or call boundless to actually prove anything

bin/citrea/tests/bitcoin/batch_prover_test.rs
```diff
@@ -1374,9 +1377,15 @@ impl TestCase for BatchProverCreateInputTest {
...
+            let proof = proof.proof;
+            let output: BatchProofCircuitOutput = Risc0Host::extract_output(&proof).unwrap();
+
             // Verify the proof
-            Risc0Host::verify(&proof.proof.as_slice(), &code_commitment)
+            Risc0Host::verify(proof.as_slice(), &code_commitment)
                 .expect("Proof verification failed");
+
+            assert_eq!(output.last_l2_height(), max_l2_blocks_per_commitment);
+            assert_eq!(output.sequencer_commitment_index_range(), (1, 1));
```
eyusufatik, line 1385: add a state root assert also

eyusufatik (changes requested):

jfldde (approved):

eyusufatik (approved):

PR #2350, opened 2025-05-13
Tests for get_last_l1_hash_on_contract

### LINKED ISSUE #2330: Test incorrect last_l1_hash witness for batch proof
Create a proving input that has a wrongful l1_hash witness and see that it fails. 

by the way this can be done by invoking `get_last_l1_hash_on_contract` in a test the regular way, don't need to run the zk program.

rakanalh:
# Description

## Linked Issues
- Fixes #2330


bin/citrea/tests/bitcoin/batch_prover_test.rs
```diff
@@ -1414,3 +1418,77 @@ async fn batch_prover_create_input_test() -> Result<()> {
...
+        });
+
+        // Assert that the call fails
+        assert!(
+            result.is_err(),
+            "Expected failure for incorrect L1 hash witness"
+        );
+
+        sequencer.wait_until_stopped().await?;
+        batch_prover.wait_until_stopped().await?;
```
jfldde: Not needed

eyusufatik: the pr will go into a unit test way actually so doesn't matter

jfldde: Another good test would be the current failure on rollback refactor `test_out_of_range_proof`, i.e. create a proof, rollback bitcoin, generate some blocks, re-send the proof and make sure it hits `ProofError::UnknownL1Hash`

crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
eyusufatik, line 1: I'd prefer not making any changes in-stf code for a while

crates/citrea-stf/src/verifier.rs
```diff
@@ -195,3 +195,209 @@ where
...
+        let mut bytes = [0u8; 64];
+        bytes[0..32].copy_from_slice(&U256::from(0).to_be_bytes::<32>());
+        bytes[32..64].copy_from_slice(&U256::from(1).to_be_bytes::<32>());
+        let evm_storage_slot = keccak256(bytes).into();
+        let inner_evm_key = Evm::<DefaultContext>::get_storage_address(
+            &BITCOIN_LIGHT_CLIENT_CONTRACT_ADDRESS,
+            &evm_storage_slot,
+        );
+        let key = StorageKey::new(&prefix, &inner_evm_key, &BorshCodec);
+        working_set.set(&key, StorageValue::new(&U256::from(0), &BorshCodec));
```
eyusufatik: it'd be easier to distinguish between existing and non-existing values for a height by just setting a different value then 0 here

crates/citrea-stf/src/verifier.rs
```diff
@@ -195,3 +195,209 @@ where
...
+        let result = get_last_l1_hash_on_contract::<DefaultContext>(
+            state_log,
+            prover_storage,
+            &mut witness,
+            final_state_root,
+        );
+
+        // Assert the result is as expected (mocked value)
+        assert_eq!(result, [0u8; 32], "Expected default hash value");
+    }
```
eyusufatik, line 672: let's add three more tests:
- height is cached but hash is not
- height is not cached but hash is
- make get_and_prove() call fail.
  - this one will require a more complex test though. instead of calling `get_last_l1_hash_on_contract` with ProverStorage (so DefaultContext), we'll call it with a ZkStorage with witness received from ProverStorage, so that the call can make the JMT proof verifications inside. to make those fail, you can check out my latest LCP state tests pr.

eyusufatik (changes requested):

eyusufatik (approved):

PR #2353, opened 2025-05-13
Add --all-features to make lint-fix

yaziciahmet:
# Description

Latest nightly `make lint-fix` cmd fails due to missing `--testing` feature. Fixes it.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,8 +91,8 @@ lint:  ## cargo check and clippy. Skip clippy on guest code since it's not suppo
 
 lint-fix:  ## dprint fmt, cargo fmt, fix and clippy. Skip clippy on guest code since it's not supported by risc0
 	dprint fmt
-	cargo fix --allow-dirty
-	SKIP_GUEST_BUILD=1 cargo clippy --fix --allow-dirty
+	cargo fix --allow-dirty --all-features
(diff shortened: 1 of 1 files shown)


jfldde (approved):

kpp (approved):

eyusufatik:
why did this change though? what did we change?

yaziciahmet:
> why did this change though? what did we change?

The problem occurred in full node bitcoin e2e test `FullNodeLcpChunkProofTest` due to usage of method `BitcoinService::test_send_separate_chunk_transaction_with_fee_rate` that is guarded behind `--testing` flag. I believe we never had test utility method within the actual crate implementation but now we do.

jfldde:
Maybe we should keep this `test_send_separate_chunk_transaction_with_fee_rate` in full_node test file instead and remove the need for testing flag

yaziciahmet:
> Maybe we should keep this `test_send_separate_chunk_transaction_with_fee_rate` in full_node test file instead and remove the need for testing flag

this will require exposing inner fields of `BitcoinService`, which I am not really fond of either.

eyusufatik (approved):

PR #2354, opened 2025-05-13
Test: Undecompressable data on DA

### LINKED ISSUE #2325: Test undecompressable data in complete proof
Somehow make bitcoin-da send undecompressable data to DA. Make sure light client prover can process the L1 block containing the undecompressable tx.

jfldde:
# Description

- Send an undecompressable tx to bitcoin da and make sure that light client prover is able to process the block and ignores the tx

## Linked Issues
- Fixes #2325

### DIFF
--- a/bin/citrea/tests/bitcoin/utils.rs
+++ b/bin/citrea/tests/bitcoin/utils.rs
@@ -1,12 +1,8 @@
 use std::path::PathBuf;
-use std::str::FromStr;
 use std::sync::Arc;
 
-use bitcoin::secp256k1::SecretKey;
-use bitcoin_da::helpers::builders::body_builders::{create_inscription_type_0, DaTxs};
 use bitcoin_da::service::{BitcoinService, BitcoinServiceConfig};
 use bitcoin_da::spec::RollupParams;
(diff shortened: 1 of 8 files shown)


bin/citrea/tests/bitcoin/utils.rs
```diff
@@ -65,3 +70,44 @@ pub(super) async fn spawn_bitcoin_da_service(
 
     bitcoin_da_service
 }
+
+pub async fn create_complete_tx_with_prefix(
```
kpp: Why put this fn here if it's used only in a specific test case?

kpp (approved):

bin/citrea/tests/bitcoin/utils.rs
jfldde: No good reason, switching it back

eyusufatik (changes requested):
looks good. 

can we also do the same for undecompressable chunks?

make random vec of size n

divide to chunks and send with aggregator. it should be ignored again.

eyusufatik (approved):

PR #2357, opened 2025-05-14
LCP storage proof tests

### LINKED ISSUE #2047: Test that tries to mislead light client proof circuit about state
Previously when we had MMR proofs in the light client proof circuit, we had tests that played around with MMR proofs to mislead the LCP. Now we use storage api and we don't have any tests (outside the storage api) that tries to mislead the circuit.

Let's implement one under `light-client-prover/tests/`

eyusufatik:
## Linked Issues
Closes #2047

### DIFF
--- a/crates/light-client-prover/src/tests/mod.rs
+++ b/crates/light-client-prover/src/tests/mod.rs
@@ -2360,3 +2360,117 @@ fn test_lcp_input_values_cant_be_tampered() {
 // with valid read and update proofs
 // essentially trying to hack the circuit by providing values and proofs from a different tree
 // we make it similar so the circuit won't panic but will follow until the storage verification part
+#[should_panic = "Witness prev root is wrong!"]
+#[test]
+fn test_lcp_cant_be_passed_roots_from_a_different_tree() {
+    // set up the test environment with
+    // a few sequencer commitments
+    // and batch proofs that will only process the first sequencer commitment
+    // is enough
+    let db_dir = tempdir().unwrap();
+    let native_circuit_runner = NativeCircuitRunner::new(db_dir.path().to_path_buf());
+    let zk_circuit_runner = LightClientProofCircuit::<ZkStorage, MockDaSpec, MockZkGuest>::new();
+
+    let light_client_proof_method_id = [1u32; 8];
+    let da_verifier = MockDaVerifier {};
+
+    let block_header_1 = MockBlockHeader::from_height(1);
+
+    let seq_comm_1 = create_mock_sequencer_commitment(1, 2, [2u8; 32]);
+    let seq_comm_2 = create_mock_sequencer_commitment(2, 3, [3u8; 32]);
+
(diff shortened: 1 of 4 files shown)


crates/light-client-prover/src/tests/mod.rs
```diff
@@ -2157,3 +2159,207 @@ fn wrong_pubkey_sequencer_commitment_should_not_work() {
     // ignored commitment from wrong pubkey
     assert_eq!(commitment, None);
 }
+
+// For some reason, even though macro is used, it sees it as unused
+#[allow(unused)]
+macro_rules! assert_panic {
+    // Match a single expression
```
yaziciahmet: I believe its because macro definitions should actually lie in lib.rs of the crate, and there is some weird rules around defining macros except there. Thats why a lot of external crates have separate crate for defining macros.

crates/sovereign-sdk/rollup-interface/src/state_machine/witness.rs
```diff
@@ -40,4 +40,17 @@ impl Witness {
...
+    pub fn get_hints(&self) -> VecDeque<Vec<u8>> {
+        self.hints.clone()
+    }
+}
+
+#[cfg(feature = "testing")]
+impl From<VecDeque<Vec<u8>>> for Witness {
+    fn from(hints: VecDeque<Vec<u8>>) -> Self {
+        Self { hints }
+    }
```
yaziciahmet, line 55: imo makes sense to guard get_hints with testing feat, but From impl is fine to expose

yaziciahmet (commented):
Made initial comments, will recheck once the last part of the test is implemented

crates/light-client-prover/src/tests/mod.rs
```diff
@@ -2157,3 +2159,207 @@ fn wrong_pubkey_sequencer_commitment_should_not_work() {
     // ignored commitment from wrong pubkey
     assert_eq!(commitment, None);
 }
+
+// For some reason, even though macro is used, it sees it as unused
+#[allow(unused)]
+macro_rules! assert_panic {
```
kpp: https://doc.rust-lang.org/book/ch11-01-writing-tests.html#checking-for-panics-with-should_panic

kpp (dismissed):

crates/light-client-prover/src/tests/mod.rs
eyusufatik: I know about that. not sure how many individual cases I'll have, so maybe setting up different tests for all maybe cumbersome, keeping it this way for now. If wee end up with < 3 cases I'll set up seperate tests with should panic.

crates/sovereign-sdk/rollup-interface/src/state_machine/witness.rs
eyusufatik: we can take it out of testing if we ever need it

yaziciahmet (approved):

PR #2360, opened 2025-05-15
Test SetBlockInfo for ShortHeaderProofNotFound and ShortHeaderProofVerificationError

### LINKED ISSUE #2332: Test EVM SetBlockInfo system tx short header proof verification fail
SetBlockInfo should be rejected

kpp:
## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/2332


eyusufatik (approved):
exactly what we need. thanks.

can you also make asserts after EVM calls that fail to check that contract is really not updated? (this is pretty much given but does not hurt :) )

eyusufatik (approved):

PR #2377, opened 2025-05-19
Full node halts on merkle root mismatch to behave like lcp

### LINKED ISSUE #2349: Full node should halt as lcp does after a sequencer commitment with wrong merkle root
Currently there is no way to verify the merkle root of a sequencer commitment in the lcp circuit, when it sees a sequencer commitment with an index it has never seen before it will just save it to its state. When the time comes to process a proof with that false commitment it will not be able to verify the proof thus the lcp will halt.

According to this we expect the full node to behave the same way so it should also halt in that situation

ercecan:
# Description

## Linked Issues
- Fixes #2349

### DIFF
--- a/crates/fullnode/src/da_block_handler.rs
+++ b/crates/fullnode/src/da_block_handler.rs
@@ -24,7 +24,7 @@ use tokio::sync::Mutex;
 use tokio::time::Duration;
 use tracing::{debug, error, info, instrument, warn};
 
-use crate::error::ProofError;
+use crate::error::{CommitmentError, HaltingError, ProcessingError, ProofError, SkippableError};
 use crate::metrics::FULLNODE_METRICS;
 
 enum ProcessingResult {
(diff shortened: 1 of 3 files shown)


crates/fullnode/src/da_block_handler.rs
```diff
@@ -501,11 +538,11 @@ where
                 })?;
 
         if start_state_root.as_ref() != initial_state_root.as_ref() {
-            return Err(anyhow!(
-                    "Proof verification: For a known and verified sequencer commitment. Pre state root mismatch - expected 0x{} but got 0x{}. Skipping proof.",
-                    hex::encode(initial_state_root),
-                    hex::encode(start_state_root)
-                ).into());
+            return Err(SkippableError::Proof(ProofError::PreStateRootMismatch(
```
eyusufatik, line 541: should this really be skippable? what happens on lcp

eyusufatik (approved):

crates/fullnode/src/da_block_handler.rs
ercecan: I mean this check is after the previous commitment merkle root verification and previous commitment hash check
If all are valid this normally should not happen

PR #2382, opened 2025-05-20
Fix vscode build (missing cfg feature guard)

kpp:
# Description

test full_node.rs uses `test_send_separate_chunk_transaction_with_fee_rate` which is defined for `feature = "testing"` only.


eyusufatik (commented):
don't we need to run cargo test with --all-features?

kpp:
That's true. However our code must be able to compile without features. I have a compilation error in vscode without this PR:

image

eyusufatik (approved):

PR #2383, opened 2025-05-21
Change proving_session_cycle_count from hist to gauge

### LINKED ISSUE #2372: Session cycle counts stats not usable

kpp:
# Description

Here is how gauge looks in prometheus and grafana:

image

image

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/2372

### DIFF
--- a/crates/risc0/src/host/bonsai.rs
+++ b/crates/risc0/src/host/bonsai.rs
@@ -4,7 +4,7 @@ use std::time::Duration;
 use anyhow::{anyhow, Context};
 use bonsai_sdk::blocking::{Client, SessionId, SnarkId};
 use bonsai_sdk::responses::SessionStats;
-use metrics::histogram;
+use metrics::gauge;
 use risc0_zkvm::{
     compute_image_id, AssumptionReceipt, Digest, InnerAssumptionReceipt, Receipt, VerifierContext,
 };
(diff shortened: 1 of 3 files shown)


rakanalh (commented):
Makes sense for it to be a gauge. I believe it was before but wasn't showing us what we expected it to show.

@eyusufatik Can we test this on Devnet before merging this PR?

eyusufatik:
> Makes sense for it to be a gauge. I believe it was before but wasn't showing us what we expected it to show.
> 
> @eyusufatik Can we test this on Devnet before merging this PR?

we can test it with 0.7.2-rc.1 that will go to devnet. 

maybe the gauge was not working as expected before as we were not calling .set bug .increase / .decrease?

kpp:
```
pub trait GaugeFn {
    /// Increments the gauge by the given amount.
    fn increment(&self, value: f64);

    /// Decrements the gauge by the given amount.
    fn decrement(&self, value: f64);

    /// Sets the gauge to the given amount.
    fn set(&self, value: f64);
}
```

I believe increment/decrement is good for let's say increment(new_active_sessions)/decrement(terminated_sessions). For absolute values we have to use set.

eyusufatik (approved):

Issue #2031, opened 2025-03-05
Support more evm tracers


eyusufatik:
now that we have DbRef implemented thanks to #2029 we can implement these tracers:
- MuxTracer
- FlatCallTracer
- PrestateTracer

PR #2386, opened 2025-05-22
More tracer support for transactions

rakanalh:
# Description

## Linked Issues
- Related to #2031


crates/ethereum-rpc/src/lib.rs
```diff
@@ -372,6 +375,9 @@ where
         )
         .map_err(to_eth_rpc_error)?;
 
+        if traces.is_empty() {
```
eyusufatik: this must be infallible as we return error above on tx not found

crates/ethereum-rpc/src/lib.rs
```diff
@@ -364,6 +366,7 @@ where
 
         let traces = debug_trace_by_block_number(
             block_number,
+            Some(tx_hash),
```
eyusufatik: there is already tx index, why pass tx_hash?

crates/ethereum-rpc/src/trace.rs
```diff
@@ -243,6 +277,31 @@ fn get_traces_with_requested_tracer_and_config(
...
+                            )?;
+                            localized_call_frames.push(new_flat_call_frame);
+                        }
+                    }
+                    new_traces.push(TraceResult::new_success(
+                        GethTrace::FlatCallTracer(localized_call_frames),
+                        tx_hash,
+                    ));
+                    Ok(new_traces)
+                }
```
eyusufatik: @ercecan pls check if this conversion is correct.

crates/ethereum-rpc/src/trace.rs
```diff
@@ -243,6 +277,31 @@ fn get_traces_with_requested_tracer_and_config(
...
+                GethDebugBuiltInTracerType::PreStateTracer => Ok(traces),
```
eyusufatik: as we skip cache on prestatetracer, we should never reach here, just like jstracer

crates/ethereum-rpc/src/trace.rs
```diff
@@ -275,6 +334,93 @@ fn get_traces_with_requested_tracer_and_config(
     }
 }
 
+fn convert_call_trace_into_flatcall_frame(
```
eyusufatik, line 333: if this code is taken from somewhere pls give reference.

crates/ethereum-rpc/src/trace.rs
```diff
@@ -201,6 +231,10 @@ fn get_traces_with_requested_tracer_and_config(
     traces: Vec<TraceResult>,
     tracer: GethDebugTracerType,
     tracer_config: GethDebugTracerConfig,
+    block_number: u64,
+    block_hash: Option<BlockHash>,
+    tx_hash: Option<TxHash>,
```
eyusufatik: tx_hash is never used.

bin/citrea/tests/mock/evm/tracing.rs
eyusufatik, line 1: test debug_trace_block_by_number with new tracers

eyusufatik (changes requested):

crates/ethereum-rpc/src/lib.rs
rakanalh: It is needed here:

https://github.com/chainwayxyz/citrea/pull/2386/files#diff-4cde329fa6d8b252aace344bef21935e2b3847909574a425d596d65883b0b6a0R419-R420

crates/ethereum-rpc/src/trace.rs
rakanalh: It is, please check again.

eyusufatik: but wherever it's passed to the tx_hash is actually extracted from tradce result

crates/ethereum-rpc/src/trace.rs
ercecan: after reading the code from 
https://github.com/ethereum/go-ethereum/blob/20ad4f500e7fafab93f6d94fa171a5c0309de6ce/eth/tracers/native/call_flat.go#L250

I can see there are some missing stuff to be done if I'm not mistaken,
1- Recursive traversal of callframe calls are missing, in this line we can see a recursive approach to get every call flattened out of each frame:
 https://github.com/ethereum/go-ethereum/blob/20ad4f500e7fafab93f6d94fa171a5c0309de6ce/eth/tracers/native/call_flat.go#L281
 2-  Trace adress can be propagated I guess because currently they are all empty, I believe it can be done by adding a parameter like `trace_address: Vec<usize>` to the fn and can take the impl from 
https://github.com/ethereum/go-ethereum/blob/20ad4f500e7fafab93f6d94fa171a5c0309de6ce/eth/tracers/native/call_flat.go#L381

3- also the blockhash and blocknumber etc I believe should be injected in callframe just like in this function:
https://github.com/ethereum/go-ethereum/blob/20ad4f500e7fafab93f6d94fa171a5c0309de6ce/eth/tracers/native/call_flat.go#L348

@rakanalh Please correct me if I’m wrong — I’m just making these assumptions based on the differences I see in the code

bin/citrea/tests/mock/evm/tracing.rs
rakanalh: Added

crates/ethereum-rpc/src/trace.rs
rakanalh: Good feedback... i missed the 1st part.

Regarding
> 3- also the blockhash and blocknumber etc I believe should be injected in callframe just like in this function:
https://github.com/ethereum/go-ethereum/blob/20ad4f500e7fafab93f6d94fa171a5c0309de6ce/eth/tracers/native/call_flat.go#L348

That's the context which is passed as a struct from outside. The context is initialized with block number / hash .. etc. The `CallFrame` itself doesn't contain these values, otherwise i would have used them as you suggested.

rakanalh: I addressed the first two points

crates/ethereum-rpc/src/trace.rs
rakanalh: I am not sure i follow. Can you please elaborate?

crates/ethereum-rpc/src/trace.rs
```diff
@@ -275,6 +333,116 @@ fn get_traces_with_requested_tracer_and_config(
...
+        }
+    };
+
+    let frame = LocalizedTransactionTrace {
+        trace,
+        block_hash,
+        block_number,
+        transaction_hash: tx_hash,
+        transaction_position: tx_index.map(|i| i as u64),
+    };
```
ercecan: You also need to assign the trace_address to the top level trace otherwise it's always empty

```suggestion
    let mut frame = LocalizedTransactionTrace {
        trace,
        block_hash,
        block_number,
        transaction_hash: tx_hash,
        transaction_position: tx_index.map(|i| i as u64),
    };
    
    frame.trace.trace_address = trace_address.clone();
```

ercecan (changes requested):

crates/ethereum-rpc/src/trace.rs
ercecan: > That's the context which is passed as a struct from outside. The context is initialized with block number / hash .. etc. The CallFrame itself doesn't contain these values, otherwise i would have used them as you suggested.

You're right I missed that

Changes looking good, just one small fix:
https://github.com/chainwayxyz/citrea/pull/2386#discussion_r2106801235

crates/ethereum-rpc/src/trace.rs
eyusufatik: https://github.com/chainwayxyz/citrea/pull/2386/commits/55c1463feb48fe89a364f175083d07d851f5538d

crates/ethereum-rpc/src/lib.rs
eyusufatik: https://github.com/chainwayxyz/citrea/pull/2386/commits/55c1463feb48fe89a364f175083d07d851f5538d

eyusufatik (commented):
for flat call let's deploy a contract and trace that pls

crates/ethereum-rpc/src/trace.rs
rakanalh: https://github.com/chainwayxyz/citrea/pull/2386/commits/dc7de77b70863f38ae4d2760f7d35f2c59c9d7e0

rakanalh:
> for flat call let's deploy a contract and trace that pls

It should already be there:
https://github.com/chainwayxyz/citrea/blob/dc7de77b70863f38ae4d2760f7d35f2c59c9d7e0/bin/citrea/tests/mock/evm/tracing.rs#L527-L592

eyusufatik (approved):

ercecan (approved):
~~approved but Do not merge 1 second please~~
Okay lgtm

Issue #2090, opened 2025-03-14
Maybe remove sov-tx chainid field


eyusufatik:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #2387, opened 2025-05-23
Remove sov-tx chainid, Fix sov-tx signature

### LINKED ISSUE #2370: Transaction stores borsh serialized signature
> Why do we store borsh(K256Signature) in TransactionV1 instead of K256Signature itself?

@kpp 

[https://github.com/chainwayxyz/citrea/blob/de51548b7cd6b6ff61b02491a4e59097877e38c[…]sovereign-sdk/rollup-interface/src/state_machine/transaction.rs](https://github.com/chainwayxyz/citrea/blob/de51548b7cd6b6ff61b02491a4e59097877e38c2/crates/sovereign-sdk/rollup-interface/src/state_machine/transaction.rs#L42)

We probably missed this we can store signature it self not the serialized version of it

kpp:
# Description

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/2370 https://github.com/chainwayxyz/citrea/issues/2090

### DIFF
--- a/crates/sequencer/src/runner.rs
+++ b/crates/sequencer/src/runner.rs
@@ -746,7 +746,7 @@ where
         // TODO: figure out what to do with sov-tx fields
         // chain id gas tip and gas limit
 
-        let tx = Transaction::new_signed_tx(&self.sov_tx_signer_priv_key, raw_message, 0, nonce);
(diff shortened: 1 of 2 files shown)


eyusufatik:
Good solution. Should we make this when we do a network upgrade?

PR #2388, opened 2025-05-23
Refactor sequencer commitment

yaziciahmet:
# Description

Sequencer commitment logic had couple of bugs and off by one errors. This is a fix for those.

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/sequencer/src/commitment/helpers.rs
```diff
@@ -1,21 +1,25 @@
...
-    let max_db = max_pending.max(max_last);
-    if let Some(max_db) = max_db {
-        max_db + 1
-    } else {
-        // if comms are empty, then index is 1
-        1
+
+    match max_commitment {
+        Some(commitment) => (commitment.index + 1, commitment.l2_end_block_number + 1),
+        None => (1, get_tangerine_activation_height_non_zero()),
```
eyusufatik, line 23: @ercecan is the height correct?

crates/sequencer/src/commitment/service.rs
```diff
@@ -284,7 +225,7 @@ where
                 .ledger_db
                 .get_last_commitment()?
                 .map(|c| c.l2_end_block_number)
-                .unwrap_or(1);
+                .unwrap_or(0);
```
eyusufatik: why change this?

crates/sequencer/src/commitment/controller.rs
```diff
@@ -10,163 +10,139 @@ use sov_db::schema::types::L2BlockNumber;
...
+            Some(range) => range,
+            None => match self.check_max_l2_blocks(l2_height) {
+                Some(range) => {
+                    // we clear the state diff when max l2 blocks triggers commitment
+                    *self.state_diff.lock() = vec![];
+                    range
+                }
+                None => return Ok(None),
+            },
+        };
```
eyusufatik, line 77: should the order be reverse?

yaziciahmet: no, if max comm l2 blocks is 100, and 100th block just arrived and causes state diff to exceed limit, it should create commitment 1-99. if we were to check max l2 blocks first it would have created 1-100.

crates/sequencer/src/commitment/helpers.rs
ercecan: yea looks good, maybe a unit test here could be nice as well

ercecan: yea looks good, maybe a unit test here could be nice as well

crates/sequencer/src/commitment/service.rs
yaziciahmet: I actually am not 100% sure about this one. But my thought process was, if we are setting the evm state to last commitment end block's state, if there are no commitments found, we should use state of block 0.

eyusufatik: no it should stay 1. if that's 0, there is no hash on the bitcoin light client contract. add some docs while fixing it pls

eyusufatik (approved):

PR #2390, opened 2025-05-23
Big witness cache prune test

### LINKED ISSUE #2336: Invoke `CacheLog::prune_half` in an e2e workflow

yaziciahmet:
# Description

Trying to create a test with big enough witness so that it triggers cache pruning. Found out off by one bug in the cache pruning logic. Also increased the cache size to 512mb as now zkvm memory limit is 3gb

## Linked Issues
- Fixes #2336 

## Testing

## Docs


bin/citrea/tests/bitcoin/batch_prover_test.rs
```diff
@@ -1320,3 +1326,141 @@ async fn batch_prover_create_input_test() -> Result<()> {
...
+        batch_prover
+            .wait_for_l1_height(finalized_height, None)
+            .await?;
+
+        // Wait for batch proof transactions to hit the mempool
+        // In this proof, cache limit of 8MB will be hit and pruning will occur.
+        // If the proving session ended successfully, we are gucci
+        da.wait_mempool_len(2, None).await?;
+
+        Ok(())
```
eyusufatik: make sure a full node can verify the proof. if we change behaviour of the full node in the future for example (e.g. no panic on error in proof) this test might become useless.

and wait_mempool_len is not enough as sequencer can commit.

so the best way to go is to have the full node verify the proof. (you can ask citrea_getLastProvenHeight etc.)

yaziciahmet: added full node and asserted the last proven l2 height on full node.

I think for this case wait_mempool_len is sufficient as sequencer is in test mode and can't really produce commitments. currently we produce 60 blocks, and 55th block hits state diff and produces commitment. in order for this test to be invalid, we would have to make such changes that it will hit state diff before 30th block, causing 2 commitments, but that big of a change will require a lot of changes in a lot of tests anyways.

eyusufatik (approved):

PR #2391, opened 2025-05-23
feat: Boundless adapter + rust 1.88 + risc0 3.0.3

### LINKED ISSUE #1308: Risc0 Boundless adapter

ercecan:
# Description
Implements boundless adapter for proving
**Tested light client proving with batch proofs** ✅ 
**Tested session recovery** ✅ 
What's left:
- **API WIP on another repo, will be integrated in a different PR**~~Deciding on how to handle pricing for each proof request.~~
- **Implemented** ✅  ~~Also might want to wait on https://github.com/boundless-xyz/boundless/pull/515~~
- ~~Need to manually test price bumping on timedout requests~~ Good thing I tested this, there were bugs and now they are fixed ✅ 

## Linked Issues
- Fixes #1308


crates/risc0/Cargo.toml
yaziciahmet, line 1: even though the newly added dependencies are only used here, we basically put almost everything in the workspace dependencies. maybe its better to move them to root?

Cargo.lock
```diff
@@ -8940,6 +10726,25 @@ dependencies = [
...
+ "anyhow",
+ "cfg-if",
+ "hex",
+ "risc0-aggregation",
+ "risc0-zkvm",
+ "serde",
+ "thiserror 2.0.12",
+ "tracing",
+]
+
```
yaziciahmet, line 10889: this adds waay too many dependencies. I see that 2.0.2 has alloy 0.13, same as us. are there any features in 2.1.0 that we specifically require? if not maybe we can drop this to 2.0.2? if that doesn't sound ok, we should create an issue for upgrading our alloy dependencies to match with this.

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,351 @@
...
+            boundless_market::alloy::providers::fillers::WalletFiller<
+                boundless_market::alloy::network::EthereumWallet,
+            >,
+        >,
+        boundless_market::balance_alerts_layer::BalanceAlertProvider<
+            boundless_market::alloy::providers::RootProvider,
+        >,
+    >,
+    boundless_market::storage::BuiltinStorageProvider,
+>;
```
yaziciahmet: can you explain this to me

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,351 @@
...
+    order_stream_url: Option<String>,
+}
+
+impl citrea_common::FromEnv for BoundlessConfig {
+    fn from_env() -> anyhow::Result<Self> {
+        let wallet_private_key = read_env("WALLET_PRIVATE_KEY")?;
+        let rpc_url = read_env("RPC_URL")?;
+        let boundless_market_address = read_env("BOUNDLESS_MARKET_ADDRESS")?;
+        let set_verifier_address = read_env("SET_VERIFIER_ADDRESS")?;
+        let order_stream_url = read_env("ORDER_STREAM_URL").ok();
```
yaziciahmet: can you add BOUNDLESS prefix to each of these envs? otherwise it becomes a bit confusing (e.g. just RPC_URL)

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,351 @@
...
+    async fn boundless_client() -> anyhow::Result<BoundlessClient> {
+        let config = BoundlessConfig::from_env().expect("Failed to load boundless config");
+
+        let local_signer = PrivateKeySigner::from_str(&config.wallet_private_key)
+            .context("Failed to parse wallet private key")?;
+
+        // If in dev mode, uses a temporary file as storage provider
+        // Otherwise first tries to parse pinata env variables
+        // If fails then tries to parse s3 env variables
+        let storage_provider = BuiltinStorageProvider::from_env().await.ok();
```
yaziciahmet: can we either document or pre-verify the environment variables that this requires?

yaziciahmet: also in the prove fn we have an no dev mode assertion, hence, we shouldn't actually build this with dev mode as far as I understand

yaziciahmet (changes requested):

crates/risc0/src/host/boundless.rs
eyusufatik: I'm sure there is another way to do this

ercecan: Will be cleaner with the next release

Cargo.lock
ercecan: It was not actually used so I removed `risc0-ethereum-contracts` and `risc0-build-ethereum`but the boundless-market crate unfortunately has all the bloat still

crates/risc0/src/host/boundless.rs
ercecan: > also in the prove fn we have an no dev mode assertion, hence, we shouldn't actually build this with dev mode as far as I understand

Yes

ercecan: [10e7078](https://github.com/chainwayxyz/citrea/pull/2391/commits/10e7078cd06fa4c5b7d4e220770adaf08a0233df)
Is this good enough docs or should I still pre-verify all the env vars of boundless at the start of batch prover, like when creating prover service or somewhere at that point, what do you reckon?

crates/risc0/Cargo.toml
```diff
@@ -27,12 +31,14 @@ sov-rollup-interface = { path = "../sovereign-sdk/rollup-interface" }
 thiserror = { workspace = true }
 tokio = { workspace = true, optional = true }
 tracing = { workspace = true }
+url = { version = "2.5.4", optional = true }
```
eyusufatik: workspace

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,355 @@
...
+    Groth16Receipt, InnerReceipt, Journal, MaybePruned, Receipt, ReceiptClaim,
+};
+use sov_db::ledger_db::{BoundlessLedgerOps, LedgerDB};
+use sov_db::schema::types::BoundlessSession;
+use sov_rollup_interface::zk::{ProofWithJob, ReceiptType};
+use tokio::sync::oneshot;
+use url::Url;
+use uuid::Uuid;
+
+type BoundlessClient = Client<
```
eyusufatik: I'm still not confused that this is required

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,355 @@
...
+>;
+
+#[derive(Debug, Clone)]
+pub struct BoundlessConfig {
+    wallet_private_key: String,
+    rpc_url: String,
+    boundless_market_address: String,
+    set_verifier_address: String,
+    order_stream_url: Option<String>,
+}
```
eyusufatik: instead of having strings here, use the types that the strings are converted to where applicable. that way, if the conversions fail, you can fail on the FromEnv call.

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,355 @@
...
+        let local_signer = PrivateKeySigner::from_str(&config.wallet_private_key)
+            .context("Failed to parse wallet private key")?;
+
+        // If in dev mode, uses a temporary file as storage provider
+        // Otherwise first tries to parse pinata env variables
+        // If fails then tries to parse s3 env variables
+        // If the environment variable `RISC0_DEV_MODE` is set, a temporary file storage provider is used.
+        // Otherwise, the following environment variables are checked in order:
+        // - `PINATA_JWT`, `PINATA_API_URL`, `IPFS_GATEWAY_URL`: Pinata storage provider;
+        // - `S3_ACCESS`, `S3_SECRET`, `S3_BUCKET`, `S3_URL`, `AWS_REGION`: S3 storage provider.
```
eyusufatik: if we are not letting boundless be used with dev mode, why would this be necessary?

eyusufatik: if we can choose what provider to use, let's use s3 provider. make sure from_env checks the necessary env vars, matter of fact if you can prefix them with BOUNDLESS_ it'd be nice.

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,355 @@
...
+        job_id: Uuid,
+        elf: Vec<u8>,
+        input: Vec<u8>,
+        assumptions: Vec<AssumptionReceipt>,
+        receipt_type: ReceiptType,
+    ) -> anyhow::Result<oneshot::Receiver<ProofWithJob>> {
+        // Upload image id
+        let image_id = compute_image_id(&elf).expect("Invalid elf program");
+
+        assert!(!risc0_zkvm::is_dev_mode(), "Boundless should not be run with dev mode as provers do not accept fake receipt requests" );
```
eyusufatik: replace env var checks on RISC0_DEV_MODE in other provers also, this one's a better usage.

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,355 @@
...
+            .with_image_url(image_url.to_string())
+            .with_input(input_url.clone())
+            .with_requirements(
+                Requirements::new(image_id, Predicate::digest_match(journal.clone().digest()))
+                    .with_groth16_proof(),
+            )
+            .with_offer(
+                Offer::default()
+                    // TODO: Get this from config maybe? or dynamically fetch from some 3rd party api
+                    .with_min_price_per_mcycle(parse_ether("0.00001")?, mcycles_count)
```
eyusufatik: create a tracking issue for this:
- building the estimator api
- integrating it to boundless.rs etc.

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,355 @@
...
+                        "Boundless proving job finished: {} | Boundless request id: {}",
+                        job_id, request_id
+                    );
+
+                }
+                Err(e)=>tracing::error!(
+                    "Failed to handle Boundless proving session job: {} | Boundless request id: {} | err={}",
+                    job_id, request_id, e
+                ),
+            }
```
eyusufatik: I assume we fall here if the proof was not locked, right? let's have a preset mechanism to send the same proving request with start price = 1.5 * old start price and end price = 2 * old end price.

eyusufatik (changes requested):

crates/risc0/src/host/boundless.rs
eyusufatik: done https://github.com/chainwayxyz/citrea/pull/2391/commits/bfd9936f5cd90da52c9456865b1a39115536cfb6

Cargo.lock
yaziciahmet: is it feasible to drop the version of boundless-market? or do you think its best to stay on the latest one despite duplicate deps?

crates/risc0/src/host/boundless.rs
yaziciahmet: I think we should pre-verify. This is the first time we add these envs, and almost everytime we added new config we had some problems, hence it might be best to catch them as early as possible.

crates/risc0/src/host/boundless.rs
ercecan: > if we are not letting boundless be used with dev mode, why would this be necessary?

dev mode will store the input in a tmp file, in non-dev mode we have two options pinata(ipfs) or s3

Cargo.lock
ercecan: I believe it is best to stay on the latest, after we update reth this will aslo be fixed, so this is a temporary issue

crates/risc0/src/host/boundless.rs
ercecan: #2395

ercecan:
Also from recent talks with Risc0 team they confirmed [#515](https://github.com/boundless-xyz/boundless/pull/515) is on the way so we can add that too in this pr

crates/risc0/src/config.rs
```diff
@@ -0,0 +1,125 @@
...
+    pub(crate) deployment: Deployment,
+}
+
+impl citrea_common::FromEnv for BoundlessConfig {
+    fn from_env() -> anyhow::Result<Self> {
+        let wallet_private_key = read_env("BOUNDLESS_WALLET_PRIVATE_KEY")?;
+        let rpc_url = read_env("BOUNDLESS_RPC_URL")?;
+
+        // TODO: Switch to Deployment::builder after boundless 1.0 release to switch between base mainnet and sepolia
+        let deployment = SEPOLIA;
```
eyusufatik: what is this?

eyusufatik: only sepolia right now as explained by @ercecan irl

eyusufatik (approved):

crates/risc0/src/config.rs
yaziciahmet, line 1: I think this is better in host/ directory

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,502 @@
...
+            let assumptions = assumptions.clone();
+
+            move || -> anyhow::Result<(Journal, u64)> {
+                let mut env = ExecutorEnvBuilder::default();
+                for assumption in assumptions {
+                    env.add_assumption(assumption);
+                }
+                let env = env.write_slice(&input).build()?;
+
+                let session_info = default_executor().execute(env, &elf)?;
```
yaziciahmet, line 190: maybe we can explicitly check and expect RISC0_EXECUTOR env as we do within our local prover RISC0_PROVER? just to catch any mistakes at the start, as we only support ipc type.

yaziciahmet (approved):
LGTM, approving right away and just left 2 nit suggestions

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,502 @@
...
+pub struct BoundlessProver {
+    pub client: Client,
+    pub ledger_db: LedgerDB,
+}
+
+impl BoundlessProver {
+    pub async fn new(ledger_db: LedgerDB) -> Self {
+        let client = Self::boundless_client()
+            .await
+            .expect("Failed to create boundless client");
```
rakanalh, line 66: `boundless_client` returns `anyhow::Result`... why not do the same here?

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,502 @@
...
+            }
+        })
+        .await??;
+
+        let request = self.build_proof_request(
+            image_id,
+            journal.digest(),
+            image_url,
+            input_url,
+            // TODO: Receive from third party api
```
rakanalh: Maybe create an issue for this and add it here?

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,502 @@
...
+        job_id: Uuid,
+        receipt_type: ReceiptType,
+        request_id: String,
+        image_id: Digest,
+        request_expiry: u64,
+        mcycles_count: u64,
+    ) -> oneshot::Receiver<ProofWithJob> {
+        let this = self.clone();
+        let (tx, rx) = oneshot::channel();
+        tokio::spawn(async move {
```
rakanalh, line 387: How can we make this conform to cancellation rules?

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,502 @@
...
+        ClientBuilder::new()
+            .with_deployment(config.deployment)
+            .with_rpc_url(config.rpc_url)
+            .with_storage_provider(Some(storage_provider))
+            .with_private_key(config.wallet_private_key)
+            .build()
+            .await
+    }
+
+    pub async fn prove(
```
rakanalh, line 115: Add instrument here

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,502 @@
...
+                                    request_id,
+                                    e
+                                );
+                                break;
+                            }
+                        }
+                    }
+                }
+            }
+        });
```
rakanalh:
```suggestion
        }).instrument(....);
```

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,502 @@
...
+        mcycles_count: u64,
+        image_id: Digest,
+        receipt_type: ReceiptType,
+    ) -> anyhow::Result<()> {
+        // Remove failed job from pending boundless sessions
+        this.ledger_db
+            .remove_pending_boundless_session(job_id)
+            .expect("Failed to remove pending boundless session on error");
+
+        // TODO: Handle slashing offchain
```
rakanalh: Create an issue for this

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,502 @@
...
+        else {
+            tracing::error!(
+                "Failed to fetch failed order for job: {} request_id: {}",
+                job_id,
+                request_id
+            );
+            return Ok(());
+        };
+
+        // TODO: Set a max limit for min and max price per mcycle when integrating the pricing api
```
rakanalh: Create an issue for this

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,502 @@
...
+                    }
+                }
+            }
+        });
+
+        rx
+    }
+
+    async fn handle_resubmit_on_failed_request(
+        this: &Self,
```
rakanalh: We can still call this self, no?

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,502 @@
...
+                    session.mcycles_count,
+                )
+                .await;
+            rxs.push(rx);
+        }
+        Ok(rxs)
+    }
+}
+
+fn now_timestamp() -> u64 {
```
rakanalh: Should this be moved to common or primitives?

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
```diff
@@ -399,6 +404,11 @@ define_table_with_default_codec!(
     (PendingBonsaiSessionByJobId) Uuid => BonsaiSession
 );
 
+define_table_with_default_codec!(
+    /// Pending Boundless proving sessions by job id
+    (PendingBoundlessSessionByJobId) Uuid => BoundlessSession
```
rakanalh, line 397: I believe this table should be included in rollback?

rakanalh (changes requested):

crates/risc0/src/host/boundless.rs
eyusufatik: It is created and being worked on another repo

crates/risc0/src/host/boundless.rs
eyusufatik: I think you can just make an error log here @ercecan

crates/risc0/src/host/boundless.rs
ercecan: not sure if we need that as this already exists:
```rust
 pub async fn new(ledger_db: LedgerDB, network: Network) -> Self {
        let prover = match std::env::var("RISC0_PROVER") {
            Ok(prover) => match prover.as_str() {
                "boundless" => Prover::Boundless(BoundlessProver::new(ledger_db).await),
                "bonsai" => Prover::Bonsai(BonsaiProver::new(ledger_db)),
                "ipc" => Prover::Local(LocalProver::new(network)),
                _ => panic!("Invalid prover specified: {}", prover),
            },
            Err(_) => {
                debug!("No prover specified.");
                Prover::Local(LocalProver::new(network))
            }
        };

        Self {
            env: Default::default(),
            assumptions: vec![],
            prover,
        }
    }
```

crates/risc0/src/host/boundless.rs
ercecan: I believe it should just panic if it cannot even create the client, no point in running prover without the prover client IMO

crates/risc0/src/host/boundless.rs
ercecan: we cannot, as there is no proof cancellation once it is on the market

crates/risc0/src/host/boundless.rs
ercecan: I believe this comment refers to this one:
https://github.com/chainwayxyz/citrea/pull/2391#discussion_r2124454901

ercecan: Correct me if not

crates/risc0/src/host/boundless.rs
ercecan: I log error right before entering and right after exiting this function

ercecan: Added issue just in case

crates/risc0/src/host/boundless.rs
ercecan: Linked the issue just in case

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
ercecan: I believe it is not necessary as we do not rollback pending bonsai sessions anyways

crates/risc0/src/host/boundless.rs
ercecan: I added this anyway is that good?
```rust
impl BoundlessProver {
    pub async fn new(ledger_db: LedgerDB) -> Self {
        assert!(
            std::env::var("RISC0_PROVER").is_ok_and(|prover| prover == "boundless"),
            "RISC0_PROVER must be explicitly set to boundless"
        );
```
I did not wanna add RISC0_EXECUTOR because it would be specific to boundless and there are already a lot of env vars specific to boundless

crates/risc0/src/host/boundless.rs
rakanalh: Yea probably forgot to remove this one.

crates/risc0/src/host/boundless.rs
rakanalh: I meant just a future cancellation so that we can exit the `loop` gracefully

rakanalh (dismissed):

crates/risc0/src/host/boundless.rs
yaziciahmet: if you think we shouldn't add RISC0_EXECUTOR check that's fine by me.

crates/risc0/src/host/boundless.rs
ercecan: got it, looking into it

ercecan: lmk wdyt about this:
https://github.com/chainwayxyz/citrea/pull/2391/commits/2d45037c1a65390ab853af0253c5ad8a7d703b03

ercecan: https://github.com/chainwayxyz/citrea/issues/2420

crates/common/src/utils.rs
```diff
@@ -154,3 +154,10 @@ pub async fn decode_sov_tx_and_update_short_header_proofs<Da: DaService, DB: Sha
 pub fn read_env(key: &str) -> anyhow::Result<String> {
     env::var(key).map_err(|_| anyhow::anyhow!("Env {} missing or invalid UTF-8", key))
 }
+
+pub fn now_timestamp() -> u64 {
```
kpp: I don't see this as a "common" fn given it's used once in boundless

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,508 @@
...
+        let request = self.build_proof_request(
+            image_id,
+            journal.digest(),
+            image_url,
+            input_url,
+            // TODO: https://github.com/chainwayxyz/citrea/issues/2397
+            parse_ether("0.000001")?,
+            parse_ether("0.001")?,
+            mcycles_count,
+            43200,
```
kpp: What's that magic?

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,508 @@
...
+        })
+        .await??;
+
+        let request = self.build_proof_request(
+            image_id,
+            journal.digest(),
+            image_url,
+            input_url,
+            // TODO: https://github.com/chainwayxyz/citrea/issues/2397
+            parse_ether("0.000001")?,
```
kpp: What's that magic?

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,508 @@
...
+            parse_ether("0.001")?,
```
kpp: What's that magic?

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,508 @@
...
+            mcycles_count,
+        };
+        self.ledger_db
+            .upsert_pending_boundless_session(job_id, db_session)
+            .context("Failed to upsert boundless session")?;
+
+        Ok((req_id.to_string(), request_expiry))
+    }
+
+    async fn spawn_handler(
```
kpp: This shouldn't be async because the body is wrapped in tokio::spawn.

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,508 @@
...
+        ));
+        let full_snark_receipt = Receipt::new(inner, journal.to_vec());
+        full_snark_receipt.verify(image_id).unwrap();
+
+        Ok(full_snark_receipt)
+    }
+
+    // Starts the recovery of proving jobs from db by starting a background task, returning list of
+    /// receiver channels that return the associated job id and proof result on finish.
+    pub async fn start_recovery(&self) -> anyhow::Result<Vec<oneshot::Receiver<ProofWithJob>>> {
```
kpp: This shouldn't be async because spawn_handler shouldn't be async

crates/risc0/src/host/mod.rs
```diff
@@ -168,4 +178,22 @@ pub enum Prover {
     Local(LocalProver),
     /// Bonsai prover
     Bonsai(BonsaiProver),
+    /// Boundless prover network
+    Boundless(BoundlessProver),
+}
+
+impl Prover {
+    /// Start recovery for prover if it supports it
+    pub async fn start_prover_session_recovery(
```
kpp: This shouldn't be async because bboundless recovery shouldn't be async

crates/risc0/src/host/boundless.rs
ercecan: nice catch fixing all unnecessary async fns

crates/common/src/utils.rs
ercecan: @rakanalh suggested this, we may use it in the future

crates/risc0/src/host/boundless.rs
ercecan: explained all magic more clearly

ercecan:
They just released boundless v0.10 might as well update to that as well

crates/common/src/utils.rs
jfldde: `timestamp_as_secs` ? 
Feel like now is already conveyed by timestamp

ercecan: how about `current_timestamp_as_secs` ? Because I feel that it does not completely convey that it is `now`

also most sql dbs have the function current_timestamp

eyusufatik (dismissed):
did you test this?

eyusufatik:
> did you test this?

sorry this was meant for #2571

crates/risc0/src/host/boundless.rs
```diff
@@ -262,6 +263,7 @@ impl BoundlessProver {
         );
         let (req_id, request_expiry) = match self.client.offchain_client {
             Some(_) => {
+                // TODO: i think requst id can be set to job id
```
ercecan: Job id is a Uuid type, should I change it just for this? I don't think so

eyusufatik:
don't merge without https://github.com/chainwayxyz/citrea/issues/2821

guests/risc0/batch-proof/bitcoin/Cargo.lock
```diff
@@ -2760,13 +2781,22 @@ dependencies = [
  "getrandom 0.2.15",
 ]
 
+[[package]]
```
eyusufatik, line 2783: maybe we can find a workaround for these useless rand crates

guests/risc0/light-client-proof/bitcoin/Cargo.lock
```diff
@@ -129,6 +129,21 @@ dependencies = [
  "syn 2.0.98",
 ]
 
+[[package]]
+name = "android-tzdata"
```
eyusufatik, line 131: how did this come here :D

guests/risc0/light-client-proof/bitcoin/Cargo.lock
```diff
@@ -1282,24 +1332,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
...
 
 [[package]]
 name = "hashbrown"
-version = "0.14.5"
+version = "0.13.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e5274423e17b7c9fc20b6e7e208532f9b19825d82dfd615708b70edd83df41f1"
+checksum = "43a3c133739dddd0d2990f9a4bdf8eb4b21ef50e4851ca85ab661199821d510e"
 dependencies = [
  "ahash",
```
eyusufatik, line 1349: interesting these are downgraded

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,710 @@
...
+    }
+
+    async fn boundless_client() -> anyhow::Result<Client> {
+        let config = BoundlessConfig::from_env().expect("Failed to load boundless config");
+
+        // If in dev mode, uses a temporary file as storage provider
+        // Otherwise first tries to parse pinata env variables
+        // If fails then tries to parse s3 env variables
+        // If the environment variable `RISC0_DEV_MODE` is set, a temporary file storage provider is used.
+        // Otherwise, the environment variables in `BoundlessPinataStorageConfig` or `BoundlessS3StorageConfig` is checked
```
jfldde: Where is that done ?

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,710 @@
...
+            matches!(receipt_type, ReceiptType::Groth16),
+            "Currently, only Groth16 receipts are supported for boundless"
+        );
+
+        // TODO: Can be done better?
+        let s3_url = read_env("BOUNDLESS_S3_URL")?;
+
+        let s3_use_presigned = read_env("BOUNDLESS_S3_NO_PRESIGNED")
+            .map(|_| false)
+            .unwrap_or(true);
```
jfldde:
```suggestion
        let s3_use_presigned = read_env("BOUNDLESS_S3_NO_PRESIGNED").is_err()
```

Or even better, switching to `BOUNDLESS_S3_USE_PRESIGNED` and use is_ok() instead

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,710 @@
...
+            .context("Failed to encode input for boundless proving")?;
+
+        // Deposit to contract
+        // let deposit_amount = U256::from(1e16 as u64); // 0.01eth
+        // let market = self.client.boundless_market.clone();
+        // market.deposit(deposit_amount).await?;
+        // tracing::info!(
+        //     "Successfully deposited {} ETH",
+        //     alloy_primitives::utils::format_units(deposit_amount, "ether")?
+        // );
```
jfldde: What is that ?

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,710 @@
...
+            .map(|_| false)
+            .unwrap_or(true);
+
+        // Upload the program(elf) to the boundless storage provider
+        let mut image_url = self.client.upload_program(&elf).await?;
+        tracing::info!("Image URL: {}", image_url);
+
+        // If we are not using presigned:
+        if !s3_use_presigned {
+            let s3_path = image_url.as_str().to_string().replace("s3://", "");
```
jfldde: Maybe strip_prefix instead to make sure we actually remove the s3:// part

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,710 @@
...
+                    .segments
+                    .iter()
+                    .map(|segment| 1 << segment.po2)
+                    .sum::<u64>();
+                let mcycles_count = total_cycles_approx.div_ceil(1_000_000);
+                tracing::info!(
+                    "Boundless proving session with job id: {} takes {} cycles",
+                    job_id,
+                    total_cycles_approx
+                );
```
jfldde, line 200:
```suggestion
                tracing::info!(
                    "Boundless proving session with job id: {job_id} takes {total_cycles_approx} cycles"
                );
```

In line with the rest of the PR formatting

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,710 @@
...
+        .await
+        .map_err(|e| {
+            anyhow::anyhow!(
+                "Failed to get price from pricing service for job: {}  | err={}",
+                job_id,
+                e
+            )
+        })?;
+
+        let lock_timeout = cmp::max(lock_timeout, 200); // at least 200 seconds
```
jfldde: Can extract 200 to a constant and maybe explain the choice as it's not ovbious

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,710 @@
...
+                Err(e) => {
+                    tracing::error!(
+                        "Failed to check if request is locked for job: {} request_id: {} | err={}",
+                        job_id,
+                        request_id,
+                        e
+                    );
+                    return Ok(ResubmitResult::Retry);
+                }
+            };
```
jfldde, line 566: let else guard would be clearer

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,710 @@
...
+                (min_price_per_mcycle, max_price_per_mcycle, lock_timeout)
+            } else {
+                // If not locked, that means the request was never taken by a prover.
+                // Increase the min and max price per mcycle.
+                let min_price_per_mcycle = min_price_per_mcycle
+                    .saturating_mul(U256::from(15))
+                    .div_ceil(U256::from(10))
+                    .min(U256::from(max_possible_price));
+                let max_price_per_mcycle = max_price_per_mcycle
+                    .saturating_mul(U256::from(2))
```
jfldde: Should these values be configurable ? At least extracted to constant

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,710 @@
...
+            Ok((req_id, exp_time)) => (req_id, exp_time),
+            Err(e) => {
+                tracing::error!(
+                    "Failed to resubmit boundless proving session retrying, job_id={} request_id={} | err={}",
+                    job_id,
+                    request_id,
+                    e
+                );
+                return Ok(ResubmitResult::Retry);
+            }
```
jfldde, line 636: let else guard would be better

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,710 @@
...
+    }
+}
+
+fn is_dev_mode() -> bool {
+    std::env::var("RISC0_DEV_MODE")
+        .ok()
+        .map(|x| x.to_lowercase())
+        .filter(|x| x == "1" || x == "true" || x == "yes")
+        .is_some()
+}
```
jfldde: Should use shared `is_dev_mode_enabled_via_environment` instead

crates/risc0/src/host/config.rs
```diff
@@ -0,0 +1,134 @@
...
+impl citrea_common::FromEnv for BoundlessS3StorageConfig {
+    fn from_env() -> anyhow::Result<Self> {
+        let s3_access_key = read_env("BOUNDLESS_S3_ACCESS_KEY")?;
+        let s3_secret_key = read_env("BOUNDLESS_S3_SECRET_KEY")?;
+        let s3_bucket = read_env("BOUNDLESS_S3_BUCKET")?;
+        let s3_url = read_env("BOUNDLESS_S3_URL")?;
+        let aws_region = read_env("BOUNDLESS_AWS_REGION")?;
+        let s3_use_presigned = read_env("BOUNDLESS_S3_NO_PRESIGNED")
+            .map(|_| false)
+            .unwrap_or(true);
```
jfldde: Same as above, rename to `BOUNDLESS_S3_USE_PRESIGNED` (matching the struct field) and use is_ok instead

crates/risc0/src/host/pricing_service.rs
```diff
@@ -0,0 +1,91 @@
...
+
+        info!("Fetching price for {} cycles", cycles);
+
+        let response = self
+            .client
+            .get(&url)
+            .query(&[("cycles", cycles)])
+            .send()
+            .await
+            .context("Failed to send request to pricing service")?;
```
jfldde, line 70: No timeout handling

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
```diff
@@ -389,6 +394,11 @@ define_table_with_default_codec!(
     (PendingBonsaiSessionByJobId) Uuid => BonsaiSession
 );
 
+define_table_with_default_codec!(
```
jfldde: Do we want it sorted ?  seek_key_codec might be better suited to leverage uuid v7 timestamp ordering

jfldde (commented):
Looking good but for a few nits. Also, boundless related changes don't seem to be linted according to Rust version bump.

crates/risc0/src/host/boundless.rs
ercecan: Nowhere, fixing comment

crates/risc0/src/host/boundless.rs
ercecan: Only used it this confusing way to make it same with boundless

crates/risc0/src/host/boundless.rs
ercecan: A deposit script that I use in a hacky way sometimes will remove

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
ercecan: I don't think we need it sorted but good call

jfldde: yes, tbh default of handling sorted and using borsh for perf explicitly would have been better but that's something else

guests/risc0/light-client-proof/bitcoin/Cargo.lock
eyusufatik: it's fine

guests/risc0/light-client-proof/bitcoin/Cargo.lock
eyusufatik: checked, it's fine

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
```diff
@@ -44,7 +46,8 @@ pub enum ReceiptType {
 
 /// A trait implemented by the prover ("host") of a zkVM program.
```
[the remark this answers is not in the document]
rakanalh: No need

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,711 @@
...
+            ),
+        ));
+
+        rx
+    }
+
+    async fn handle_resubmit_on_failed_request(
+        &self,
+        job_id: Uuid,
+        request_id: &mut String,
```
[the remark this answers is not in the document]
rakanalh: We will still panic, as agreed with Esad.

crates/risc0/src/host/boundless.rs
```diff
@@ -0,0 +1,711 @@
...
+
+    pub async fn prove(
+        &self,
+        job_id: Uuid,
+        elf: Vec<u8>,
+        input: Vec<u8>,
+        assumptions: Vec<AssumptionReceipt>,
+        receipt_type: ReceiptType,
+    ) -> anyhow::Result<oneshot::Receiver<ProofWithJob>> {
+        // Upload image id
```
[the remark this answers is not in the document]
rakanalh: Done in #2972

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
rakanalh: Done, instead of using default codec, i am using `seek_key_codec`

rakanalh (dismissed):

eyusufatik:
TODO: update v0.8.25 alloy crates to v0.8.26

also see https://github.com/alloy-rs/core/releases

bin/citrea/src/rollup/bitcoin.rs
```diff
@@ -280,7 +281,8 @@ impl RollupBlueprint for BitcoinRollup {
         proof_sampling_number: usize,
         is_light_client_prover: bool,
     ) -> ParallelProverService<Self::DaService, Self::Vm> {
-        let vm = Risc0Host::new(ledger_db.clone(), self.network);
+        let risc0_config = Risc0HostConfig::from_env().expect("Failed to load risc0 config");
```
eyusufatik, line 284: make issue for also adding the Risc0HostConfig to the ProverConfig we read from toml files

deny.toml
```diff
@@ -38,14 +38,16 @@ github = [
...
 ignore = [
   "RUSTSEC-2024-0388", # Unmaintained derivative crate
-  "RUSTSEC-2024-0370", # Unmaintained proc-macro-error crate
   "RUSTSEC-2024-0384", # Unmaintained instant crate
   "RUSTSEC-2024-0436", # Unmaintained paste crate
   "RUSTSEC-2025-0012", # Unmaintained backoff crate
   "RUSTSEC-2025-0055", # Ignore until risc0 update
+  "RUSTSEC-2023-0071", # Ignore rsa 0.9.8 until a version that fixes the issue is available.
+  "RUSTSEC-2025-0052", # Ignore async-std being discontinued. httpmock uses it, which in turn is used by boundless market.
+  "RUSTSEC-2025-0073", # Ignore alloy-dyn-abi until we are able to update to 1.4.1
```
eyusufatik, line 52: remove this one and merge nightly, I fixed it with a version upgrade

crates/common/src/config/mod.rs
```diff
@@ -156,6 +156,7 @@ impl<DaC: FromEnv> FromEnv for FullNodeConfig<DaC> {
 pub struct BatchProverConfig {
     /// Prover run mode
     pub proving_mode: ProverGuestRunConfig,
+    // TODO: handle this correctly in
```
eyusufatik: can you just remove this todo?

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
```diff
@@ -62,7 +65,7 @@ pub trait ZkvmHost: Zkvm + Clone {
     /// This runs the guest binary compiled for the zkVM target, optionally
     /// creating a SNARK of correct execution. Running the true guest binary comes
     /// with some mild performance overhead and is not as easy to debug as [`simulate_with_hints`](ZkvmHost::simulate_with_hints).
-    fn run(
+    async fn run(
```
eyusufatik, line 68: since this is now async, make issue to update bonsai adapter with async client

eyusufatik (changes requested):

bin/citrea/tests/bitcoin/batch_prover_test.rs
```diff
@@ -1266,7 +1268,8 @@ impl TestCase for BatchProverCreateInputTest {
         let network = Network::Nightly;
 
         let ledger_db = LedgerDB::with_config(&rocksdb_config).unwrap();
-        let mut risc0_host = Risc0Host::new(ledger_db, network);
+        let risc0_config = Risc0HostConfig::from_env().expect("Failed to load risc0 config");
```
exeokan, line 1271: i think Risc0HostConfig::default() makes more sense here

crates/common/src/utils.rs
```diff
@@ -140,6 +140,13 @@ pub fn read_env(key: &str) -> anyhow::Result<String> {
     env::var(key).map_err(|_| anyhow::anyhow!("Env {} missing or invalid UTF-8", key))
 }
 
+pub fn current_timestamp_as_secs() -> u64 {
```
exeokan: where is this used?

exeokan (changes requested):

deny.toml
eyusufatik: the alloy-dyn-abi one

bin/citrea/tests/bitcoin/batch_prover_test.rs
rakanalh: Not really, we implement from_env for all configs:
```rust
impl citrea_common::FromEnv for Risc0HostConfig {
        let prover = Risc0ProverConfig::from_env()?;
        let tx_backup_dir = std::env::var("TX_BACKUP_DIR").ok().map(PathBuf::from);

        Ok(Self {
            prover,
            tx_backup_dir,
        })
    }
}
```

crates/common/src/utils.rs
rakanalh: Removed

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/mod.rs
rakanalh: No need for an issue... made the client async now.

eyusufatik (approved):

rakanalh (approved):

exeokan (approved):

jfldde (approved):

PR #2394, opened 2025-05-26
remove invalid sender tracking as reth now supports it

eyusufatik:
# Description
at that time we couldn't update reth due to risc0 rust versions. now removing unnecessary code for tracking invalid senders in the context of a block.

### DIFF
--- a/crates/sequencer/src/runner.rs
+++ b/crates/sequencer/src/runner.rs
@@ -245,8 +245,9 @@ where
                             L2BlockModuleCallError::EvmTransactionExecutionError(_) => {
                                 transactions.mark_invalid(
                                     &evm_tx,
-                                    // don't really have a way to know the cost right now
-                                    // passing 1 & 0 as it doesn't matter (the kind field is never used)
+                                    // don't really have a way to know the underlying EVM error due to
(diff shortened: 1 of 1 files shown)


crates/sequencer/src/runner.rs
```diff
@@ -250,7 +243,15 @@ where
...
-                                invalid_senders.insert(evm_tx.transaction_id.sender);
+                                transactions.mark_invalid(
+                                    &evm_tx,
+                                    // don't really have a way to know the cost right now
+                                    // passing 1 & 0 as it doesn't matter (the kind field is never used)
+                                    InvalidPoolTransactionError::Overdraft {
+                                        cost: U256::from(1),
+                                        balance: U256::ZERO,
+                                    },
+                                );
```
yaziciahmet: why is this Overdraft error? execution error could be anything

yaziciahmet (approved):
LGTM, left one question

crates/sequencer/src/runner.rs
eyusufatik: ah yeah I should update that comment there, the thing is I cant extract the actual error at that point.

PR #2405, opened 2025-06-01
chore: return error for no bytecode

### LINKED ISSUE #2399: `crates/evm/src/evm/db.rs:100` shouldn't be Default::default
## Summary
Not a serious bug, however, probably should be an error.

varun-doshi:
# Description
Adds a new error type to `DBError`:`BytecodeNotAvailable` and returns in place of `default` at `evm/src/evm/db.rs:100`

## Linked Issues
- Fixes #2399


eyusufatik (changes requested):
cannot directly return not available error btw. maybe we should check for keccak_empty? or maybe revm is already doing that and this is fine?

eyusufatik (changes requested):
I checked revm, and yes it does check for keccak empty before calling the trait implementation. so I can aprove the pr if you change the error name and message to UnknownCodeHash.

varun-doshi:
> I checked revm, and yes it does check for keccak empty before calling the trait implementation. so I can aprove the pr if you change the error name and message to UnknownCodeHash.

Understood
Will update

eyusufatik (approved):

jfldde (approved):

PR #2406, opened 2025-06-02
Exclude failed deposit transactions

### LINKED ISSUE #2343: Sequencer should skip failed deposits
It is possible that an attacker deploy a malicious contract, and deposit to that contract through the bridge, and make a deposit fail via consuming a lot of gas. This may stop the sequencer, so sequencer should skip failed deposits.

yaziciahmet:
# Description

Exclude failed deposit transactions as deposits to non-EOA (e.g. through receive fn of contracts)

## Linked Issues
- Fixes #2343 

## Testing

## Docs


eyusufatik (commented):
I'll need a test that tries to run two txs after the failing deposit, a full node should reach the same state root afterwards.

crates/sequencer/src/runner.rs
```diff
@@ -1006,6 +1010,19 @@ where
...
+                // as deposits to non-EOA addresses can revert
+                if matches!(
+                    e,
+                    StateTransitionError::ModuleCallError(
+                        L2BlockModuleCallError::EvmSystemTransactionNotSuccessful
+                    )
+                ) && is_deposit
+                {
+                    warn!("Deposit transaction failed: {:?}", e);
+                    working_set_to_discard = working_set.revert().to_revertable();
```
kpp, line 1214: and it will be overwritten with line 1028 `working_set_to_discard = working_set.checkpoint().to_revertable();` with next sys tx

eyusufatik: by doing this, we are basically reverting state changes caused by the unsuccessful deposit.

eyusufatik:
just realised, even without this pr. things like double deposits won't break the sequencer. because it will consume the deposits and then the sequencer will retry producing the block. but ofc, you can spam double deposits or any non-working deposit to create a dos attack.

eyusufatik (approved):
tested on clementine repo with double deposit scenario thanks to @aoengin and this works.

PR #2422, opened 2025-06-04
Add MINIMUM_WITNESS_COMMITMENT_SIZE check when looking for commitment index in verify_transactions

jfldde:
# Description

- Adds additional `MINIMUM_WITNESS_COMMITMENT_SIZE` check when looking for commitment index in line with bitcoin core `GetWitnessCommitmentIndex` method :
```
/** Compute at which vout of the block's coinbase transaction the witness commitment occurs, or -1 if not found */
inline int GetWitnessCommitmentIndex(const CBlock& block)
{
    int commitpos = NO_WITNESS_COMMITMENT;
    if (!block.vtx.empty()) {
        for (size_t o = 0; o < block.vtx[0]->vout.size(); o++) {
            const CTxOut& vout = block.vtx[0]->vout[o];
            if (vout.scriptPubKey.size() >= MINIMUM_WITNESS_COMMITMENT &&
                vout.scriptPubKey[0] == OP_RETURN &&
                vout.scriptPubKey[1] == 0x24 &&
                vout.scriptPubKey[2] == 0xaa &&
                vout.scriptPubKey[3] == 0x21 &&
                vout.scriptPubKey[4] == 0xa9 &&
                vout.scriptPubKey[5] == 0xed) {
                commitpos = o;
            }
        }
    }
    return commitpos;
}
```

A malicious miner could add an output which starts with `WITNESS_COMMITMENT_PREFIX` but doesn't meet the `MINIMUM_WITNESS_COMMITMENT` requirement. This would then pass through bitcoin core `CheckWitnessMalleation` without validating the stack size.
It would then fail on `verify_transactions` expecting a correct witness structure but would panic when not having any actual witness on next() at 
```
let input_witness_value = coinbase_tx.input[0].witness.iter().next().unwrap();
```

- Remove unwrap and replace it by new `InvalidWitnessCommitmentStructure` `ValidationError` variant as code hygiene


crates/bitcoin-da/src/verifier.rs
```diff
@@ -162,10 +164,11 @@ impl DaVerifier for BitcoinVerifier {
         // the one with highest output index is assumed to be the commitment.
         // That  is why the iterator is reversed.
         let commitment_idx = coinbase_tx.output.iter().rev().position(|output| {
```
kpp: In the PR description indices go from 0 to .. N: `for (size_t o = 0; o < block.vtx[0]->vout.size(); o++)`. Why is it reversed?

eyusufatik: at one point @ercecan said bitcoin core was traversing in reverse but yeah the code jeremy shared traverses normally

eyusufatik (approved):
let's not merge without resolving what @kpp pointed out

crates/bitcoin-da/src/verifier.rs
jfldde: Wouldn't change much since we're re-assigning to commitment_idx in the Some(_) branch. 
I'll switch it so we keep it in line with bitcoin core implementation.

jfldde: Though I might need @ercecan to shed some light on the re-assignement itself

eyusufatik: I think it's important. what if the coinbase tx has 2 outputs with the correct prefix? then which order you traverse makes a difference.

eyusufatik: oh wait. so since the for loop does not break when it finds the index with the witness commitment prefix, the last one is the actual commitment, yeah?

jfldde:
```
static bool CheckWitnessMalleation(const CBlock& block, bool expect_witness_commitment, BlockValidationState& state)
{
    if (expect_witness_commitment) {
        if (block.m_checked_witness_commitment) return true;
        
        int commitpos = GetWitnessCommitmentIndex(block);
        if (commitpos != NO_WITNESS_COMMITMENT) {
            assert(!block.vtx.empty() && !block.vtx[0]->vin.empty());
            const auto& witness_stack{block.vtx[0]->vin[0].scriptWitness.stack};

            if (witness_stack.size() != 1 || witness_stack[0].size() != 32) {
                return state.Invalid(
                    /*result=*/BlockValidationResult::BLOCK_MUTATED,
                    /*reject_reason=*/"bad-witness-nonce-size",
                    /*debug_message=*/strprintf("%s : invalid witness reserved value size", __func__));
            }

            // The malleation check is ignored; as the transaction tree itself
            // already does not permit it, it is impossible to trigger in the
            // witness tree.
            uint256 hash_witness = BlockWitnessMerkleRoot(block, /*mutated=*/nullptr);

            CHash256().Write(hash_witness).Write(witness_stack[0]).Finalize(hash_witness);
            if (memcmp(hash_witness.begin(), &block.vtx[0]->vout[commitpos].scriptPubKey[6], 32)) {
                return state.Invalid(
                    /*result=*/BlockValidationResult::BLOCK_MUTATED,
                    /*reject_reason=*/"bad-witness-merkle-match",
                    /*debug_message=*/strprintf("%s : witness merkle commitment mismatch", __func__));
            }

            block.m_checked_witness_commitment = true;
            return true;
}
```
As long as there's 1 witness commitment matching the expected length and prefix, it will go through the check. 
In our case, we reversing but then re-reversing on re-assignment here `commitment_idx = coinbase_tx.output.len() - commitment_idx - 1;`

ercecan: the reason behind starting from the highest index (reversing) is

> The wTXID commitment can be contained inside any of the outputs of the coinbase transaction. If there are multiple outputs with this structure for some reason, the highest [output index number](https://learnmeabitcoin.com/technical/transaction/input/vout/) that contains this wTXID commitment structure will be taken as the commitment.

ref: https://learnmeabitcoin.com/technical/transaction/wtxid/#commitment

> If there are more than one scriptPubKey matching the pattern, the one with
highest output index is assumed to be the commitment

ref: https://www.bitcoin.org.hk/media/presentations/2016-03-16/2016-03-16-Segregated_Witness.pdf

ercecan: > oh wait. so since the for loop does not break when it finds the index with the witness commitment prefix, the last one is the actual commitment, yeah?

yeah

ercecan (approved):

kpp (approved):

ercecan (approved):

PR #2427, opened 2025-06-05
Test: Use serialized fake proofs in full_node e2e tests

jfldde:
# Description

- Remove batch prover dependency and use serialized fake proofs instead.

## Linked Issues
- Related to #2321. Fixes issue with test failing now that we properly rollback l1 known hashes


eyusufatik (changes requested):
If you added LCP to these requests, it makes sense to compare results of it with full nodes.

jfldde:
> If you added LCP to these requests, it makes sense to compare results of it with full nodes.

It was only really used for the methods ids but it's a good addition, I'll add these asserts

eyusufatik (approved):

eyusufatik (approved):

rakanalh (approved):

Issue #2433, opened 2025-06-10
feat: `expose_rpc` macro can acess ledger db


eyusufatik:
Make rpcs created through the `expose_rpc` macro (e.g. evm rpc) have access to the ledgerdb.

This will be used in another issue.

eyusufatik:
#2439

PR #2439, opened 2025-06-11
LedgerDB in rpc macros

kpp:
# Description

- Reduce cyclomatic complexity for rpc macro (To help with LedgerDB arg)
- implementation of C::LedgerDb arg in rpc macro

Usage:

```
#[rpc_gen(server, namespace = "myNamespace")]
impl<C: Context> MyModule<C> {
    #[rpc_method(name = "withLedger")]
    fn with_ledger(&self, _ledger_arg: &C::LedgerDB) -> RpcResult<u32> {
        Ok(1)
    }
}
```

## Linked Issues
- Related to https://github.com/chainwayxyz/citrea/issues/2433


crates/sovereign-sdk/module-system/sov-modules-core/src/module/spec.rs
```diff
@@ -53,6 +53,9 @@ pub trait Spec: BorshDeserialize + BorshSerialize {
     /// Authenticated state storage used by the rollup. Typically some variant of a merkle-patricia trie.
     type Storage: Storage + Send + Sync;
 
+    /// The type of LedgerDB (empty for zk).
+    type LedgerDB: Send + Sync + Clone;
```
kpp: It can be a problem because the type must implement `SharedLedgerOps` or something like this. Please review and say how to implement it better.

jfldde:
Does it make sense to have rpc related code as part of stf ? 
Raising this as I guess that's one of the issue with https://github.com/chainwayxyz/citrea/pull/2439#discussion_r2143979332 and we could use LedgerDb concretly instead

crates/sovereign-sdk/module-system/sov-modules-macros/src/rpc/rpc_gen.rs
eyusufatik, line 1: I ain't reading this

eyusufatik (approved):

crates/sovereign-sdk/module-system/sov-modules-core/src/module/spec.rs
rakanalh: I agree it's going to be a problem. Since we're passing the generic type, i.e:
https://github.com/chainwayxyz/citrea/pull/2439/files#diff-f2899e9dfce24f4daea4e4c5c0ffd80d25a4507512aab6216c2fda717545cb9bR21

AFAICS We're not going to be able to use methods from that apart from `clone`... i think instead of defining an associated type here, we just use concrete type (LedgerDB itself).

That is, unless we can verify what happens by implementing a case that uses ledger db inside the `rpc_methods` method.

ercecan: I mean we can pass LedgerDB itself but is it a problem for the type to implement SharedLedgerOps? It should be the bare minimum I guess so it may be fine no?

ercecan (approved):
LGTM when this is resolved:
https://github.com/chainwayxyz/citrea/pull/2439/files#r2143979332

crates/sovereign-sdk/module-system/sov-modules-core/src/module/spec.rs
rakanalh: To come to think of it, yes it should be fine.

jfldde (approved):

eyusufatik (approved):

PR #2441, opened 2025-06-11
HOTFIX - Do not return error on unknown function selector for system txs

ercecan:
# Description
We always check for the latest selector because the contracts are compiled, if the incoming selector does not match with the latest versions of selectors we throw error which halted the node
The issue arises because we update the deposit call signature
To fix, we will not return error in that case and just log an informative warning that we got an unexpected function selector

**Tested locally, it works**
**Ran tests locally, they pass**

### DIFF
--- a/crates/common/src/utils.rs
+++ b/crates/common/src/utils.rs
@@ -104,9 +104,10 @@ async fn update_short_header_proof_from_sys_tx<Da: DaService, DB: SharedLedgerOp
         BridgeContract::depositCall::SELECTOR => {
             tracing::info!("Deposit system tx found inside block");
         }
+        // TODO: https://github.com/chainwayxyz/citrea/issues/2442
         unexpected_selector => {
             tracing::warn!(
-                "Invalid system tx, unexpected function selector: {unexpected_selector:?} , tx input: {:?}, tx hash: {:?}, tx nonce: {:?}", tx.inner().transaction().input(), tx.inner().hash(), tx.inner().transaction().nonce()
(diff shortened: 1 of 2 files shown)


jfldde (dismissed):

crates/common/src/utils.rs
```diff
@@ -104,8 +104,10 @@ async fn update_short_header_proof_from_sys_tx<Da: DaService, DB: SharedLedgerOp
         BridgeContract::depositCall::SELECTOR => {
             tracing::info!("Deposit system tx found inside block");
         }
-        _ => {
-            return Err(anyhow!("Invalid system tx"));
+        unexpected_selector => {
+            tracing::warn!(
+                "Invalid system tx, unexpected function selector: {unexpected_selector:?} , tx input: {:?}, tx hash: {:?}, tx nonce: {:?}", tx.inner().transaction().input(), tx.inner().hash(), tx.inner().transaction().nonce()
```
eyusufatik: Don't call this invalid system tx. it's not a part of our rules that the system tx must have certain function selector. call it something else

eyusufatik (changes requested):

eyusufatik:
it's going to be a pain in the ass to release 0.7.2 while the risc0 2.0.2 crates are yanked...

kpp:
So that's for fullnodes to sync only? Let's say in mainnet we'll need to actually return error?

eyusufatik:
> So that's for fullnodes to sync only? Let's say in mainnet we'll need to actually return error?

Not this is fine as is

eyusufatik (dismissed):

eyusufatik (approved):

PR #2465, opened 2025-06-18
Handle potential mutex poisoning in native short header proof provider

ercecan:
# Description
Handles potential mutex poisoning in native short header proof provider

The NativeShortHeaderProofProviderService is vulnerable to a denial-of-service attack due to improper handling of mutex locks. Panics within critical sections, such as from out-of-memory errors during vector allocations, can poison a mutex, causing subsequent lock attempts to panic and crash the service
The NativeShortHeaderProofProviderService utilizes a mutexed hashmap named queried_and_verified_hashes to store verified block hashes. Critical sections accessing this shared data employ lock() .unwrap() or lock(). expect() for acquiring the mutex. If a thread panics while holding this lock, for instance, due to an out-of-memory (OOM) error during a vector allocation, the mutex becomes poisoned. Consequently, any subsequent attempt by another thread to acquire the lock via unwrap() or expect() will also panic, leading to a halt in the service's operation and a denial of service for the node.

### DIFF
--- a/crates/evm/src/evm/executor.rs
+++ b/crates/evm/src/evm/executor.rs
@@ -200,6 +200,11 @@ fn verify_system_tx<C: sov_modules_api::Context>(
             Err(ShortHeaderProofProviderError::ShortHeaderProofNotFound) => {
                 return Err(L2BlockModuleCallError::ShortHeaderProofNotFound);
             }
+            #[cfg(feature = "native")]
+            Err(ShortHeaderProofProviderError::MutexPoisoned(e))
+            | Err(ShortHeaderProofProviderError::VectorAllocationFailed(e)) => {
+                return Err(L2BlockModuleCallError::ShortHeaderProofOtherError(e));
+            }
         }
     }
 
--- a/crates/evm/src/evm/executor.rs
+++ b/crates/evm/src/evm/executor.rs
@@ -200,9 +200,8 @@ fn verify_system_tx<C: sov_modules_api::Context>(
             Err(ShortHeaderProofProviderError::ShortHeaderProofNotFound) => {
                 return Err(L2BlockModuleCallError::ShortHeaderProofNotFound);
             }
-            #[cfg(feature = "native")]
             Err(ShortHeaderProofProviderError::VectorAllocationFailed(e)) => {
-                return Err(L2BlockModuleCallError::ShortHeaderProofOtherError(e));
+                return Err(L2BlockModuleCallError::ShortHeaderProofAllocationError(e));
             }
         }
     }
--- a/crates/evm/src/evm/executor.rs
+++ b/crates/evm/src/evm/executor.rs
@@ -201,8 +201,7 @@ fn verify_system_tx<C: sov_modules_api::Context>(
                 return Err(L2BlockModuleCallError::ShortHeaderProofNotFound);
             }
             #[cfg(feature = "native")]
-            Err(ShortHeaderProofProviderError::MutexPoisoned(e))
-            | Err(ShortHeaderProofProviderError::VectorAllocationFailed(e)) => {
+            Err(ShortHeaderProofProviderError::VectorAllocationFailed(e)) => {
                 return Err(L2BlockModuleCallError::ShortHeaderProofOtherError(e));
             }
         }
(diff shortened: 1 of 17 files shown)


crates/sequencer/src/runner.rs
```diff
@@ -347,6 +347,9 @@ where
                             L2BlockModuleCallError::EvmSystemTransactionNotSuccessful => {
                                 panic!("System tx failed")
                             }
+                            L2BlockModuleCallError::ShortHeaderProofOtherError(e) => {
+                                panic!("Short header proof error: {:?}", e);
```
ercecan, line 351: Should I still panic here?

jfldde:
Why not switching to parking_lot which doesn't poison in native ?

ercecan:
> Why not switching to parking_lot which doesn't poison in native ?

Let me do that makes sense

crates/short-header-proof-provider/Cargo.toml
```diff
@@ -14,6 +14,7 @@ resolver = "2"
 borsh = { workspace = true }
 futures = { workspace = true, optional = true }
 once_cell = { workspace = true, default-features = false }
+parking_lot = { workspace = true, default-features = false }
```
jfldde: Optional and only under native

crates/evm/src/evm/executor.rs
```diff
@@ -200,6 +200,10 @@ fn verify_system_tx<C: sov_modules_api::Context>(
             Err(ShortHeaderProofProviderError::ShortHeaderProofNotFound) => {
                 return Err(L2BlockModuleCallError::ShortHeaderProofNotFound);
             }
+            #[cfg(feature = "native")]
+            Err(ShortHeaderProofProviderError::VectorAllocationFailed(e)) => {
+                return Err(L2BlockModuleCallError::ShortHeaderProofOtherError(e));
+            }
```
eyusufatik: isn't this basically unreachable here? no `get_and_verify_short_header_proof_by_l1_hash` impl. returns this error

crates/short-header-proof-provider/Cargo.toml
eyusufatik: native feature?

crates/sovereign-sdk/rollup-interface/src/state_machine/stf.rs
```diff
@@ -139,6 +139,9 @@ pub enum L2BlockModuleCallError {
     ShortHeaderProofNotFound,
     /// Short Header Proof Verification Error
     ShortHeaderProofVerificationError,
+    /// Some other error related to short header proof
+    /// This will occur if mutex is poisoned or vector reservation fails in native side
+    ShortHeaderProofOtherError(String),
```
eyusufatik: let's not go Other(String) way

guests/risc0/batch-proof/bitcoin/Cargo.lock
eyusufatik, line 1: these would go away if you native feature gate parking_lot

crates/short-header-proof-provider/src/lib.rs
```diff
@@ -14,6 +14,9 @@ pub use zk::*;
 pub enum ShortHeaderProofProviderError {
     #[error("Short header proof not found")]
     ShortHeaderProofNotFound,
+    #[cfg(feature = "native")]
```
eyusufatik: no need for this

eyusufatik: I mean the feature gating

eyusufatik (changes requested):

crates/evm/src/evm/executor.rs
ercecan: The native one can return that error

crates/short-header-proof-provider/src/native.rs
```diff
@@ -58,12 +59,20 @@ impl<Da: DaSpec> ShortHeaderProofProvider for NativeShortHeaderProofProviderServ
...
+                            const BLOCK_HASH_SIZE: usize = 32;
+                            occ.get_mut().try_reserve(BLOCK_HASH_SIZE).map_err(|e| {
+                                ShortHeaderProofProviderError::VectorAllocationFailed(e.to_string())
+                            })?;
+                            occ.get_mut().push(block_hash);
+                        }
+                        std::collections::hash_map::Entry::Vacant(vac) => {
+                            vac.insert(vec![block_hash]);
+                        }
+                    }
```
jfldde, line 75: Seems a bit over the top for an alloc of 32

jfldde (approved):

eyusufatik (approved):

PR #2466, opened 2025-06-18
Remove redundant sequencer pubkey check in begin_l2_block

exeokan:
# Description
The check in `begin_l2_block` was redundant since we pass the same public key to both `l2_block_info` and `sequencer_public_key` parameters. The actual check happens in `verify_l2_block` where we verify the l2 block signature with the `sequencer_public_key`

Also changed the test `test_wrong_sequencer_public_key` to `test_wrong_l2_block_signature`, now testing the `verify_l2_block`.


crates/sovereign-sdk/rollup-interface/src/state_machine/stf.rs
```diff
@@ -70,19 +70,10 @@ pub struct L2BlockResult<Cs, W, SL> {
 #[derive(Debug, PartialEq)]
 /// Error in the l2 block itself
 pub enum L2BlockError {
-    /// The public key of the sequencer (known by a full node or prover) does not match
```
eyusufatik, line 73: these were unused?

eyusufatik (approved):

crates/sovereign-sdk/rollup-interface/src/state_machine/stf.rs
exeokan: yes, and `SequencerPublicKeyMismatch` just became one

ercecan (approved):

PR #2468, opened 2025-06-19
Pass LedgerDb to get_sealed_block_by_number

### LINKED ISSUE #2434: feat: Allow `safe` and `finalized` for ethereum rpcs
can be done after #2433

The namings will be a bit confusing, but we'll write this down in our documentation.

For the `safe` block use the result of the result of `citrea_getLastCommittedL2Height`.
For the `finalized` block use the result of the result of `citrea_getLastProvenL2Height`.

These will only work on full nodes but that's okay.

kpp:
# Description

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/2434


crates/evm/src/query.rs
eyusufatik, line 1: functions should be updated:
- set_state_to_end_of_evm_block_by_block_id
- block_number_for_id

eyusufatik: check everywhere that has `BlockNubmerOrTag`

eyusufatik (changes requested):
write new tests that show:
- block number for safe and finalized are always same with committed and proven.
- things that accept safe or finalized as their block tag behave correctly


I think the best place to write these would be mock_e2e evm

crates/evm/src/query.rs
```diff
@@ -1709,6 +1750,36 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
             )),
+            Some(BlockNumberOrTag::Safe) => {
+                let block_number = ledger_db
+                    .get_highest_l2_height_for_status(L2HeightStatus::Committed, None)
+                    .map_err(|e| EthApiError::InvalidParams(e.to_string()))?;
+                let Some(block_number) = block_number else {
+                    return Err(EthApiError::InvalidParams(
+                        "safe block is not found".to_string(),
+                    ));
+                };
```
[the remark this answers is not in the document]
eyusufatik: we may use 0 (genesis) for safe if the get_highest_l2_height_for_status functions return None for when there is no committed / proven

bin/citrea/tests/mock/evm/mod.rs
```diff
@@ -1264,3 +1267,256 @@ async fn eip7702_tx_test() -> Result<(), anyhow::Error> {
...
+    async fn compare_with_numbered_params(
+        test_client: &TestClient,
+        expected_safe_block_height: u64,
+        expected_finalized_block_height: u64,
+        test_address: alloy_primitives::Address,
+    ) {
+        println!(
+            "Comparing safe and finalized blocks with expected heights: safe={}, finalized={}",
+            expected_safe_block_height, expected_finalized_block_height
+        );
```
eyusufatik: remove or use tracing::info! so that we don't see this while running tests

crates/evm/src/query.rs
```diff
@@ -1642,15 +1682,26 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
                     Err(EthApiError::HeaderNotFound((*block_id).into()))
                 }
             }
-            _ => Err(EthApiError::InvalidParams(
-                "Please provide a number or earliest/latest/pending tag".to_string(),
-            )),
+            BlockNumberOrTag::Safe => {
+                let block_number = ledger_db
+                    .get_highest_l2_height_for_status(L2HeightStatus::Committed, None)
+                    .map_err(|e| EthApiError::InvalidParams(e.to_string()))?;
```
eyusufatik: change error type to internal error somehow 😅

crates/evm/src/query.rs
```diff
@@ -1642,15 +1682,26 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+                let block_number = block_number.map(|b| b.height).unwrap_or_default();
+                Ok(block_number)
+            }
+            BlockNumberOrTag::Finalized => {
+                let block_number = ledger_db
+                    .get_highest_l2_height_for_status(L2HeightStatus::Proven, None)
+                    .map_err(|e| EthApiError::InvalidParams(e.to_string()))?;
```
eyusufatik: same

crates/evm/src/query.rs
```diff
@@ -1709,14 +1762,34 @@ impl<C: sov_modules_api::Context> Evm<C> {
                     .last(&mut working_set.accessory_state())
                     .expect("Head block must be set"),
             )),
+            Some(BlockNumberOrTag::Safe) => {
+                let block_number = ledger_db
+                    .get_highest_l2_height_for_status(L2HeightStatus::Committed, None)
+                    .map_err(|e| EthApiError::InvalidParams(e.to_string()))?;
```
eyusufatik: error

crates/evm/src/query.rs
```diff
@@ -1709,14 +1762,34 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+                let block_number = block_number.map(|b| b.height).unwrap_or_default();
+
+                Ok(self
+                    .blocks
+                    .get(block_number as usize, &mut working_set.accessory_state()))
+            }
+            Some(BlockNumberOrTag::Finalized) => {
+                let block_number = ledger_db
+                    .get_highest_l2_height_for_status(L2HeightStatus::Proven, None)
+                    .map_err(|e| EthApiError::InvalidParams(e.to_string()))?;
```
eyusufatik: error

bin/citrea/tests/mock/evm/mod.rs
```diff
@@ -1264,3 +1267,256 @@ async fn eip7702_tx_test() -> Result<(), anyhow::Error> {
...
+    let _ = test_client
+        .send_eth(Address::random(), None, None, None, 1_000_000)
+        .await
+        .unwrap();
+
+    test_client.send_publish_batch_request().await;
+    test_client.send_publish_batch_request().await;
+    wait_for_l2_block(&full_node_client, 4, None).await;
+
+    compare_with_numbered_params(&full_node_client, 0, 0, tx_sender).await;
```
eyusufatik: add comment that no commitment or proof has been sent, so safe = 0 finalized = 0

eyusufatik: you can move this just before the commitment is sent or test will be flaky

bin/citrea/tests/mock/evm/mod.rs
```diff
@@ -1264,3 +1267,256 @@ async fn eip7702_tx_test() -> Result<(), anyhow::Error> {
...
+
+    // wait for prover to see commitment, since sampling is too high, proving won't be triggered here
+    wait_for_prover_l1_height(&prover_client, 3, None)
+        .await
+        .unwrap();
+    wait_for_prover_l1_height(&full_node_client, 3, None)
+        .await
+        .unwrap();
+
+    compare_with_numbered_params(&full_node_client, 4, 0, tx_sender).await;
```
eyusufatik, line 1483: comment

bin/citrea/tests/mock/evm/mod.rs
```diff
@@ -1264,3 +1267,256 @@ async fn eip7702_tx_test() -> Result<(), anyhow::Error> {
...
+
+    let proofs = full_node_client
+        .ledger_get_verified_batch_proofs_by_slot_height(4)
+        .await
+        .unwrap();
+    assert_eq!(proofs.len(), 1);
+    let proof = &proofs[0];
+    assert_eq!(proof.proof_output.last_l2_height.to::<u64>(), 4);
+
+    compare_with_numbered_params(&full_node_client, 4, 4, tx_sender).await;
```
eyusufatik, line 1506: comment

eyusufatik (changes requested):

crates/evm/src/query.rs
exeokan: decided panicking with expect is easier

eyusufatik (approved):

crates/evm/src/query.rs
```diff
@@ -636,15 +645,23 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+            fork_from_block_number,
+        )
     }
 
     pub(crate) fn create_access_list_inner(
         &self,
         request: TransactionRequest,
         block_number: Option<BlockNumberOrTag>,
         working_set: &mut WorkingSet<C::Storage>,
+        ledger_db: &crate::LedgerDB,
```
yaziciahmet, line 664: why is this imported as `crate::`?

yaziciahmet (commented):
My only concern with this PR is that we enforce ledger_db into Evm methods even though except RPC it is not needed. I think it would be better if we create a new wrapper structure called `BlockProvider` that contains ledger_db and evm instance in it, and orchestrate the querying of blocks based on the corresponding tag. And other components such as sequencer etc. can still just use raw evm and wouldn't need to pass in ledger db.

Let me know if this not feasible due to something I am missing.

jfldde:
I second Ahmet suggestion.
We could have an `EthRpcProvider` which holds both `evm` and `ledgerdb` on which we would derive the `#[rpc_gen(client, server)]` macros in `query.rs` (could be renamed to `rpc.rs` in line with the other files as well). 
This should tidy up passing down ledger_db everywhere

PR #2469, opened 2025-06-19
prev hash proof

eyusufatik:
## TODOs
- [x] Remove index gapped proving
- [x] Use the new input field in the prover
- [x] Assert equality of the init root and state root from the prev_hash's header 
- [x] Write unit test for prev_hash proof and edge cases
- [x] Remove index gap using tests


crates/batch-prover/src/prover.rs
```diff
@@ -1076,6 +1084,55 @@ fn generate_cumulative_witness<Da: DaService, DB: BatchProverLedgerOps>(
...
+            .l2_end_block_number
+            + 1
+    };
+
+    let blocks = ledger_db
+        .get_l2_block_range(
+            &(L2BlockNumber(prev_commitment_start_height)
+                ..=L2BlockNumber(previous_commitment.l2_end_block_number)),
+        )
+        .unwrap();
```
yaziciahmet, line 1110: we do this twice in proving process (I think first one is when regenerating witness) and this is quite mem intensive as it could load about 1000 blocks into memory at once. can we get l2 blocks beforehand and pass as arg into both places?

yaziciahmet (changes requested):
haven't looked into stf blueprint yet, checked everything else. leaving one review right away.

crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
```diff
@@ -528,11 +528,17 @@ where
                     "Prev hash proof must be valid"
                 );
 
-                last_header_hash
+                Some(last_header_hash)
             }
             None => {
                 assert!(prev_hash_proof.is_none());
-                [0; 32] // This is the going to start from the first l2 block
+                // If the chain starts with Tangerine fork or higher, we know that the previous hash must be [0; 32],
```
eyusufatik: you can this if will be entered on mainnet, local tests with tangerine only, or local dev launches etc. 

None only applies to networks that were not initially deployed with tangerine

crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
```diff
@@ -638,11 +644,15 @@ where
                     "L2 block height is not equal to the expected height"
                 );
 
-                assert_eq!(
-                    l2_block.prev_hash(),
-                    prev_l2_block_hash,
-                    "L2 block previous hash must match the hash of the block before"
-                );
+                // There is no previous l2 block hash. For mainnet this is going to be the case for the 1st block.
+                // But for testnet it will be the Tangerine fork start, hence, we will have to have trust in the first proof.
```
eyusufatik, line 649: kinda sounds like mainnet won't have the first prev_hash, let's fix language

crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
ercecan: Makes sense

ercecan (approved):
Logic seems sound to me

crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
yaziciahmet: updated

eyusufatik:
batch prover related tests are too flaky. put hold merge label

crates/batch-prover/src/prover.rs
yaziciahmet: talked offline with @eyusufatik , this is the blocks of prev commitment so this can't be done.

crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
```diff
@@ -485,7 +487,62 @@ where
...
+                    prev_hash_proof.last_header.state_root(),
+                    *initial_state_root,
+                    "Initial state root must match the last header state root"
+                );
+
+                // TODO: Is this ok???
+                let index = (commitment.l2_end_block_number
+                    - prev_hash_proof.prev_sequencer_commitment_start)
+                    as usize;
+                let count = index + 1;
```
yaziciahmet: index seems wrong. we can just do:
```
let count = commitment.l2_end_block_number - prev_hash_proof.prev_sequencer_commitment_start + 1;
```

also remove todo

yaziciahmet (approved):
LGTM, approving with one nit

crates/sovereign-sdk/module-system/sov-modules-stf-blueprint/src/lib.rs
eyusufatik: well index is correct actually.

though my comment (// TODO: is this ok???) is actually regarding the safety of the operation.

the rs_merkle::MerkleProof::verify() function needs the leaf, proof, leaf count and leaf index to verify.

from what  I understand, given a merkle root, we can't fool the verify function by giving different index or different start block height, as that breaks the index etc.

yaziciahmet: aah its merkle tree index, my mind went to commitment index. then ok, my bad

eyusufatik: checked with @ozankaymak and the comment is fine, I removed it

PR #2470, opened 2025-06-19
The current timestamp of the block should be strictly greater than thee timestamp of the last block

ercecan:
# Description
The current timestamp of the block should be strictly greater than thee timestamp of the last block


eyusufatik:
@ercecan realized we have a bunch of consecutive blocks with the same timestamp. this is due to how we handled setblockinfo events before. so we won't fix this for now.

PR #2471, opened 2025-06-19
Non empty batch proof initial method ids check

exeokan:
# Description
Adds `non_empty_slice` const fn to light client circuit's `initial_values` module.


eyusufatik (changes requested):
add unit test showing the function works. you can use should_panic()

+ pls try compiling the project while one of the slices are actually empty

eyusufatik (approved):

yaziciahmet (approved):

kpp (approved):

PR #2482, opened 2025-06-20
Limit max decompressed blob size

ercecan:
# Description
Limits max decompressed blob size

changes how Complete proofs are compressed and serialized in order to make the decompression last step of the data extraction, so that we can filter blobs by checking pubkeys.


crates/primitives/src/constants.rs
```diff
@@ -39,3 +39,12 @@ pub const PRE_TANGERINE_BRIDGE_INITIALIZE_PARAMS: &[u8] = &[
     10, 8, 0, 0, 0, 0, 59, 154, 202, 0, 104, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
     0, 0, 0, 0,
 ];
+
+/// Maximum size of a decompressed blob in bytes.
+/// This is set to 1 MB for testing and 100 MB for production to allow larger blobs in real scenarios.
+/// This limit is enforced during decompression to prevent excessive memory usage
+/// and potential denial of service attacks like decompression bombs.
+#[cfg(feature = "testing")]
+pub const MAX_DECOMPRESSED_BLOB_SIZE: usize = 1024 * 1024; // 1 MB
```
eyusufatik, line 48: do we need to seperate testing vs non testing?

crates/primitives/src/compression.rs
```diff
@@ -8,14 +10,95 @@ pub fn compress_blob(blob: &[u8]) -> Result<Vec<u8>, io::Error> {
...
+
+    let mut buf = [0u8; 4096];
+
+    let mut decompressed_data = Vec::new();
+
+    loop {
+        match reader.read(&mut buf) {
+            Err(e) => {
+                if let io::ErrorKind::Interrupted = e.kind() {
+                    continue;
```
eyusufatik, line 27: continue feels wrong here idk

crates/primitives/src/compression.rs
```diff
@@ -8,14 +10,95 @@ pub fn compress_blob(blob: &[u8]) -> Result<Vec<u8>, io::Error> {
...
-            "Brotli decompression failure",
-        )),
+    let mut reader = brotli::Decompressor::new(
+        blob,
+        4 * 1024, // 4kb buffer size
+    );
+
+    let mut buf = [0u8; 4096];
+
+    let mut decompressed_data = Vec::new();
```
eyusufatik: should we pre-allocate some size? something like 200kB makes sense imo

eyusufatik (commented):
changes look good. just have a few questions

crates/primitives/src/compression.rs
ercecan: I believe this is the only error type where we can continue because in the docs it says:
`/// Interrupted operations can typically be retried.`

crates/primitives/src/constants.rs
ercecan: Feels better for testing 100mb sounds big for testing but I can rm if u want

guests/risc0/batch-proof/bitcoin/Cargo.lock
eyusufatik, line 1: shouldn't change maybe?

crates/bitcoin-da/src/service.rs
```diff
@@ -965,21 +965,22 @@ impl DaService for BitcoinService {
                         if complete.public_key() == prover_da_pub_key
                             && complete.get_sig_verified_hash().is_some()
                         {
-                            // push only when signature is correct
-                            let Ok(body) = decompress_blob(&complete.body) else {
-                                warn!("{tx_id}: Failed to decompress blob");
+                            let Ok(data) = DataOnDa::borsh_parse_complete(&complete.body) else {
```
eyusufatik: @kpp why does this function exist?

guests/risc0/batch-proof/bitcoin/Cargo.lock
ercecan: `brotli-7.0.0`  has `brotli-decompressor version = "~4.0.0"` that is probably why it updated from `4.0.2` to `4.0.3`

ercecan: it also is a rather recent update as well

crates/bitcoin-da/src/service.rs
kpp: Because it used to be under Fork switch for backwards compat

eyusufatik: how to make this better now?

crates/bitcoin-da/src/service.rs
```diff
@@ -1467,19 +1464,17 @@ impl From<TxidWrapper> for [u8; 32] {
 }
 
 /// This function splits Proof based on its size. It is either:
-/// 1: compress(borsh(DataOnDa::Complete(Proof)))
+/// 1: borsh(DataOnDa::Complete(compress(Proof)))
 /// 2:
-///   let compressed = compress(borsh(Proof))
```
kpp, line 1472: Why `borsh` removed?

eyusufatik: was just borshing a Vec<u8> i think

eyusufatik: which is already borsh(InnerReceipt)

kpp: Ok, I see, but this is a hardfork

eyusufatik: yeah, we are going to reprove everything anyways

eyusufatik: due to not wanting to make backwards compatibile upgrade to risc0 2.1.0

crates/bitcoin-da/src/service.rs
```diff
@@ -965,21 +965,22 @@ impl DaService for BitcoinService {
...
-                                warn!("{tx_id}: Failed to parse complete data");
+                            let DataOnDa::Complete(compressed_zk_proof) = data else {
+                                warn!("{}: Complete: unexpected kind", tx_id);
                                 continue;
                             };
 
-                            let DataOnDa::Complete(zk_proof) = data else {
-                                warn!("{}: Complete: unexpected kind", tx_id);
+                            // push only when signature is correct
+                            let Ok(zk_proof) = self.decompress_chunks(&compressed_zk_proof) else {
```
kpp, line 979: decompress_chunks no longer have to borsh::from_slice for `Vec<u8>`?

crates/bitcoin-da/src/service.rs
kpp: Now we no longer need it. We can do borsh::from_slice instead, it's gonna be similar.

Cargo.toml
```diff
@@ -63,7 +63,7 @@ bitcoin = { version = "0.32.5", features = ["serde", "rand"] }
 bitcoincore-rpc = "0.18.0"
 bcs = "0.1.6"
 bech32 = { version = "0.9.1", default-features = false }
-brotli = "7"
+brotli = "7.0.0"
```
kpp: Why? 7 and 7.0.0 is same

crates/bitcoin-da/src/service.rs
ercecan: no, as we do not borsh complete proof now

Cargo.toml
ercecan: Just made some copy pasterino, will revert

eyusufatik (approved):

crates/primitives/src/compression.rs
```diff
@@ -8,14 +10,95 @@ pub fn compress_blob(blob: &[u8]) -> Result<Vec<u8>, io::Error> {
...
+        assert_eq!(
+            decompressed_blob, blob,
+            "Decompressed blob should match original"
+        );
+    }
+
+    #[test]
+    fn test_decompress_exceeds_max_size() {
+        let huge_blob = vec![0u8; MAX_DECOMPRESSED_BLOB_SIZE * 10]; // 10 times the max size
+        let compressed_huge_blob = compress_blob(&huge_blob).expect("Compression failed");
```
jfldde, line 83: Shouldn't we fail here then ? Do we want to be able to compress blobs that will be un-decompressable ?

eyusufatik: this is to test the limit on compressability, othewise the zk vm would run out of memory

crates/primitives/src/compression.rs
```diff
@@ -8,14 +10,95 @@ pub fn compress_blob(blob: &[u8]) -> Result<Vec<u8>, io::Error> {
...
-            "Brotli decompression failure",
-        )),
+    let mut reader = brotli::Decompressor::new(
+        blob,
+        4 * 1024, // 4kb buffer size
+    );
+
+    let mut buf = [0u8; 4096];
+
+    let mut decompressed_data = Vec::with_capacity(200 * 1024); // Preallocate for 200 KB
```
yaziciahmet: Maybe we can make an optimistic guess here. we know that on avg brotli compresses our proof by about 66%. we can assume a compression of 70% to be safe and just preallocate `blob.len() * 10 / 3` bytes.

Also please check if brotli has any "approximate_size" like apis that could help for preallocation size.

eyusufatik: makes sense, though i think this is better:

```
max(blob.len() * 10 / 3, 400kB)
```

jfldde (approved):

yaziciahmet (approved):
LgTm

crates/primitives/src/compression.rs
eyusufatik: should be min as @ercecan stated

PR #2484, opened 2025-06-22
Use consistent endianess in sov tx serialization

### LINKED ISSUE #2370: Transaction stores borsh serialized signature

kpp:
# Description

Change the chain_id and nonce serialisation to big-endian to keep endianess consistent with sov-tx::compute_digest.

Fixes: https://github.com/chainwayxyz/citrea/issues/2370


crates/sovereign-sdk/rollup-interface/src/state_machine/transaction.rs
```diff
@@ -129,17 +185,23 @@ impl Transaction {
...
     #[cfg(feature = "native")]
     pub fn new_signed_tx(
+        spec: SpecId,
         priv_key: &K256PrivateKey,
         runtime_msg: Vec<u8>,
         chain_id: u64,
         nonce: u64,
     ) -> Self {
-        Self::V1(TransactionV1::new(priv_key, runtime_msg, chain_id, nonce))
+        if spec < SpecId::Tangerine {
```
eyusufatik: why tangerine? we already deployed it. I think we can define a fork4

eyusufatik: so if equal to or greater than fork4, we use tx type v2

eyusufatik (changes requested):

eyusufatik:
maybe we should turn all into little endian to have less cycles inside risc0 zkvm?

crates/sovereign-sdk/rollup-interface/src/spec.rs
```diff
@@ -35,7 +35,6 @@ pub enum SpecId {
     /// Tangerine spec
     Tangerine = 2,
     /// Fork3 spec
-    #[cfg(feature = "testing")]
```
eyusufatik, line 38: i think you should also update nightly and testnetworkwithforks Fork3

kpp:
All little endian is great if only it mattered more than 2 cycles per 1 sov tx (thousands of evm txs). I can do that if you want to

eyusufatik:
> All little endian is great if only it mattered more than 2 cycles per 1 sov tx (thousands of evm txs). I can do that if you want to

for real? then let's not do that because all other stuff like blocks is be already

crates/sequencer/src/commitment/service.rs
```diff
@@ -294,7 +294,7 @@ where
             } else {
                 // Submit commitment
                 let l2_start_block_number = if pending_db_comm.index == 1 {
-                    get_tangerine_activation_height_non_zero()
+                    get_fork3_activation_height_non_zero()
```
eyusufatik: this shouldn't have changed

crates/primitives/src/forks.rs
```diff
@@ -82,18 +83,18 @@ const _CHECK_FORKS: () = {
...
     let forks = get_forks();
     let fork = forks
         .iter()
-        .find(|f| f.spec_id == SpecId::Tangerine)
-        .expect("Tangerine should exist");
+        .find(|f| f.spec_id == SpecId::Fork3)
+        .expect("Fork3 should exist");
     if fork.activation_height == 0 {
         return 1;
     }
```
eyusufatik: revert

eyusufatik (changes requested):

crates/sovereign-sdk/rollup-interface/src/state_machine/transaction.rs
```diff
@@ -58,12 +62,75 @@ impl TransactionV1 {
...
+        hasher.update(self.chain_id.to_be_bytes());
+        hasher.update(self.nonce.to_be_bytes());
+        hasher.finalize()
+    }
+}
+
+/// A Transaction object that is compatible with the module-system/sov-default-stf.
+#[derive(Debug, PartialEq, Eq, Clone, borsh::BorshDeserialize, borsh::BorshSerialize)]
+pub struct TransactionV2 {
+    signature: Vec<u8>,
```
eyusufatik: we can do the #2370 here btw

eyusufatik:
add why we have fork3 on forks.rs or somewhere relevant. (fixes for vulnerabilities that need forking on existing networks)

crates/sovereign-sdk/rollup-interface/src/spec.rs
jfldde: Should update `from_u8` accordingly

crates/sovereign-sdk/rollup-interface/src/state_machine/transaction.rs
```diff
@@ -8,13 +8,17 @@ use sov_keys::default_signature::{K256PublicKey, K256Signature};
 use sov_keys::PrivateKey;
 use sov_keys::Signature;
 
-const EXTEND_MESSAGE_LEN: usize = 2 * core::mem::size_of::<u64>();
+#[cfg(feature = "native")]
+use crate::spec::SpecId;
+
+const EXTEND_MESSAGE_LEN: usize = 1 + 2 * core::mem::size_of::<u64>();
```
jfldde:
```suggestion
const EXTEND_MESSAGE_LEN: usize = 1 + 2 * core::mem::size_of::<u64>(); // version (u8) + chain_id (u64) + nonce (u64) 
```

jfldde (approved):

eyusufatik (approved):

PR #2485, opened 2025-06-23
Implement halt/resume commitments

### LINKED ISSUE #2481: feat: Sequencer stop committing RPC
Can be useful if we need to stop commitment sending. When a bug is found etc.

Upon restart the committing can continue normally.

rakanalh:
# Description

## Linked Issues
- Fixes #2481


jfldde (approved):

crates/sequencer/src/types.rs
```diff
@@ -0,0 +1,8 @@
+/// An enum representing the possible messages to send
+/// from the RPC handler into the sequencer
+pub enum SequencerRpcMessage {
+    /// Produce a test block in the sequencer
+    ProduceTestBlock,
+    /// Halt or resume sequencer commitments.
+    HaltCommitments(bool),
```
jfldde: Maybe split to two variants `HaltCommitments` && `ResultCommitments` instead of holding a bool

crates/sequencer/src/commitment/service.rs
```diff
@@ -41,6 +41,10 @@ where
     sequencer_da_pub_key: Vec<u8>,
     /// Maximum number of L2 blocks that can be included in a single commitment
     max_l2_blocks: u64,
+    /// Channel for receiving halt signals from the runner
+    halt_rx: mpsc::UnboundedReceiver<bool>,
+    /// Current running state
+    is_running: bool,
```
kpp: `is_producing_commitments`

crates/sequencer/src/commitment/service.rs
```diff
@@ -100,9 +111,37 @@ where
...
+                            warn!("CommitmentService: Halt signal channel closed");
+                            return;
+                        }
+                    }
+                },
                 _ = check_new_block_tick.tick() => {
+                    // Skip commitment processing if not running
+                    if !self.is_running {
+                        debug!("CommitmentService: Skipping commitment processing (halted)");
+                        continue;
```
kpp, line 142: This is very interesting. Shall we stop them and ignore new blocks OR we shall keep those blocks and continue producing commitments after we resumed from the pause point?

kpp (approved):

ercecan (approved):

crates/sequencer/src/commitment/service.rs
rakanalh: The commitments service has been made separate from block signals... so this would just halt and resume commitments from the last point we committed.

bin/citrea/tests/mock/sequencer_behaviour.rs
```diff
@@ -678,3 +678,118 @@ fn find_subarray(haystack: &[u8], needle: &[u8]) -> Option<usize> {
...
+    // Resume commitments via RPC
+    seq_test_client
+        .sequencer_resume_commitments()
+        .await
+        .unwrap();
+
+    // Wait a bit for the resume signal to be processed
+    tokio::time::sleep(Duration::from_millis(100)).await;
+
+    // Create one more L2 block to trigger commitment processing
```
eyusufatik: do we need this? as commitment service now checks on its own, should commit without this.

eyusufatik (approved):

PR #2491, opened 2025-06-23
Refactor verify_tx_merkle_root

kpp:
# Description

- Get rid of Spec::Hasher (use sha256 everywhere)
- Use common code to build/verify merkle root
- Add separators to merkle tree leaves

### DIFF
--- a/crates/primitives/src/merkle.rs
+++ b/crates/primitives/src/merkle.rs
@@ -1,27 +1,60 @@
-use rs_merkle::algorithms::Sha256;
+use rs_merkle::algorithms::Sha256 as Sha256WithoutSeparator;
 use rs_merkle::MerkleTree;
+use sha2::digest::{Digest, FixedOutput};
+use sov_rollup_interface::spec::SpecId;
 use sov_rollup_interface::transaction::Transaction;
 
 use crate::EMPTY_TX_ROOT;
 
-pub fn compute_tx_hashes(txs: &[Transaction]) -> Vec<[u8; 32]> {
-    txs.iter().map(|tx| tx.compute_digest()).collect()
+pub fn compute_tx_hashes(txs: &[Transaction], spec: SpecId) -> Vec<[u8; 32]> {
+    let digest_fn = if spec >= SpecId::Fork3 {
+        |tx: &Transaction| {
+            // rehash with separator
+            let hash = tx.compute_digest();
+            let mut hasher = sha2::Sha256::new_with_prefix([0]);
+            hasher.update(hash);
+            <[u8; 32]>::from(hasher.finalize_fixed())
+        }
+    } else {
+        |tx: &Transaction| tx.compute_digest()
+    };
+    txs.iter().map(digest_fn).collect()
 }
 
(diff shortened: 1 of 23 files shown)


crates/primitives/src/merkle.rs
```diff
@@ -0,0 +1,27 @@
...
+        .root()
+        .expect("Couldn't compute merkle root")
+}
+
+pub fn verify_tx_merkle_root(txs: &[Transaction], root: [u8; 32]) -> bool {
+    // Calculate tx hashes for merkle root
+    let tx_hashes = compute_tx_hashes(txs);
+    let tx_merkle_root = compute_tx_merkle_root(&tx_hashes);
+
+    tx_merkle_root != root
```
eyusufatik: shouldn't this be == ?

crates/primitives/src/merkle.rs
```diff
@@ -0,0 +1,27 @@
+use rs_merkle::algorithms::Sha256;
+use rs_merkle::MerkleTree;
+use sov_rollup_interface::transaction::Transaction;
+
+use crate::EMPTY_TX_ROOT;
+
+pub fn compute_tx_hashes(txs: &[Transaction]) -> Vec<[u8; 32]> {
```
eyusufatik: rehash tx hashes by concatting with a prefix? like the recommendation? I don't know how we are going to the internal nodes though

ercecan (approved):
LGTM

crates/primitives/src/merkle.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+pub fn compute_digest_post_fork3(tx: &Transaction) -> [u8; 32] {
+    let mut hasher = sha2::Sha256::new_with_prefix([0]);
+    hasher.update([TxVersion::V2 as u8]);
+    hasher.update(tx.runtime_msg());
+    hasher.update(tx.chain_id().to_be_bytes());
+    hasher.update(tx.nonce().to_be_bytes());
+    <[u8; 32]>::from(hasher.finalize_fixed())
+}
```
eyusufatik: I don't like these being here tbh. they should be under transaction.rs

eyusufatik (changes requested):

eyusufatik:
looks clean

crates/primitives/src/merkle.rs
kpp: Yes and no. Shall Sha256WithSeparator be inside transaction.rs? The implementation of this method lives here becase no-one else uses it, its also highly specific to merkle tree because of separators.
What do you think?

eyusufatik: but tx.compute_digest() might need to be called everywhere

imo revert the removal of compute_digest these from transaction.rs

and then on `compute_tx_hashes` in primitives, just call Transaction::compute_digest, which calls the underlying hash functions

eyusufatik (changes requested):
prefix should be added to tx hash not tx itself as we discussed on the call

ercecan (approved):

eyusufatik (approved):

PR #2504, opened 2025-06-30
bitcoin-da: Queue Txs before sending to da

### LINKED ISSUE #2461: Queue transactions before sending to DA
Rationale for queueing txs is that we chunk over MAX_TX_BODY_SIZE value of 397000 which is just under default mempool policy limit DEFAULT_DESCENDANT_SIZE_LIMIT_KVB of 101kvb. That means that on mainnet network, a single chunk pair with aggregate (smallest possible chunking result) would be rejected. We need to send chunk part at block N and aggregate at block N+1 (where N is a block which includes chunks, which would free up space for subsequent txs).

- Prevent having package stuck with `package-tool-large` mempool rejection

jfldde:
# Description

## Explanation
Rationale for queueing txs is that we chunk over `MAX_TX_BODY_SIZE` value of `397000` which is just under default mempool policy limit `DEFAULT_DESCENDANT_SIZE_LIMIT_KVB` of 101kvb. That means that on mainnet network, a single chunk pair with aggregate (smallest possible chunking result) would be rejected. We need to send chunk part at block N and aggregate at block N+1 (where N is a block which includes chunks, which would free up space for subsequent txs).

- Add an intermediate queueing step before sending transactions to DA.
Flow changes : 

Old: Create → Sign → Send immediately
New: Create → Sign → Queue → Try sending immediately -> Process queue on each new block

- Enhance mempool rejection handling. Add  `PackageTooLarge`, `PackageTooManyTransactions` and `PackageMempoolLimits` variants. These variants should be queued to be tried at a later point after mempool state updates. 

---

#### Remarks:
- Queue is not persistent
- No re-queuing mechanism
- Remove one of the two usage of `itertools` dependency. Only remaining usage is `zip_eq` which can probably be replaced by an additional len check

## Linked Issues
- Fixes #2461 

## Testing
- Update `ChunkingPackageTooBigTest` and showcase how txs going above the mempool limit would be handle.
A chunked proof going above the limit in an already filled mempool would be queued and it's chunks be broadcasted up to the first error. The remaining chunks and aggregate will be processed on subsequent block.
A single proof going above the mempool policy limit would be chunked, queued and chunks will be broadcasted up to the first mempool policy error. The remaining chunks and aggregate will be sent to the next block


crates/bitcoin-da/src/test_utils.rs
```diff
@@ -1,154 +1,146 @@
...
-                    }
-                }
-            }
-            _ => {
-                return Err(BitcoinServiceError::InvalidTransaction(
-                    "Expected chunk zk proof".to_owned(),
-                ))
-            }
-        }
+        // let network = self.network;
```
kpp: Is there any need in this test?

crates/bitcoin-da/src/service.rs
```diff
@@ -338,322 +357,162 @@ impl BitcoinService {
...
-                            commit,
-                            reveal,
-                            self.tx_backup_dir.clone(),
-                        )
-                        .await
-                    }
-                    _ => panic!("ZKProof tx must be either complete or chunked"),
-                }
-            }
+        let prefix = self.reveal_tx_prefix.clone();
```
kpp, line 409: Move it down near the usage.

crates/bitcoin-da/src/service.rs
```diff
@@ -338,322 +357,162 @@ impl BitcoinService {
...
-        histogram!("da_transaction_size").record(raw_txs_size_sum as f64);
+    /// Send transaction out of the queue to DA until the first error.
+    /// Returns the successfully sent txs.
+    pub(crate) async fn process_transaction_queue(&self) -> Result<Vec<Txid>> {
+        let mut queue = self.tx_queue.lock().await;
 
-        let txids = self.send_raw_transactions(&raw_txs).await?;
+        let mut txids = Vec::new();
+        while let Some(tx) = queue.front() {
+            if self.test_mempool_accept(&tx.as_raw_txs()).await.is_err() {
```
kpp: So if we got an error in the middle of chunks, once sequencer receives another request (zkproof/seqcom), we will try to send it again and fail.

Instead we can do a queue [pack1, pack2, ...] where a pack is [bitcoin tx1, bitcoin tx2, ...]. And instead of picking the front pack, we pop the front pack. So at least we will be able to put seqcomms on DA if we stuck in the middle of chunk sending.

crates/bitcoin-da/src/service.rs
```diff
@@ -338,322 +357,162 @@ impl BitcoinService {
...
-                errors
-            )));
+        match &tx.kind {
+            TransactionKind::Complete
+            | TransactionKind::BatchProofMethodId
+            | TransactionKind::SequencerCommitment => {
+                info!("Blob inscribe tx sent. Hash: {}", tx.reveal_txid())
+            }
+            TransactionKind::Chunks => {
+                histogram!("da_transaction_size").record(raw_txs_size_sum);
```
kpp: I believe these should be gauge too. See https://github.com/chainwayxyz/citrea/pull/2383

cc @eyusufatik

crates/bitcoin-da/src/service.rs
```diff
@@ -338,322 +357,162 @@ impl BitcoinService {
...
     }
 
+    /// Test whether signed transactions should be accepted to the queue.
+    /// Any error recoverable by mempool state changes should be queued, such as package too large or package too many transactions.
+    /// When the mempool state changes, on every new block, the package limitations change accordingly.
+    /// The queued transactions will be retried on every block until the transaction is accepted to mempool.
+    async fn test_mempool_accept_queue_tx(&self, txs: &[SignedTxPair]) -> Result<()> {
+        let raw_txs: Vec<&Vec<u8>> = txs.iter().flat_map(|v| v.as_raw_txs()).collect();
+
+        match self.test_mempool_accept(raw_txs.as_slice()).await {
```
kpp: &raw_txs

kpp (approved):

crates/bitcoin-da/src/service.rs
```diff
@@ -265,6 +272,18 @@ impl BitcoinService {
...
+    /// Queue and try sending transaction to DA
+    pub async fn send_transaction_with_fee_rate(
+        &self,
+        tx_request: DaTxRequest,
+        fee_sat_per_vbyte: u64,
+    ) -> Result<Vec<Txid>> {
+        self.queue_da_transaction_with_fee_rate(tx_request, fee_sat_per_vbyte)
+            .await?;
+
+        self.process_transaction_queue().await
```
kpp: Also I don't like this one. What's the point of a queue if the queue is processed only after queue_da_transaction_with_fee_rate. So in theory it will be filled, read, and empty after send_transaction_with_fee_rate call is over. So... it's useless?

jfldde: It's not only processed here, it's also processed on every new block, see `run_da_queue` changes :
```
new_height_opt = new_block_rx.recv() => {
                    if let Some(new_height) = new_height_opt {
                        trace!("New da block height {new_height}. Processing transaction queue.");
                        if let Err(e) = self.process_transaction_queue().await {
                            error!(?e, "Error processing queue on new block");
                        }
                    }
                }
```

The idea is that we now face an issue where chunked proof cannot be sent as a single package since they are over mempool policy limit. We here send chunks up to the limit (process queue right after queuing), then send the remaining queued txs on new block trigger when and if mempool space is freed up (if our in mempool chunks are part of the last block)

crates/bitcoin-da/src/service.rs
jfldde: We already test for test_mempool_accept and only queue txs that passed validation or were rejected with `MempoolRejection::PackageTooLarge | MempoolRejection::PackageTooManyTransactions | MempoolRejection::PackageMempoolLimits`. These are mempool rejection that are dependent on mempool state and will evolve based on our txs being mined and freeing up our max vsize usage.

crates/bitcoin-da/src/service.rs
kpp: Ah, I see. Thanks

crates/bitcoin-da/src/service.rs
kpp: Thank you!

crates/bitcoin-da/src/test_utils.rs
jfldde: Restored

ercecan (approved):
LGTM

crates/bitcoin-da/src/helpers/backup.rs
```diff
@@ -0,0 +1,101 @@
...
+        TransactionKind::Chunks => "chunks",
+        TransactionKind::Aggregate => "aggregate",
+        TransactionKind::Unknown(_) => "unknown",
+    }
+}
+
+pub(crate) fn backup_txs_to_file(path: &Path, txs: &[SignedTxPair]) -> anyhow::Result<()> {
+    if let Some(tx) = txs.first() {
+        match &tx.kind {
+            TransactionKind::Complete
```
eyusufatik, line 24: an extra length check here to make sure we don't call this with > 2 txs

bin/citrea/tests/bitcoin/queue.rs
eyusufatik, line 1: rename file

crates/bitcoin-da/src/service.rs
```diff
@@ -267,6 +281,43 @@ impl BitcoinService {
         }
     }
 
+    /// Queue and try sending transaction to DA
+    pub async fn send_transaction_with_fee_rate(
+        &self,
+        tx_request: DaTxRequest,
+        fee_sat_per_vbyte: u64,
+    ) -> Result<Vec<[TxWithId; 2]>> {
+        // Prevent sending tx to DA while transaction queue is not empty
```
eyusufatik, line 290: add comment stating otherwise we'd try double spending some utxos etc.

crates/bitcoin-da/src/service.rs
```diff
@@ -340,337 +405,122 @@ impl BitcoinService {
...
-            if let Some(errors) = signed_raw_commit_tx.errors {
-                return Err(BitcoinServiceError::InvalidTransaction(format!(
-                    "Failed to sign commit transaction: {:?}",
-                    errors
-                )));
+        let mut txids = Vec::new();
+        while let Some(tx) = queue.front() {
+            if let Err(e) = self.test_mempool_accept(&tx.as_raw_txs()).await {
+                debug!(?e, "Rejected by mempool");
+                break;
```
ercecan: If a tx is rejected because it's min-relay-fee is not met, what happens? Does this halt here because we keep breaking out of the loop?

eyusufatik (approved):

PR #2506, opened 2025-06-30
Remove L2BlockByNumber table from light client prover

ercecan:
# Description
Fixes this todo:
https://github.com/chainwayxyz/citrea/blob/855c723e810975b73d3826e799b91518b41f44bd/crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs#L138

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


Cargo.toml
```diff
@@ -165,7 +165,7 @@ alloy-eips = { version = "0.13", default-features = false }
 alloy-consensus = { version = "0.13", default-features = false, features = ["serde", "serde-bincode-compat"] }
 alloy-network = { version = "0.13", default-features = false }
 
-citrea-e2e = { git = "https://github.com/chainwayxyz/citrea-e2e", rev = "4aad695" }
+citrea-e2e = { git = "https://github.com/chainwayxyz/citrea-e2e", rev = "70b1a75237c74e3def82cc06e2cd90086b236670" }
```
eyusufatik: use short form

crates/common/src/rpc/mod.rs
```diff
@@ -71,6 +75,49 @@ pub fn register_healthcheck_rpc<T: Send + Sync + 'static>(
...
+                })
+            }
+        })
+        .await;
+        match res {
+            Ok(_) => Ok::<(), ErrorObjectOwned>(()),
+            Err(e) => Err(error(&format!(
+                "Failed to retrieve head block header: {}",
+                e
+            ))),
```
eyusufatik, line 114: i wonder if we can somehow check on if the proving job is still working 🤔

crates/sovereign-sdk/module-system/sov-modules-rollup-blueprint/src/lib.rs
```diff
@@ -65,18 +65,26 @@ pub trait RollupBlueprint: Sized + Send + Sync {
...
+    fn register_ethereum_rpc(
+        &self,
+        da_service: Arc<Self::DaService>,
+        storage: ProverStorage,
+        ledger_db: LedgerDB,
+        methods: &mut jsonrpsee::RpcModule<()>,
+        sequencer_client_url: Option<String>,
+        l2_block_rx: Option<broadcast::Receiver<u64>>,
+    ) -> Result<(), anyhow::Error>;
+
```
eyusufatik: should we maybe take this out of the trait and then register the etheruem rpcs like we register the healthcheck endpoints?

eyusufatik: bc it does not make sense as a trait object

crates/common/src/rpc/mod.rs
ercecan: Well if there are no new blocks this rpc will keep failing

jfldde: Is there really value in this ? Why not using `ledger_getLastScannedL1Height` as health check endpoint directly ?

ercecan: > Is there really value in this ? Why not using ledger_getLastScannedL1Height as health check endpoint directly ?

I agree, @eyusufatik what do you say

eyusufatik: are we going to wait for that value to change?

maybe we should re-think about this

eyusufatik (approved):

jfldde (approved):

crates/common/src/rpc/mod.rs
ercecan: I will open an issue about this for now #2510

Issue #2507, opened 2025-07-01
Inconsistent implementation between NodeType, NodeKind and StorageNodeTypeArg


jfldde:
## Summary

- Citrea-cli :
```
#[derive(Clone, Debug, ValueEnum)]
#[value(rename_all = "kebab-case")]
enum NodeKind {
    BatchProver,
    Sequencer,
    FullNode,
    LightClientProver,
}

impl std::fmt::Display for NodeKind {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        match self {
            NodeKind::BatchProver => write!(f, "batch-prover"),
            NodeKind::Sequencer => write!(f, "sequencer"),
            NodeKind::FullNode => write!(f, "full-node"),
            NodeKind::LightClientProver => write!(f, "light-client-prover"),
        }
    }
}
```

citrea (fullnode vs full-node above)
```
pub(crate) enum NodeType {
    Sequencer(SequencerConfig),
    FullNode,
    BatchProver(BatchProverConfig),
    LightClientProver(LightClientProverConfig),
}

impl std::fmt::Display for NodeType {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        match self {
            NodeType::Sequencer(_) => {
                write!(f, "sequencer")
            }
            NodeType::FullNode => {
                write!(f, "fullnode")
            }
            NodeType::BatchProver(_) => {
                write!(f, "batch-prover")
            }
            NodeType::LightClientProver(_) => {
                write!(f, "light-client-prover")
            }
        }
    }
}
```

Internally used by rollback : (will use light-client vs light-client-prover for the two above)

```
#[derive(Copy, Clone, ValueEnum)]
pub enum StorageNodeTypeArg {
    Sequencer,
    FullNode,
    BatchProver,
    LightClient,
}

impl From<StorageNodeTypeArg> for StorageNodeType {
    fn from(value: StorageNodeTypeArg) -> Self {
        match value {
            StorageNodeTypeArg::Sequencer => StorageNodeType::Sequencer,
            StorageNodeTypeArg::FullNode => StorageNodeType::FullNode,
            StorageNodeTypeArg::BatchProver => StorageNodeType::BatchProver,
            StorageNodeTypeArg::LightClient => StorageNodeType::LightClient,
        }
    }
}

[...]

PR #2511, opened 2025-07-01
Use enum for node kind instead of string

ercecan:
# Description
`node_kind` in backup manager used to be String, now it is enum.

Fixes TODO:
https://github.com/chainwayxyz/citrea/blob/f91916d9bc0a96e200698af168410899f85cd7f6/crates/common/src/backup/manager.rs#L87

### DIFF
--- a/crates/storage-ops/src/types.rs
+++ b/crates/storage-ops/src/types.rs
@@ -7,3 +7,22 @@ pub enum StorageNodeType {
     BatchProver,
     LightClient,
 }
+
+#[derive(Copy, Clone, Debug)]
+pub enum NodeKind {
+    Sequencer,
+    FullNode,
+    BatchProver,
+    LightClientProver,
+}
+
+impl std::fmt::Display for NodeKind {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            NodeKind::BatchProver => write!(f, "batch-prover"),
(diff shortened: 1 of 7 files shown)


crates/storage-ops/src/types.rs
```diff
@@ -7,3 +7,22 @@ pub enum StorageNodeType {
...
+
+impl std::fmt::Display for NodeKind {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            NodeKind::BatchProver => write!(f, "batch-prover"),
+            NodeKind::Sequencer => write!(f, "sequencer"),
+            NodeKind::FullNode => write!(f, "full-node"),
+            NodeKind::LightClientProver => write!(f, "light-client-prover"),
+        }
+    }
```
jfldde, line 27: Take a look at https://github.com/chainwayxyz/citrea/issues/2507, maybe we can consolidate further

ercecan: maybe I can use NodeKindArg and NodeKind instead of StorageNodeTypeArg and StorageNodeType? or vice versa might be faster.

eyusufatik (approved):

crates/storage-ops/src/types.rs
ercecan: https://github.com/chainwayxyz/citrea/issues/2507#issuecomment-3023857295
added this comment, we can handle this on that issue

jfldde (approved):

PR #2518, opened 2025-07-02
Fix lcp rollback issue

### LINKED ISSUE #2440: LCP rollback issues
## Summary

LCP hits 
```
        let Some(l2_block_number) = ledger_db.get_head_l2_block_height()? else {
            return Ok(());
        };
```
On rollback command entry and exit early without being able to complete the rollback

ercecan:
# Description
Normally we pass `l2_height + 1` as db version for target of native and state db rollback, however in lcp we need to pass L1 height, also since L1 height does not start from 0 we pass the difference of `last_scanned_l1_height - target_l1_height` as the db version

## Linked Issues
- Fixes #2440


crates/storage-ops/src/rollback/mod.rs
```diff
@@ -64,18 +69,70 @@ impl Rollback {
...
+                    if last_scanned_l1_height == 0 {
+                        // If no slots have been scanned, we can rollback to the target version
+                        return Ok(());
+                    }
+
+                    // TODO: Should I check this?
+                    if last_scanned_l1_height <= state_db_last_version {
+                        return Ok(());
+                    }
+
```
eyusufatik: imo this should never happen. panic if this happens

eyusufatik (approved):

rakanalh (approved):

crates/storage-ops/src/rollback/mod.rs
```diff
@@ -64,18 +67,70 @@ impl Rollback {
...
+                    if last_scanned_l1_height <= state_db_last_version {
+                        unreachable!("Last scanned L1 height should not be less than or equal to the last state DB version");
+                    }
+
+                    // This is done because the state DB version will start from 0 but the L1 height starts from a random value
+                    // We take the difference with the last scanned l1 height and the target l1 height we want to rollback to
+                    // and subtract it from the latest state DB version to get the target state version.
```
jfldde, line 115: What happens if `initial_da_height` config changes between restart ? Might not be important since rolling back to `initial_da_height` could never be used and one could spawn a new lcp with this `initial_da_height` instead but we might want to pin the first value in case. 
Could we imagine rolling back to a target N then restart with a `initial_da_height` value of N + M to skip blocks ?

eyusufatik: that scenario only works on regtest though 

in devnet testnet or mainnet, to change the inital da height, you'd need a completely new zkvm elf

jfldde (approved):

PR #2523, opened 2025-07-03
Ignore eip-4844 transactions

rakanalh:
# Description

This PR makes sure that EIP-4844 are ignored. 

Even though EIP-4844 transactions are ignored through the RPC endpoint, this PR builds upon the check for EIP-4844 transactions in the EVM by not panicing but rather just ignores the transaction.

However, the caveat is, when an EIP-4844 transaction is ignored... the result of that transaction is HALT and the nonce for the user account is still incremented though nothing was executed.

### DIFF
--- a/crates/evm/src/evm/executor.rs
+++ b/crates/evm/src/evm/executor.rs
@@ -3,10 +3,10 @@ use alloy_eips::Typed2718;
 use alloy_primitives::{keccak256, U256};
 use alloy_sol_types::SolCall;
 use reth_primitives::{Recovered, TransactionSigned};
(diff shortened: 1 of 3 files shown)


eyusufatik:
just doing this will cause a lot of problems where the execute_multiple_tx function is called. 

please read code that calls this function carefully

```
  for (evm_tx_recovered, result) in users_txs.into_iter().zip(results.into_iter()) {
```

rakanalh:
This is not the complete impl... i was going to write a test to verify.

eyusufatik:
let's not fix this at this time. the relaxing of this rule in the future if needed would be backwards compatible anyways

PR #2527, opened 2025-07-03
Use NonEmptySlice in initial method ids

exeokan:
# Description

Implements `NonEmptySlice` to return from `InitialValueProvider::initial_batch_proof_method_ids`. By making const vars of the type `NonEmptySlice`, we can also enforce that the method ids are not empty at compile time.

### DIFF
--- a/crates/light-client-prover/src/circuit/initial_values.rs
+++ b/crates/light-client-prover/src/circuit/initial_values.rs
@@ -1,13 +1,12 @@
 #[cfg(feature = "native")]
 use bitcoin_da::spec::BitcoinSpec;
+use non_empty_slice::NonEmptySlice;
 #[cfg(feature = "native")]
 use sov_mock_da::MockDaSpec;
 #[cfg(feature = "native")]
 use sov_modules_api::DaSpec;
 #[cfg(feature = "native")]
 use sov_rollup_interface::Network;
-
-use non_empty_slice::NonEmptySlice;
 
 /// Genesis root for the Light Client Prover's Jellyfish Merkle Tree.
 pub(crate) const LCP_JMT_GENESIS_ROOT: [u8; 32] = match const_hex::const_decode_to_array(
@@ -70,8 +69,8 @@ pub mod mockda {
 
 /// Module containing initial values for the Bitcoin DA (Data Availability) specification.
 pub mod bitcoinda {
-    use super::non_empty_slice::NonEmptySlice;
     use super::decode_to_u32_array;
+    use super::non_empty_slice::NonEmptySlice;
 
     /// Genesis L2 root for the Bitcoin DA on Mainnet.
     pub const MAINNET_GENESIS_ROOT: [u8; 32] = match const_hex::const_decode_to_array(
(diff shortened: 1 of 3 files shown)


crates/light-client-prover/src/circuit/initial_values.rs
```diff
@@ -470,24 +471,42 @@ impl InitialValueProvider<BitcoinSpec> for Network {
...
-mod tests {
-    #[test]
-    fn test_non_empty_slice_check() {
-        let slice: &[(u64, [u32; 8])] = &[];
-        let result = std::panic::catch_unwind(|| super::non_empty_slice(slice));
-        assert!(result.is_err());
+/// Module for NonEmptySlice, so that it cannot be constructed like `NonEmptySlice(&[])`.
+pub mod non_empty_slice {
+    /// A wrapper around a slice to ensure that it is never empty.
+    pub struct NonEmptySlice<'a, T>(&'a [T]);
```
ercecan, line 477: Instead use #[non_exhaustive]

ercecan (changes requested):

crates/light-client-prover/src/circuit/initial_values.rs
ercecan: nvm does not work in the same module

ercecan (approved):
Maybe move NonEmptySlice to primitives LGTM

eyusufatik (changes requested):
I guess for the proposed fix you'd need to change the type of the circuit function input to NonEmptyVec otherwise devs can add a const of empty slice (or vec) still.

crates/light-client-prover/src/da_block_handler.rs
```diff
@@ -246,7 +246,7 @@ where
             l1_block.header().clone(),
             previous_lcp_output,
             self.network.get_l2_genesis_root(),
-            self.network.initial_batch_proof_method_ids(),
+            self.network.initial_batch_proof_method_ids().to_vec(),
```
kpp, line 249: I wonder if we really have to pass a Vec into this method. A slice would be better to avoid OOMs.

crates/light-client-prover/src/circuit/initial_values.rs
```diff
@@ -470,24 +471,42 @@ impl InitialValueProvider<BitcoinSpec> for Network {
...
+        pub const fn new(slice: &'a [T]) -> Self {
+            assert!(!slice.is_empty(), "Empty slice passed to NonEmptySlice");
+            Self(slice)
+        }
 
-        let slice: &[u32] = &[];
-        let result = std::panic::catch_unwind(|| super::non_empty_slice(slice));
-        assert!(result.is_err());
+        /// Returns the inner slice.
+        pub const fn inner(&self) -> &'a [T] {
```
kpp, line 486: Please use AsRef/Deref traits.

kpp (approved):

eyusufatik (approved):

eyusufatik:
as this was reviewed and accepted by the audit team I'll merge it as is

PR #2554, opened 2025-07-08
Verify chunk wtxid when aggregating

### LINKED ISSUE #2378: Full node retrieves chunks by txid but lcp queries locally with wtxid
In full node when we detect a chunked part of a proof we skip it and do nothing, when we see an aggregate proof (the data of chunks which are txids and corresponding wtxids) we start iterating over txids and query bitcoin rpc with txid to get the chunk transaction, however we never check if the wtxid matches to the corresponding txid from the aggregate.

In lcp what we do is we iterate over all relevant txs and if the data is a chunked part we store it in jmt with wtxid to chunk mapping. This is better because we can directly get the chunk data from our storage instead of asking from a network.

Currently one possible attack vector on full nodes is that some aggregate proof might have the list of txids in the following way:
`[(txid_x,wtxid_1), (txid_y,wtxid_2), (txid_z,wtxid_3)]`
The batch prover may send arbitrary txids that do not correspond to the wtxids of the tuples that they are in, and this makes full nodes unable to correctly identify the chunked parts from txids and skip that chunked batch proof. However because lcp tracks chunked parts by wtxid only, it will be able to construct the complete proof and process it correctly.

In order for full node to behave similarly it should also track chunked parts and store them by their wtxid and when an aggregate comes it should fetch them locally with wtxids rather than via rpc network with txids.

rakanalh:
# Description

## Linked Issues
- Fixes #2378 

## Notes 
- Hold merge until after 0.7.3 is released.


crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
```diff
@@ -985,6 +986,27 @@ impl NodeLedgerOps for LedgerDB {
...
+        Ok(())
+    }
+
+    #[instrument(level = "trace", skip(self), err)]
+    fn get_chunk(&self, wtxid: Wtxid) -> anyhow::Result<Option<Vec<u8>>> {
+        self.db.get::<ChunksByWtxid>(&wtxid)
+    }
+
+    #[instrument(level = "trace", skip(self), err)]
+    fn delete_chunk(&self, wtxid: Wtxid) -> anyhow::Result<()> {
```
kpp: Unused?

kpp (dismissed):

loopghost:
Hi, I was looking through older open PRs and wanted to ask about the current status of this one. Is `#2554` still expected to land, or has it been superseded elsewhere since `0.7.3`?

If it is still active, do you have a rough sense of whether it is expected to be merged or closed in the near term?

loopghost:
cc @rakanalh @kpp @eyusufatik @ldiego08 @jfldde in case any of you knows: since `v0.7.3` was already released on 2025-08-28, is there a reason this PR is still open rather than being merged or closed/superseded?

PR #2558, opened 2025-07-08
Build mdbook from docs/

kpp:
# Description

To build the book: `mdbook build --open`.

# How it looks 

image
 
image


eyusufatik:
should we also handle code documentation here?

ercecan (approved):

kpp:
> should we also handle code documentation here?

No, they are built different. `cargo doc` scans the .rs code instead of .md files. So `cargo doc` is for API docs, while `mdbook` is for a wise knowledge. In the future we can build both and put them under different /endpoints/ on citrea.xyz

eyusufatik (approved):

PR #2572, opened 2025-07-10
chore: Remove CommitmentMerkleRoots table from Ledger DB

### LINKED ISSUE #2556: Remove CommitmentMerkleRoots table from Ledger DB
`get_l2_range_by_commitment_merkle_root` method of ledger db is never called, it was used in the old batch proof processing function.

varun-doshi:
# Description
Removes:
- `set_l2_range_by_commitment_merkle_root`
- `get_l2_range_by_commitment_merkle_root`
- `CommitmentMerkleRoots`

## Linked Issues
- Fixes #2556

### DIFF
--- a/crates/fullnode/src/da_block_handler.rs
+++ b/crates/fullnode/src/da_block_handler.rs
@@ -475,11 +475,6 @@ where
             sequencer_commitment.clone(),
         )?;
 
-        self.ledger_db.set_l2_range_by_commitment_merkle_root(
-            sequencer_commitment.merkle_root,
-            (L2BlockNumber(start_l2_height), L2BlockNumber(end_l2_height)),
-        )?;
-
         self.ledger_db
             .put_commitment_by_index(sequencer_commitment)?;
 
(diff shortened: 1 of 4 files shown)


eyusufatik:
I think just outright removing tables break nodes when they upgrade

can you test that? start a full node with v0.7.2, then use the same db with the your branch's version

eyusufatik:
see this pr for a migration https://github.com/chainwayxyz/citrea/pull/2582

varun-doshi:
> see this pr for a migration #2582

Looks like #2582 is running `CommitmentMerkleRoots` in the migration to drop tables. Which means I can revert the removal of the table in this PR, and only remove the 2 methods.
Works?

kpp (approved):

eyusufatik:
> > see this pr for a migration #2582
> 
> Looks like #2582 is running `CommitmentMerkleRoots` in the migration to drop tables. Which means I can revert the removal of the table in this PR, and only remove the 2 methods. Works?

that only does it for the sequencer though, need migrations for full node, batch prover, and light client prover too

eyusufatik:
will you do the above comment?

PR #2574, opened 2025-07-10
Track LCP status in lcp health RPC

### LINKED ISSUE #2510: Better health check for light client prover
Right now light client prover health check will only check if it's rpc is available and whether or not if it can access to bitcoin, there is no check if the prover service is still working.
We should implement a better health check that will also check prover service

kpp:
# Description

Set lcp_finished=true when lcp stops. Maybe you guys have a better idea?

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/2510

### DIFF
--- a/crates/common/src/rpc/mod.rs
+++ b/crates/common/src/rpc/mod.rs
@@ -1,4 +1,5 @@
 //! Common RPC crate provides helper methods that are needed in rpc servers
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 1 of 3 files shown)


crates/common/src/rpc/mod.rs
```diff
@@ -91,19 +94,20 @@ pub fn register_healthcheck_rpc_light_client_prover<T: Send + Sync + 'static, Da
...
-                da_service.get_head_block_header().await.map_err(|e| {
-                    let e = e;
-                    backoff::Error::transient(e)
-                })
-            }
+        let res = retry_backoff(exponential_backoff.clone(), async || {
+            da_service.get_head_block_header().await.map_err(|e| {
+                let e = e;
+                backoff::Error::transient(e)
+            })
```
kpp, line 110: That's just a refactoring.

rakanalh (commented):
What's the reason we're doing this?

ercecan:
> What's the reason we're doing this?

Before this light client provers health check was useless, we need a new health check logic for light client prover node

rakanalh (approved):
LGTM

ercecan (approved):

PR #2576, opened 2025-07-11
New metrics

ercecan:
# Description

Screenshot 2025-07-24 at 10 20 55
Screenshot 2025-07-24 at 10 21 09
Screenshot 2025-07-24 at 10 21 16

List of metrics:
### General

- RPC response times ✔️
- cache hit / miss on db accesses? (is this possible) ✔️
- gas usage ✔️
- bitcoin da queue processing times ✔️
- current bitcoinda queue size ✔️

### Sequencer 

- dry run txs & the actual block times ✔️
    - maybe per tx time could be logged in dry run as debug! or could be saved in a prometheus histogram ✔️
- sequencer commitment service loop times ✔️
- bitcoin da queue processing times ✔️
- More metrics ⇒ db save block time, more mempool metrics etc. stf apply block time?  ✔️
- Deposit tx count, tx size, deposit tx eth_call duration, unaccepted req count, dep tx rate over a minute ✔️ 

### Full node

- block processing time (already exists) ✔️
- block size?? ✔️
- L1 processing time per block (already exists) ✔️
- sequencer commitment processing time ✔️
- batch proof processing time ✔️
- eth_call & eth_estimateGas method running times (they are covered in rpc response times already and can be filtered with method name) ✔️
- committed and proven heights ✔️

### Batch prover

- input preparation time ✔️
    - total input generation time ✔️
    - cumulative witness generation time ✔️
- State and offchain log cache size in witness generation ✔️
- proving time ✔️
- ongoing proof job count ✔️ 
    - proving ✔️ 
    - in queue etc. ✔️ 

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


crates/evm/src/metrics.rs
```diff
@@ -0,0 +1,26 @@
...
+/// It is unused because we directly use gauge and histogram macros since that is the
+/// only way in which we can provide additional labels to the metric.
+/// However, deriving `Metrics` here is convenient to provide descriptions for each of
+/// the metrics.
+#[allow(unused)]
+#[derive(Metrics)]
+#[metrics(scope = "evm")]
+pub struct EvmMetrics {
+    #[metric(describe = "Current Block Gas Usage")]
+    pub(crate) gas_usage: Gauge,
```
ercecan: maybe histogram here

crates/sequencer/src/runner.rs
```diff
@@ -559,6 +567,11 @@ where
         );
 
         let state_diff = self.save_l2_block(l2_block, l2_block_result, tx_hashes, blobs)?;
+        histogram!(
```
eyusufatik: include maintain mempool also

crates/evm/src/hooks.rs
```diff
@@ -252,6 +254,16 @@ impl<C: sov_modules_api::Context> Evm<C> {
             let sealed_block = block.seal();
 
             self.blocks.push(&sealed_block, accessory_working_set);
+
+            #[cfg(feature = "native")]
+            {
+                // Update the metrics with the new block count
```
eyusufatik: maybe do this in the finalize hook?

crates/sequencer/src/commitment/service.rs
```diff
@@ -168,6 +172,18 @@ where
                                 // continue functioning correctly. We just need to resubmit the failed commitment to DA.
                                 error!("Failed to submit commitment: {:?}", e);
                             }
+                            histogram!(
+                                "sequencer_commitment_entire_process_time",
+                                "index" => index.to_string(),
```
eyusufatik: nice usage

crates/evm/src/hooks.rs
ercecan: this is the finalize hook?

ercecan (commented):
add l1 fee rate metric

ercecan (commented):
initialize full node metrics like highest commtted index l2 height etc.

ercecan (commented):
check cache hit/miss metrics

ercecan (commented):
this should be gauge:
pub highest_proven_l2_height: Counter,

Use std lazy

bin/citrea/src/main.rs
```diff
@@ -255,6 +255,8 @@ where
 
     let task_executor = task_manager.executor();
 
+    let node_type_str = node_type.to_string();
```
jfldde: Can pass node_type down itself

crates/bitcoin-da/src/helpers/builders/body_builders.rs
```diff
@@ -278,7 +278,7 @@ pub fn create_inscription_type_0(
                     commit_tx_address
                 );
 
-                histogram!("mine_da_transaction").record(
+                histogram!("type_0_mine_da_transaction").record(
```
jfldde: Can we use something clearer such as Complete instead of type_0 ? Same below

crates/bitcoin-da/src/service.rs
```diff
@@ -436,6 +443,7 @@ impl BitcoinService {
 
     async fn queue_transactions(&self, txs: Vec<SignedTxPair>) {
         self.tx_queue.lock().await.extend(txs);
+        gauge!("bitcoin_da_transaction_queue_size").set(self.tx_queue.lock().await.len() as f64);
```
jfldde: increment(txs.len()) maybe ?

crates/sequencer/src/commitment/service.rs
```diff
@@ -163,11 +165,23 @@ where
...
+                                "l2_start_height" => commitment_range.start().0.to_string(),
+                                "l2_end_height" => commitment_range.end().0.to_string()
+                            ).record(
+                                std::time::Instant::now()
+                                    .saturating_duration_since(start_commitment_processing)
+                                    .as_secs_f64(),
+                            );
+                            // Reset the start time for the next commitment processing
```
jfldde: Can we have helpers that handle these long calls and reduce the clutter as much as possible ?

bin/citrea/src/main.rs
ercecan: I will probably revert this instead as we will be able to query by node type as @Chinacolt said

crates/batch-prover/src/metrics.rs
```diff
@@ -26,7 +26,27 @@ pub struct BatchProverMetrics {
 
     /// Histogram tracking the time taken to scan and process L1 blocks
     #[metric(describe = "The duration of scanning and processing a single L1 block")]
-    pub scan_l1_block: Histogram,
+    pub scan_l1_block: Gauge,
```
rakanalh: Are we sure this is correct? i feel like anything measuring time should be a histogram... which is also what the comment above it says

crates/bitcoin-da/src/service.rs
```diff
@@ -324,6 +325,12 @@ impl BitcoinService {
         // Process transaction queue.
         self.process_transaction_queue().await?;
 
+        histogram!("bitcoin_da_transaction_queue_processing_time").record(
```
rakanalh: We don't seem to be defining any labels here... which means that we can use `BITCOIN_DA_METRICS`

crates/bitcoin-da/src/service.rs
```diff
@@ -435,7 +442,9 @@ impl BitcoinService {
     }
 
     async fn queue_transactions(&self, txs: Vec<SignedTxPair>) {
+        let txs_len = txs.len();
         self.tx_queue.lock().await.extend(txs);
+        gauge!("bitcoin_da_transaction_queue_size").increment(txs_len as f64);
```
rakanalh: same as above

crates/bitcoin-da/src/service.rs
```diff
@@ -458,6 +467,7 @@ impl BitcoinService {
             match self.send_signed_transaction(tx).await {
                 Ok(ids) => {
                     queue.pop_front();
+                    gauge!("bitcoin_da_transaction_queue_size").decrement(1);
```
rakanalh: same as above

crates/evm/src/hooks.rs
```diff
@@ -252,6 +254,16 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
 
             self.blocks.push(&sealed_block, accessory_working_set);
+
+            #[cfg(feature = "native")]
+            {
+                let base_fee_gwei = sealed_block.header.base_fee_per_gas.unwrap_or_default() as f64
+                    / 1_000_000_000.0; // Convert to Gwei
+                                       // Update the metrics with the new block count
+                gauge!("evm_gas_usage").set(sealed_block.header.gas_used as f64);
+                gauge!("evm_base_fee_per_gas").set(base_fee_gwei);
```
rakanalh: Same... am i missing something here?

crates/fullnode/src/metrics.rs
```diff
@@ -22,11 +22,35 @@ pub struct FullnodeMetrics {
 
     /// Histogram tracking the time taken to scan and process L1 blocks
     #[metric(describe = "The duration of scanning and processing a single L1 block")]
-    pub scan_l1_block: Histogram,
+    pub scan_l1_block: Gauge,
```
rakanalh: Same comment as in the batch prover

crates/light-client-prover/src/metrics.rs
```diff
@@ -16,7 +16,10 @@ pub struct LightClientProverMetrics {
     pub current_l1_block: Gauge,
     #[metric(describe = "The duration of scanning and processing a single L1 block")]
     /// The duration of scanning and processing a single L1 block
-    pub scan_l1_block: Histogram,
+    pub scan_l1_block: Gauge,
+    /// Tracking the time taken to prove a state transition, gauge because one proof is generated per l1 block
+    #[metric(describe = "The duration of generating a light client proof")]
+    pub proving_time: Gauge,
```
rakanalh, line 22: I believe these should be histograms

crates/sequencer/src/commitment/service.rs
```diff
@@ -22,7 +22,7 @@ use tokio::sync::{mpsc, oneshot};
 use tracing::{debug, error, info, instrument, warn};
 
 use super::controller::CommitmentController;
-use crate::metrics::SEQUENCER_METRICS;
+use crate::metrics::SEQUENCER_METRICS as sm;
```
rakanalh:
```suggestion
use crate::metrics::SEQUENCER_METRICS as SM;
```
I am not sure if clippy is going to whine about this but since we're importing a constant / static, we should capitalize it.

crates/sequencer/src/runner.rs
```diff
@@ -231,6 +233,12 @@ where
                     da_blocks,
                     &mut nonce,
                 )?;
+            let dry_run_system_txs_duration = Instant::now()
+                .saturating_duration_since(start_dry_run_system_txs)
+                .as_secs_f64();
+            sm.dry_run_system_txs_time
+                .record(dry_run_system_txs_duration);
+            gauge!("sequencer_dry_run_system_txs_time_gauge").set(dry_run_system_txs_duration);
```
rakanalh: What does this mean when you're setting the value via record and then using the gauge?

crates/sequencer/src/runner.rs
```diff
@@ -550,13 +552,18 @@ where
...
         // TODO: cleanup l2 block structure once we decide how to pull data from the running sequencer in the existing form
-        let l2_block = L2Block::new(signed_header, txs);
+        let l2_block = L2Block::new(signed_header, signed_txs);
 
         info!(
             "New block #{}, Tx count: #{}",
             l2_block.height(),
             evm_txs_count
         );
+        gauge!("sequencer_sign_l2_block_header_time_gauge").set(
```
rakanalh: How come instances like this are not using sm?

crates/batch-prover/src/metrics.rs
ercecan: We are now using it to get exact amount of time it takes to scan an L1 block, so in grafana I can put L1 block height on the X axis and the time it takes to scan that block in the Y axis

However, when syncing really fast for example on startup, there will be data losses if the l1 block or any other gauge that I use is updated more than once before the scrape interval of prometheus. So, I believe I can either keep both histogram and gauge for these kinds of stuff so we can see both the average and the exact values, or pick one of histogram or gauge

cc: @eyusufatik @jfldde

eyusufatik: yeah both histogram and gauge seems like a better deal

crates/evm/src/hooks.rs
ercecan: Nope, I missed those, I will use the structure for them as well

crates/batch-prover/src/metrics.rs
rakanalh: Would something like this work?
```
metrics::gauge!(
    "scan_l1_block_time",
    scan_duration.as_secs_f64(),
    "block" => block_number
);
```
where the block label is used as the X-axis... this would give us exacts even if we are syncing.

I haven't tried it though.

ercecan: I have tried that at first but I could not use label as x axis

ercecan: I will give it another shot, I just saw label to fields transformation let me try that real quick

ercecan: nope pretty sure I can't do that

crates/evm/src/hooks.rs
```diff
@@ -252,6 +254,16 @@ impl<C: sov_modules_api::Context> Evm<C> {
             let sealed_block = block.seal();
 
             self.blocks.push(&sealed_block, accessory_working_set);
+
+            #[cfg(feature = "native")]
+            {
```
eyusufatik: already in native, does not need the feature gating

eyusufatik (changes requested):

eyusufatik (approved):

eyusufatik (approved):

PR #2571, opened 2025-07-10
Pricing service for Boundless

yaziciahmet:
# Description

PricingService client to fetch price information to use in Boundless proof requests.

## Linked Issues
- Fixes #2397 

## Testing
Describe how these changes were tested. If you've added new features, have you added unit tests?

## Docs
Describe where this code is documented. If it changes a documented interface, have the docs been updated?

PR #2578, opened 2025-07-11
Boundless: Adjust offer params on expiry

### LINKED ISSUE #2428: In boundless bump price if request is expired
In boundless only bump price if the request is expired, not if some prover failed to deliver on time
Also if prover failed to deliver on time only bump lock timeout

exeokan:
# Description
- Check the error from `handle_session`, do not resubmit on network errors
- If the request is expired, check if it is locked to see if it was taken by a prover
- If taken, only bump lock timeout, else, bump price. 

## Linked Issues
- Fixes #2428

### DIFF
--- a/crates/risc0/src/host/boundless.rs
+++ b/crates/risc0/src/host/boundless.rs
@@ -160,14 +160,7 @@ impl BoundlessProver {
 
         // Start boundless proving session
         let (req_id, request_expiry) = self
-            .send_request(
-                request,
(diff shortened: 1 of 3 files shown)


ercecan (approved):
Approved but let's wait on @yaziciahmet's pr first #2571  because when bumping price, we will need the price to be never more than 5usd per billion cycles

ercecan (approved):

PR #2581, opened 2025-07-14
Do not check for capacity in hashmap before try reserve

ercecan:
# Description
try_reserve() does not reserve more memory if the capacity is already sufficient. It checks for this internally, so no need to check for the capacity of the hashmap.

### DIFF
--- a/crates/short-header-proof-provider/src/native.rs
+++ b/crates/short-header-proof-provider/src/native.rs
@@ -60,11 +60,11 @@ impl<Da: DaSpec> ShortHeaderProofProvider for NativeShortHeaderProofProviderServ
 
                 if return_cond {
                     let mut queried_hashes_map = self.queried_and_verified_hashes.lock();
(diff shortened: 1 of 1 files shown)


kpp (commented):
Well, yes, but this makes code highly dependent on the structure (hashbrown::HashMap). If we switch to another one, this will grow the capacity every time even for already existing hashes.

On the other hand, if we can assume that the lenght of queried_hashes_map is < 10, then we don't care and we can do something like Vec with a linear search instead of HashMap.

ercecan:
> Well, yes, but this makes code highly dependent on the structure (hashbrown::HashMap). If we switch to another one, this will grow the capacity every time even for already existing hashes.

Do you reckon I should close the pr so this works even if we used another hashmap implementation? extra one if check here is not the biggest issue anyways

rakanalh (approved):

eyusufatik (approved):

PR #2587, opened 2025-07-16
upgrade risc0 to 2.3.0

eyusufatik:
# Description
Upgrades to risc0 v2.3.0

In this upgrade "dev mode" (fake receipt verification) is enabled programatically in the runtime by setting a field in the Receipt::verify_with_context functions. 

We must make sure we are not calling the regular `Receipt::verify` function as it will default to old the environment variable method.

### DIFF
--- a/crates/light-client-prover/src/circuit/mod.rs
+++ b/crates/light-client-prover/src/circuit/mod.rs
@@ -8,6 +8,7 @@ use accessors::{
     VerifiedStateTransitionForSequencerCommitmentIndexAccessor,
 };
 use borsh::BorshDeserialize;
+use citrea_primitives::network_to_dev_mode;
 use initial_values::LCP_JMT_GENESIS_ROOT;
 use sov_modules_api::da::BlockHeaderTrait;
 use sov_modules_api::{BlobReaderTrait, DaSpec, WorkingSet, Zkvm};
@@ -243,6 +244,7 @@ impl<S: Storage, DS: DaSpec, Z: Zkvm> LightClientProofCircuit<S, DS, Z> {
         proof: &[u8],
         last_l2_height: u64,
         last_sequencer_commitment_index: u32,
+        network: Network,
         working_set: &mut WorkingSet<S>,
     ) -> Result<(), CircuitError> {
         let Ok(journal) = Z::extract_raw_output(proof) else {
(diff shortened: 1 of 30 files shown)


jfldde (approved):
LGTM

crates/risc0/src/host/local.rs
```diff
@@ -120,11 +117,6 @@ impl LocalProver {
             }
         }
 
-        if self.dev_mode {
```
exeokan, line 123: looks like we can remove `dev_mode` from `LocalProver`

exeokan (changes requested):

crates/risc0/src/host/local.rs
exeokan: nvm the check is still relevant

eyusufatik: I like this sanity check right here:

Screenshot 2025-07-16 at 15 30 32

exeokan (approved):
LGTM. We should also update docs/run_testnet.md and docker/docker-compose.yml and remove RISC0_DEV_MODE for full node

crates/light-client-prover/src/circuit/mod.rs
```diff
@@ -243,6 +244,7 @@ impl<S: Storage, DS: DaSpec, Z: Zkvm> LightClientProofCircuit<S, DS, Z> {
         proof: &[u8],
         last_l2_height: u64,
         last_sequencer_commitment_index: u32,
+        network: Network,
```
ercecan, line 248: update function arguments comments

crates/light-client-prover/src/circuit/mod.rs
```diff
@@ -363,6 +370,7 @@ impl<S: Storage, DS: DaSpec, Z: Zkvm> LightClientProofCircuit<S, DS, Z> {
     /// * `RunL1BlockResult` - The result of running the L1 block, contains updates to the L2 state and the light client's JMT state.
     pub fn run_l1_block(
         &self,
+        network: Network,
```
ercecan, line 375: update function arguments comments

crates/primitives/src/zk_dev_mode.rs
```diff
@@ -0,0 +1,17 @@
+use sov_rollup_interface::Network;
+
+/// Allow dev mode usage for proof verification, only in testing or dev networks.
+/// So we return false for mainnet and true for all other networks.
+pub fn network_to_dev_mode(network: Network) -> bool {
```
ercecan, line 5: I prefer
```suggestion
pub fn is_risc0_dev_mode_by_network(network: Network) -> bool {
```

ercecan (changes requested):

crates/primitives/src/zk_dev_mode.rs
eyusufatik: I went for this as we can support multiple zkvms

ercecan (approved):

PR #2610, opened 2025-07-28
Fix lagged blocks recovery

### LINKED ISSUE #2612: Fix citrea subscribe lag recovery

ercecan:
# Description
I believe that in the first implementation If we had lagged responses twice in a row we would skip those lagged blocks because we return an empty vector, with this implementation there is an atomic variable for each subscriber that tracks the last successfully sent block and acts accordingly in lagged responses

## Linked Issues
- Fixes #2612 
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/sequencer/src/rpc.rs
+++ b/crates/sequencer/src/rpc.rs
@@ -402,9 +402,9 @@ async fn handle_l2_block_subscription(
     ledger: LedgerDB,
 ) {
     let head_block_num = ledger.get_head_l2_block_height().unwrap_or(0);
-    let last_sent_block = Arc::new(AtomicU64::new(head_block_num));
+    let last_sent_block = AtomicU64::new(head_block_num);
     loop {
-        match receive_next_blocks(rx, last_sent_block.clone()).await {
+        match receive_next_blocks(rx, &last_sent_block).await {
             BlockReceiveResult::HighestBlock(highest_block_height) => {
                 let last_sent_block_num = last_sent_block.load(Ordering::SeqCst);
                 for block_height in last_sent_block_num + 1..=highest_block_height {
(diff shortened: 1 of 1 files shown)


crates/sequencer/src/rpc.rs
```diff
@@ -400,13 +401,17 @@ async fn handle_l2_block_subscription(
...
-            BlockReceiveResult::Blocks(blocks) => {
-                for block_height in blocks {
+        match receive_next_blocks(rx, last_sent_block.clone()).await {
+            BlockReceiveResult::HighestBlock(highest_block_height) => {
+                let last_sent_block_num = last_sent_block.load(Ordering::SeqCst);
+                for block_height in last_sent_block_num + 1..=highest_block_height {
                     if !send_block_notification(&subscription, block_height, &ledger).await {
                         return;
                     }
+                    last_sent_block.store(block_height, Ordering::SeqCst);
```
jfldde: Can store `highest_block_height` once

crates/sequencer/src/rpc.rs
```diff
@@ -400,13 +401,17 @@ async fn handle_l2_block_subscription(
     rx: &mut tokio::sync::broadcast::Receiver<u64>,
     ledger: LedgerDB,
 ) {
+    let head_block_num = ledger.get_head_l2_block_height().unwrap_or(0);
+    let last_sent_block = Arc::new(AtomicU64::new(head_block_num));
```
jfldde: Don't need the Arc

jfldde (approved):

crates/sequencer/src/rpc.rs
ercecan: AtomicU64 does not have the clone so not sure how to handle it in the loop

PR #2618, opened 2025-07-29
Rollback prover jobs with commitments

yaziciahmet:
# Description

We observed that Testnet Prover was stuck in recovery for long time even though we rolled back the state. We were sending proofs that we couldn't query from RPC. It seems that `PendingL1SubmissionJobs` and `ProofByJobId` tables are not rolled back and caused Prover to go into recovery even though it shouldn't.

There was also a bug in rollback of commitments, where if a job contains the commitment index whole job is deleted. But this was problematic for the case where the job is for indices [9, 10, 11] but the rollback is for index 10, we also delete the job that proves 9 and 10 even though we shouldn't. This PR also fixes that

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/crates/storage-ops/src/rollback/node/batch_prover.rs
+++ b/crates/storage-ops/src/rollback/node/batch_prover.rs
@@ -4,8 +4,8 @@ use std::sync::Arc;
 use sov_db::schema::tables::{
     CommitmentIndicesByJobId, CommitmentIndicesByL1, JobIdOfCommitment, L2BlockByHash,
     L2BlockByNumber, PendingL1SubmissionJobs, ProofByJobId, ProverLastScannedSlot,
-    ProverPendingCommitments, ProverStateDiffs, SequencerCommitmentByIndex, ShortHeaderProofBySlotHash,
-    SlotByHash,
+    ProverPendingCommitments, ProverStateDiffs, SequencerCommitmentByIndex,
+    ShortHeaderProofBySlotHash, SlotByHash,
 };
 use sov_db::schema::types::{L2BlockNumber, SlotNumber};
 use sov_schema_db::{ScanDirection, SchemaBatch, DB};
--- a/crates/storage-ops/src/rollback/node/batch_prover.rs
+++ b/crates/storage-ops/src/rollback/node/batch_prover.rs
@@ -3,8 +3,9 @@ use std::sync::Arc;
 
 use sov_db::schema::tables::{
     CommitmentIndicesByJobId, CommitmentIndicesByL1, JobIdOfCommitment, L2BlockByHash,
-    L2BlockByNumber, ProverLastScannedSlot, ProverPendingCommitments, ProverStateDiffs,
-    SequencerCommitmentByIndex, ShortHeaderProofBySlotHash, SlotByHash,
+    L2BlockByNumber, PendingL1SubmissionJobs, ProofByJobId, ProverLastScannedSlot,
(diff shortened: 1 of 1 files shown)


crates/storage-ops/src/rollback/node/batch_prover.rs
```diff
@@ -83,26 +86,58 @@ impl BatchProverLedgerRollback {
...
+                jobs_to_check.push((job_id, commitment_indices));
             }
         }
 
+        // Process collected jobs
+        for (job_id, commitment_indices) in jobs_to_check {
+            // If ANY index should be preserved, skip the whole job
+            if commitment_indices
+                .iter()
+                .any(|&idx| idx <= last_sequencer_commitment_index)
```
rakanalh: I think this whole loop can be removed. See below comment

crates/storage-ops/src/rollback/node/batch_prover.rs
```diff
@@ -83,26 +86,58 @@ impl BatchProverLedgerRollback {
...
+
+        // Collect jobs that have any indices above our rollback point
+        for job_record in jobs_iter {
+            let job_record = job_record?;
+            let job_id = job_record.key;
+            let commitment_indices = job_record.value;
+
+            if commitment_indices
+                .iter()
+                .any(|&idx| idx > last_sequencer_commitment_index)
```
rakanalh: If we replace any with all to make sure that all indices belonging to a job are larger than the last sequencer commitment index.

crates/storage-ops/src/rollback/node/batch_prover.rs
```diff
@@ -104,27 +103,21 @@ impl BatchProverLedgerRollback {
...
-                .iter()
-                .any(|&idx| idx <= last_sequencer_commitment_index)
-            {
-                tracing::warn!(
-                    "Preserving job {} that spans rollback boundary. Job indices: {:?}, rollback target: {}",
-                    job_id,
-                    commitment_indices,
-                    last_sequencer_commitment_index
-                );
+                if commitment_indices
```
rakanalh: I don't think this is necessary. and you're not pushing to `jobs_to_check` any more.

My suggestion:
```rust
if commitment_indices.iter()
    .all(|&idx| idx > last_sequencer_commitment_index) { 
        jobs_to_check.push((job_id, commitment_indices));
}
                 
```

rakanalh: Scratch this... now i understand what you did.

rakanalh: I don't think this if statement is necessary. The outer one will do the job for the tracing warn

yaziciahmet: my reasoning was that, if the rollback index is 10, but a job has 1,2,3 indices, then we don't print it as its completely out of the bounds. but if a job has 9,10,11, only then we print this as its transgressing on both sides. wdyt?

rakanalh: IMO, if NOT all commitment indices are above the target rollback index, then we print the trace warn. So the all does exactly what you described. We will ignore any job which has 1,2,3 and 9

yaziciahmet: Done https://github.com/chainwayxyz/citrea/pull/2618/commits/2b2e53ec28d14ef0c55f929833ed0753315db2d9

rakanalh (approved):

eyusufatik (approved):

PR #2620, opened 2025-07-30
Cycle counts & state diff workflow

exeokan:
# Description

This PR introduces a new github action: comparing cycle counts and state diff with the nightly.
In order to use the same circuit input, we use frozen sequencer DB and bitcoin data directory.

**Freezing the sequencer DB and bitcoin data:**
Wrote the `generate_proving_stats_db` e2e test (added ignore tagged so it doesn't run with cargo test)
that reads test transactions from file, generates two commitments and mines the commitments.

**Action flow**
The action needs docker-setup and citrea build.

1. Starts bitcoin-regtest container, loads wallets.
2. Starts sequencer and batch prover.
3. `proving-stats/get-proving-stats.py` waits for proofs, extracts cycle counts from batch prover's stdout, and gets state diff size via RPC, and saves results to file. (batch prover produces two proofs, we sum the cycle counts and state diff sizes)
4. Restore nightly stats from cache.
5. If cache miss, checkout to nightly, then repeat the same steps above for nightly and cache the result.
6. Compare results with `proving-stats/compare-results.py` and report the results as a comment to the PR.

TODO:

- [x] Convert base branch to nightly before merging


bin/citrea/tests/proving-stats/dbs/sequencer-db/ledger/000009.log
kpp, line 1: Do we need to keep those files like logs, locks, states?

.github/scripts/run-test.sh
```diff
@@ -0,0 +1,22 @@
...
+
+mkdir -p $TEST_DIR/results
+python3 bin/citrea/tests/proving-stats/get-proving-stats.py batch-prover.log $TEST_DIR/results/$OUT_FILE_NAME
+
+pkill citrea
+docker compose -f $TEST_DIR/docker-compose.regtest.yml down
+
+sleep 2 # Give some time for the processes to terminate
+
+git reset --hard
```
kpp: Wow, that's tough

bin/citrea/tests/proving-stats/dbs/sequencer-db/ledger/000009.log
exeokan: I am not very familiar with the rocksdb files, let me see what I can remove after I get the action working

.github/scripts/run-test.sh
exeokan: I think I will remove this since actions/checkout does this by default

bin/citrea/tests/proving-stats/.gitignore
```diff
@@ -0,0 +1,3 @@
+.venv
+logs/
+results/
```
jfldde:
```suggestion
.venv
logs/
results/
bitcoin/
```

exeokan: what would be the benefit🤔

bin/citrea/tests/bitcoin/guest_cycles.rs
```diff
@@ -194,3 +201,126 @@ async fn guest_cycles() {
 
     println!("Execution stats: {:?}", stats);
 }
+
+/// This test generates a proving stats database by running transactions through the sequencer
+/// and ensuring two commitments are published to Bitcoin before freezing the state.
+/// We use the sequencer DB and bitcoin data dir in the proving stats workflow.
+///
+/// Unfortunately, before running the test, we have to remove citrea-primitives from dev-dependencies
```
exeokan: I don't like this solution, but couldn't come up with a better one

.github/scripts/run-test.sh
exeokan: actions/checkout does git clean -ffdx and cleans the ignored files as well, which is no good here. I am keeping this.

.github/workflows/checks.yml
```diff
@@ -509,4 +511,128 @@ jobs:
...
+      - name: Get nightly commit hash
+        id: nightly-commit
+        run: echo "nightly-hash=$(git rev-parse HEAD)" >> $GITHUB_OUTPUT
+
+      - name: Restore nightly test cache
+        id: restore-nightly
+        uses: actions/cache@v4
+        with:
+          path: bin/citrea/tests/proving-stats/results/nightly-stats.json
+          key: nightly-cycle-counts-${{ steps.nightly-commit.outputs.nightly-hash }}
```
jfldde: This could be higher up

.github/workflows/checks.yml
```diff
@@ -509,4 +511,128 @@ jobs:
...
+          ./.github/scripts/run-proving-test.sh patch-stats.json
+        shell: bash
+
+      - name: Get current commit hash
+        id: current-commit
+        run: |
+          echo "patch-hash=${{ github.event.pull_request.head.sha }}" >> $GITHUB_OUTPUT
+
+      - name: Checkout nightly branch
+        run: git fetch origin ege/mock-nightly && git checkout ege/mock-nightly
```
jfldde:
```suggestion
        run: git fetch origin nightly && git checkout nightly
```

bin/citrea/tests/bitcoin/guest_cycles.rs
jfldde: Running with docker image citrea-dev (which runs without testing features)

exeokan: nice, updating the comment

.github/workflows/checks.yml
exeokan: before checking out to nightly?

bin/citrea/tests/bitcoin/guest_cycles.rs
```diff
@@ -194,3 +201,127 @@ async fn guest_cycles() {
...
     println!("Execution stats: {:?}", stats);
 }
+
+/// This test generates a proving stats database by running transactions through the sequencer
+/// and ensuring two commitments are published to Bitcoin before freezing the state.
+/// We use the sequencer DB and bitcoin data dir in the proving stats workflow.
+///
+/// As we use the citrea artifact built without the testing feature in CI,
+/// this test should be run without the testing feature as well.
+/// This can be achieved by running with docker image citrea-dev
```
jfldde, line 211: Where is that done ? 
I'm not sure it matters anyway since testing feature shouldn't impact cycle count

jfldde (approved):

bin/citrea/tests/bitcoin/guest_cycles.rs
exeokan: The reason is that in proving-stats job, we make use of the citrea build artifacts that are built without the testing feature. If we run the `generate_proving_stats_db` regularly, the sequencer commitments will have the shorter reveal tx prefix, and they wont be recognized by the batch prover when we run the job.

We can also modify the build job to build with testing feature, but I'm not sure if we should.

ercecan (approved):

bin/citrea/tests/proving-stats/requirements.txt
eyusufatik, line 1: can you move this to the root directory?

we don't need this to be under tests

.github/scripts/run-proving-test.sh
```diff
@@ -0,0 +1,25 @@
+TEST_DIR=bin/citrea/tests/proving-stats
+OUT_FILE_NAME="$1"
+
+docker compose -f $TEST_DIR/docker-compose.regtest.yml up -d
+sleep 5
+
+docker exec bitcoin-regtest bitcoin-cli -regtest -rpcuser=citrea -rpcpassword=citrea loadwallet sequencer
```
eyusufatik, line 10: I'm so used to using docker exec only for loading bash on the container I wasn't aware you can do such a thing lol

eyusufatik (changes requested):

eyusufatik (approved):

Issue #1987, opened 2025-02-26
Listen mode sequencer


eyusufatik:
A type of sequencer node which can be used as backup on our infra.

- [ ] Are we sure it's gonna work with the middle man? Originally posted by @kpp in https://github.com/chainwayxyz/citrea/pull/2622#discussion_r2246706954
- [ ] Bench L2Syncer against nightly

rakanalh:
Can you please shed more light on this?

We have sequencer replacement tests which ascertain the fact that we can use fullnode databases in the sequencer which practically means that we have sequencer backups.

Am i missing something?

jfldde:
> Are we sure it's gonna work with the middle man? 
cc @eyusufatik  @Chinacolt

_Originally posted by @kpp in https://github.com/chainwayxyz/citrea/pull/2622#discussion_r2246706954_

PR #2622, opened 2025-07-31
L2Syncer: Add citrea_newL2Blocks notification support

### LINKED ISSUE #2604: Enhance L2Syncer to use new citrea_subscribe
- Add support for citrea_suscribe that can be used once at tip.
- Pre-req for listen mode sequencer
- Might need a refactor, had a glance and it seems that it could already be shared by fullnode and batch-prover

jfldde:
# Description

- Add support for syncing via subscription

## Linked Issues
- Fixes #2604

### DIFF
--- a/crates/common/src/l2.rs
+++ b/crates/common/src/l2.rs
@@ -94,11 +94,15 @@ impl SequentialL2BlockBuffer {
         self.notify.notify_one();
     }
 
-    fn extend_blocks(&mut self, mut blocks: BTreeMap<u64, L2BlockResponse>) {
-        blocks.retain(|&k, _| k >= self.next_expected_height);
-        let should_notify = !blocks.is_empty();
+    fn extend_blocks(&mut self, blocks: BTreeMap<u64, L2BlockResponse>) {
+        let mut should_notify = false;
 
-        self.blocks.extend(blocks);
+        for (height, block) in blocks {
+            if height >= self.next_expected_height && !self.blocks.contains_key(&height) {
+                self.blocks.insert(height, block);
+                should_notify = true;
+            }
+        }
 
         if should_notify {
             self.notify.notify_one();
--- a/crates/common/src/l2.rs
+++ b/crates/common/src/l2.rs
@@ -605,14 +605,18 @@ async fn run_subscription_task(
     block_buffer: Arc<Mutex<SequentialL2BlockBuffer>>,
 ) {
     loop {
-        match subscribe_to_new_l2_blocks(&sequencer_ws_url, block_buffer.clone()).await {
-            Ok(_) => {}
-            Err(e) => {
-                error!("Subscription error: {}", e);
-            }
-        }
-
-        // Retry after 1 sec
-        sleep(Duration::from_secs(1)).await;
+        let exponential_backoff = ExponentialBackoff::default();
+        let _ = retry_backoff(exponential_backoff, || async {
+            subscribe_to_new_l2_blocks(&sequencer_ws_url, block_buffer.clone())
+                .await
+                .map_err(|e| {
+                    error!("Subscription error: {}", e);
+                    backoff::Error::Transient {
+                        err: e,
+                        retry_after: None,
+                    }
+                })
+        })
+        .await;
     }
 }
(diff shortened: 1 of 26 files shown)


crates/common/src/l2.rs
```diff
@@ -490,3 +568,53 @@ where
...
+
+    Ok(())
+}
+
+async fn run_subscription_task(
+    sequencer_ws_url: String,
+    block_buffer: Arc<Mutex<SequentialL2BlockBuffer>>,
+) {
+    loop {
+        match subscribe_to_new_l2_blocks(&sequencer_ws_url, block_buffer.clone()).await {
```
rakanalh:
```suggestion
        if let Err(e) = subscribe_to_new_l2_blocks(&sequencer_ws_url, block_buffer.clone()).await {
```

rakanalh: Instead of a loop here, we can use exponential backoff retry

crates/common/src/l2.rs
```diff
@@ -413,27 +485,34 @@ where
...
-                Some(l2_blocks) = l2_rx.recv() => {
-                    // While syncing, we'd like to process L2 blocks as they come without any delays.
-                    for l2_block in l2_blocks {
+                _ = notifier.notified() => {
+                    let blocks_to_process = {
+                        let mut queue = self.block_queue.lock().await;
+                        queue.drain_sequential()
+                    };
+
+                    for l2_block in blocks_to_process {
```
rakanalh, line 525: I believe we have to rethink this approach.

Notice that we drain all the blocks, for loop through them and for us to be stuck at processing the current l2_block, we `loop` indefinitely inside.

Instead of all this, we use the queue to get first unprocessed block, try processing it, if it fails, we don't pop it or drain it out of the queue, it just stays there. This would introduce yet another problem which is the queue getting larger as the operating sequencer gets more blocks. So if the current block fails to process 3 times, we should fail and stop the node.

rakanalh (changes requested):

crates/common/src/l2.rs
ercecan: I don't mind draining all of it but I agree that we should refactor the loop to panic or sth after couple retries

jfldde: >I don't mind draining all of it but I agree that we should refactor the loop to panic or sth after couple retries

We could add a max retry but do note that this follows the existing behaviour, the only difference being that instead of processing the blocks from the channel, they were drained from the queue

crates/common/src/l2.rs
```diff
@@ -57,6 +60,61 @@ enum SyncError {
...
+            self.notify.notify_one();
+        }
+    }
+
+    /// Drain from next_expected_height until the first gap or until the buffer is fully consumed
+    /// Returns an ordered vector of `L2BlockResponse`
+    fn drain_sequential(&mut self) -> Vec<L2BlockResponse> {
+        let mut sequential_blocks = Vec::new();
+
+        while let Some(block) = self.blocks.remove(&self.next_expected_height) {
```
kpp, line 117: I don't know if we have a 100% guarantee that there will be NO gaps when we `.add_block`. But if there are, then it's gonna be a big problem.

crates/common/src/l2.rs
```diff
@@ -490,3 +568,53 @@ where
...
+) -> anyhow::Result<()> {
+    debug!(
+        "Connecting to sequencer subscription at {}",
+        sequencer_ws_url
+    );
+
+    let ws_client = WsClientBuilder::default().build(&sequencer_ws_url).await?;
+    let mut subscription = ws_client
+        .subscribe(
+            "citrea_subscribe",
```
kpp, line 594: Are we sure it's gonna work with the middle man? 
cc @eyusufatik  @Chinacolt

crates/common/src/l2.rs
jfldde: A gap through subscription would be filled via polling

ercecan (approved):

crates/common/src/l2.rs
jfldde: Noted https://github.com/chainwayxyz/citrea/issues/1987#issuecomment-3150342491

crates/common/src/l2.rs
jfldde: Not taking care of it here :
Follow-up issue : https://github.com/chainwayxyz/citrea/issues/2629

ercecan (approved):

PR #2626, opened 2025-08-01
Listen mode sequencer: Commitments tracking

### LINKED ISSUE #2611: Listen mode sequencer: Track sent commitments
When running as listen mode sequencer, it needs to tracks pending commitments and commitment by index so that when restarted as producer sequencer, it goes through `load_next_commitment_index_and_start_height` and resolves correctly the pending/existing commitment ordering

rakanalh:
# Description

## Linked Issues
- Fixes #2611

### DIFF
--- a/crates/sequencer/src/l1_syncer.rs
+++ b/crates/sequencer/src/l1_syncer.rs
@@ -1,0 +1,181 @@
+//! DA block syncing for the sequencer
+//!
+//! This module is responsible for processing L1 blocks, extracting and storing
+//! sequencer commitments for the sequencer in listen mode.
+
+use std::collections::VecDeque;
+use std::sync::Arc;
+
+use anyhow::bail;
+use citrea_common::backup::BackupManager;
+use citrea_common::cache::L1BlockCache;
+use citrea_common::da::{extract_sequencer_commitments, sync_l1};
+use citrea_common::RollupPublicKeys;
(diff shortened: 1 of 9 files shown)


crates/sequencer/src/l1_syncer.rs
```diff
@@ -0,0 +1,181 @@
...
+    /// 4. Maintains metrics about syncing progress
+    #[instrument(name = "SequencerL1Syncer", skip_all)]
+    pub async fn run(mut self, mut shutdown_signal: GracefulShutdown) {
+        let l1_sync_worker = sync_l1(
+            self.scan_l1_start_height,
+            self.da_service.clone(),
+            self.pending_l1_blocks.clone(),
+            self.l1_block_cache.clone(),
+            None,
+        );
```
jfldde, line 108: We need a way to sync up to latest block and not only up to finalized block height as is the current sync_l1 behaviour or we might miss non-finalized/pending commitments.

Not sure if we want to do it here or restore pending commitment from DA on startup (instead of relying solely on `PendingSequencerCommitment` table. 
cc/ @eyusufatik

ercecan: I believe the commitment controller's `resubmit_pending_commitment` will handle all the non-finalized and in mempool commitments and also it will create pending commitments it self because it is deterministic

ercecan (approved):

PR #2628, opened 2025-08-04
chore: update to rust 1.88

ercecan:
# Description
Updates rust version to 1.88

## Linked Issues
- Related to # (issue) 

## Testing

## Docs


jfldde (approved):

Makefile
```diff
@@ -90,7 +90,7 @@ lint:  ## cargo check and clippy. Skip clippy on guest code since it's not suppo
 	dprint check
 	cargo +nightly fmt --all --check
 	cargo check --all-targets --all-features
-	SKIP_GUEST_BUILD=1 cargo clippy --all-targets --all-features
+	SKIP_GUEST_BUILD=1 cargo clippy --all-targets --all-features -- --allow clippy::doc_overindented_list_items
```
kpp, line 93: What's that?

kpp: It's a little bit unexpected to see this in Makefile instead of in clippy.toml (so it's applied to both `make` and `cargo`) or in the code itself where we need it.

kpp (approved):

Cargo.toml
```diff
@@ -49,6 +49,9 @@ homepage = "https://citrea.xyz"
 publish = false
 repository = "https://github.com/chainwayxyz/citrea"
 
+[workspace.lints]
+clippy.doc_overindented_list_items = "allow"
```
kpp, line 53: The question remains: do we need it? What if we don't allow it? How many lines of doc comments will change?

PR #2643, opened 2025-08-08
Implement listen mode sequencer mempool subscriber

### LINKED ISSUE #2630: Listen mode sequencer: Mempool tx subscription

ercecan:
# Description
Implements listen mode sequencer mempool subscriber
Has a small buffer before saving the txs to db (3 seconds)

## Linked Issues
- Fixes #2630

### DIFF
--- a/bin/citrea/src/main.rs
+++ b/bin/citrea/src/main.rs
@@ -265,7 +265,7 @@ where
                 is_listen_mode,
                 sequencer_config
             );
-            let (mempool_transaction_tx, _mempool_transaction_rx) = broadcast::channel(10);
+            let (mempool_transaction_tx, _mempool_transaction_rx) = broadcast::channel(1000);
             match rollup_blueprint
                 .create_sequencer(
                     genesis_config,
--- a/bin/citrea/src/main.rs
+++ b/bin/citrea/src/main.rs
@@ -37,6 +37,7 @@ use sov_rollup_interface::Network;
 use sov_state::storage::NativeStorage;
 use tokio::signal;
 use tokio::signal::unix::{signal, SignalKind};
+use tokio::sync::broadcast;
 use tracing::{debug, error, info, instrument};
 
 use crate::cli::{node_type_from_args, Args, NodeWithConfig, SupportedDaLayer};
@@ -264,6 +265,7 @@ where
                 is_listen_mode,
                 sequencer_config
             );
+            let (mempool_transaction_tx, _mempool_transaction_rx) = broadcast::channel(10);
             match rollup_blueprint
                 .create_sequencer(
                     genesis_config,
@@ -273,6 +275,7 @@ where
                     ledger_db,
                     storage_manager,
                     l2_block_tx,
+                    mempool_transaction_tx,
                     rpc_module,
                     backup_manager,
                     task_executor.clone(),
--- a/crates/sequencer/src/lib.rs
+++ b/crates/sequencer/src/lib.rs
@@ -48,6 +48,7 @@ use jsonrpsee::RpcModule;
 use l1_syncer::L1Syncer;
 use listen_mode::ListenModeSequencer;
 use mempool::CitreaMempool;
+use mempool_syncer::MempoolSyncer;
 use parking_lot::Mutex;
 use reth_tasks::TaskExecutor;
 pub use rpc::SequencerRpcClient;
@@ -56,6 +57,7 @@ use sov_db::ledger_db::{LedgerDB, SequencerLedgerOps};
 use sov_modules_stf_blueprint::StfBlueprint;
 use sov_prover_storage_manager::ProverStorageManager;
 use sov_rollup_interface::fork::ForkManager;
+use sov_rollup_interface::rpc::MempoolTransactionSignal;
 use sov_rollup_interface::services::da::DaService;
 use tokio::sync::mpsc::unbounded_channel;
 use tokio::sync::{broadcast, Mutex as AsyncMutex};
@@ -77,6 +79,8 @@ mod l1_syncer;
 mod listen_mode;
 /// Module containing mempool functionality for transaction management
 mod mempool;
+/// Module containing mempool synchronization functionality for listen mode sequencer
+mod mempool_syncer;
 /// Module containing metrics collection and reporting functionality
 mod metrics;
 /// Provides access to sequencer RPC functionality
@@ -129,6 +133,7 @@ pub fn build_services<Da>(
     ledger_db: LedgerDB,
     storage_manager: ProverStorageManager,
     l2_block_tx: broadcast::Sender<u64>,
+    mempool_transaction_tx: broadcast::Sender<MempoolTransactionSignal>,
     fork_manager: ForkManager<'static>,
     rpc_module: RpcModule<()>,
     backup_manager: Arc<BackupManager>,
@@ -158,6 +163,8 @@ where
         ledger_db.clone(),
         sequencer_config.test_mode,
         l2_block_tx.subscribe(),
+        mempool_transaction_tx.clone(),
+        mempool_transaction_tx.subscribe(),
     );
     let rpc_module = rpc::register_rpc_methods(rpc_context, rpc_module)?;
 
@@ -169,7 +176,7 @@ where
             .expect("Listen Mode Config must be set in listen mode");
 
         let l2_syncer = L2Syncer::new(
-            listen_mode_config.sequencer_client_url,
+            listen_mode_config.sequencer_client_url.clone(),
             listen_mode_config.sync_blocks_count,
             init_params,
             native_stf,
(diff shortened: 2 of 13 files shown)


bin/citrea/src/main.rs
```diff
@@ -264,6 +265,7 @@ where
                 is_listen_mode,
                 sequencer_config
             );
+            let (mempool_transaction_tx, _mempool_transaction_rx) = broadcast::channel(10);
```
jfldde: Why 10 ? Feels low

crates/sequencer/src/mempool_syncer.rs
```diff
@@ -0,0 +1,156 @@
...
+where
+    DB: SequencerLedgerOps + Clone + Send + Sync + 'static,
+{
+    /// Database for ledger operations
+    ledger_db: DB,
+    /// Buffer for mempool transactions before storing into the ledger db
+    /// Mapping: Tx hash to rlp encoded transaction
+    transactions_buffer: Arc<Mutex<HashMap<Vec<u8>, Vec<u8>>>>,
+    /// Transaction hashes to be removed from mempool ledger db
+    transactions_to_remove_buffer: Arc<Mutex<HashSet<Vec<u8>>>>,
```
jfldde, line 28: Why not `CitreaMempool` ?

ercecan: Do you mean why not use mempool implementation instead of hashmaps? If that is the case I believe this would be more efficient as it has other checks

crates/sequencer/src/listen_mode.rs
```diff
@@ -100,6 +105,12 @@ where
                 |shutdown_signal| async move { self.l1_syncer.run(shutdown_signal).await },
             );
 
+        self.task_executor
+            .spawn_critical_with_graceful_shutdown_signal(
```
rakanalh: Don't use spawn critical... this is just for the sequencer runner. 
```suggestion
            .spawn_with_graceful_shutdown_signal(
```

crates/sequencer/src/mempool_syncer.rs
```diff
@@ -0,0 +1,156 @@
...
+            sequencer_ws_endpoint,
+        }
+    }
+
+    #[instrument(name = "MempoolSyncer", skip_all)]
+    pub async fn run(self, shutdown_signal: GracefulShutdown) {
+        let this = self.clone();
+        tokio::spawn(async move {
+            // Start the subscription task
+            this.run_subscription_task().await;
```
rakanalh: Pass down the shutdown signal to the spawned tasks so that they can be gracefully shutdown

crates/sequencer/src/mempool_syncer.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub async fn update_mempool_transaction_db_task(&self) {
+        loop {
+            // Waiting at least 2 seconds here so that txs that got in block are removed so we do less db ops
+            tokio::time::sleep(Duration::from_secs(3)).await;
+            {
+                let transactions_buffer = self.transactions_buffer.clone();
+                let mut txs_buffer = transactions_buffer.lock();
+
+                let txs = txs_buffer.drain().collect();
+                if let Err(e) = self.ledger_db.batch_insert_mempool_txs(txs) {
```
rakanalh: Just wondering why save the transactions into mempool specific tables while we can just insert them directly to the listen mode mempool. When the listen mode is converted into an actual sequencer, the transactions will be there.

crates/sequencer/src/lib.rs
```diff
@@ -129,6 +133,7 @@ pub fn build_services<Da>(
     ledger_db: LedgerDB,
     storage_manager: ProverStorageManager,
     l2_block_tx: broadcast::Sender<u64>,
+    mempool_transaction_tx: broadcast::Sender<MempoolTransactionSignal>,
```
rakanalh, line 136: Since we created `SequencerRpcMessage` which would let us send messages from Rpc handlers into the sequencer, we can generalize this as well so that the sequencer sends to the mempool ... not specifically for mempool.

I am just thinking that:
1. We have `l2_block_tx/rx` to notify subscriptions about new l2 blocks
2. We have `rpc_message_tx/rx`
3. and now this.

That basically move stuff from/to sequencer / listen mode sequencer.

Wondering if we can refactor a bit to make use multiple receivers in this case so that we publish with a single message variant (enum) and receivers would choose to process or ignore the message based on specialty.

No pressure though, let's just think about this to improve the code base instead of adding more tx/rx combinations.

ercecan: yeah makes sense let me think about this

crates/sequencer/src/mempool_syncer.rs
ercecan: mempool is not persistent so we will lose txs after restart, that's why I went with this table. Upon restart sequencer will check this table and insert them to mempool.

bin/citrea/src/main.rs
```diff
@@ -264,6 +265,7 @@ where
                 is_listen_mode,
                 sequencer_config
             );
+            let (mempool_transaction_tx, _mempool_transaction_rx) = broadcast::channel(1000);
```
jfldde, line 268: Wouldn't it be safe to unbound it if the tx is sent post mempool validation on the main sequencer ?

ercecan: makes sense, but I see that tokio does not support unbounded mpmc, I saw that there flume might be a good alternative should I use that?
https://crates.io/crates/flume

ercecan: oh nvm, it works differently than broadcast it picks up consumber by consumer, not like broadcast where all consumers receive all messages

crates/sequencer/src/lib.rs
ercecan: Okay I see that I can drop one of the broadcast receivers and however I cannot do one sender one receiver directly because rpc_message sender and other senders are different one is mpsc and the other is broadcast.
After dropping one receiver I can do something like a RpcChannelManager where I collect all the senders and receivers under one structure which would make it a bit more concise maybe?
But I wanna make this a new issue, I'll add it to the sub issues of listen mode sequencer

ercecan: https://github.com/chainwayxyz/citrea/issues/2648

crates/sequencer/src/mempool_syncer.rs
```diff
@@ -0,0 +1,178 @@
...
+                        let txs_to_remove_buffer = self.transactions_to_remove_buffer.clone();
+                        let mut txs_to_remove = txs_to_remove_buffer.lock();
+
+                        let txs = txs_to_remove.drain().collect();
+                        if let Err(e) = self.ledger_db.remove_mempool_txs(txs) {
+                            error!("Failed to batch remove mempool transactions: {}", e);
+                        }
+                    }
+                }
+            }
```
jfldde: We should be able to remove from `txs` first before going through the ledger db ops. This should be a common occurence

crates/sequencer/src/mempool_syncer.rs
```diff
@@ -0,0 +1,194 @@
...
+        // Waiting at least 2 seconds here so that txs that got in block are removed so we do less db ops
+        let mut interval = tokio::time::interval(Duration::from_secs(3));
+        interval.tick().await;
+
+        loop {
+            tokio::select! {
+                _ = shutdown_signal.clone() => {
+                    info!("Shutting down mempool transaction update task");
+                    return;
+                }
```
jfldde: Let's update the mempool tx once more as part of the shutdown so we don't miss any

crates/sequencer/src/mempool_syncer.rs
```diff
@@ -0,0 +1,194 @@
...
+                    }
+                })
+            })
+            .await;
+        }
+    }
+
+    pub async fn update_mempool_transaction_db_task(&self, shutdown_signal: GracefulShutdown) {
+        // Waiting at least 2 seconds here so that txs that got in block are removed so we do less db ops
+        let mut interval = tokio::time::interval(Duration::from_secs(3));
```
jfldde, line 93: If we update_mempool_transactions as part of the shutdown, I think we can have a higher interval without loss ?

crates/sequencer/src/mempool_syncer.rs
```diff
@@ -0,0 +1,194 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    fn update_mempool_transactions(&self) -> anyhow::Result<()> {
+        let mut txs = {
+            let mut guard = self.transactions_buffer.lock();
+            std::mem::take(&mut *guard).into_iter().collect::<Vec<_>>() // moves out, leaves empty HashMap
```
jfldde: Clearer with drain(..) instead

crates/sequencer/src/mempool_syncer.rs
```diff
@@ -0,0 +1,197 @@
...
+                }
+            }
+        }
+    }
+
+    fn update_mempool_transactions(&self) -> anyhow::Result<()> {
+        let mut txs = {
+            let mut guard = self.transactions_buffer.lock();
+            guard.drain().collect::<Vec<_>>()
+        };
```
jfldde:
```suggestion
let mut txs = self.transactions_buffer.lock().drain().collect::<Vec<_>>();
```

jfldde (approved):

rakanalh (approved):
LGTM

PR #2644, opened 2025-08-08
Send proofs to DA as soon as they are proved

### LINKED ISSUE #2317: Debug why DA transaction sending delays when high number of proving happens
When we reindexed the chain from commitment 1, we saw that the prover did not send the proof results despite logs showing them as completed. There were no errors or warnings, and after couple of hours/days, it started sending all of them quickly and catched up. Even though it is not a "bug", its best to look into this when we have time. I suspect tokio/futures related quirk that is causing this, so debugging with tokio console would be our best option.

kpp:
# Description

Push (job_id, rx(proof)) in the stream and process it in parallel instead of pushing them to a vec. The order is FIFO, so [j1, j2, j3] will be sent to DA in that order even if j2 finishes faster than j1.

## Linked Issues
- Fixes https://github.com/chainwayxyz/citrea/issues/2317

### DIFF
--- a/crates/batch-prover/src/prover.rs
+++ b/crates/batch-prover/src/prover.rs
@@ -296,7 +296,12 @@ where
             return Ok(vec![]);
         }
 
-        let mut proving_jobs = Vec::with_capacity(partitions.len());
+        let (proving_jobs_tx, proving_jobs_rx) = mpsc::unbounded_channel();
+
+        // start watching the proving jobs to finish in the background
+        self.watch_proving_jobs(proving_jobs_rx);
+
+        let mut job_ids = Vec::with_capacity(partitions.len());
         for partition in partitions {
             let id = Uuid::now_v7();
             let input = self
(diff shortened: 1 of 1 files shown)


crates/batch-prover/src/prover.rs
```diff
@@ -322,10 +330,7 @@ where
                 .context("Failed to delete pending commitments")?;
         }
 
-        let job_ids = proving_jobs.iter().map(|job| job.0).collect();
-
-        // start watching the proving jobs to finish in the background
-        self.watch_proving_jobs(proving_jobs);
+        assert!(!job_ids.is_empty(), "received empty jobs list");
```
eyusufatik, line 333: is this useful here

kpp: Same behavior as before, but I can’t check if a stream rx is empty

eyusufatik (approved):

ercecan (approved):
Watched the loom video, really nice and clean solution @kpp 🫡

rakanalh (approved):

PR #2650, opened 2025-08-11
Proving stats: Trigger on label

exeokan:
# Description
Only runs the `proving-stats` job if the PR has guest-code label

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -518,7 +518,7 @@ jobs:
     name: Proving stats
     runs-on: ubicloud-standard-2
     needs: [build, docker-setup]
-    if: github.event_name == 'pull_request' && github.event.pull_request.draft == false && contains(github.event.pull_request.labels.*.name, 'guest-code')
+    if: github.event_name == 'pull_request' && github.event.pull_request.draft == false && contains(github.event.pull_request.labels.*.name, 'guest-code') 
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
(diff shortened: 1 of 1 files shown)


.github/workflows/checks.yml
```diff
@@ -518,7 +518,7 @@ jobs:
     name: Proving stats
     runs-on: ubicloud-standard-2
     needs: [build, docker-setup]
-    if: github.event_name == 'pull_request' && github.event.pull_request.draft == false
+    if: github.event.pull_request.draft == false && contains(github.event.pull_request.labels.*.name, 'guest-code')
```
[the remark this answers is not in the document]
exeokan: on the second thought, keeping it may be a good idea, because we know that `github.event.pull_request.draft == false` evaluates to true on push??

.github/workflows/checks.yml
```diff
@@ -518,7 +518,7 @@ jobs:
     name: Proving stats
     runs-on: ubicloud-standard-2
     needs: [build, docker-setup]
-    if: github.event_name == 'pull_request' && github.event.pull_request.draft == false
+    if: github.event_name == 'pull_request' && github.event.pull_request.draft == false && contains(github.event.pull_request.labels.*.name, 'guest-code')
```
jfldde: Let's remove draft. We then have the option to set the label in draft if that can help

jfldde (approved):

.github/workflows/checks.yml
exeokan: We can remove it👍🏻. But probably won't run when draft since it depends on the build job.

eyusufatik (approved):

Issue #2646, opened 2025-08-11
Listen mode sequencer: Folder restructure


ercecan:
After all tasks are done I believe we can do a listen mode module in sequencer crate and put all listen mode related sub modules there

PR #2654, opened 2025-08-13
Listen mode sequencer docs & metrics

### LINKED ISSUE #2639: Listen mode sequencer: Documentation

### LINKED ISSUE #2645: Listen mode sequencer: Metrics
We can record:
- Last scanned l2 height
- Last scanned l1 height
- Number of txs in incoming txs buffer
- Number of txs in txs to be removed buffer
- L2/L1 Block process duration

ercecan:
# Description
Adds module top level doc
Adds metrics and grafana dashboard setup.
This pr also fixes: https://github.com/chainwayxyz/citrea/pull/2654#discussion_r2273530265
and tests that it is working as expected

## Linked Issues
- Fixes #2645 
- Fixes #2639 

Screenshot 2025-08-13 at 16 42 29


crates/sequencer/src/l1_syncer.rs
```diff
@@ -198,6 +202,14 @@ where
                 .set_last_scanned_l1_height(SlotNumber(l1_height))?;
             pending_l1_blocks.pop_front();
 
+            SM.listen_mode_l1_block_process_duration_secs.record(
+                Instant::now()
+                    .saturating_duration_since(start)
+                    .as_secs_f64(),
+            );
```
jfldde, line 209: Shouldn't this be optional and only if running as listen_monde ?

crates/sequencer/src/listen_mode.rs
```diff
@@ -25,8 +45,10 @@ where
         Ok(())
```
jfldde: Unrelated to this PR but don't we need to `set_l2_state_diff` here ?

crates/sequencer/src/l1_syncer.rs
ercecan: This module is only used by listen mode sequencer, will make more sense after #2646

crates/sequencer/src/listen_mode.rs
ercecan: yes! I was thinking we should have done sth here but what?? Thanks for reminding me

crates/sequencer/src/l1_syncer.rs
jfldde: Oh, right

crates/sequencer/src/listen_mode.rs
jfldde: Let's add this to the test as well somewhere (Have a state diff trigger a commitment after restarted as main sequencer) to make sure this is correctly handled (should fail for now)

bin/citrea/tests/bitcoin/listen_mode_sequencer.rs
```diff
@@ -827,3 +828,175 @@ impl TestCase for MempoolSyncerTest {
...
+        }
+    }
+
+    fn scan_l1_start_height() -> Option<u64> {
+        Some(147)
+    }
+
+    fn sequencer_config() -> SequencerConfig {
+        SequencerConfig {
+            // Keeping this high to generate commitments from state diff check
```
jfldde, line 850: I think this would be clearer if we bump this value much higher, so that we're sure commitments are not trigger by it, and only create commitment with state diff threshold

jfldde (approved):

PR #2655, opened 2025-08-13
Replace L1 Syncer interval with notification

jfldde:
# Description

- Replaces interval with notification as soon as block is available
- Results in ~100x faster L1 sync performance locally during initial sync and catch-up scenarios


ercecan (approved):

eyusufatik (commented):
doesn't this slow down L2 sync though? have you tried against testnet

jfldde:
> doesn't this slow down L2 sync though? have you tried against testnet

I can bench more thoroughly on Monday but I just tried locally and ended up syncing testnet4 in 10min so any slowdown on L2 sync will be quite limited

eyusufatik (approved):

PR #2692, opened 2025-08-18
Track pending deposits and only allow adding the same deposit once

### LINKED ISSUE #2681: DepositDataMempool accepts a successful deposit more than once
If a deposit is simulated as successful, the sequencer accepts multiple instances of the same deposit until one is mined. (and then the rest would fail on block production)

since a deposit is 10 cBTC, I don't think we need to apply size limitations, so let's just keep a hashset of the deposit data and reject based on that.

rakanalh:
# Description

## Linked Issues
- Fixes #2681


crates/sequencer/src/deposit_data_mempool.rs
```diff
@@ -48,20 +50,118 @@ impl DepositDataMempool {
...
+    /// # Returns
+    /// `true` if the deposit was added, `false` if it was already pending
     #[instrument(level = "trace", skip_all, ret)]
-    pub fn add_deposit_tx(&mut self, req: Vec<u8>) {
+    pub fn add_deposit_tx(&mut self, req: Vec<u8>) -> bool {
+        // Check if deposit is already pending
+        if !self.pending_deposits.insert(req.clone()) {
+            tracing::debug!("Deposit already pending in mempool");
+            return false;
+        }
```
jfldde: Do we need the new set ? Wouldn't a `contains` call on `accepted_deposit_txs` be enough ?

rakanalh: You're right, though using a hashset has O(1) lookup compared to the O(N) on accepted_deposit_txs. I am guessing we are favoring performance since @eyusufatik asked for a hashset in the original issue.

crates/sequencer/src/rpc.rs
```diff
@@ -311,10 +311,20 @@ impl SequencerRpcServer for SequencerRpcServerImpl {
...
             Ok(hex_res) => {
                 tracing::debug!("Deposit tx processed successfully {}", hex_res);
-                self.context
+                let already_added = self
+                    .context
                     .deposit_mempool
                     .lock()
                     .add_deposit_tx(deposit.to_vec());
+
+                if !already_added {
```
eyusufatik: shouldn't this be already_added? not !already_added

eyusufatik (changes requested):

crates/sequencer/src/deposit_data_mempool.rs
eyusufatik: we can parse the deposit data into Bridge contract's deposit function arguments, extract the transaction, and use it's txid actually. which would be way better.

crates/sequencer/src/rpc.rs
rakanalh: The name of the var is wrong... will amend after your push

crates/sequencer/src/deposit_data_mempool.rs
```diff
@@ -60,7 +60,9 @@ impl DepositDataMempool {
 
         // Remove fetched deposits from the pending set
         for deposit in &deposits {
-            self.pending_deposits.remove(deposit);
+            if let Ok(txid) = Self::calc_tx_id(deposit) {
+                self.pending_deposits.remove(&txid.to_vec());
```
jfldde: Can keep it as array

jfldde (approved):

eyusufatik (approved):

PR #2695, opened 2025-08-19
rust bitcoin merkle tree: assert index is 0 at the end of merkle proo…

### LINKED ISSUE #2670: Bitcoin Merkle Tree verify extra bits attack
## Summary
If you supply `calculate_root_with_merkle_proof` a correct proof with an index that's correct but has extra bits on the left of the proof.len()th index, it will pass.

For instance, if your merkle proof has 4 levels, and you index is `0011` you can get away with `10011` index. 

This is due the code not restricting the index at the end:

```rust
/// It recomputes the Merkle root using
/// - a leaf hash
/// - a Merkle proof (list of sibling hashes along the path)
/// - an index indicating left/right sibling order at each level
pub fn calculate_root_with_merkle_proof(
    txid: [u8; 32],
    idx: u32,
    merkle_proof: &[[u8; 32]],
) -> [u8; 32] {
    let mut preimage: [u8; 64] = [0; 64];
    let mut combined_hash: [u8; 32] = txid;
    let mut index = idx;
    let mut level: u32 = 0;
    while level < merkle_proof.len() as u32 {
        // using rightmost bit to tell left or right
        if index % 2 == 0 {
            preimage[..32].copy_from_slice(&combined_hash);
            preimage[32..].copy_from_slice(&merkle_proof[level as usize]);
            combined_hash = calculate_double_sha256(&preimage);
        } else {
            let left = &merkle_proof[level as usize];
            if left == &combined_hash {
                panic!("Merkle proof is invalid: left hash matches combined hash");
            }
            preimage[..32].copy_from_slice(left);
            preimage[32..].copy_from_slice(&combined_hash);
            combined_hash = calculate_double_sha256(&preimage);
        }
        level += 1;
        // this is consuming bits on the right basically
        index /= 2; 
    }
    // here we should have went through all the bits
    combined_hash
}
```

We should check index == 0 at the very end.

This passes:

```rust
  #[test]
  fn test_merkle_root_with_proof() {
      let mut transactions: Vec<[u8; 32]> = vec![];
      for i in 0u8..100u8 {
          let tx = [i; 32];
          transactions.push(tx);
      }
      let tree = BitcoinMerkleTree::new(transactions.clone());
      let root = tree.root();
      let idx_path = tree.get_idx_path(0);
      let calculated_root =
          BitcoinMerkleTree::calculate_root_with_merkle_proof(transactions[0], 0, &idx_path);
      assert_eq!(root, calculated_root);

      let calculated_root =
          BitcoinMerkleTree::calculate_root_with_merkle_proof(transactions[0], 256, &idx_path);
      assert_eq!(root, calculated_root);
  }
```

eyusufatik:
…f verification

# Description
closes #2670

### DIFF
--- a/crates/bitcoin-da/src/helpers/merkle_tree.rs
+++ b/crates/bitcoin-da/src/helpers/merkle_tree.rs
@@ -294,7 +294,12 @@ mod tests {
             BitcoinMerkleTree::calculate_root_with_merkle_proof(transactions[0], 0, &idx_path);
         assert_eq!(root, calculated_root);
 
+        // with a 7 level tree like ours, represent 0 with
+        // 0000000 in bits
+        // without the assert!(index == 0) at the and of the verify function
+        // any index that has its last 7 bits 0 would pass
+        // e.g. 10000000 (128) or 100000000 (256) or even 101110000000 (2944)
         let _ =
-            BitcoinMerkleTree::calculate_root_with_merkle_proof(transactions[0], 256, &idx_path);
+            BitcoinMerkleTree::calculate_root_with_merkle_proof(transactions[0], 2944, &idx_path);
     }
 }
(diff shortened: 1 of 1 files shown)


ozankaymak (commented):
LGTM

okkothejawa (approved):

PR #2703, opened 2025-08-19
Store state diff before committing l2 block

### LINKED ISSUE #2696: Handle non existing state diff
## Summary
Since we commit the l2 block then save the state diff to ledger db, there is a chance that the prover/sequencer shuts down between the two operations. If this happens, the batch prover panics on `get_l2_state_diff` and I haven't checked the sequencer.

We can assume an empty diff for this and keep operating.

rakanalh:
# Description

## Linked Issues
- Fixes #2696

### DIFF
--- a/crates/batch-prover/src/l2_syncer.rs
+++ b/crates/batch-prover/src/l2_syncer.rs
@@ -6,15 +6,12 @@
 use std::sync::Arc;
 use std::time::Instant;
 
-use anyhow::{bail, Context as _};
 use backoff::backoff::Backoff;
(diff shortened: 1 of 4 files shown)


crates/batch-prover/src/l2_syncer.rs
eyusufatik, line 1: pls minimize the diff, don't make a new function and move stuff around

eyusufatik (approved):

jfldde (approved):

jfldde:
Need to be mindful of order when merging to #1987

PR #2714, opened 2025-08-21
incr. timestamp in evm query pending env

### LINKED ISSUE #2712: `get_pending_block_env` should update block timestamp

eyusufatik:
# Description

fixes #2712


jfldde (approved):

eyusufatik:
we may use now() actually

PR #2715, opened 2025-08-21
BitcoinService: Choose prev_utxo based off UtxoSelectionMode

jfldde:
# Description

- Introduces new enum `UtxoSelectionMode` that can be run either as `Chained` (current behaviour) or `Oldest`.
New selection mode `Oldest` will choose the UTXO with the highest amount of confirmations when prev_utxo is still part of the sending queue.
- Changes available utxos selection logic. Filters out of the `list_unspent` all txs that are in use by queued tx or non-finalized change outputs (so that we don't end up selecting potential change output from existing utxo chain). Choosing to keep only finalized as even confirmed could be impacted by re-org and end up being over mempool policy limit

### DIFF
--- a/crates/bitcoin-da/src/service.rs
+++ b/crates/bitcoin-da/src/service.rs
@@ -371,7 +371,7 @@ impl BitcoinService {
                 Err(BitcoinServiceError::QueueNotEmpty)
             }
             UtxoSelectionMode::Oldest => Ok(if prev_utxo.is_some() {
-                // Latest monitored TX has been succesfully accepted to mempool and can be used as starting point for another utxo chain
+                // Latest monitored TX has been successfully accepted to mempool and can be used as starting point for another utxo chain
                 prev_utxo
             } else {
                 // Latest monitored TX has `Queued` status and internal `get_tx_out` errors.
--- a/crates/bitcoin-da/src/service.rs
+++ b/crates/bitcoin-da/src/service.rs
@@ -360,10 +360,8 @@ impl BitcoinService {
     }
 
     async fn select_prev_utxo(&self) -> Result<Option<UTXO>> {
-        let queue = self.tx_queue.lock().await;
         let prev_utxo = self.get_prev_utxo().await;
-
-        if queue.is_empty() {
+        if self.tx_queue.lock().await.is_empty() {
             return Ok(prev_utxo);
         }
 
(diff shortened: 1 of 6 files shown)


crates/bitcoin-da/src/service.rs
```diff
@@ -297,14 +327,11 @@ impl BitcoinService {
...
-        // otherwise, the tx that will be built may use the same UTXO as the one in the queue
-        if !self.tx_queue.lock().await.is_empty() {
-            return Err(BitcoinServiceError::QueueNotEmpty);
-        }
+
+        let prev_utxo = self.select_prev_utxo().await?;
 
         let da_txs = self
-            .create_da_transactions_with_fee_rate(tx_request, fee_sat_per_vbyte)
+            .create_da_transactions_with_fee_rate(tx_request, fee_sat_per_vbyte, prev_utxo)
```
eyusufatik, line 334: but we may end up choosing other utxos that are in the mempool to pay for the tx, prev_utxo may be too small

crates/bitcoin-da/src/service.rs
```diff
@@ -110,6 +130,11 @@ impl citrea_common::FromEnv for BitcoinServiceConfig {
             tx_backup_dir: read_env("TX_BACKUP_DIR")?,
             monitoring: MonitoringConfig::from_env().ok(),
             mempool_space_url: read_env("MEMPOOL_SPACE_URL").ok(),
+            utxo_selection_mode: serde_json::from_str(&format!(
+                "\"{}\"",
```
eyusufatik, line 134: why the \" ?

crates/bitcoin-da/src/service.rs
```diff
@@ -332,6 +359,25 @@ impl BitcoinService {
...
+            return Ok(prev_utxo);
+        }
+
+        match self.utxo_selection_mode {
+            UtxoSelectionMode::Chained => {
+                // Prevent UTXO conflicts when queue is not empty and running UtxoSelectionMode::Chained mode
+                Err(BitcoinServiceError::QueueNotEmpty)
+            }
+            UtxoSelectionMode::Oldest => Ok(if prev_utxo.is_some() {
+                prev_utxo
```
eyusufatik, line 375: this seems weird?

eyusufatik: so in oldest mode, you still use the prev_utxo from the tx chain, right?

eyusufatik (changes requested):

crates/bitcoin-da/src/service.rs
jfldde: Taken it from existing `BatchProverConfig` but I'll see if I can clean it up

crates/bitcoin-da/src/service.rs
jfldde: I'll document it.
prev_utxo uses monitoring service to resolve using the latest monitored tx. If the latest TX is queued, internal `get_tx_out` call will error and it will resolve to None. 
If it resolves to Some, it means that there are some TX queued, but a new chain with valid utxos has already been started and it should continue from it.

tl;dr :
None => Start a new utxo chain, 
Some => continue latest utxo chain

eyusufatik: > Some => continue latest utxo chain

but for oldest do we want this case?

jfldde: Behaviour is this :

1 - Initial UTXO is selected until it hits the first mempool policy limit, then we start queueing. 
2- When queued, first call after queuing (in that case None) starts a new utxo chain
3 - Next call, queue is not empty but latest UTXO is part of the new UTXO chain (doesn't share mempool policy limit with initial utxo chain, we want to continue from there), this is the Some case.
4 - Continue this chain as long as we don't hit mempool policy limit. If we hit, start again from 1

eyusufatik: okay that's great

eyusufatik (approved):

PR #2716, opened 2025-08-22
Check existing commitments before CreateInput

### LINKED ISSUE #2707: `create_circuit_input` to check for empty commitments
## Summary

if we send nonexisting commitments to the `createCircuitInput` RPC, the call panics in `filter_unsynced_commitments` down the line bc it expects a non-empty list. So we should make sure there is at least one commitment pulled from db in the RPC handler

rakanalh:
# Description

## Linked Issues
- Fixes #2707


eyusufatik (approved):
looking good.

jfldde (approved):

PR #2729, opened 2025-08-26
BitcoinService: Sanity check fee calculation before queueing

jfldde:
# Description

- Make sure every queued txs meet minimum relay fee of 1sat/vb

### DIFF
--- a/crates/bitcoin-da/src/fee.rs
+++ b/crates/bitcoin-da/src/fee.rs
@@ -12,8 +12,10 @@ use bitcoincore_rpc::json::{
 use bitcoincore_rpc::{Client, RpcApi};
 use tracing::{debug, instrument, trace, warn};
 
+use crate::error::BitcoinServiceError;
 use crate::monitoring::{MonitoredTx, MonitoredTxKind};
 use crate::spec::utxo::UTXO;
+use crate::tx_signer::SignedTxPair;
 
 const DEFAULT_MEMPOOL_SPACE_URL: &str = "https://mempool.space/";
 const MEMPOOL_SPACE_RECOMMENDED_FEE_ENDPOINT: &str = "api/v1/fees/recommended";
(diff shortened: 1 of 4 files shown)


crates/bitcoin-da/src/service.rs
```diff
@@ -329,13 +331,25 @@ impl BitcoinService {
...
+                tx_request,
+                fee_sat_per_vbyte,
+                utxos.clone(),
+                prev_utxo.clone(),
+            )
             .await?;
         let signed_txs = self.tx_signer.sign_da_txs(da_txs).await?;
+
+        // Test whether signed_txs should be accepted in queue
         self.test_mempool_accept_queue_tx(&signed_txs).await?;
```
eyusufatik: wouldn't this already reject by min relay fee not met? what's the point of calling `validate_txs_fee_rate` afterwards?

crates/bitcoin-da/src/fee.rs
```diff
@@ -213,6 +215,62 @@ pub(crate) async fn get_fee_rate_from_mempool_space(
...
+                utxo_map
+                    .get(&(input.previous_output.txid, input.previous_output.vout))
+                    .cloned()
+            })
+            .sum();
+        let output_amount = commit_tx.output.iter().map(|tx| tx.value).sum();
+
+        if (input_amount - output_amount) < Amount::from_sat(commit_tx.vsize() as u64) {
+            return Err(BitcoinServiceError::FeeCalculation(fee_rate));
+        }
```
eyusufatik, line 251: imo put some kind of min relay fee multiplier here, it's not always 1 sat/vbyte, right?

crates/bitcoin-da/src/fee.rs
```diff
@@ -213,6 +215,62 @@ pub(crate) async fn get_fee_rate_from_mempool_space(
...
+            .flat_map(|input| {
+                utxo_map
+                    .get(&(input.previous_output.txid, input.previous_output.vout))
+                    .cloned()
+            })
+            .sum();
+        let output_amount = commit_tx.output.iter().map(|tx| tx.value).sum();
+
+        if (input_amount - output_amount) < Amount::from_sat(commit_tx.vsize() as u64) {
+            return Err(BitcoinServiceError::FeeCalculation(fee_rate));
```
eyusufatik, line 250: I think the commit_tx is not signed at this point, it's signed at the end when we are actually sending the txs, no?

eyusufatik (changes requested):

crates/bitcoin-da/src/service.rs
jfldde: This fixes the issue where mempool policy limit checks are done first, so it passes `test_mempool_accept_queue_tx` as a mempool policy limit failure that should be queued up even if part of the package is not meeting min relay fee. 
This is a safeguard for this odd case where the package tested is both failing mempool policy limit and min-relay-fee at the same time.

crates/bitcoin-da/src/fee.rs
jfldde: We're doing this check post-signing

crates/bitcoin-da/src/service.rs
jfldde: I'll modify to only use this safeguard when we hit a mempool policy limit error

eyusufatik (approved):

PR #2733, opened 2025-08-27
Improve choose_utxos based on findings

rakanalh:
# Description

Based on suggested implementation in https://cantina.xyz/code/49b9e08d-4f8f-4103-b6e5-f5f43cf9faa1/findings/544#comment-ce5d3f62-4193-49cf-a1c1-0c660cc5debe, though it had it's own issues, i have somehow merged the good parts between that and the existing solution.

This new version:
1. Doesn't rely on caller to pre-filter, but also doesn't break if caller does pre-filter
2. Uses tx_id AND vout for comparison
3. Uses filter instead of HashSet difference
4. Always filters out already chosen UTXOs when selecting more

### DIFF
--- a/crates/bitcoin-da/src/helpers/builders/mod.rs
+++ b/crates/bitcoin-da/src/helpers/builders/mod.rs
@@ -10,7 +10,6 @@ mod tests;
 
 use core::fmt;
 use core::result::Result::Ok;
-use std::collections::HashSet;
 
 use anyhow::anyhow;
 use bitcoin::absolute::LockTime;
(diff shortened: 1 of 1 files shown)


eyusufatik:
I kinda hesitate on such a big change on utxo selection right before mainnet :)

crates/bitcoin-da/src/helpers/builders/mod.rs
```diff
@@ -420,58 +412,85 @@ fn choose_utxos(
...
+    // Check if we already have enough with just the required UTXO
     if sum >= amount {
-        return Ok((chosen_utxos, sum, utxos.to_vec()));
+        // Filter out the chosen UTXOs (just the required one) from the original list
+        let leftovers = utxos
+            .iter()
+            .filter(|u| {
+                !chosen_utxos
+                    .iter()
+                    .any(|chosen| chosen.tx_id == u.tx_id && chosen.vout == u.vout)
```
jfldde, line 429: A Vec might not be the right structure if we need to filter/any so much, it has to be unique by (txid, vout), maybe we can do something with that

eyusufatik:
I'm closing this pr because I don't want to make changes to an important part of the repo at this stage

PR #2736, opened 2025-08-27
Handle SkippapleError in fullnode pending proofs

### LINKED ISSUE #2735: Fullnode: Pending proof unrecoverable error is not discarded
Fails with :`Failed to process pending proof with index 2403-2403: Other(Failed to verify proof: claim digest does not match the expected digest 70106ee09fea6dfd808f2523d71489578c3d150a0b2f3da546322ce27628d7c7; expected 698f1b8829804c110f339a7f2c45f7dba75ba3689e244609dc922e84c80955f4. Skipping it...)` but is kept as pending

jfldde:
# Description

- Discard errors in `process_zk_proof`

## Linked Issues
- Fixes #2735

### DIFF
--- a/crates/fullnode/src/da_block_handler.rs
+++ b/crates/fullnode/src/da_block_handler.rs
@@ -561,8 +561,9 @@ where
         tracing::trace!("ZK proof: {:?}", proof);
 
         // Extract and verify the proof using the appropriate ZKVM
-        let batch_proof_output = Vm::extract_output::<BatchProofCircuitOutput>(&proof)
-            .map_err(|e| anyhow!("Failed to extract batch proof output from proof: {:?}", e))?;
+        let Ok(batch_proof_output) = Vm::extract_output::<BatchProofCircuitOutput>(&proof) else {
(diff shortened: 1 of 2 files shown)


crates/fullnode/src/da_block_handler.rs
```diff
@@ -577,7 +577,9 @@ where
             code_commitment,
             network_to_dev_mode(self.network),
         )
-        .map_err(|err| anyhow!("Failed to verify proof: {:?}. Skipping it...", err))?;
+        .map_err(|err| {
```
eyusufatik: should we maybe return Ok(Discarded) like we do on sequencer commitments 🤔

eyusufatik (approved):
I mean it still works with the error type ofc

crates/fullnode/src/da_block_handler.rs
jfldde: Yes, switching to this instead

ercecan (approved):

PR #2742, opened 2025-08-28
feat: Integrate Reth's Mempool maintenance task.

### LINKED ISSUE #2697: Proper mempool maintenance
## Summary
Can't really run reth's mempool maintenance task and we have our own to some extent, but we can improve it.

Plus there is a finding about not removing all mempool txs saved to ledger db, which also should be added to the maintenance.

rakanalh:
# Description

Dashboard changes should look something like this
Screenshot_20250903_191446

## Performance

The notification time and bundle extraction logic shows numbers around the following:
```
2025-09-10T11:23:01.984627Z  INFO citrea_sequencer::runner: Bundle state extraction took 0.137ms
2025-09-10T11:23:02.019260Z  INFO citrea_sequencer::runner: Canonical notification preparation took 0.305ms
```
Meaning that they will appear as 0 in the "Stacked Sequencer Block Production Flow" histogram

## Linked Issues
- Fixes #2697


bin/citrea/tests/common/client.rs
```diff
@@ -880,6 +882,64 @@ impl TestClient {
...
+            .map(|account_txs| account_txs.len())
+            .sum::<usize>();
+
+        Ok(pending_count + queued_count)
+    }
+
+    /// Check if a specific transaction is in the mempool
+    pub(crate) async fn is_transaction_in_mempool(
+        &self,
+        tx_hash: TxHash,
```
eyusufatik, line 917: how are these two not comparable

crates/sequencer/src/db_provider/mod.rs
```diff
@@ -328,30 +446,43 @@ impl BlockIdReader for DbProvider {
...
+
     fn finalized_block_number(&self) -> ProviderResult<Option<BlockNumber>> {
-        unimplemented!("finalized_block_number")
+        // Finalized blocks are those included in commitments sent to DA layer
+        match self.ledger_db.get_last_commitment() {
+            Ok(Some(commitment)) => Ok(Some(commitment.l2_end_block_number)),
+            Ok(None) => Ok(None),
+            Err(_) => Ok(None),
+        }
     }
```
eyusufatik, line 404: where is reth maintenance using this exactly? we may actually be better off just saying the last block is finailized directly

crates/sequencer/src/db_provider/mod.rs
eyusufatik, line 1: I need to review function body implementations, just went through which functions was implemented

crates/sequencer/src/runner.rs
```diff
@@ -565,6 +579,74 @@ where
...
+            self.db_provider.cfg().base_fee_params,
+        );
+
+        self.mempool.set_block_info(BlockInfo {
+            block_gas_limit: latest_header.gas_limit,
+            last_seen_block_number: l2_height,
+            last_seen_block_hash: B256::from_slice(&self.l2_block_hash),
+            pending_basefee: next_base_fee,
+            pending_blob_fee: None,
+        });
```
eyusufatik: I wonder if this should be done? Why are we sending the chain update seperately?

I guess I don't know how reth expects the mempool maintenance task to be used. Can you eloborate?

crates/sequencer/src/runner.rs
```diff
@@ -769,23 +851,115 @@ where
...
+        for (cache_key, cache_value) in state_log.iter_ordered_writes() {
+            if cache_key.key.starts_with(b"E/i/") {
+                if let Some(value) = cache_value {
+                    let encoded_addr = &cache_key.key[4..];
+                    if let Ok(addr_bytes) = borsh::from_slice::<[u8; 20]>(encoded_addr) {
+                        let address = Address::from_slice(&addr_bytes);
+                        if let Ok(account_id) = borsh::from_slice::<u64>(&value.value) {
+                            account_id_to_address.insert(account_id, address);
+                        }
+                    }
```
eyusufatik: i think we can panic on borsh::from_slice returning Err here. because that means there is some huge inconsistency in the system

crates/sequencer/src/runner.rs
```diff
@@ -769,23 +851,115 @@ where
...
+            }
+        }
+
+        for (account_id, account_info) in account_id_to_info {
+            if let Some(&address) = account_id_to_address.get(&account_id) {
+                let revm_info = ReVmAccountInfo {
+                    balance: account_info.balance,
+                    nonce: account_info.nonce,
+                    code_hash: account_info.code_hash.unwrap_or_else(|| keccak256([])),
+                    code: None,
```
eyusufatik, line 938: if code_hash is Some() you can pull the code from offchain_code

crates/sequencer/src/runner.rs
```diff
@@ -769,23 +851,115 @@ where
...
+                };
+
+                let bundle_account = BundleAccount {
+                    info: Some(revm_info),
+                    storage: Default::default(),
+                    original_info: None,
```
eyusufatik, line 950: I guess original_info is prev_state, should we supply it?

crates/sequencer/src/runner.rs
```diff
@@ -769,23 +851,115 @@ where
...
+                    status: Default::default(),
```
eyusufatik: what is status field?

crates/sequencer/src/runner.rs
```diff
@@ -769,23 +851,115 @@ where
...
+        let block = self
+            .db_provider
+            .block_by_id(BlockId::Number(BlockNumberOrTag::Number(l2_height)))?
+            .ok_or(anyhow!("Block {} must exist after saving", l2_height))?;
+
+        let sealed_block = SealedBlock::new_unchecked(block, block_hash);
+        let recovered_block = RecoveredBlock::new_sealed(sealed_block, senders);
+
+        let execution_outcome =
+            ExecutionOutcome::new(bundle_state, vec![receipts], l2_height, vec![]);
```
eyusufatik, line 1021: what's the last arg of vec here?

crates/sequencer/src/runner.rs
```diff
@@ -769,23 +851,115 @@ where
...
 
-        // Remove processed/failed transactions from mempool
-        self.mempool.remove_transactions(txs_to_remove.clone());
-
-        // Update account states in mempool
-        let account_updates = self.get_account_updates()?;
+        // Remove L1 fee failed transactions from the mempool
+        // These are not handled by the maintenance task
+        if !l1_fee_failed_txs.is_empty() {
+            self.mempool.remove_transactions(l1_fee_failed_txs.clone());
```
eyusufatik, line 1036: I wonder if we can get rid of the metrics we handle, as reth also has mempool metrics

bin/citrea/tests/mock/mempool_maintenance.rs
eyusufatik, line 1: Also need to review test in my second review

eyusufatik (changes requested):

crates/sequencer/src/db_provider/mod.rs
rakanalh: This is called inside the task itself

Initially here and keeps updating 
https://github.com/paradigmxyz/reth/blob/4a28cf42811158ab965e5766efab3981227da6f8/crates/transaction-pool/src/maintain.rs#L161-L163

I will update to use latest l2 block.

crates/sequencer/src/runner.rs
rakanalh: You're probably right. After checking, the `Commit` notification sets the block info. I thought the process was separate since `set_block_info` sets the metadata for validation and the notification basically helps the task to cleanup. But it's redundant here.

crates/sequencer/src/runner.rs
rakanalh: Since the maintenance task is only interested in ChangedAccount

https://github.com/paradigmxyz/reth/blob/4a28cf42811158ab965e5766efab3981227da6f8/crates/evm/execution-types/src/execution_outcome.rs#L24-L31

No need to provide the code.

bin/citrea/tests/common/client.rs
rakanalh: Fixed

crates/sequencer/src/runner.rs
rakanalh: Apparently, it's only used in places (Reth RPC endpoints) not related to pool maintenance task.

crates/sequencer/src/runner.rs
rakanalh: Should have set that... it's done.

crates/sequencer/src/runner.rs
rakanalh: It's EIP-7685 requests, i believe we don't support this yet.

crates/sequencer/src/runner.rs
ercecan: Yeah they should be seen now

crates/sequencer/src/runner.rs
```diff
@@ -525,7 +536,7 @@ where
         let (signed_txs, blobs) = self.encode_and_sign_evm_txs_into_sov_txs(
             &mut working_set,
             &l2_block_info,
-            txs_to_run,
+            txs_to_run.clone(),
```
ercecan, line 555: why clone here?

crates/sequencer/src/lib.rs
```diff
@@ -127,10 +129,28 @@ where
...
+    {
+        let pool = mempool.inner_pool().clone();
+        let client = db_provider.clone();
+        let events_stream = UnboundedReceiverStream::new(canon_state_rx);
+        let maintenance_future = reth_transaction_pool::maintain::maintain_transaction_pool_future(
+            client,
+            pool,
+            events_stream,
+            task_executor.clone(),
+            Default::default(), // Use default MaintainPoolConfig
```
ercecan: I believe MaintainPoolConfig should be modifiable from env variables and config just like mempool config

bin/citrea/tests/mock/mempool_maintenance.rs
```diff
@@ -0,0 +1,638 @@
+use std::path::PathBuf;
```
ercecan, line 1: After adding maintain pool config, can you write a test with a low value for max_tx_lifetime and test if stale txs are actually evicted? Doing this with e2e framework might be easier, but you will also need to add the MaintainPool config to e2e repo

ercecan (changes requested):

crates/sequencer/src/runner.rs
rakanalh: I have replaced our own mempool metrics with Reth's

crates/common/src/config/mod.rs
```diff
@@ -314,6 +320,9 @@ impl Default for SequencerMempoolConfig {
             base_fee_tx_limit: 100000,
             base_fee_tx_size: 200,
             max_account_slots: 16,
+            max_update_depth: None,
+            max_reload_accounts: None,
+            max_tx_lifetime_secs: None,
```
ercecan: Why are these None? I believe we should use reth's default values here

bin/citrea/tests/mock/mempool_maintenance.rs
rakanalh: Added

rakanalh: https://github.com/chainwayxyz/citrea-e2e/pull/134

crates/sequencer/src/runner.rs
```diff
@@ -769,23 +839,130 @@ where
...
+                        .expect("Failed to borsh deserialize account ID");
+                    account_id_to_address.insert(account_id, address);
+                }
+            } else if cache_key.key.starts_with(b"E/a/") && cache_value.is_some() {
+                let encoded_id = &cache_key.key[4..];
+                let account_id =
+                    borsh::from_slice::<u64>(encoded_id).expect("Failed to parse account ID");
+                // Account was read from state, so it's Loaded
+                account_id_to_status.insert(account_id, AccountStatus::Loaded);
+            }
```
eyusufatik: I believe no need to chaeck for accounts from the reads, only indices should suffice

crates/sequencer/src/db_provider/mod.rs
```diff
@@ -420,16 +559,47 @@ impl TransactionsProvider for DbProvider {
...
+        let mut accessory_state = working_set.accessory_state();
+
+        // Get the transactions in the range
+        let transactions = self
+            .evm
+            .get_block_transactions(start, end, &mut accessory_state);
+
+        // Extract senders
+        let senders: Vec<Address> = transactions.iter().map(|tx| tx.signer).collect();
+
```
eyusufatik: this will be slow, we can keep track of senders when we are including txs into the block i believe

crates/sequencer/src/runner.rs
```diff
@@ -769,23 +839,130 @@ where
...
+        block_hash: alloy_primitives::B256,
+        senders: Vec<alloy_primitives::Address>,
+        receipts: Vec<Receipt>,
+        bundle_state: BundleState,
+    ) -> anyhow::Result<Chain> {
+        let block = self
+            .db_provider
+            .block_by_id(BlockId::Number(BlockNumberOrTag::Number(l2_height)))?
+            .ok_or(anyhow!("Block {} must exist after saving", l2_height))?;
+
```
eyusufatik: you've read this block multiple times, if you can reuse the same working_set all through the notification creation, you'll speed up things quite a lot

bin/citrea/tests/bitcoin/batch_prover_test.rs
```diff
@@ -1366,6 +1366,9 @@ impl TestCase for InvokeCachePruningTest {
                 base_fee_tx_limit: 1_000_000,
                 base_fee_tx_size: 100_000_000,
                 max_account_slots: 1_000_000,
+                max_update_depth: None,
+                max_reload_accounts: None,
+                max_tx_lifetime_secs: None,
```
jfldde:
```suggestion
                ..Default::default()
```

crates/common/src/config/mod.rs
```diff
@@ -328,6 +337,15 @@ impl FromEnv for SequencerMempoolConfig {
             base_fee_tx_limit: read_env("BASE_FEE_TX_LIMIT")?.parse()?,
             base_fee_tx_size: read_env("BASE_FEE_TX_SIZE")?.parse()?,
             max_account_slots: read_env("MAX_ACCOUNT_SLOTS")?.parse()?,
+            max_update_depth: std::env::var("MEMPOOL_MAX_UPDATE_DEPTH")
```
jfldde: Maybe prefix it as SEQUENCER_MEMPOOL so it's not confusing with da mempool.

crates/sequencer/src/db_provider/mod.rs
```diff
@@ -173,33 +189,123 @@ impl BlockBodyIndicesProvider for DbProvider {
...
+                BlockNumberOrTag::Number(n) => n,
+                BlockNumberOrTag::Latest => self.best_block_number()?,
+                BlockNumberOrTag::Finalized => match self.finalized_block_number()? {
+                    Some(n) => n,
+                    None => return Ok(None),
+                },
+                BlockNumberOrTag::Safe => match self.finalized_block_number()? {
+                    Some(n) => n,
+                    None => return Ok(None),
+                },
```
jfldde:
```suggestion
                BlockNumberOrTag::Finalized | BlockNumberOrTag::Safe => match self.finalized_block_number()? {
                    Some(n) => n,
                    None => return Ok(None),
                },
```
Makes it explicit that we use the same logic for Finalized and Safe

crates/sequencer/src/db_provider/mod.rs
```diff
@@ -173,33 +189,123 @@ impl BlockBodyIndicesProvider for DbProvider {
...
+                ) {
+                    Ok(Some(block)) => Ok(Some(block.inner.header.into_consensus())),
+                    Ok(None) => Ok(None),
+                    Err(_) => Ok(None),
+                }
+            }
+            BlockNumberOrTag::Latest | BlockNumberOrTag::Safe | BlockNumberOrTag::Finalized => {
+                let mut working_set = WorkingSet::new(self.storage.clone());
+                match self.evm.get_block_by_number(
+                    Some(BlockNumberOrTag::Latest),
```
jfldde: Why do we use `Latest` for `Finalized` and `Safe` here ?

crates/sequencer/src/db_provider/mod.rs
jfldde: Agree, we can remove the hoops of re-fetching block and going through `senders_by_tx_range` as we can get senders from `dry_run_transactions`

crates/sequencer/src/db_provider/mod.rs
```diff
@@ -173,33 +189,123 @@ impl BlockBodyIndicesProvider for DbProvider {
...
+        let citrea_block = match self
+            .evm
+            .get_block_by_height(block_number, &mut accessory_state)
+        {
+            Some(block) => block,
+            None => return Ok(None),
+        };
+
+        let transactions = self
+            .transactions_by_block(BlockHashOrNumber::Number(block_number))?
```
jfldde: We already have the `SealedBlock` here, we should use `transactions_by_block_range`

crates/sequencer/src/db_provider/mod.rs
```diff
@@ -441,33 +611,77 @@ impl TransactionsProvider for DbProvider {
...
+
     fn transaction_id(&self, _tx_hash: TxHash) -> ProviderResult<Option<TxNumber>> {
         unimplemented!("transaction_id")
     }
+
     fn transaction_sender(&self, _id: TxNumber) -> ProviderResult<Option<Address>> {
         unimplemented!("transaction_sender")
     }
+
     fn transactions_by_block(
```
jfldde, line 537: We should use `transactions_by_block_range` as we have the `SealedBlock` at the unique call site

crates/evm/src/hooks.rs
```diff
@@ -183,7 +183,9 @@ impl<C: sov_modules_api::Context> Evm<C> {
         self.should_be_end_of_sys_txs = false;
 
         #[cfg(not(feature = "native"))]
-        pending_transactions.clear();
+        {
+            pending_transactions.clear();
+        }
```
eyusufatik: revert :D

crates/evm/src/lib.rs
```diff
@@ -1,32 +1,39 @@
...
+    create_initial_system_events, populate_deposit_system_events, populate_set_block_info_event,
+};
+#[cfg(feature = "native")]
+pub use crate::metrics::EVM_METRICS;
+pub use crate::system_events::SYSTEM_SIGNER;
+pub use crate::EvmConfig;
+
+mod call;
+use evm::db::EvmDb;
+mod evm;
```
eyusufatik, line 4: revert

crates/evm/src/lib.rs
```diff
@@ -202,6 +196,59 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
         EvmDb::new(self, working_set)
     }
+
+    /// Get transactions for a block by transaction index range
+    #[cfg(feature = "native")]
+    pub fn get_block_transactions(
+        &self,
+        start_idx: u64,
+        end_idx: u64,
+        accessory_state: &mut sov_modules_api::AccessoryWorkingSet<C::Storage>,
```
eyusufatik: use workingset then convert to accessory inside, that's what we do everywhere. it's a cheap operation

crates/citrea-stf/Cargo.toml
```diff
@@ -16,6 +16,7 @@ borsh = { workspace = true }
 # will be needed after we update the storage system
 # tokio = { workspace = true, features = ["sync"], optional = true }
 jsonrpsee = { workspace = true, features = ["http-client", "server"], optional = true }
+reth-primitives = { workspace = true }
```
eyusufatik: can be removed now

crates/sovereign-sdk/module-system/sov-modules-api/Cargo.toml
eyusufatik, line 1: revert

guests/risc0/batch-proof/bitcoin/Cargo.lock
eyusufatik, line 1: these cargo locks need to be reverted

Cargo.toml
```diff
@@ -83,10 +83,10 @@ k256 = { version = "0.13.4" }
 lru = "0.13"
 hex = { version = "0.4.3", default-features = false, features = ["alloc", "serde"] }
 log-panics = { version = "2", features = ["with-backtrace"] }
-metrics = "0.23.0"
+metrics = "0.24.0"
 metrics-derive = "0.1.0"
-metrics-exporter-prometheus = "0.15.3"
-metrics-util = "0.17.0"
+metrics-exporter-prometheus = "0.17.2"
+metrics-util = "0.20.0"
```
eyusufatik: these upgrades should be done in another pr

crates/sequencer/src/db_provider/mod.rs
```diff
@@ -304,21 +350,26 @@ impl BlockNumReader for DbProvider {
     fn best_block_number(&self) -> ProviderResult<BlockNumber> {
         unimplemented!("best_block_number")
     }
+
```
eyusufatik, line 352: revert these formatting changes

crates/sequencer/src/lib.rs
```diff
@@ -127,10 +129,42 @@ where
...
+        // Build MaintainPoolConfig from mempool config
+        let mut maintain_config = reth_transaction_pool::maintain::MaintainPoolConfig::default();
+        if let Some(max_update_depth) = sequencer_config.mempool_conf.max_update_depth {
+            maintain_config.max_update_depth = max_update_depth;
+        }
+        if let Some(max_reload_accounts) = sequencer_config.mempool_conf.max_reload_accounts {
+            maintain_config.max_reload_accounts = max_reload_accounts;
+        }
+        if let Some(max_tx_lifetime_secs) = sequencer_config.mempool_conf.max_tx_lifetime_secs {
+            maintain_config.max_tx_lifetime = std::time::Duration::from_secs(max_tx_lifetime_secs);
```
eyusufatik: I don't like how different this config set happens from the others btw, we may need to change this

crates/sequencer/src/runner.rs
```diff
@@ -779,23 +873,166 @@ where
...
+                bundle_state.state.insert(address, bundle_account);
+            }
+        }
+
+        bundle_state
+    }
+
+    /// Build block data from in-memory transactions without DB reads
+    fn build_reth_block_data(
```
eyusufatik, line 962: couldn't the operations done here be done like the receipts?

crates/sequencer/src/runner.rs
```diff
@@ -575,6 +612,36 @@ where
...
+
+        // Use the transactions we already have from dry_run (txs_to_run)
+        // and build block structure without DB reads
+        let (reth_block, reth_receipts) =
+            self.build_reth_block_data(l2_height, &txs_to_run, &senders, &receipts)?;
+
+        // Create the Chain notification with the produced block data
+        if let Ok(chain) = self.create_chain_notification(
+            l2_height,
+            B256::from_slice(&self.l2_block_hash),
```
eyusufatik: this actually expects evm block hash, not l2 block hash

crates/sequencer/src/runner.rs
```diff
@@ -779,23 +873,166 @@ where
...
+    /// Creates a Chain notification from the produced L2 block
+    fn create_chain_notification(
+        &self,
+        l2_height: u64,
+        block_hash: alloy_primitives::B256,
+        block: reth_primitives::Block,
+        senders: Vec<alloy_primitives::Address>,
+        receipts: Vec<Receipt>,
+        bundle_state: BundleState,
+    ) -> anyhow::Result<Chain> {
```
eyusufatik: doesn't need to be `Result<T>`

crates/sequencer/src/runner.rs
```diff
@@ -779,23 +873,166 @@ where
...
-        // Remove processed/failed transactions from mempool
-        self.mempool.remove_transactions(txs_to_remove.clone());
-
-        // Update account states in mempool
-        let account_updates = self.get_account_updates()?;
+        // Remove L1 fee failed transactions from the mempool
+        // These are not handled by the maintenance task
+        if !l1_fee_failed_txs.is_empty() {
+            self.mempool.remove_transactions(l1_fee_failed_txs.clone());
+        }
```
eyusufatik, line 1038: should we get rid of the below lines? I mean the successful tx removal?

eyusufatik (changes requested):

crates/sequencer/src/runner.rs
rakanalh: Lines below that:
```
        // Clean up persistent storage for both included and failed transactions
        let mut txs_to_remove = self.db_provider.last_block_tx_hashes()?;
... etc
```
Remove failed transactions not the successful ones... and they get removed from ledger not the mempool

crates/sequencer/src/db_provider/mod.rs
rakanalh: Leaving a new line between definitions? Why? There are some with a space in between and some that don't which is inconsistent. IMO, we should keep them.

eyusufatik:
performance approved on devnet

rakanalh:
@eyusufatik @jfldde 

Regarding the cache pruning test:
```
2025-09-22T18:07:27.889060Z  INFO L2Syncer: citrea_common::l2: New State Root after l2 block #55 is: 0x5179deb8616c156f475cfad892c3db77c1bbb2d7cfbb6091f5318eac60e96add
2025-09-22T18:07:34.021310Z  INFO L1Syncer: citrea_batch_prover::l1_syncer: Processed L1 block 172
2025-09-22T18:07:34.022344Z  INFO L1Syncer: citrea_batch_prover::l1_syncer: Processed L1 block 173
2025-09-22T18:07:34.023281Z  INFO L1Syncer: citrea_batch_prover::l1_syncer: Processed L1 block 174
2025-09-22T18:07:34.024255Z  INFO L1Syncer: citrea_batch_prover::l1_syncer: Processed L1 block 175
2025-09-22T18:07:34.026869Z  INFO L1Syncer: citrea_batch_prover::l1_syncer: Got commitment with index 1 in L1 block 176
2025-09-22T18:07:34.027021Z  INFO L1Syncer: citrea_batch_prover::l1_syncer: Processed L1 block 176
2025-09-22T18:07:34.027153Z  INFO BatchProver: citrea_batch_prover::prover: Got 1 pending commitment(s)
2025-09-22T18:07:34.027229Z  INFO BatchProver: citrea_batch_prover::prover: Got 1 synced commitment(s)
2025-09-22T18:07:34.027242Z  INFO BatchProver: citrea_batch_prover::prover: Got 1 provable commitment(s)
2025-09-22T18:07:34.037396Z  INFO BatchProver: citrea_batch_prover::partition: Adding commitment partition: indices=[1,1] blocks=[1,55] reason=Finish
2025-09-22T18:07:34.037494Z  INFO BatchProver: citrea_batch_prover::prover: Partitioned commitments into 1 parts
2025-09-22T18:07:34.042678Z  INFO BatchProver:create_circuit_input{job_id="0199729c-0e95-7f12-829f-ce7bf23780a8"}: citrea_batch_prover::prover: Processing commitment 1 with 55 L2 blocks
2025-09-22T18:07:37.574618Z  INFO BatchProver:create_circuit_input{job_id="0199729c-0e95-7f12-829f-ce7bf23780a8"}: citrea_batch_prover::prover: Cache pruning triggered at L2 block 42
```
Even in failing runs, the cache is pruned consistently at L2 block 42.

Instead of creating 60 blocks, where we know that the state diff based commit is going to happen at block 55 by the sequencer, we create 55 blocks and we have to make sure that the batch prover syncs up to 55 and proves the last L2 block which is 55 in this case.

This should make the test consistently pass.

resources/grafana/sequencer.dashboard.json
```diff
@@ -1085,6 +1085,40 @@
...
+            "uid": "${DS_PROMETHEUS}"
+          },
+          "disableTextWrap": false,
+          "editorMode": "builder",
+          "expr": "sequencer_mempool_extract_bundle_state_time",
+          "fullMetaSearch": false,
+          "hide": false,
+          "includeNullMetadata": true,
+          "instant": false,
+          "legendFormat": "sequencer_mempool_extract_bundle_state_time",
```
eyusufatik: can you sue __auto like the other items?

crates/sequencer/src/db_provider/mod.rs
```diff
@@ -160,10 +161,23 @@ impl OmmersProvider for DbProvider {
...
+        // Get the block from EVM
+        let block = match self.evm.get_block_by_height(num, &mut working_set) {
+            Some(block) => block,
+            None => return Ok(None),
+        };
+
+        // Create StoredBlockBodyIndices from the block's transaction range
+        Ok(Some(reth_db::models::StoredBlockBodyIndices {
+            first_tx_num: block.transactions.start,
+            tx_count: block.transactions.end - block.transactions.start,
```
eyusufatik, line 177: are we sure about the end - start here

should it be end - start + 1 maybe?

gotta check

crates/evm/src/lib.rs
```diff
@@ -202,6 +196,44 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+        working_set: &mut WorkingSet<C::Storage>,
+    ) -> Vec<CitreaReceiptWithBloom> {
+        let mut accessory_state = working_set.accessory_state();
+        let mut receipts = Vec::new();
+        for idx in start_idx..end_idx {
+            if let Some(receipt) = self.receipts.get(idx as usize, &mut accessory_state) {
+                receipts.push(receipt);
+            }
+        }
+        receipts
```
eyusufatik, line 223: pls double check there is not +-1 mistake here

crates/common/src/config/mod.rs
```diff
@@ -284,6 +284,72 @@ impl FromEnv for SequencerConfig {
...
+    #[serde(default = "default_max_reload_accounts")]
+    pub max_reload_accounts: usize,
+    /// Maximum lifetime for non-executable transactions in seconds
+    #[serde(default = "default_max_tx_lifetime_secs")]
+    pub max_tx_lifetime_secs: u64,
+}
+
+const fn default_max_update_depth() -> u64 {
+    64
+}
```
eyusufatik: we don't need to support this, always supply default to reth, don't parse from env or expect this in the config.tomls

bin/citrea/tests/mock/mempool_maintenance.rs
```diff
@@ -0,0 +1,791 @@
...
+        "Mempool should have 1 transaction (nonce 4) after nonce 2 was published"
+    );
+
+    seq_task.graceful_shutdown();
+}
+
+/// Test that transactions with insufficient L1 fees are removed after failed inclusion
+///
+/// This test verifies that Reth's mempool maintenance handles transactions
+/// that fail during block production due to L1 fee validation or other execution errors.
```
eyusufatik: but we handle that, right? not reth-mempool

eyusufatik (changes requested):

crates/sequencer/src/db_provider/mod.rs
rakanalh: Reth expects this as it reconstructs the range as `first_tx_num..first_tx_num + tx_count`, where tx_count should be the difference without adding 1.
`block.transactions` is a `Range<u64>` 
Example: Range 10..15 are transactions 10, 11, 12, 13, 14 (5 total)
    - end - start = 15 - 10 = 5

eyusufatik: and we are also doing the same in our evm crate?

crates/evm/src/lib.rs
```diff
@@ -202,6 +203,44 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
+    }
+
+    /// Get the current head block from working set
+    #[cfg(feature = "native")]
+    pub fn get_head_block(
+        &self,
+        working_set: &mut WorkingSet<C::Storage>,
+    ) -> Option<Block<AlloyHeader>> {
+        self.head.get(working_set)
+    }
```
eyusufatik, line 244: no need for these two functions

crates/evm/src/lib.rs
```diff
@@ -48,11 +48,12 @@ use alloy_consensus::Header as AlloyHeader;
 use alloy_primitives::{Address, TxHash, B256};
 use evm::db::EvmDb;
 use sov_modules_api::{L2BlockModuleCallError, ModuleInfo, SpecId as CitreaSpecId, WorkingSet};
+#[cfg(feature = "native")]
+use sov_modules_api::{StateValueAccessor, StateVecAccessor};
 use sov_state::codec::{BcsCodec, RlpCodec};
 
+use crate::evm::primitive_types::Block;
 #[cfg(feature = "native")]
```
eyusufatik, line 55: seems like the revert wasn't done correctly

eyusufatik (changes requested):

crates/sequencer/src/db_provider/mod.rs
rakanalh: Yes... i double checked that as well.

crates/common/src/config/mod.rs
```diff
@@ -284,6 +284,60 @@ impl FromEnv for SequencerConfig {
...
+                .and_then(|v| v.parse().ok())
+                .unwrap_or_else(default_max_tx_lifetime_secs),
+        })
+    }
+}
+
+impl From<MempoolMaintenanceConfig> for reth_transaction_pool::maintain::MaintainPoolConfig {
+    fn from(config: MempoolMaintenanceConfig) -> Self {
+        Self {
+            max_update_depth: Default::default(),
```
ercecan, line 334: Why is max update depth not configurable?

rakanalh: @eyusufatik said to remove it.

eyusufatik: no reorgs

ercecan (approved):

eyusufatik (approved):

crates/sequencer/src/runner.rs
```diff
@@ -790,23 +881,164 @@ where
...
+            }
+        }
+
+        bundle_state
+    }
+
+    /// Build block data from in-memory transactions without DB reads
+    fn build_reth_block_data(
+        &self,
+        l2_height: u64,
```
jfldde, line 964: Couldn't we pass the block itself instead of l2_height and re-fetching ?

rakanalh:
```rust
        let citrea_block = self
            .db_provider
            .evm
            .get_block_by_height(l2_height, &mut working_set)
            .unwrap_or_else(|| panic!("Block {} must exist after saving", l2_height));
```
This actually returns SealedBlock which `l2_block` doesn't have.  The rest is to construct the reth types

eyusufatik: another optimization is to read it from the working_set, just like we did for receipts.

how about that?

eyusufatik: I think I meant the same thing here: https://github.com/chainwayxyz/citrea/pull/2742#discussion_r2355370223

rakanalh: Will implement this in a separate PR, issue: #2907

PR #2745, opened 2025-08-31
fix: broken reference link in filter.rs

sashass1315:
https://github.com/paradigmxyz/reth/blob/main/crates/rpc/rpc-types/src/eth/filter.rs - 404
https://reth.rs/docs/src/reth_rpc/eth/filter.rs - correct link

### DIFF
--- a/crates/evm/src/rpc_helpers/filter.rs
+++ b/crates/evm/src/rpc_helpers/filter.rs
@@ -1,4 +1,4 @@
-// https://github.com/paradigmxyz/reth/blob/main/crates/rpc/rpc-types/src/eth/filter.rs
+// https://reth.rs/docs/src/reth_rpc/eth/filter.rs
 
 use std::env;
 use std::iter::StepBy;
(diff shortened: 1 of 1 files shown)


kpp (dismissed):

crates/evm/src/rpc_helpers/filter.rs
```diff
@@ -1,4 +1,4 @@
-// https://github.com/paradigmxyz/reth/blob/main/crates/rpc/rpc-types/src/eth/filter.rs
+// https://reth.rs/docs/src/reth_rpc/eth/filter.rs
```
eyusufatik: pls put in github link

eyusufatik (changes requested):

crates/evm/src/rpc_helpers/filter.rs
sashass1315: > pls put in github link

done

kpp (approved):

PR #2752, opened 2025-09-01
fix: ensure database backups are created atomically under L1/L2 locks

VolodymyrBg:
Previously, L1/L2 processing locks were released immediately after spawning backup tasks, allowing block processing to resume before all database snapshots were actually created. This could result in inconsistent backup state across databases.

Changes:
- Hold L1/L2 processing locks until all database backup operations complete
- Use futures::future::join_all to wait for all spawn_blocking tasks atomically
- Update misleading comment to accurately reflect the synchronization behavior
- Maintain parallel backup execution while ensuring atomic snapshot timing

This ensures all databases have consistent snapshots taken at the same point in time relative to L1/L2 block processing, eliminating potential backup inconsistencies under high load or aggressive RocksDB background operations.

### DIFF
--- a/crates/common/src/backup/manager.rs
+++ b/crates/common/src/backup/manager.rs
@@ -4,6 +4,7 @@ use std::sync::{Arc, RwLock};
 use std::time::{Instant, SystemTime, UNIX_EPOCH};
 
 use anyhow::{bail, ensure, Context};
+use futures::future;
 use rocksdb::backup::BackupEngineInfo;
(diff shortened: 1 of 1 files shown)


jfldde:
RocksDB creates a snapshot while under lock and then backup from the snapshot. No need to hold locks longer than that. Moreover, holding the lock would stop block production and this is not acceptable as it's a farily long process

jfldde:
Closing as not pertinent

PR #2760, opened 2025-09-01
fix: Hide key from rpc request for backup endpoints

### LINKED ISSUE #2731: dont log rpc api key
## Summary
in the rpc logger, we log the whole request, which will also log the api key used by full nodes for protected methods.

it shouldn't be a problem normally, but it's best practice not to log these type of values

kpp:
Fixes https://github.com/chainwayxyz/citrea/issues/2731

I tested this fn locally with an rpc parsed from string. It worked. I didn't launch a node with DEBUG to inspect logs though.

### DIFF
--- a/crates/common/src/rpc/auth.rs
+++ b/crates/common/src/rpc/auth.rs
@@ -1,8 +1,11 @@
+use std::borrow::Cow;
+
 use futures::future::BoxFuture;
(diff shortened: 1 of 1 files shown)


eyusufatik (approved):
tested locally

PR #2765, opened 2025-09-02
fix: Notify subscribers about missing L2 blocks

### LINKED ISSUE #2758: send websocket notifications on missed da blocks
## Summary
cantina 227

kpp:
Fixes https://github.com/chainwayxyz/citrea/issues/2758

### DIFF
--- a/crates/sequencer/src/runner.rs
+++ b/crates/sequencer/src/runner.rs
@@ -399,21 +399,19 @@ where
         }
     }
 
-    /// Produces a new L2 block with the given DA blocks
+    /// Produces a new L2 block with the given DA blocks.
+    /// Notifies subscribers of the new block height.
     ///
     /// # Arguments
     /// * `da_blocks` - Data availability blocks to process
     /// * `l1_fee_rate` - Current L1 fee rate
     /// * `last_used_l1_height` - Last processed L1 block height
-    ///
-    /// # Returns
-    /// The height of the produced L2 block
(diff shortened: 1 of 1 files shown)


rakanalh (commented):
Sorry, i don't get how this notifies about missing L2 blocks.

rakanalh (approved):

crates/sequencer/src/runner.rs
```diff
@@ -436,7 +434,17 @@ where
...
+        match result {
+            Ok(l2_height) => {
+                // Only errors when there are no receivers
+                let _ = self.l2_block_tx.send(l2_height);
+            }
+            Err(e) => {
+                error!("Sequencer error: {}", e);
+            }
+        }
+
```
eyusufatik: should we have the `l2_block_tx.send()` before the metrics are saved maybe? @ercecan

ercecan: I guess it makes sense, maybe we can also separately time the duration of sending process just in case, and add it to the very top of the stacked block production graph in our grafana dashboard

eyusufatik: let's include the sending, but seperate tracking is overkill

kpp: Done

ercecan (approved):

eyusufatik (approved):

PR #2768, opened 2025-09-02
feat: Limit compressed aggregate size while collecting chunks

### LINKED ISSUE #2756: Limit size of compressed data while aggregating chunks
## Summary

cantina 273

kpp:
Fixes https://github.com/chainwayxyz/citrea/issues/2756

### DIFF
--- a/crates/bitcoin-da/src/service.rs
+++ b/crates/bitcoin-da/src/service.rs
@@ -1120,12 +1120,12 @@ impl DaService for BitcoinService {
                             continue 'aggregate;
                         };
 
-                        body.extend(chunk);
-
-                        if body.len() > MAX_COMPRESSED_BLOB_SIZE {
+                        if chunk.len() + body.len() > MAX_COMPRESSED_BLOB_SIZE {
                             warn!("{tx_id}: Compressed aggregate too large");
                             continue 'aggregate;
                         }
+
+                        body.extend(chunk);
                     }
                     ParsedTransaction::Complete(_)
                     | ParsedTransaction::Aggregate(_)
--- a/crates/bitcoin-da/src/service.rs
+++ b/crates/bitcoin-da/src/service.rs
@@ -25,7 +25,7 @@ use bitcoincore_rpc::{Client, Error as BitcoinError, Error, RpcApi, RpcError};
 use borsh::BorshDeserialize;
 use citrea_common::utils::read_env;
 use citrea_primitives::compression::{compress_blob, decompress_blob};
-use citrea_primitives::MAX_TX_BODY_SIZE;
+use citrea_primitives::{MAX_COMPRESSED_BLOB_SIZE, MAX_TX_BODY_SIZE};
 use lru::LruCache;
 use reth_tasks::shutdown::GracefulShutdown;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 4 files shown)


crates/bitcoin-da/src/service.rs
```diff
@@ -1000,6 +1000,11 @@ impl DaService for BitcoinService {
                                 continue;
                             };
 
+                            if compressed_zk_proof.len() > MAX_COMPRESSED_BLOB_SIZE {
```
rakanalh: Can we guarantee that our compressed proof size is always less than this?

crates/primitives/src/constants.rs
```diff
@@ -48,3 +48,8 @@ pub const PRE_TANGERINE_BRIDGE_INITIALIZE_PARAMS: &[u8] = &[
 pub const MAX_DECOMPRESSED_BLOB_SIZE: usize = 1024 * 1024; // 1 MB
 #[cfg(not(feature = "testing"))]
 pub const MAX_DECOMPRESSED_BLOB_SIZE: usize = 1024 * 1024 * 100; // 100 MB
+
+/// Maximum size of a compressed blob in bytes.
+/// This limit is enforced during chunk aggregation to prevent excessive memory usage
+/// and potential denial of service attacks.
+pub const MAX_COMPRESSED_BLOB_SIZE: usize = MAX_DECOMPRESSED_BLOB_SIZE;
```
rakanalh, line 55: Instead of doing this, we might as well just rename it to `MAX_BLOB_SIZE` for both compressed and decompressed

eyusufatik (changes requested):
gotta update verifier part too

crates/bitcoin-da/src/service.rs
```diff
@@ -1000,6 +1000,11 @@ impl DaService for BitcoinService {
                                 continue;
                             };
```
eyusufatik, line 1005: a compresed complete will always be <= due to 4mb bitcoin block size, no need for this check here

crates/bitcoin-da/src/service.rs
```diff
@@ -1114,7 +1119,13 @@ impl DaService for BitcoinService {
                             warn!("{tx_id}: Chunk: unexpected kind",);
                             continue 'aggregate;
                         };
+
                         body.extend(chunk);
+
+                        if body.len() > MAX_COMPRESSED_BLOB_SIZE {
+                            warn!("{tx_id}: Compressed aggregate too large");
+                            continue 'aggregate;
+                        }
```
eyusufatik, line 1126: the problem is not limiting the individual chunks, but the total of chunks from an aggregate

eyusufatik (changes requested):

crates/bitcoin-da/src/service.rs
kpp: It's the total of chunks from an aggregate

crates/bitcoin-da/src/service.rs
kpp: Yes, because of brotli: `compressed(data) <= data`. `MAX_COMPRESSED_BLOB_SIZE = MAX_DECOMPRESSED_BLOB_SIZE`, so we can't push to da anything bigger.

crates/bitcoin-da/src/service.rs
kpp: No need, I agree. But extra check is better than no check, right?

eyusufatik: no we shouldn't check, no need to

crates/primitives/src/constants.rs
kpp: What shall I do with it?

eyusufatik (approved):

eyusufatik (changes requested):
need tests

eyusufatik (approved):

PR #2771, opened 2025-09-02
fix: fullnode pending commitment/proof handling

### LINKED ISSUE #2781: pending proofs should not be read from db all at once
## Summary
cantina 720

jfldde:
# Description

- Break early in `process_pending_proofs`.
- Pending proofs are sorted by `min_index`, we can safely break on the first pending one as the subsequent will at least depend on this first one and be kept as pending a well
- No intermediate collection for pending proof/commitment

## Linked Issues
- Fullnode testnet sync resources usage
- Cantina 720 https://cantina.xyz/code/49b9e08d-4f8f-4103-b6e5-f5f43cf9faa1/findings/720
- Requires rolling back L1


ercecan (approved):

eyusufatik (changes requested):
this is a good improvement, however, I think we should stop reading all pending proofs at once tbh. the way we are doing it rn causing lots of disk usage

jfldde:
Yes, this let's us sort it by min_index but let's make it iterate in order instead and use the iterator directly.
I will make the fix on this PR

eyusufatik:
proposed feature also fixes cantina 720 btw

crates/fullnode/src/da_block_handler.rs
```diff
@@ -847,12 +845,10 @@ where
...
         current_l1_block_height: u64,
     ) -> Result<(), ProcessingError> {
-        let pending_proofs = self.ledger_db.get_pending_proofs()?;
-        if pending_proofs.is_empty() {
-            return Ok(());
-        }
+        let mut pending_proofs = self.ledger_db.get_pending_proofs()?;
 
-        for ((min_index, max_index), proof, found_in_l1_height) in pending_proofs {
+        while let Some(Ok(item)) = pending_proofs.next() {
```
kpp: There is `impl<S> Iterator for SchemaIterator<'_, S>`, so I don't see why we can't use for instead of while let some.

crates/fullnode/src/da_block_handler.rs
```diff
@@ -774,13 +774,11 @@ where
...
     ) -> Result<(), ProcessingError> {
-        let pending_commitments = self.ledger_db.get_pending_commitments()?;
-        if pending_commitments.is_empty() {
-            return Ok(());
-        }
+        let mut pending_commitments = self.ledger_db.get_pending_commitments()?;
 
         // Try to process each pending commitment in order
-        for (index, commitment, found_in_l1_height) in pending_commitments {
+        while let Some(Ok(item)) = pending_commitments.next() {
```
kpp: Try use `for`

crates/storage-ops/src/rollback/node/fullnode.rs
```diff
@@ -200,12 +201,14 @@ impl FullNodeLedgerRollback {
...
         let mut batch = SchemaBatch::new();
 
-        let pending_sequencer_commitments = self
+        let mut pending_sequencer_commitments = self
             .ledger_db
             .iter_with_direction::<PendingSequencerCommitments>(
                 Default::default(),
                 ScanDirection::Backward,
             )?;
+        pending_sequencer_commitments.seek_to_last();
```
kpp, line 210: This looks a little bit awkward. There is RawDbReverseIterator, which does seek_to_last, so I think for Forward we have to return a ForwardIterator, while Backward has to be RawDbReverseIterator inside, so we don't need to manually seek_to_last.

Also I believe their impl of forward/backward iterators is faulty because RawDbReverseIterator::seek must use seek_for_prev (and they do), but in common version they don't check for it.

jfldde: We shouldn't need to but if we don't, this doesn't enter the loop and pending commitments/proofs were never actually rolled back. This is a common pattern across the codebase btw, this is the only place missing seek_to_last

jfldde: The underlying issue should ideally be fixed. This is a footgun and easy to overlook

crates/fullnode/src/da_block_handler.rs
jfldde: This is a bit inconsistent across, we should go through and clean it up. I'll change it here but do note that this required unwrapping of the item

eyusufatik (approved):

PR #2774, opened 2025-09-02
fix: possible logs subscription panic for blocks over 1k logs

### LINKED ISSUE #2755: `l2_block_event_handler` shouldn't be limited by MAX_LOGS_PER_RESPONSE
## Summary
as it's the notifier for all log subscriptions, it shouldn't be limited by the 1000 logs limit. we should let it extract all logs, then the consumer of the logs will filter by the subscriptions' filters. 

however it's a good idea to check out what other node clients are doing.

this is particularly dangerous because the get_logs call made in the event handler expects the log exctraction to succeeed.

found by me, while reading cantina 327

ercecan:
# Description
Copied logic from reth

## Linked Issues
- Fixes #2755

### DIFF
--- a/crates/ethereum-rpc/src/subscription.rs
+++ b/crates/ethereum-rpc/src/subscription.rs
@@ -170,7 +170,7 @@ pub async fn l2_block_event_handler<C: sov_modules_api::Context>(
                 &Filter::default(),
                 height,
                 height,
-                Some(usize::MAX),
+                usize::MAX,
             )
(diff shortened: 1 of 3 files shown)


eyusufatik (changes requested):
I'd just parametrize max logs in get_logs_in_block_range and pass max from the subscriptions manager.

eyusufatik (approved):

PR #2776, opened 2025-09-02
perf(memory): Do not save l1 hashes to memory unnecessarily

ercecan:
# Description
Do not save l1 hashes to memory unnecessarily by shp provider  for sequencer and full node
Creates another native short header proof provider to be used by full node and sequencer, does not save queried hashes on those nodes as it won't need them

## Linked Issues
- Related to # (issue)

### DIFF
--- a/crates/short-header-proof-provider/tests/shp_integration.rs
+++ b/crates/short-header-proof-provider/tests/shp_integration.rs
@@ -39,6 +39,32 @@ fn test_proof_not_found() {
 }
 
 #[test]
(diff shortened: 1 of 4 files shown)


crates/short-header-proof-provider/tests/shp_integration.rs
eyusufatik, line 1: add a test with save_hashes = false

eyusufatik (changes requested):

eyusufatik (approved):

PR #2778, opened 2025-09-02
feat: Separate logs and heads in eth subscribe

### LINKED ISSUE #2773: Completely seperate newHeads and logs subscription handling
## Summary
cantina 327

kpp:
- Separate logs and heads tx.send() to separate tokio threads

Fixes: https://github.com/chainwayxyz/citrea/issues/2773

### DIFF
--- a/crates/ethereum-rpc/src/subscription.rs
+++ b/crates/ethereum-rpc/src/subscription.rs
@@ -189,6 +189,9 @@ pub async fn l2_block_event_handler<C: sov_modules_api::Context>(
             Ok(height) => height,
         };
 
+        let new_heads_tx = new_heads_tx.clone();
+        let logs_tx = logs_tx.clone();
+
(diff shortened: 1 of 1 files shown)


kpp:
However I don't think we should really separate tx.send() into different tokio threads, because subscription.send in a different thread already should be enough. Please test this PR and PR #2764 on /heads performance. It should be fine in #2764 too.

eyusufatik:
I'm unsure about limiting addresses per filter

kpp:
Why not? If someone needs more, they can add another filter with other addresses?

eyusufatik (approved):

eyusufatik (changes requested):
let's remove address per filter limit

kpp:
> let's remove address per filter limit

Done

eyusufatik (approved):

PR #2787, opened 2025-09-04
refactor(sequencer): Remove pending commitments 

### LINKED ISSUE #2679: `SequencerLedgerOps::get_pending_commitments` fixes
## Summary
the get_data_range() misses the last index because we pull

(start..end)

also the `put_pending_commitment` call does not save to the SequencerCommitmentsByIndex table, so we always return empty vec from the `get_pending_commitments` call. 

After the latest rewrite of the commitment service, I think we don't even need to handle pending commitments on the sequencer side. L1 scan and mempool scan should be enough.

We may fix these two issues or just remove it altogether.

exeokan:
# Description
Pending commitments logic was not working correctly.
Even if we got it to work, we would not gain much:
Commitments are stored as pending only if the DA service dies, or the sequencer crashes while submitting the tx to DA.
When we restart, using the ready-made commitments wouldn't fasten the process much, as the bottleneck will be the DA submission.

Now, when the sequencer restarts, we still store the commitments that are mined or in mempool, to not submit them again. 

TODO:

- [x] Check if migration needed
- [x] Add test for commitments from DA

## Linked Issues
- Fixes #2679


crates/sequencer/src/commitment/service.rs
```diff
@@ -91,9 +91,9 @@ where
             let prestate = storage_manager.create_final_view_storage();
             let working_set = WorkingSet::new(prestate.clone());
 
-            // Resubmit if there were pending commitments on restart, skip it on first init
-            if let Err(e) = self.resubmit_pending_commitments(working_set).await {
-                error!("Could not resubmit pending commitments: {:?}", e);
+            // Store commitments on DA to db, skip it on first init
+            if let Err(e) = self.store_commitments_on_da(working_set).await {
```
jfldde: `store_commitments_from_da` maybe ? Wasn't clear

crates/sequencer/src/commitment/service.rs
```diff
@@ -310,52 +288,53 @@ where
...
 
-        pending_commitments_to_remove.sort();
-        pending_commitments_to_remove.dedup();
-        info!(
-            "Pending commitments to remove: {:?}",
-            pending_commitments_to_remove
+        commitments_to_store.sort();
+        commitments_to_store.dedup();
+
+        let last_index = last_commitment.map(|c| c.index).unwrap_or(0);
```
jfldde: `.map_or(0, |c| c.index)`

jfldde (approved):
LGTM, good simplification for the listen mode sequencer as well. cc/ @ercecan

crates/sequencer/src/commitment/service.rs
```diff
@@ -164,11 +164,9 @@ where
...
                             .expect("Commitment criteria check failed")
                         {
-                            if let Err(e) = self.commit(index, commitment_range.clone()).await {
-                                // We just log error and continue here as the controller updated its internal state and it can
-                                // continue functioning correctly. We just need to resubmit the failed commitment to DA.
-                                error!("Failed to submit commitment: {:?}", e);
-                            }
+                            self.commit(index, commitment_range.clone())
+                                .await
+                                .expect("Failed to submit commitment");
```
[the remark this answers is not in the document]
exeokan: the error comes from the DA service not the commitment, so not needed I think.

crates/sequencer/src/commitment/service.rs
```diff
@@ -310,52 +288,53 @@ where
...
+            commitments_to_store
+                .first()
+                .is_none_or(|c| c.index == last_index + 1),
+            "First commitment to store must be the next after last stored commitment"
+        );
+        assert!(
+            commitments_to_store
+                .windows(2)
+                .all(|w| w[0].index + 1 == w[1].index),
+            "Commitments to store must be consecutive"
```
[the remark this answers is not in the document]
exeokan: added more info, but I want to panic instead of returning err

eyusufatik (approved):
test if this needs a migration

crates/sequencer/src/commitment/service.rs
```diff
@@ -91,9 +91,9 @@ where
             let prestate = storage_manager.create_final_view_storage();
             let working_set = WorkingSet::new(prestate.clone());
 
-            // Resubmit if there were pending commitments on restart, skip it on first init
-            if let Err(e) = self.resubmit_pending_commitments(working_set).await {
-                error!("Could not resubmit pending commitments: {:?}", e);
+            // Store commitments on DA to db, skip it on first init
```
jfldde: Comment should be adapted accordingly, i.e. `from` instead of `to`

crates/sequencer/src/db_migrations/drop_pending_commitments.rs
```diff
@@ -0,0 +1,29 @@
...
+    fn identifier(&self) -> (String, u64) {
+        ("drop_pending_commitments".to_string(), 1)
+    }
+
+    fn execute(
+        &self,
+        _ledger_db: Arc<LedgerDB>,
+        tables_to_drop: &mut Vec<String>,
+    ) -> anyhow::Result<()> {
+        let fullnode_tables_to_drop = vec!["PendingSequencerCommitment"];
```
eyusufatik: rename the var pls

eyusufatik (changes requested):

crates/sequencer/src/db_migrations/drop_pending_commitments.rs
exeokan: oops

eyusufatik (approved):

PR #2797, opened 2025-09-04
fix: Assert `depositAmount` / `SAT_TO_WEI` not overflowing `uint64`

### LINKED ISSUE #2676: Assert `depositAmount / SAT_TO_WEI` is not overflowing `uint64`
For this to occur `depositAmount` needs to be 184 billion BTC which is impossible but we can fix for code correctness.

If `depositAmount` was to be that high, it would overflow `uint64` in the following line and thus would lead to revert in signature verification.

```
        bytes32 shaAmounts = sha256(abi.encodePacked(bytes8(BTCUtils.reverseUint64(uint64(depositAmount/(SAT_TO_WEI)))))); // 1000000000 in LE
```

It must be noted that `depositAmount` being that high would render the `Bridge` useless as it starts with only 21 million cBTC balance so the chain being initialized with that large `depositAmount` would require a chain restart anyway so this issue has no impact.

Cantina 112

okkothejawa:
# Description
Fixes an operationally impossible issue that results in signature verification problems due to `uint64` downcasting in `shaAmounts` calculation if the deposit amount is 184 billion BTC or more for code correctness.

## Linked Issues
- Fixes #2676

## Testing
Added a unit test.

### DIFF
--- a/crates/evm/src/evm/system_contracts/src/Bridge.sol
+++ b/crates/evm/src/evm/system_contracts/src/Bridge.sol
@@ -97,6 +97,7 @@ contract Bridge is Ownable2StepUpgradeable {
         require(_depositAmount != 0, "Deposit amount cannot be 0");
         require(_depositPrefix.length != 0, "Deposit script cannot be empty");
         require(_depositAmount % SAT_TO_WEI == 0, "Deposit amount must have valid satoshi value");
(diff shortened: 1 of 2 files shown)


eyusufatik (approved):
Good thing this is fixed, people with 184 billion btc would be so angry!!

PR #2800, opened 2025-09-04
fix: Assert withdrawal UTXOs are not used more than once in `Bridge`

### LINKED ISSUE #2672: `withdraw` and `batchWithdraw` UTXO already spent check in `Bridge.sol`
Check if the passed UTXO is used before to prevent users from burning their cBTC away with used UTXOs.

Cantina 70 (also many more)

okkothejawa:
## Linked Issues
- Fixes #2672

## Testing
Added unit tests.

### DIFF
--- a/crates/evm/src/evm/system_contracts/src/Bridge.sol
+++ b/crates/evm/src/evm/system_contracts/src/Bridge.sol
@@ -62,6 +62,7 @@ contract Bridge is Ownable2StepUpgradeable {
     bytes32[] public depositTxIds;
 
     mapping(bytes32 => bool) public processedTxIds;
+    mapping(bytes32 => bool) public usedWithdrawalUTXO;
     
     event Deposit(bytes32 wtxId, bytes32 txId, address recipient, uint256 timestamp, uint256 depositId);
     event Withdrawal(UTXO utxo, uint256 index, uint256 timestamp);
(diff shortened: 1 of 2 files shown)


eyusufatik (approved):

crates/evm/src/evm/system_contracts/src/Bridge.sol
```diff
@@ -215,6 +216,11 @@ contract Bridge is Ownable2StepUpgradeable {
     /// @param outputId The outputId of the output in the withdrawal transaction
     function withdraw(bytes32 txId, bytes4 outputId) public payable {
         require(msg.value == depositAmount, "Invalid withdraw amount");
+
+        bytes32 utxoKey = sha256(abi.encodePacked(txId, outputId));
```
kpp, line 220: Isnt txId and outputId already natural ids that dont need to be hashed? Sha256 seems to be an overkill. Concat should give same guarantees but way cheaper in cycles

eyusufatik: cant use 36 bytes as a map key in evm (or solidity, I'm not sure if it's a limitation of solidity

okkothejawa: An EVM limitation since storage slot addresses are 32 bytes.

ekrembal (approved):

crates/evm/src/evm/system_contracts/src/Bridge.sol
kpp: I see. We could've had `map(bytes32 => map(bytes4 => bool))` but I believe that Solidity way of hashing keys for multihash is close to sha256 (we can investigate but at this point I am 95% convinced it does not make much difference).

PR #2802, opened 2025-09-05
feat: add limit/skip for batchProver_getProvingJobs

### LINKED ISSUE #2779: Sanitize count in batchProver_getProvingJobs
- Can crash the batch prover with allocation failure

kpp:
Fixes: https://github.com/chainwayxyz/citrea/issues/2779

cc @Chinacolt : Added an optional `RPC_PROVING_JOBS_LIMIT` env var which is bound to optional config setting `rpc.proving_jobs_limit`. Default value is 100.


crates/batch-prover/src/rpc.rs
```diff
@@ -221,12 +221,17 @@ pub trait BatchProverRpc {
...
+    /// * `limit` - The number of latest proving jobs to retrieve (maximum 50).
+    /// * `skip` - The number of latest proving jobs to skip for pagination (default is 0).
     ///
     /// # Returns
     /// A vector of `ProvingJobResponse` containing job IDs and their statuses.
     #[method(name = "getProvingJobs")]
-    async fn get_proving_jobs(&self, count: usize) -> RpcResult<Vec<ProvingJobResponse>>;
+    async fn get_proving_jobs(
+        &self,
+        limit: usize,
```
eyusufatik: pls use U64 or U32 from alloy, it enables us to use hex str if we want

crates/batch-prover/src/rpc.rs
```diff
@@ -568,11 +573,17 @@ where
         }))
     }
 
-    async fn get_proving_jobs(&self, count: usize) -> RpcResult<Vec<ProvingJobResponse>> {
+    async fn get_proving_jobs(
+        &self,
+        limit: usize,
+        skip: Option<usize>,
+    ) -> RpcResult<Vec<ProvingJobResponse>> {
+        let skip = skip.unwrap_or(50);
```
eyusufatik: skip.unwrap_or(0) makes much more sense doesn't it?

crates/batch-prover/src/rpc.rs
eyusufatik, line 1: add some limitation to limit

like cant be more than 100 or smth.

should be configured from the batch prover config

eyusufatik (changes requested):
add a test, or extend one of the tests related to batch prover jobs

crates/batch-prover/src/rpc.rs
kpp: Added to runner config

crates/batch-prover/src/rpc.rs
kpp: Yes. It was a stupid mistake of mine. Thanks

crates/batch-prover/src/rpc.rs
kpp: Done

crates/common/src/config/mod.rs
```diff
@@ -51,6 +56,9 @@ pub struct RunnerConfig {
     pub pruning_config: Option<PruningConfig>,
     /// The DA block number to start L1 sync from
     pub scan_l1_start_height: u64,
+    /// Maximum number of responded proving jobs in RPC batchProver_getProvingJobs
+    #[serde(default = "default_max_rpc_proving_jobs_limit")]
+    pub max_rpc_proving_jobs_limit: usize,
```
eyusufatik: make optional

because we don't want full node runners to be concerned with this config

also name should contain batch prover somehow

maybe batch_prover_rpc_proving_jobs_limit or smth

eyusufatik (changes requested):

eyusufatik (approved):

jfldde (approved):

PR #2816, opened 2025-09-10
fix: add mantissa size check

### LINKED ISSUE #2814: Add size check to mantissa
Add [this](https://github.com/rust-bitcoin/rust-bitcoin/blob/a0cb9bfc648ded8839b7897e81d7a9d73623a636/bitcoin/src/pow.rs#L187) check to [here](https://github.com/chainwayxyz/citrea/blob/7b7c9dc1ab7ab3c419eb36aee8ae077eee450b04/crates/bitcoin-da/src/verifier.rs#L604)

exeokan:
# Description
Sets the target to zero if mantissa sign bit is set.

## Linked Issues
- Fixes #2814

### DIFF
--- a/crates/bitcoin-da/src/verifier.rs
+++ b/crates/bitcoin-da/src/verifier.rs
@@ -605,7 +605,8 @@ fn bits_to_target(bits: u32) -> [u8; 32] {
     let size = (bits >> 24) as usize;
     let mantissa = bits & 0x00ffffff;
 
-    // Mantissa should not be negative
+    // Mantissa is signed in Bitcoin core, if the sign bit is set, the target is set to 0.
+    // https://github.com/bitcoin/bitcoin/blob/ee42d59d4de970769ebabf77b89ff4269498f61e/src/arith_uint256.cpp#L175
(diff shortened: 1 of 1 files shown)


crates/bitcoin-da/src/verifier.rs
```diff
@@ -605,6 +605,11 @@ fn bits_to_target(bits: u32) -> [u8; 32] {
     let size = (bits >> 24) as usize;
     let mantissa = bits & 0x00ffffff;
 
+    // Mantissa should not be negative
```
ozankaymak: I believe we need to explain why there is a difference here for Rust Bitcoin and Bitcoin core implementation (u32, i32) because u32 can never be negative. Maybe also a link for CompactTarget etc.

ozankaymak (changes requested):

eyusufatik (approved):

ozankaymak (approved):

crates/bitcoin-da/src/verifier.rs
exeokan: edited comment

PR #2818, opened 2025-09-10
chore: give more detail on proof processing logs

ercecan:
# Description
The full node used to only log:
`Processing zk proof at height: n`
I added some more info

Batch prover logs the same exact thing for recovered and normal proofs, I added "Recovered" prefix to that log

### DIFF
--- a/crates/bitcoin-da/src/service.rs
+++ b/crates/bitcoin-da/src/service.rs
@@ -990,6 +990,7 @@ impl DaService for BitcoinService {
                         if complete.public_key() == prover_da_pub_key
                             && complete.get_sig_verified_hash().is_some()
                         {
+                            tracing::info!("Found complete tx with tx id: {}", tx_id);
                             let Ok(data) = DataOnDa::try_from_slice(&complete.body) else {
                                 warn!("{tx_id}: Failed to parse complete data");
                                 continue;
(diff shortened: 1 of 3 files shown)


rakanalh (approved):

crates/fullnode/src/da_block_handler.rs
```diff
@@ -565,6 +565,13 @@ where
             return Ok(ProcessingResult::Discarded);
         };
 
+        tracing::info!(
```
eyusufatik, line 568: can you log L1 block height and if posssible txid of the batch proof?

eyusufatik (changes requested):

crates/fullnode/src/da_block_handler.rs
ercecan: The l1 height is logged couple lines above, added l1 txid logs in bitcoinda

crates/bitcoin-da/src/service.rs
```diff
@@ -990,6 +990,7 @@ impl DaService for BitcoinService {
                         if complete.public_key() == prover_da_pub_key
                             && complete.get_sig_verified_hash().is_some()
                         {
+                            tracing::info!("Found complete tx with tx id: {}", tx_id);
```
eyusufatik, line 993: are these hex or bytes

ercecan: /// serialized in reverse byte order when converted to a hex string via [`std::fmt::Display`]

eyusufatik (approved):

PR #2825, opened 2025-09-11
fix: Decrement ongoing proving job count on error

### LINKED ISSUE #2824: decrement job count on `ParallelProverService::start_proving` err paths

ercecan:
# Description
The ongoing_proving_jobs is incremented in reserve_proof_slot after that this is the only path it can error out?

## Linked Issues
- Fixes #2824

### DIFF
--- a/crates/prover-services/src/parallel.rs
+++ b/crates/prover-services/src/parallel.rs
@@ -138,7 +138,11 @@ where
         // start proof immediately in the background
         let proof_start_time = std::time::Instant::now();
         let proof_rx = make_proof(vm, job_id, elf, self.proof_mode, receipt_type)
-            .context("Failed to start proving")?;
+            .context("Failed to start proving")
+            .map_err(|e| {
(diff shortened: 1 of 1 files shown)


eyusufatik (approved):
Aproving, but please spend some time thinking on different paths this can happen, or even other metrics this could happen

jfldde (approved):

crates/prover-services/src/parallel.rs
```diff
@@ -179,18 +180,20 @@ where
...
+            proof_was_in_queue = true;
             warn!("Reached parallel proof limit, waiting for one of the proving tasks to finish");
         }
 
         let permit = self.proof_semaphore.clone().acquire_owned().await?;
 
         PARALLEL_PROVER_METRICS.ongoing_proving_jobs.increment(1);
 
-        if available_permits == 0 {
+        if proof_was_in_queue {
```
ercecan: Same thing but easier to understand at first glance

PR #2826, opened 2025-09-11
chore: remove unnecessary use core::panic;

MozirDmitriy:
Delete redundant use core::panic; from crates/evm/src/call.rs.
The code only uses the panic! macro, which is available without importing core::panic.
Suppresses unused import warning and aligns with Rust idioms.


eyusufatik:
we are not touching STF level code ATM, sorry

PR #2828, opened 2025-09-11
feat: Restore bitcoin-da tx queue from tx_backup_dir

jfldde:
# Description

- Queued txs that were never broadcasted to bitcoin mempool (due to mempool policy limit) were not properly restored on startup via `listunspent`.
- Add an additional step that goes through tx_backup_dir, parses all the txs and put back tx in queue that are not found in `listtransactions` nor mempool


eyusufatik:
closing this in favor of #2841

PR #2830, opened 2025-09-12
fix: Re-add deposit txs

### LINKED ISSUE #2705: On block production error, all deposits are dropped
## Summary

If a block fails to be produced for any reason, we drop the deposits we drain from the deposit mempool. Instead, we can readd the deposits.

Cantina 159

rakanalh:
# Description

## Linked Issues
- Fixes #2705

### DIFF
--- a/crates/sequencer/src/runner.rs
+++ b/crates/sequencer/src/runner.rs
@@ -431,14 +431,26 @@ where
         let start: Instant = Instant::now();
         let l2_height = self.ledger_db.get_head_l2_block_height()?.unwrap_or(0) + 1;
         self.fork_manager.register_block(l2_height)?;
+
+        let deposit_data = self
+            .deposit_mempool
+            .lock()
+            .fetch_deposits(self.config.deposit_mempool_fetch_limit);
+
         let result = {
             if da_blocks.len() == 1 && da_blocks[0].header().height() == *last_used_l1_height {
                 // If we are producing regular blocks, not for missed da blocks, and if the last used L1 block is the same as the last finalized block
                 // then there is no need to pass da data to the sequencer
                 da_blocks.clear();
             }
-            self.produce_l2_block_inner(da_blocks, l1_fee_rate, l2_height, last_used_l1_height)
-                .await
+            self.produce_l2_block_inner(
+                da_blocks,
+                l1_fee_rate,
+                l2_height,
+                last_used_l1_height,
+                deposit_data.clone(),
+            )
+            .await
         };
 
         match result {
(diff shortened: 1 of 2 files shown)


crates/sequencer/src/runner.rs
```diff
@@ -450,6 +462,13 @@ where
             }
             Err(e) => {
                 error!("Sequencer error: {}", e);
+                if !deposit_data.is_empty() {
```
eyusufatik: maybe some of the deposits failed and some of them did not

crates/sequencer/src/runner.rs
```diff
@@ -431,18 +432,38 @@ where
...
+                deposit_data.clone(),
+            )
+            .await
         };
 
         match result {
-            Ok(l2_height) => {
+            Ok((l2_height, failed_deposits)) => {
+                // Restore only the deposits that failed to process
+                if !failed_deposits.is_empty() {
```
kpp, line 460: No need to check if they are empty

crates/sequencer/src/runner.rs
```diff
@@ -481,7 +502,8 @@ where
         l1_fee_rate: u128,
         l2_height: u64,
         last_used_l1_height: &mut u64,
-    ) -> anyhow::Result<u64> {
+        deposit_data: Vec<Vec<u8>>,
+    ) -> anyhow::Result<(u64, Vec<Vec<u8>>)> {
```
kpp, line 506: Missing fix in doc comment to this method. Also having a newtype or at least an alias to `Vec<u8>` is better than raw vec.

crates/sequencer/src/runner.rs
```diff
@@ -1511,14 +1533,19 @@ where
...
 
-        // Store deposit txs by index
-        let is_deposit_tx = system_events
-            .iter()
-            .map(|ev| matches!(ev, SystemEvent::BridgeDeposit(_)))
-            .collect::<Vec<_>>();
+        let mut deposit_indices = Vec::new();
+        for (deposit_idx, ev) in system_events.iter().enumerate() {
+            if matches!(ev, SystemEvent::BridgeDeposit(_)) {
+                deposit_indices.push(Some(deposit_idx));
```
kpp, line 1539: I would expect it to be `deposit_indices.push(Some(i)); i = i+1;`

crates/sequencer/src/runner.rs
```diff
@@ -1551,9 +1578,11 @@ where
                     StateTransitionError::ModuleCallError(
                         L2BlockModuleCallError::EvmSystemTransactionNotSuccessful
                     )
-                ) && is_deposit
+                ) && deposit_indices[i].is_some()
```
kpp, line 1581: https://doc.rust-lang.org/edition-guide/rust-2024/let-chains.html

PR #2837, opened 2025-09-16
feat: Add check for tx len in verifier

kpp:
Cantina 450


eyusufatik:
We already do zip_eq. We panicn the circuit to make proving fail

PR #2838, opened 2025-09-16
chore: remove SafeMath import from WitnessUtils.sol

kpp:
Cantina 451.


kpp:
Duplicate in https://github.com/chainwayxyz/citrea/pull/2786/

PR #2844, opened 2025-09-16
feat: Add enable JsTracer option

### LINKED ISSUE #2667: Apply timeout to `JsTacer` debug_* calls
The js tracers can be used to execute any js script, which may have an infinite loop or memory heavy operations, this is DOS through RPC endpoints. We should apply timeouts for these calls. I remember somewhere in the tracer config structs, there was such an option.

Cantina 228

exeokan:
# Description
Rather than applying a timeout, which seems like a difficult task, this PR adds `enable_js_tracer` to the `RpcConfig`.
Also moved `debug_traceCall` handler to `EthereumRpcServerImpl` to check the newly added option.

## Linked Issues
- Fixes #2667

### DIFF
--- a/crates/ethereum-rpc/src/trace.rs
+++ b/crates/ethereum-rpc/src/trace.rs
@@ -147,7 +147,7 @@ pub fn debug_trace_by_block_number<C: sov_modules_api::Context, Da: DaService>(
         None => false,
         Some(GethDebugTracerType::JsTracer(_)) => {
             if disable_js_tracer {
-                return Err(EthApiError::Unsupported("JS tracer is disabled").into());
+                return Err(EthApiError::Unsupported("JsTracer is disabled on this node").into());
             }
             true
         }
--- a/crates/ethereum-rpc/src/trace.rs
+++ b/crates/ethereum-rpc/src/trace.rs
@@ -141,8 +141,7 @@ pub fn debug_trace_by_block_number<C: sov_modules_api::Context, Da: DaService>(
     enable_js_tracer: bool,
 ) -> Result<Vec<TraceResult>, ErrorObjectOwned> {
     let is_js_tracer = matches!(
-        opts.as_ref()
-            .and_then(|o| o.tracer.as_ref()),
+        opts.as_ref().and_then(|o| o.tracer.as_ref()),
         Some(GethDebugTracerType::JsTracer(_))
     );
     if is_js_tracer && !enable_js_tracer {
@@ -154,9 +153,9 @@ pub fn debug_trace_by_block_number<C: sov_modules_api::Context, Da: DaService>(
         o.tracer.as_ref().is_none_or(|inner| match inner {
             GethDebugTracerType::JsTracer(_) => true,
             GethDebugTracerType::BuiltInTracer(bit) => matches!(
-            bit,
-            GethDebugBuiltInTracerType::MuxTracer | GethDebugBuiltInTracerType::PreStateTracer
-        ),
+                bit,
+                GethDebugBuiltInTracerType::MuxTracer | GethDebugBuiltInTracerType::PreStateTracer
+            ),
         })
     });
     if skip_cache {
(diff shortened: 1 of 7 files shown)


eyusufatik (changes requested):
naming should be "enable" not disable. we have other params that use the enable convention

ercecan (approved):

crates/ethereum-rpc/src/trace.rs
```diff
@@ -136,18 +138,25 @@ pub fn debug_trace_by_block_number<C: sov_modules_api::Context, Da: DaService>(
...
-            GethDebugTracerType::JsTracer(_) => true,
-            GethDebugTracerType::BuiltInTracer(bit) => matches!(
-                bit,
-                GethDebugBuiltInTracerType::MuxTracer | GethDebugBuiltInTracerType::PreStateTracer
-            ),
-        })
-    });
+
+    let skip_cache = match opts.as_ref().and_then(|o| o.tracer.as_ref()) {
+        None => false,
```
eyusufatik: wrong

crates/ethereum-rpc/src/trace.rs
```diff
@@ -136,18 +138,25 @@ pub fn debug_trace_by_block_number<C: sov_modules_api::Context, Da: DaService>(
...
-        o.tracer.as_ref().is_none_or(|inner| match inner {
-            GethDebugTracerType::JsTracer(_) => true,
-            GethDebugTracerType::BuiltInTracer(bit) => matches!(
-                bit,
-                GethDebugBuiltInTracerType::MuxTracer | GethDebugBuiltInTracerType::PreStateTracer
-            ),
-        })
-    });
+
+    let skip_cache = match opts.as_ref().and_then(|o| o.tracer.as_ref()) {
```
eyusufatik: imo instead of changing here, just make another if check above or below

eyusufatik (changes requested):

crates/ethereum-rpc/src/trace.rs
exeokan: you're right, made changes

crates/ethereum-rpc/src/trace.rs
```diff
@@ -136,16 +138,25 @@ pub fn debug_trace_by_block_number<C: sov_modules_api::Context, Da: DaService>(
...
     // If tracer option is not specified, or it is JsTracer, then do not check cache or insert cache, just perform the operation
     // Skip cache from JsTracer, MuxTracer and PreStateTracer
     let skip_cache = opts.as_ref().is_none_or(|o| {
         o.tracer.as_ref().is_none_or(|inner| match inner {
             GethDebugTracerType::JsTracer(_) => true,
             GethDebugTracerType::BuiltInTracer(bit) => matches!(
-                bit,
-                GethDebugBuiltInTracerType::MuxTracer | GethDebugBuiltInTracerType::PreStateTracer
-            ),
+            bit,
```
eyusufatik: formatting

crates/ethereum-rpc/src/trace.rs
```diff
@@ -136,16 +138,25 @@ pub fn debug_trace_by_block_number<C: sov_modules_api::Context, Da: DaService>(
...
     opts: Option<GethDebugTracingOptions>,
+    enable_js_tracer: bool,
 ) -> Result<Vec<TraceResult>, ErrorObjectOwned> {
+    let is_js_tracer = matches!(
+        opts.as_ref()
+            .and_then(|o| o.tracer.as_ref()),
+        Some(GethDebugTracerType::JsTracer(_))
+    );
+    if is_js_tracer && !enable_js_tracer {
+        return Err(EthApiError::Unsupported("JsTracer is disabled on this node").into());
```
eyusufatik: just say JsTracer is disabled

eyusufatik (approved):

Issue #2744, opened 2025-08-29
Fix test_queue_da_transactions_oldest_mode flakiness


jfldde:
https://github.com/chainwayxyz/citrea/actions/runs/17320970723/job/49174065398?pr=2743

PR #2853, opened 2025-09-17
fix: increase mempool wait timeouts in test_queue_da_transactions_oldest_mode

Forostovec:
- Replace None timeouts with TEN_MINS (10 minutes) for all wait_mempool_len calls
- Aligns with other E2E tests that use explicit timeouts
- Prevents premature timeouts in slow CI environments
- Addresses flakiness where test times out waiting for mempool length 10

Fixes test_queue_da_transactions_oldest_mode timeout failures in make coverage #2744


jfldde:
Default timeout is 5mins. Not sure this addresses the flakiness

eyusufatik:
as shown by the last ci run, the flakiness is still there

PR #2855, opened 2025-09-18
chore: Parse SequencerConfig::bridge_initialize_params as hex

kpp:
# Description
In case bridge_initialize_params is not hex it will fail on `::deserialize()` on startup, not inside of `create_initial_system_events` in guest code.


rustyrustician (approved):

rakanalh (approved):

jfldde (dismissed):

crates/evm/src/evm/system_events.rs
```diff
@@ -59,7 +59,7 @@ fn system_event_to_transaction(event: SystemEvent, nonce: u64, chain_id: u64) ->
         },
         SystemEvent::BridgeInitialize(params) => TxEip1559 {
             to: TxKind::Call(BridgeWrapper::address()),
-            input: BridgeWrapper::initialize(params.as_slice()),
+            input: BridgeWrapper::initialize(&params),
```
eyusufatik: hmm, maybe revert this? to not make changes on audited code
xd

eyusufatik (changes requested):

eyusufatik:
btw, this doesn't need guest-code

crates/evm/src/evm/system_events.rs
eyusufatik: @kpp if you revert this, we can merge this to nightly as well

eyusufatik (approved):

PR #2857, opened 2025-09-18
pruning: add runtime-gated state DB pruning

VolodymyrBg:
- Add enable_state_pruning to PruningConfig (default false, supports ENABLE_STATE_PRUNING env).
- Propagate flag into Pruner; conditionally spawn prune_state_db at runtime.
- Remove #[allow(dead_code)] from state_db; keep progress semantics unchanged.
- Replace “TODO: Fix me” with explicit gated state pruning logic and join handling.
- Update callers to use PruningConfig { distance, ..Default::default() }.


eyusufatik:
pruning does not work correctly right now.

Issue #2841, opened 2025-09-16
bitcoin-da Job service


jfldde:
## Issues with current design 

- Stale fee issue. We fund and sign the tx at the time of request, but any queued tx will have a non-adaptive fee rate (either too high or too low) as it's not being re-funded on each block according to mempool changes.

- Complexity of two queues, one for the request, one for chunked txs sequential processing. These can be both replaced by a single job queue instead.

- UTXO selection is not done against its current state but against the state at the point of request.

- In-memory queues. This leads to complex recovery mechanism

## Proposed changes
- Replace bitcoin service queues with a single job queuing mechanism.
- Existing `UnboundedSender<TxRequestWithNotifier<TxidWrapper>>` queue would now be replaced by the job queuing mechanism itself. Entry point to the service is sending a TxRequest and receives a `jobId` to track request status.
- Existing `da_queue` which keeps tracks of `SignedTx` would be also handled by the job service. On queue processing, split up the payload and try sending against current mempool state. Any mempool policy limit would be handled by existing behaviour defined by `UtxoSelectionMode`.
- This changes the tx funding/signing to when the queue is actually processed. Before, we created the txs, selected the utxos, signed the txs then queued it. We would now queue the chunked payload and fund it as mempool space is freed up

- Chunked txs can be over mempool policy limit. In that case, we need to split the payload and send it partially. This will be represented by an intermediate state `InProgress` that keeps track of the already sent/remaining chunks to be sent.

Main job states :

```
enum JobStatus {
    Completed { txs: Vec<SignedTxPair> },
    InProgress { sent: Vec<SignedTxPair>, remaining: Vec<Vec<u8>> },
    Processing,
    Pending,
    Cancelled
}
```

- Restore LedgerDB in bitcoin-da to be used for job storage and recovery.
- Replace complex recovery mechanism of https://github.com/chainwayxyz/citrea/pull/2828 by restoring pending job status instead.

- `batchProver_submitFakeProof` will now return a JobId instead of an optional `l1_tx_id`
- Sequencer `send_transaction` will be replaced by `submit_commitment`, which returns the job itself and a signal to wait for the commitment to be sent to DA.
- Proving service entry point signature `submit_proof` will be changed to return a `JobId` instead of a `TxId`. Proving service won't wait for the `TxId` (as this would require the tx to be funded and signed at this point) but will be able to query status via jobId instead

PR #2863, opened 2025-09-20
feat: resend commitment RPC

exeokan:
# Description
Adds an endpoint for resubmitting an existing commitment to DA queue.

## Testing
Added mock-e2e test.


exeokan:
We won't be needing this with #2841, closing the pr.

PR #2882, opened 2025-09-23
fix: Readd deposit transactions

### LINKED ISSUE #2705: On block production error, all deposits are dropped

rakanalh:
# Description

## Linked Issues
- Fixes #2705

### DIFF
--- a/crates/sequencer/src/deposit_data_mempool.rs
+++ b/crates/sequencer/src/deposit_data_mempool.rs
@@ -80,20 +80,20 @@ impl DepositDataMempool {
         // Calculate txids for the deposits to remove
         let mut txids_to_remove = HashSet::new();
         for deposit in deposits_to_remove {
-            if let Ok(txid) = Self::calc_tx_id(deposit) {
-                txids_to_remove.insert(txid.to_vec());
-            }
+            let txid = Self::calc_tx_id(deposit)
+                .expect("calc_tx_id should never be called on non-deposit");
+            txids_to_remove.insert(txid.to_vec());
         }
 
         // Retain only deposits that are not in the removal set
         self.accepted_deposit_txs.retain(|deposit| {
-            if let Ok(txid) = Self::calc_tx_id(deposit) {
-                if txids_to_remove.contains(txid.as_slice()) {
-                    // Remove from pending set
-                    self.pending_deposits.remove(txid.as_slice());
-                    removed_count += 1;
-                    return false;
-                }
+            let txid = Self::calc_tx_id(deposit)
+                .expect("calc_tx_id should never be called on non-deposit");
+            if txids_to_remove.contains(txid.as_slice()) {
+                // Remove from pending set
+                self.pending_deposits.remove(txid.as_slice());
+                removed_count += 1;
+                return false;
             }
             true
         });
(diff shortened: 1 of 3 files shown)


crates/sequencer/src/deposit_data_mempool.rs
```diff
@@ -42,30 +45,71 @@ impl DepositDataMempool {
...
+            .collect();
+
+        self.accepted_deposit_txs = remaining_deposits;
+
+        // Update metrics
+        SM.deposit_data_mempool_txs
+            .set(self.accepted_deposit_txs.len() as f64);
+
+        debug!("Removed {} deposits from mempool", removed_count);
+        removed_count
```
eyusufatik, line 106: since this is a queue, only the count is enough to remove the deposits from the mempool. so you don't need to the following:

```rust
self
    .accepted_deposit_txs
    .drain(..)
    .filter(|deposit| {
        if let Ok(txid) = Self::calc_tx_id(deposit) {
            if txids_to_remove.contains(txid.as_slice()) {
                // Remove from pending set
                self.pending_deposits.remove(txid.as_slice());
                removed_count += 1;
                return false;
            }
        }
        true
    })
    .collect();
```

but just `.drain(..deposits.len())`

eyusufatik: but you still have to remove form txid set that's correct

eyusufatik (changes requested):

crates/sequencer/src/deposit_data_mempool.rs
rakanalh: An even better approach would be to call `retain` let me update accordingly

eyusufatik: keeping the retain approach as it's a safer API

eyusufatik (approved):

kpp (approved):

crates/sequencer/src/deposit_data_mempool.rs
```diff
@@ -42,30 +45,65 @@ impl DepositDataMempool {
...
             if let Ok(txid) = Self::calc_tx_id(deposit) {
-                self.pending_deposits.remove(txid.as_slice());
+                txids_to_remove.insert(txid.to_vec());
             }
         }
 
-        deposits
+        // Retain only deposits that are not in the removal set
+        self.accepted_deposit_txs.retain(|deposit| {
+            if let Ok(txid) = Self::calc_tx_id(deposit) {
```
jfldde: Shouldn't we return the error here as that would only happen in a corrupted state ? Above as well

jfldde (approved):

crates/sequencer/src/deposit_data_mempool.rs
eyusufatik: yeah, good point.

@rakanalh we can panic on calc_tx_id errors in this function. it should never be called with non-deposit Vec

rakanalh: I checked. `calc_tx_id` doesn't really return an error... it always returns `Ok`. We should change the signature of the method.

PR #2893, opened 2025-09-24
feat: bridge pausability

### LINKED ISSUE #2675: Make `deposit` and `withdraw` functions pausable in `Bridge`

eyusufatik:
# Description
closes #2675

### DIFF
--- a/crates/evm/src/evm/system_contracts/src/Bridge.sol
+++ b/crates/evm/src/evm/system_contracts/src/Bridge.sol
@@ -6,11 +6,12 @@ import "bitcoin-spv/solidity/contracts/BTCUtils.sol";
 import "../lib/WitnessUtils.sol";
 import "./BitcoinLightClient.sol";
 import "openzeppelin-contracts-upgradeable/contracts/access/Ownable2StepUpgradeable.sol";
+import "openzeppelin-contracts-upgradeable/contracts/utils/PausableUpgradeable.sol";
 
 /// @title Bridge contract for the Citrea end of Citrea <> Bitcoin bridge
 /// @author Citrea
 
-contract Bridge is Ownable2StepUpgradeable {
+contract Bridge is Ownable2StepUpgradeable, PausableUpgradeable {
     using BTCUtils for bytes;
     using BytesLib for bytes;
     using WitnessUtils for bytes;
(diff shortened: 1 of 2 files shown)


crates/evm/src/evm/system_contracts/test/Bridge.t.sol
```diff
@@ -692,4 +705,115 @@ contract BridgeTest is Test {
         bridge.initialize(depositPrefix, depositSuffix, overflowingDepositAmount);
         vm.stopPrank();
     }
+
+    function testBridgePausability() public {
```
okkothejawa: I think it's better to split this into more tests that test a single property

crates/evm/src/evm/system_contracts/test/Bridge.t.sol
```diff
@@ -692,4 +705,115 @@ contract BridgeTest is Test {
...
+        // random user cannot pause/unpause
+        vm.prank(address(0x123));
```
okkothejawa: `makeAddr("user")` is more stylish imo

crates/evm/src/evm/system_contracts/src/Bridge.sol
```diff
@@ -240,7 +246,7 @@ contract Bridge is Ownable2StepUpgradeable {
     /// @param payoutTx Transaction parameters of the payout transaction on Bitcoin
     /// @param blockHeader Block header of the associated Bitcoin block
     /// @param withdrawalAddressPubKey The script pubkey of the user that BTC is withdrawn to, included for extra validation
-    function safeWithdraw(Transaction calldata prepareTx, MerkleProof calldata prepareProof, Transaction calldata payoutTx, bytes calldata blockHeader, bytes memory withdrawalAddressPubKey) external payable {
+    function safeWithdraw(Transaction calldata prepareTx, MerkleProof calldata prepareProof, Transaction calldata payoutTx, bytes calldata blockHeader, bytes memory withdrawalAddressPubKey) external payable whenNotPaused {
```
okkothejawa: We can remove `whenNotPaused` here as `withdraw` already has it, if we don't do this `safeWithdraw` calls in the unpaused state will be slightly more costly as pause state will be checked twice. Calling in the paused state will result in a much later revert though thus reverting txs would cost more.

eyusufatik: ok. remove

PR #2912, opened 2025-09-26
feat: max commitments per proof

### LINKED ISSUE #2896: batch prover config for max commitments per proof
limit partitions by commitment count config, implement the env var too.

rakanalh:
# Description

## Linked Issues
- Fixes #2896


crates/batch-prover/src/prover.rs
```diff
@@ -477,10 +481,12 @@ where
     /// # Arguments
     /// * `commitments` - A slice of sequencer commitments to partition
     /// * `mode` - The partition mode to use for partitioning the commitments
+    /// * `max_commitments_per_proof` - Optional maximum number of commitments per proof partition
     fn partition_commitments<'a>(
         &self,
         commitments: &'a [SequencerCommitment],
         mode: PartitionMode,
+        max_commitments_per_proof: Option<usize>,
```
eyusufatik: you can access this from self, no need to pass it as an arg

crates/batch-prover/src/prover.rs
```diff
@@ -1549,4 +1571,348 @@ mod tests {
...
+
+        // Partition 3: commitment 7
+        assert_eq!(partitions[2].commitments.len(), 1,);
+        assert_eq!(partitions[2].commitments[0].index, 7);
+        assert_eq!(partitions[2].start_height, 7);
+        assert_eq!(partitions[2].end_height, 7);
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn commitment_count_exact_limit() {
```
eyusufatik: isn't this test same with `commitment_count_max_two`

eyusufatik (changes requested):
can you update citrea-e2e with this config change?

no need to write an e2e test, however, when needed let's not spend time on it.

crates/batch-prover/src/prover.rs
rakanalh: Yea you're right... i was simulating when we have exactly the number of commitment as the max per proof... but two does do that indirectly.

rakanalh:
> can you update citrea-e2e with this config change?
> 
> no need to write an e2e test, however, when needed let's not spend time on it.

https://github.com/chainwayxyz/citrea-e2e/pull/137

crates/batch-prover/src/prover.rs
eyusufatik: or some help from AI to write tests ;)

eyusufatik (approved):

kpp (approved):

PR #2914, opened 2025-09-26
chore: update citrea e2e

rakanalh:


jfldde (approved):

bin/citrea/tests/bitcoin/bitcoin_test.rs
```diff
@@ -26,7 +28,7 @@ impl TestCase for BitcoinReorgTest {
         TestCaseConfig {
             with_sequencer: true,
             with_batch_prover: true,
-            n_nodes: 2,
+            n_nodes: HashMap::from([(NodeKind::Bitcoin, 2)]),
```
eyusufatik, line 31: this is citrea-e2e related, nevertheless, why not make different fields for this, like bitcoin_node_count, citrea_node_count etc..

eyusufatik (approved):

bin/citrea/tests/bitcoin/bitcoin_test.rs
jfldde: Yes we could have flattened it but it would require 6 additional fields (4 citrea nodes, bitcoin, clementine) and I felt it getting a bit out of hand. I'll re-assess if we can handle it some other way

PR #2915, opened 2025-09-26
Put network id in batch proof method id update

ercecan:
# Description
Puts network identifier in batch proof method id update to prevent cross chain attacks


crates/primitives/src/network.rs
```diff
@@ -0,0 +1,11 @@
...
+    network: sov_rollup_interface::Network,
+) -> u8 {
+    match network {
+        sov_rollup_interface::Network::Mainnet => 0,
+        sov_rollup_interface::Network::Testnet => 1,
+        sov_rollup_interface::Network::Devnet => 2,
+        sov_rollup_interface::Network::Nightly => 3,
+        sov_rollup_interface::Network::TestNetworkWithForks => 4,
+    }
+}
```
rakanalh: No need for this, we can use repr(u8)
```
  #[repr(u8)]
  enum MyEnum {
      First = 0,
      Second = 1,
      Third = 2,
  }

  let variant = MyEnum::Second;
  let value: u8 = variant as u8;  // value is 1
```

rakanalh (changes requested):

crates/primitives/src/network.rs
ercecan: The order and the values may change so let's not do that now

crates/light-client-prover/src/circuit/mod.rs
```diff
@@ -601,3 +610,15 @@ impl<S: Storage, DS: DaSpec, Z: Zkvm> Default for LightClientProofCircuit<S, DS,
...
+
+pub fn citrea_network_to_method_id_upgrade_identifier(
+    network: sov_rollup_interface::Network,
+) -> u8 {
+    match network {
+        sov_rollup_interface::Network::Mainnet => 0,
+        sov_rollup_interface::Network::Testnet => 1,
+        sov_rollup_interface::Network::Devnet => 2,
+        sov_rollup_interface::Network::Nightly => 3,
+        sov_rollup_interface::Network::TestNetworkWithForks => 4,
```
eyusufatik: mainnet --> 1
testnet --> 10
devnet --> 100
nightly --> 105
test network with forks --> 106

crates/light-client-prover/src/circuit/mod.rs
```diff
@@ -418,6 +419,13 @@ impl<S: Storage, DS: DaSpec, Z: Zkvm> LightClientProofCircuit<S, DS, Z> {
                         .0;
 
                     if batch_proof_method_id.body.activation_l2_height > last_activation_height {
```
eyusufatik: convert above if to early return and log as well

eyusufatik (changes requested):

eyusufatik (approved):

PR #2919, opened 2025-09-26
Use constants instead of some magic numbers in method id update

ercecan:
# Description
Uses constants instead of some magic numbers in method id update

### DIFF
--- a/crates/light-client-prover/src/circuit/mod.rs
+++ b/crates/light-client-prover/src/circuit/mod.rs
@@ -23,7 +23,7 @@ use crate::circuit::method_id_verifier::verify_method_id_security_council;
 pub const SECURITY_COUNCIL_COMPRESSED_PUBKEY_SIZE: usize = 33;
 
 /// Total number of security council members.
-pub const SECURITY_COUNCIL_MEMBERS: usize = 5;
+pub const SECURITY_COUNCIL_MEMBER_COUNT: usize = 5;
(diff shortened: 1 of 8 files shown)


crates/light-client-prover/src/circuit/mod.rs
```diff
@@ -19,6 +19,12 @@ use sov_rollup_interface::Network;
 
 use crate::circuit::method_id_verifier::verify_method_id_security_council;
 
+/// Size of a compressed public key in bytes.
+pub const SECURITY_COUNCIL_COMPRESSED_PUBKEY_SIZE: usize = 33;
+
+/// Total number of security council members.
+pub const SECURITY_COUNCIL_MEMBERS: usize = 5;
```
jfldde:
```suggestion
pub const SECURITY_COUNCIL_N_MEMBERS: usize = 5;
```
Or something better but it wasn't clear

jfldde (approved):
Small nit on naming but looks good otherwise

crates/light-client-prover/src/circuit/mod.rs
eyusufatik: SECURITY_COUNCIL_MEMBER_COUNT

eyusufatik (changes requested):

eyusufatik (approved):

PR #2920, opened 2025-09-26
More efficient duplicate index check

ercecan:
# Description
Implements more flexible duplicate index check

### DIFF
--- a/crates/light-client-prover/src/circuit/method_id_verifier.rs
+++ b/crates/light-client-prover/src/circuit/method_id_verifier.rs
@@ -22,9 +22,13 @@ pub fn verify_method_id_security_council(
     }
 
     // Make sure the indexes are in ascending order to prevent duplicates
-    for i in 1..signatures_with_idx.len() {
-        if signatures_with_idx[i].1 <= signatures_with_idx[i - 1].1 {
-            log!("Signature indices are not in ascending order");
+    for i in 0..signatures_with_idx.len() - 1 {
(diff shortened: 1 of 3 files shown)


eyusufatik (approved):

crates/light-client-prover/src/circuit/method_id_verifier.rs
```diff
@@ -21,13 +21,16 @@ pub fn verify_method_id_security_council(
...
+    // Make sure the indexes are in ascending order to prevent duplicates
+    for i in 0..signatures_with_idx.len() - 1 {
+        if signatures_with_idx[i].1 >= signatures_with_idx[i + 1].1 {
+            log!(
+                "Signature indices are not in ascending order, failing indices: {}, {}",
+                signatures_with_idx[i].1,
+                signatures_with_idx[i + 1].1
+            );
+            return false;
+        }
```
jfldde, line 40: Much more complex now imo. Are we sure that is more efficient ?

ercecan: it is for future changes on the value of signature_threshold to be verified

ercecan: If we decided to make this a 5/7 for example this will still work

jfldde: Ok, the PR title threw me off

jfldde (approved):

PR #2925, opened 2025-09-29
fix: clementine-569

eyusufatik:
# Description
https://github.com/chainwayxyz/clementine/pull/1139
https://cantina.xyz/code/ce181972-2b40-4047-8ee9-89ec43527686/findings/596


crates/bitcoin-da/src/verifier.rs
```diff
@@ -649,11 +649,30 @@ pub(crate) const fn target_to_bits(target: &[u8; 32]) -> u32 {
...
+    // We define the inverse of max as 1.
+    if target == U256::MAX {
+        return U256::ONE;
+    }
+
+    let comp = !target;
 
-    U256::MAX.wrapping_div(&target_plus_one)
+    let ret = comp.wrapping_div(&target.wrapping_add(&U256::ONE));
+    ret.wrapping_add(&U256::ONE)
```
jfldde, line 674:
```
fn target_to_work(target: &[u8; 32]) -> U256 {
    let target = U256::from_be_slice(target);
    target.inverse()
}
```
Should work

jfldde: With `inverse` being :

```

    /// Calculates 2^256 / (x + 1) where x is a 256 bit unsigned integer.
    ///
    /// 2**256 / (x + 1) == ~x / (x + 1) + 1
    ///
    /// (Equation shamelessly stolen from bitcoind)
    fn inverse(&self) -> U256 {
        // We should never have a target/work of zero so this doesn't matter
        // that much but we define the inverse of 0 as max.
        if self.is_zero() {
            return U256::MAX;
        }
        // We define the inverse of 1 as max.
        if self.is_one() {
            return U256::MAX;
        }
        // We define the inverse of max as 1.
        if self.is_max() {
            return U256::ONE;
        }

        let ret = !*self / self.wrapping_inc();
        ret.wrapping_inc()
    }
```

ozankaymak (approved):

crates/bitcoin-da/src/verifier.rs
jfldde: Resolving as we're not using bitcoin U256

jfldde (approved):

PR #2935, opened 2025-09-29
Remove unnecessary TransactionRequest clones in call/env preparation

VolodymyrBg:
- Replace request.clone() with moves in prepare_call_env and trace-call path in query.rs.
- These clones were not needed because the request is not used after create_txn_env in those contexts.
- Keeps clone in access-list flow where request is reused (access_list take/update and subsequent estimate), preserving existing behavior and avoiding accidental semantics changes.

### DIFF
--- a/crates/evm/src/query.rs
+++ b/crates/evm/src/query.rs
@@ -722,13 +722,7 @@ impl<C: sov_modules_api::Context> Evm<C> {
         let nonce = request.nonce.unwrap_or(account.nonce);
         let chain_id = cfg_env.chain_id();
 
-        let tx_env = create_txn_env(
-            &block_env,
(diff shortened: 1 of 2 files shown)


crates/evm/src/query.rs
```diff
@@ -722,13 +722,7 @@ impl<C: sov_modules_api::Context> Evm<C> {
         let nonce = request.nonce.unwrap_or(account.nonce);
         let chain_id = cfg_env.chain_id();
 
-        let tx_env = create_txn_env(
-            &block_env,
-            request.clone(),
```
ercecan, line 727: Looks like we needed this

ercecan (changes requested):

PR #2946, opened 2025-09-30
chore: Remove eth_syncing from sequencer

### LINKED ISSUE #2928: redefine `eth_syncing` in sequencer
full nodes asks the sequencer its highest height. but if the existing implementation is called on the sequencer, the request won't be responded, because the .unwrap call in the sequencer will panic.

this is not a big isssue as panicking inside RPC method call is fine, it doesn't cause trouble, you just won't reply to the reuquest.

we can remove it from ethereum-rpc like `txpool_content` when it's the sequencer, and redefine it under sequencer/rpc.rs

cantina 622

rakanalh:
# Description

## Linked Issues
- Fixes #2928


ercecan (commented):
Could you implement a simple e2e test showing these rpcs are not reachable?

kpp (approved):

crates/sequencer/src/rpc.rs
```diff
@@ -177,6 +177,10 @@ pub trait SequencerRpc {
     /// Returns the hashes of the removed transactions.
     #[method(name = "txpool_removeTransactionsBySender")]
     async fn txpool_remove_txs_by_sender(&self, sender: Address) -> RpcResult<Vec<B256>>;
+
+    /// Returns the sync status
+    #[method(name = "eth_syncing")]
+    async fn eth_syncing(&self) -> RpcResult<EthSyncStatus>;
```
jfldde, line 183: Why not returning method not found here instead ?

jfldde: As in, not having the method at all

ercecan (approved):

crates/sequencer/src/rpc.rs
rakanalh: @eyusufatik requirements says that we should define the method in the sequencer and return head block info

eyusufatik (approved):

PR #2948, opened 2025-10-01
Make method id network identifier chain id

ercecan:
# Description
We now use chain id instead of a random network identifier for batch proof method id

### DIFF
--- a/crates/bitcoin-da/tests/test_utils.rs
+++ b/crates/bitcoin-da/tests/test_utils.rs
@@ -150,7 +150,7 @@ pub async fn generate_mock_txs(
     let method_id_body = BatchProofMethodIdBody {
         method_id: [0; 8],
         activation_l2_height: 0,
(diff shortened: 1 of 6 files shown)


crates/bitcoin-da/tests/test_utils.rs
```diff
@@ -537,12 +537,18 @@ pub mod macros {
 }
 
 // This will be removed in nightly since we moved these tests under binary
```
jfldde, line 539: Is that still relevant ?

jfldde (approved):

crates/bitcoin-da/tests/test_utils.rs
ercecan: It will be gone, after cherry picking to nightly, keeping it to not break things

eyusufatik (approved):

PR #2956, opened 2025-10-03
feat: Implement geth filter rpc endpoints

### LINKED ISSUE #2924: feat: additional Geth RPC endpoints support
**I need more endpoints**

While trying to build another project that leverages Citrea's EVM, I discovered that Citrea's node RPC lacks `eth_newFilter` endpoint.

**I do like wider support of standard Geth RPC**
Here is the example of endpoint I'm reporting about
https://www.quicknode.com/docs/ethereum/eth_newFilter

**Alternatives**
I'd have to re-engineer software I am currently making to be working with Citrea dev node.

**Additional context**
I reduced my issue to a mere standard example in `rust-web3`.
 
https://github.com/tomusdrw/rust-web3/blob/master/examples/contract_log_filter.rs

It also indicated lack of support `eth_accounts` endpoint whereas I know that Citrea supports standard regtest accounts and fills them with test tokens. However, the `rust-web3` example code fails to execute.

as of now, i could summarize feature request with two additional RPC endpoints:

- [ ] `eth_newFilter`
- [ ] `eth_accounts`

ercecan:
# Description
Implements filter changes related geth rpc endpoints.
Note: pending transactions related filters are not supported

## Linked Issues
- Fixes #2924 
- Related to # (issue)


crates/evm/src/rpc_helpers/filter.rs
```diff
@@ -115,3 +131,290 @@ pub fn convert_block_number(
...
+            active_filters: ActiveFilters::new(),
+            id_provider: Arc::new(EthSubscriptionIdProvider::default()),
+            task_executor,
+            stale_filter_ttl: stale_filter_ttl
+                .map(|d| Duration::from_secs(d as u64))
+                .unwrap_or_else(|| DEFAULT_STALE_FILTER_TTL),
+        };
+
+        let this = citrea_filter.clone();
+        tracing::info!("AAStarting stale filter clearing task ");
```
rakanalh:
```suggestion
        tracing::info!("Starting stale filter clearing task ");
```

Should this be info?

resources/configs/bitcoin-regtest/sequencer_rollup_config.toml
```diff
@@ -24,3 +24,4 @@ bind_host = "127.0.0.1"
 bind_port = 12345
 enable_subscriptions = true
 max_subscriptions_per_connection = 100
+stale_filter_ttl = 300
```
rakanalh, line 27: I don't see where we read this value through env vars just like all other configs.

ercecan: It's in RpcConfig

crates/evm/src/rpc_helpers/filter.rs
ercecan: No, forgot to change it to trace

crates/ethereum-rpc/src/lib.rs
```diff
@@ -701,25 +768,21 @@ where
...
-        gas_price_oracle_config,
-        fee_history_cache_config,
-    } = eth_rpc_config;
-
     // If the node does not have a sequencer client, then it is the sequencer.
     let is_sequencer = sequencer_client_url.is_none();
     let enable_subscriptions = l2_block_rx.is_some();
 
     // If the running node is a full node rpc context should also have sequencer client so that it can send txs to sequencer
+    tracing::info!("ttl: {:?}", rpc_config.stale_filter_ttl);
```
jfldde: Seems high, trace instead ?

crates/evm/src/rpc_helpers/filter.rs
```diff
@@ -115,3 +131,290 @@ pub fn convert_block_number(
...
+        task_executor: reth_tasks::TaskExecutor,
+        stale_filter_ttl: Option<usize>,
+    ) -> CitreaFilter {
+        let citrea_filter = CitreaFilter {
+            active_filters: ActiveFilters::new(),
+            id_provider: Arc::new(EthSubscriptionIdProvider::default()),
+            task_executor,
+            stale_filter_ttl: stale_filter_ttl
+                .map(|d| Duration::from_secs(d as u64))
+                .unwrap_or_else(|| DEFAULT_STALE_FILTER_TTL),
```
jfldde:
```suggestion
                .map_or(DEFAULT_STALE_FILTER_TTL, |d| Duration::from_secs(d as u64)),
```

crates/evm/src/rpc_helpers/filter.rs
```diff
@@ -115,3 +131,290 @@ pub fn convert_block_number(
...
+        };
+
+        let this = citrea_filter.clone();
+        tracing::trace!("Starting stale filter clearing task ");
+        citrea_filter.task_executor.spawn_critical(
```
jfldde, line 202: Shouldn't be critical, any panic in a critical task propagates and shutdown the node.

crates/evm/src/rpc_helpers/filter.rs
```diff
@@ -115,3 +131,290 @@ pub fn convert_block_number(
...
+
+    /// Returns all currently active filters
+    pub fn active_filters(&self) -> &ActiveFilters {
+        &self.active_filters
+    }
+
+    /// Endless future that [`Self::clear_stale_filters`] every `stale_filter_ttl` interval.
+    /// Nonetheless, this endless future frees the thread at every await point.
+    async fn watch_and_clear_stale_filters(&self) {
+        tracing::info!(
```
jfldde:
```suggestion
        tracing::trace!(
```

crates/evm/src/rpc_helpers/filter.rs
```diff
@@ -115,3 +131,290 @@ pub fn convert_block_number(
...
+        // start_block is the block from which we should start fetching changes, the next block from
+        // the last time changes were polled, in other words the best block at last poll + 1
+        let (start_block, kind) = {
+            let mut filters = self.active_filters.inner.lock().await;
+            let filter = filters
+                .get_mut(&id)
+                .ok_or(EthFilterError::FilterNotFound(id.clone()))?;
+
+            if matches!(filter.kind, FilterKind::PendingTransaction) {
+                self.uninstall_filter(id).await.ok();
```
jfldde:
```suggestion
                let _ = self.uninstall_filter(id).await;
```

crates/evm/src/rpc_helpers/filter.rs
```diff
@@ -115,3 +131,290 @@ pub fn convert_block_number(
...
+            let mut block = latest_block_number + 1;
+            std::mem::swap(&mut filter.block, &mut block);
+            filter.last_poll_timestamp = Instant::now();
+
+            (block, filter.kind.clone())
+        };
+
+        match kind {
+            // Pending transaction filters are not supported
+            FilterKind::PendingTransaction => Ok(FilterChanges::Empty),
```
jfldde:
```suggestion
            FilterKind::PendingTransaction => return Err(EthApiError::Unsupported(
                    "Pending transaction filters are not supported",
                )
                .into());,
```

crates/evm/src/rpc_helpers/filter.rs
```diff
@@ -115,3 +131,290 @@ pub fn convert_block_number(
...
+                // [start_block..latest_block_number]
+                let end_block = latest_block_number;
+                let block_hashes = evm
+                    .sealed_headers_range(start_block..=end_block, working_set)
+                    .map_err(|_| {
+                        EthApiError::HeaderRangeNotFound(start_block.into(), end_block.into())
+                    })?
+                    .iter()
+                    .map(|h| h.header().hash_slow())
+                    .collect::<Vec<_>>();
```
jfldde, line 353: We should have the hash here (from the sealed_headers_range). Can call `hash()` directly instead of `header().hash_slow()`

crates/evm/src/rpc_helpers/filter.rs
ercecan: This may cause oom in the long run if fails, so might be a good idea to keep it critical

crates/ethereum-rpc/src/ethereum.rs
```diff
@@ -36,24 +36,26 @@ pub struct Ethereum<C: sov_modules_api::Context, Da: DaService> {
...
+    pub(crate) citrea_filter: Arc<CitreaFilter>,
 }
 
 impl<C: sov_modules_api::Context, Da: DaService> Ethereum<C, Da> {
     #[allow(clippy::too_many_arguments)]
     pub(crate) fn new(
         da_service: Arc<Da>,
-        gas_price_oracle_config: GasPriceOracleConfig,
-        fee_history_cache_config: FeeHistoryCacheConfig,
+        eth_rpc_config: EthRpcConfig,
```
kpp, line 48: How come we pass eth_rpc_config and stale_filter_ttl separately from each other?

crates/evm/src/rpc_helpers/filter.rs
```diff
@@ -115,3 +131,287 @@ pub fn convert_block_number(
...
+        tracing::info!(target: "rpc::eth", "clear stale filters");
+        self.active_filters()
+            .inner
+            .lock()
+            .await
+            .retain(|id, filter| {
+                let is_valid = (now - filter.last_poll_timestamp) < self.stale_filter_ttl;
+
+                if !is_valid {
+                    tracing::trace!(target: "rpc::eth", "evict filter with id: {:?}", id);
```
kpp: Please consider using `extract_if` to collect items that are removed, unlock active_filters and then report in trace evicted filters. This is handful to reduce lock time

crates/evm/src/rpc_helpers/filter.rs
```diff
@@ -115,3 +131,287 @@ pub fn convert_block_number(
     };
     Ok(Some(num))
 }
+
+/// All active filters
+#[derive(Debug, Clone, Default)]
+pub struct ActiveFilters {
+    inner: Arc<Mutex<HashMap<FilterId, ActiveFilter>>>,
```
kpp: Using Mutex will lock rpc on concurrent access for a small portion of time. consider RwLock

kpp (dismissed):

crates/evm/src/rpc_helpers/filter.rs
rakanalh: In our version of Rust, 1.85, extract_if is unstable.

jfldde (dismissed):

crates/evm/src/rpc_helpers/filter.rs
```diff
@@ -115,3 +131,287 @@ pub fn convert_block_number(
...
+
+        let id = match subscription_id {
+            SubscriptionId::Num(n) => FilterId::Num(n),
+            SubscriptionId::Str(s) => FilterId::Str(s.into_owned()),
+        };
+        let mut filters = self.active_filters.inner.lock().await;
+        filters.insert(
+            id.clone(),
+            ActiveFilter {
+                block: last_poll_block_number,
```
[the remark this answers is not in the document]
rakanalh: This is not an off-by-one bug. The current code is correct and matches Reth's implementation because when you install a filter at block N, you should get block N's changes on the first eth_getFilterChanges call.

The flow is:
1. Install filter at block N: filter.block = N
2. First getFilterChanges call: returns changes from block N, then updates filter.block = latest + 1
3. Next call: returns changes from the stored block onwards

If we initialize with last_poll_block_number + 1, the first call would skip block N entirely, which breaks things.

Tests confirm this is correct. With +1, test_eth_get_logs_and_filter_proofs and test_filter_changes both fail because filters return Empty when they shouldn't.

crates/ethereum-rpc/src/ethereum.rs
rakanalh: Merged

jfldde (dismissed):

resources/configs/bitcoin-regtest/sequencer_rollup_config.toml
eyusufatik, line 1: let's put the config in other configs also

crates/common/src/config/rpc.rs
```diff
@@ -133,6 +143,9 @@ impl FromEnv for RpcConfig {
             trace_chain_block_limit: read_env("RPC_TRACE_CHAIN_BLOCK_LIMIT")
                 .ok()
                 .and_then(|val| val.parse().ok()),
+            stale_filter_ttl: read_env("RPC_STALE_FILTER_TTL")
```
eyusufatik, line 146: this wasn't added to display

bin/citrea/tests/mock/filter.rs
```diff
@@ -0,0 +1,218 @@
...
+use tokio::time::sleep;
+
+use super::evm::init_test_rollup;
+use crate::common::helpers::{
+    create_default_rollup_config, start_rollup, tempdir_with_children, wait_for_l2_block, NodeMode,
+};
+use crate::common::TEST_DATA_GENESIS_PATH;
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_filter_changes() -> Result<(), anyhow::Error> {
```
eyusufatik, line 15: is the changes response per block? so in this test its emtpy response for 6 blocks? should we sprinkle some event generating txs in blocks and assert them here?

crates/evm/src/rpc_helpers/filter.rs
```diff
@@ -115,3 +131,293 @@ pub fn convert_block_number(
...
+                kind,
+            },
+        );
+        Ok(id)
+    }
+
+    /// Uninstalls a filter with the given id. Returns true if the filter was found and removed,
+    /// false otherwise.
+    pub async fn uninstall_filter(&self, id: FilterId) -> RpcResult<bool> {
+        tracing::trace!(target: "rpc::eth", "Serving eth_uninstallFilter");
```
eyusufatik: we already have serving responding rpc methods logs, please remove

crates/evm/src/rpc_helpers/filter.rs
```diff
@@ -115,3 +131,293 @@ pub fn convert_block_number(
...
+        if filters.remove(&id).is_some() {
+            tracing::trace!(target: "rpc::eth::filter", ?id, "uninstalled filter");
+            Ok(true)
+        } else {
+            Ok(false)
+        }
+    }
+
+    /// Returns all the filter changes for the given id, if any
+    pub async fn filter_changes<C: sov_modules_api::Context>(
```
eyusufatik, line 297: I am a bit concerned about the DOS possibilities here. what happens if somebody installs a filter then doesn't call it for 100k blocks? are we protected by these type of attacks through the stale ttl config?

also we had max logs per response config, does that apply here already? how will such a thing work with this api?

eyusufatik (changes requested):

bin/citrea/tests/mock/filter.rs
ercecan: This test is done using a default filter, just to see if the functionality of filtering and stale ttl works or not, a log filter test can be found in `bin/citrea/tests/mock/evm/mod.rs` `test_getlogs` function. In that test, different filters with different block ranges, addresses, topics etc. are tested and compared against eth_getLogs results.

crates/evm/src/rpc_helpers/filter.rs
ercecan: > if somebody installs a filter then doesn't call it for 100k blocks?

Then the filter will be timed out and will be removed from cache

> also we had max logs per response config, does that apply here already?

Yes it does, it basically calls the same functions as `eth_getLogs` that's why the bounds checks are the same

eyusufatik: wouldn't it be a problem for the filter user  though? how does reth handle this?

eyusufatik (dismissed):

crates/evm/src/rpc_helpers/filter.rs
jfldde: This could be re-assessed

jfldde (dismissed):

crates/evm/src/rpc_helpers/filter.rs
ercecan: [9fa0ef7](https://github.com/chainwayxyz/citrea/pull/2956/commits/9fa0ef7a344d6322a108b2990b2bd8cc940caeab)

eyusufatik (approved):

jfldde (approved):

PR #2957, opened 2025-10-04
Remove unused `pruning/components/mod.rs` module

viktorking7:
## Summary
Removes dead code module that references non-existent submodules and is no longer used.

## Changes
- Deleted `crates/storage-ops/src/pruning/components/mod.rs`
- Module contained references to non-existent `ledger_db`, `native_db`, `state_db` modules
- Actual pruning logic lives in `pruning/ledger`, `pruning/native.rs`, `pruning/state.rs`


eyusufatik:
we will go back to working on it soon.

PR #2959, opened 2025-10-07
chore(tests): Replace complex generate_block flow

jfldde:
# Description

- Replace complex `generate_block` flow with regular `generate` instead. Order of tx doesn't affect the test and `generate` pattern is the go-to in simpler case.


kpp (approved):

eyusufatik:
i think this was intentional for this test

jfldde:
> i think this was intentional for this test

Some tests were partially including txs in block with `generate_block` and I kept as it but this one shouldn't change the behaviour. All txs are included in the first block in the same order

eyusufatik (approved):
yes you are right

PR #2942, opened 2025-09-30
feat: Temporary cap bitcoin-da fee rate

jfldde:
# Description

- Cap bitcoin-da fee rate via config. Defaults to 15 sat/vb

## Linked Issues
- Fixes #2801

Issue #2930, opened 2025-09-29
Flaky test: test_queue_da_transactions_oldest_mode


rakanalh:
Failure on nightly:
https://github.com/chainwayxyz/citrea/actions/runs/18006857276/job/51250164664

jfldde:
Duplicate of #2744

PR #2960, opened 2025-10-08
feat(bitcoin-da): Job service

### LINKED ISSUE #2801: feat: batch proof sending stops over 15 sat/vbyte

### LINKED ISSUE #2841: bitcoin-da Job service

### LINKED ISSUE #2999: Bypass commit transaction change unsafeness
Current commit vout 1 is tagged as unsafe since it's parent is the reveal which isn't part of the wallet.
Need to still pick this change output as available utxo so we can chain for longer while running in chained mode

jfldde:
# Description

- Fixes the stale fee issue and utxo selection by funding the tx at the time of sending instead of the time of queueing.
- Replaces two queues with a single persistent job storage.

# Improvements

- A proving job id now has an associated da job id. A partial chunked proof can now be recovered via da job id and doesn't have to be fully resent. This saves re-sending already sent chunks. 
- Each da job now uses the current bitcoin fee rate for sending transaction. Txs are funded on the same block they are sent. This should fix the stale fee issue https://github.com/chainwayxyz/citrea/issues/2801
- Da queue is now persistent and handles recovery of pending and in-progress da jobs

## Status
- [x] Job submission and tracking
- [x] Partial sending of chunked transactions  
- [x] Recovery (missing tests)

# Missing 
- [x] Job iteration by status @jfldde 
- [x] Prevent concurrent job processing @jfldde 
- [x] RPC (Admin) @jfldde 
  - [x] Cancel a job
  - [x] Retry a job
  - [x] List jobs
  - [x] Protect with API_KEY
- [x] Fee rate per job @jfldde 
- [x] Cap fee rate, see #2942  @jfldde 
- [x] Improve UTXO selection for chained mode @jfldde 
- [x] metrics
- [x] Handle `finalize_proving_job`
- [x] Database job schemas (@rakanalh )
- [x] Tests
  - [x] RPCs
  - [x] recovery
  - [ ] Parallel proof sending
- [x] rollback handling (@rakanalh)
- [ ] documentation
- [x] Tame disk usage (de-duplicate `RawTxData`)
- [x] Fix #2999
- [x] fix queue_oldest_mode flakiness #2930 && #2744

## Linked Issues
- Fixes #2841 
- fixes #2801


crates/sovereign-sdk/rollup-interface/src/state_machine/da.rs
```diff
@@ -102,13 +104,16 @@ impl core::cmp::Ord for SequencerCommitment {
 }
 
 /// Transaction request to send to the DA queue.
+#[cfg(feature = "native")]
 #[allow(clippy::large_enum_variant)]
 #[derive(Debug, Clone, Eq, PartialEq, BorshDeserialize, BorshSerialize)]
 pub enum DaTxRequest {
```
eyusufatik: maybe we can move this to node/services/da.rs

rakanalh (dismissed):
LGTM

kpp (dismissed):

crates/sovereign-sdk/rollup-interface/src/state_machine/da.rs
```diff
@@ -102,13 +104,16 @@ impl core::cmp::Ord for SequencerCommitment {
...
     /// A commitment from the sequencer
     SequencerCommitment(SequencerCommitment),
     /// Or a zk proof and state diff
     ZKProof(Proof),
+    /// Or a job id for a stored proof
+    StoredProof(Uuid),
```
eyusufatik: should we remove the ZKProof item then?

crates/sequencer/src/commitment/service.rs
```diff
@@ -235,10 +235,10 @@ where
         let start = Instant::now();
         let ledger_db = self.ledger_db.clone();
 
-        let _tx_id = rx
+        let _txid = rx
             .await
-            .map_err(|_| anyhow!("DA service is dead!"))?
-            .map_err(|_| anyhow!("Send transaction cannot fail"))?;
+            .map_err(|_| anyhow!("DA notification channel closed"))? // Handle RecvError
+            .map_err(|e| anyhow!("DA job failed: {e}"))?;
```
eyusufatik: if we are not doing anything with txid now, do we even need to wait?

imo the sequencer can just save it as sent. maybe it can include the da job id in it's saved data for sequencer commitment as well. if you don't want to save it, please log it here along with index and start-end height

crates/sequencer/src/commitment/service.rs
```diff
@@ -235,10 +235,10 @@ where
         let start = Instant::now();
         let ledger_db = self.ledger_db.clone();
 
-        let _tx_id = rx
+        let _txid = rx
```
eyusufatik, line 238: can you also update "New commitment. L2 range: #{}-{}" to include the index?

crates/sovereign-sdk/adapters/mock-da/src/service.rs
```diff
@@ -450,21 +480,10 @@ impl DaService for MockDaService {
...
-        let this = self.clone();
-        tokio::spawn(async move {
-            while let Some(req) = rx.recv().await {
-                let res = this.send_transaction(req.tx_request).await;
-                let _ = req.notify.send(res);
-            }
-        });
-        tx
+        let _ = tx.send(Ok(MockHash([0; 32])));
+        Ok((Uuid::nil(), rx))
```
eyusufatik, line 486: why nil?

crates/sovereign-sdk/adapters/mock-da/src/service.rs
```diff
@@ -583,8 +609,8 @@ mod tests {
             let block_3_before = da.get_block_at(3).await.unwrap();
 
             // Disabling this check because our modified mock da creates blocks when a transaction is sent
-            // let result = da.get_block_at(4).await;
-            // assert!(result.is_err());
+            let result = da.get_block_at(4).await;
+            assert!(result.is_err());
```
eyusufatik, line 613: remove above comment if this works now

crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs
```diff
@@ -501,6 +512,26 @@ define_table_with_seek_key_codec!(
...
+);
+
+define_table_with_seek_key_codec!(
+    /// Da job progress by uuid
+    (DaJobProgressById) Uuid => JobProgress
+);
+
+define_table_with_seek_key_codec!(
+    /// Index by (status, jobid)
+    (DaJobStatusIndex) (u8, Uuid) => ()
```
eyusufatik, line 543: can you explain how this is used?

crates/sovereign-sdk/full-node/db/sov-db/src/schema/types/da_jobs.rs
```diff
@@ -0,0 +1,95 @@
...
+impl DaJobStatus {
+    /// u8 representation of `DaJobStatus`
+    pub fn as_u8(&self) -> u8 {
+        match self {
+            DaJobStatus::Pending => 0,
+            DaJobStatus::InProgress => 1,
+            DaJobStatus::Completed => 2,
+            DaJobStatus::Cancelled => 3,
+            DaJobStatus::Failed { .. } => 4,
+        }
```
eyusufatik, line 35: can't you already do #[repr(u8)] on the enum and that as u8 would yield the value?

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
```diff
@@ -996,3 +1015,84 @@ impl ForkMigration for LedgerDB {
         Ok(())
     }
 }
+
+impl DaLedgerOps for LedgerDB {
+    fn submit_job(
+        &self,
+        job_id: Uuid,
+        da_tx_request: &DaTxRequest,
+        progress: &JobProgress,
```
eyusufatik, line 1074: I guess the JobProgress should always be new here, so do we need the argument here?

crates/batch-prover/src/prover.rs
```diff
@@ -821,23 +836,48 @@ where
...
+                .expect("DB call shouldn't fail")
+            {
+                info!(
+                    "DA job {} already exists for proving job {}",
+                    da_job_id, proving_job_id
+                );
+                prover_service
+                    .wait_for_existing_da_job(da_job_id)
+                    .await
+                    .expect("Should recover da job receiver")
```
eyusufatik, line 859: i think you should use a tokio::spawn for this

or else we'll wait on all recoveries sequentially, which will take a long time

eyusufatik: or maybe this only returns the receiver for the txid?

crates/batch-prover/src/rpc.rs
```diff
@@ -656,15 +683,73 @@ where
...
+            .await
+            .map_err(internal_rpc_error)?;
+
+        ledger_db
+            .set_da_job_id_by_prover_job_id(proving_job_id, da_job_id)
+            .expect("Failed to save da job by id");
+
+        info!("Submitted proof from file, da job id: {da_job_id}");
+
+        Ok(da_job_id)
```
eyusufatik, line 823: should we just passed the output as the input for the rpc call maybe?

crates/common/src/rpc/auth.rs
```diff
@@ -8,7 +8,13 @@ use jsonrpsee::MethodResponse;
 use serde_json::value::RawValue;
 use serde_json::Value;
 
-const PROTECTED_METHODS: [&str; 3] = ["backup_create", "backup_validate", "backup_info"];
+const PROTECTED_METHODS: [&str; 5] = [
+    "backup_create",
+    "backup_validate",
+    "backup_info",
+    "da_job_cancel",
+    "da_job_retry",
```
eyusufatik: we don't really need to put anything that's not available on full nodes here

crates/bitcoin-da/src/helpers/mod.rs
```diff
@@ -68,6 +71,15 @@ impl TransactionKind {
...
 }
 
+/// Return UNIX timestamp in seconds
+#[cfg(feature = "native")]
+pub(crate) fn get_timestamp() -> u64 {
+    SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .expect("Cannot fail because there is always a UNIX epoch")
+        .as_secs()
+}
```
eyusufatik, line 81: why is this here?

bin/citrea/src/rollup/bitcoin.rs
```diff
@@ -85,6 +82,9 @@ impl RollupBlueprint for BitcoinRollup {
         let da_methods = create_da_rpc_module(da_service.clone());
         rpc_methods.merge(da_methods)?;
 
+        let da_methods = create_da_job_rpc_module(da_service.clone());
+        rpc_methods.merge(da_methods)?;
```
eyusufatik, line 86: should we really separate these two?

crates/bitcoin-da/src/job/rpc.rs
```diff
@@ -0,0 +1,321 @@
...
+
+    /// List jobs with optional filtering and pagination
+    ///
+    /// # Arguments
+    /// * `filter` - Optional filter criteria for jobs
+    ///
+    /// # Returns
+    /// * `Ok(Vec<JobInfoResponse>)` - List of jobs matching the filter criteria
+    /// * `Err` on database or serialization errors
+    fn list_jobs(&self, filter: JobListFilter) -> Result<Vec<JobProgress>>;
```
eyusufatik, line 47: I think if you don make JobListFilter Optional here, we would need to call this job like this:

```json
{
    "method": "daJob_listJobs",
    "params": [{}]
}
```

but maybe it works without the {}

crates/bitcoin-da/src/job/rpc.rs
```diff
@@ -0,0 +1,321 @@
...
+
+impl Default for JobListFilter {
+    fn default() -> Self {
+        Self {
+            status: None,
+            limit: Some(25),
+            offset: None,
+        }
+    }
+}
```
eyusufatik, line 79: will the param default to this Default impl.? if not, let's make it happen, bc if we accidentally called it for all proofs etc. it will be problematic :D

crates/bitcoin-da/src/job/rpc.rs
```diff
@@ -0,0 +1,321 @@
...
+    ///
+    /// # Arguments
+    /// * `job_id` - The unique identifier of the job
+    ///
+    /// # Returns
+    /// * Detailed job information including status, timestamps, and progress
+    ///
+    /// # Errors
+    /// * Database error related errors
+    #[method(name = "get")]
```
eyusufatik: imo, add ByJobId suffix to all these methods

crates/bitcoin-da/src/service.rs
```diff
@@ -1317,26 +1366,68 @@ impl DaService for BitcoinService {
...
-                notify: tx,
-            })
-            .map_err(|_| BitcoinServiceError::ChannelSendError)?;
-        Ok(rx.await?.expect("Queue never sends error"))
+    ) -> Result<(Uuid, oneshot::Receiver<Result<TxidWrapper>>)> {
+        let (tx, rx) = oneshot::channel();
+        let job_id = {
+            let job_service = self.job_service.lock().await;
+
+            // TODO handle chaining job request
```
eyusufatik, line 1281: todo

crates/bitcoin-da/src/service.rs
```diff
@@ -1317,26 +1366,68 @@ impl DaService for BitcoinService {
...
+            if self.utxo_selection_mode == UtxoSelectionMode::Chained {
+                let active_jobs = job_service.get_all_active_job_ids()?;
+                if !active_jobs.is_empty() {
+                    return Err(BitcoinServiceError::PreviousJobInProgress);
+                }
```
eyusufatik, line 1286: imo, in chained mode, if no active jobs, basically the last sent job should be the `prev_tx`

crates/bitcoin-da/src/service.rs
```diff
@@ -1317,26 +1366,68 @@ impl DaService for BitcoinService {
...
+
+        let (tx, rx) = oneshot::channel();
+
+        match progress.status {
+            DaJobStatus::Completed => {
+                // Job already finished before we subscribed
+                if let Some(last_tx) = progress.sent_chunks.reveal_txs.last() {
+                    let _ = tx.send(Ok(TxidWrapper(Txid::from_byte_array(*last_tx))));
+                } else {
+                    let _ = tx.send(Err(JobServiceError::NoTransactionsFound(job_id).into()));
```
eyusufatik: wouldn't this be a panic level inconsestincy? or should we keep going after this? not sure. let's discuss

crates/sovereign-sdk/full-node/db/sov-db/src/schema/types/da_jobs.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// Job is in progress
+    InProgress,
+    /// Job completed successfully
+    Completed,
+    /// Job was cancelled before completion
+    Cancelled,
+    /// Job failed with error
+    Failed {
+        /// Error associated with the failure
+        error: String,
```
eyusufatik, line 22: how can a job really fail?

crates/bitcoin-da/src/job/service.rs
```diff
@@ -0,0 +1,341 @@
...
+                Ok(RawTxData::BatchProofMethodId(blob))
+            }
+        }
+        .context("Failed to retrieve RawTxData from DaTxRequest")?;
+
+        self.raw_tx_data_cache
+            .lock()
+            .unwrap()
+            .push(job_id, raw_tx_data.clone());
+
```
eyusufatik, line 131: seems like this cache is a bit useless? rocksdb already caches the value when you query it

crates/bitcoin-da/src/job/service.rs
```diff
@@ -0,0 +1,341 @@
...
+    pub(crate) fn get_pending_chunks(&self) -> Result<HashSet<Txid>> {
+        let mut txids = HashSet::new();
+
+        let active_job_ids = self.get_all_active_job_ids()?;
+        for job_id in active_job_ids {
+            if let Some(JobProgress {
+                status: DaJobStatus::InProgress,
+                sent_chunks,
+                ..
+            }) = self.get_progress(&job_id)?
```
eyusufatik, line 203: again if we can't get job here, a big inconsestency in the system, no?

at least log it imo

crates/sovereign-sdk/full-node/db/sov-db/src/schema/types/da_jobs.rs
```diff
@@ -0,0 +1,95 @@
...
+pub enum DaJobStatus {
+    /// Job is queued and waiting to be processed
+    Pending,
+    /// Job is in progress
+    InProgress,
+    /// Job completed successfully
+    Completed,
+    /// Job was cancelled before completion
+    Cancelled,
+    /// Job failed with error
```
eyusufatik: please clarify the JobStatuses. 

For example, when is the job completed? when it reaches bitcoin mempool?

crates/bitcoin-da/src/service.rs
```diff
@@ -489,8 +618,7 @@ impl BitcoinService {
 
     /// Returns the UTXO with the highest number of confirmations
     #[instrument(level = "trace", skip_all, ret)]
-    async fn get_highest_confirmation_utxo(&self) -> Result<Option<UTXO>> {
-        let mut utxos = self.get_utxos().await?;
+    async fn get_highest_confirmation_utxo(&self, mut utxos: Vec<UTXO>) -> Result<Option<UTXO>> {
```
eyusufatik: why did you pass this down?

bin/citrea/tests/bitcoin/bitcoin_test.rs
```diff
@@ -489,89 +489,89 @@ async fn test_cpfp_fee_bump() -> Result<()> {
...
-}
-
-#[tokio::test]
-async fn test_min_relay_fee_handling() -> Result<()> {
-    TestCaseRunner::new(MinRelayFeeTest)
-        .set_citrea_path(get_citrea_path())
-        .run()
-        .await
-}
+// struct MinRelayFeeTest;
```
eyusufatik, line 494: why is this gone?

bin/citrea/tests/bitcoin/mod.rs
```diff
@@ -15,6 +15,8 @@ pub mod bitcoin_service;
 pub mod bitcoin_test;
 pub mod bitcoin_verifier;
 #[cfg(feature = "testing")]
```
eyusufatik, line 17: why do we have these feature flags in this file? isn't this already a testing module?

bin/citrea/tests/bitcoin/da_queue.rs
```diff
@@ -533,27 +522,25 @@ impl DaTransactionQueueingUtxoSelectionModeOldestTest {
...
         assert!(res.is_ok());
 
         da.wait_mempool_len(18 * 2, None).await?;
         assert_eq!(da.get_raw_mempool().await?.len(), 18 * 2);
 
-        // Assert that all queued txs are monitored
+        // Assert that all sent txs are monitored
         let monitored_txs = da_service.monitoring.get_monitored_txs().await;
-        assert_eq!(monitored_txs.len(), 88);
+        assert_eq!(monitored_txs.len(), 76);
```
eyusufatik, line 507: why did this change?

crates/bitcoin-da/src/monitoring.rs
eyusufatik, line 1: do we still need this?

we could extend DaJobService with JobTypes:
- Send
- SendForChunked
- Monitor

the last job type would be repeated until the tx it monitors is finalized for example.

crates/bitcoin-da/src/service.rs
eyusufatik, line 1: do you think we should let the job service to handle its own loop?

the loop here would be simplified a lot.

could be a good refactor.

eyusufatik: another general commetn I have is the DaJobService is now designed to handle a single job.

I propose something like what's mentioned in this comment: https://github.com/chainwayxyz/citrea/pull/2960#discussion_r2494156449

wdyt?

the chunk sending can also be simplified that way.

when we need to send chunks, we put n of:

```rust
ChunkJob {
    prev_chunk_job_id: _,
    next_chunk_job_id: _,
    this_chunk: Vec<u8>, // just data, no txs yet
}
```

and a single:

```rust
AggregateJob {
    chunk_job_ids: Vec<JobId>
}
```


The completes and commitments would be a regular job, and then the JobService decides on how to handle the both.

I understand monitoring job would be harder to write as of now, because it can't really be a chronogically ordered job, it's better in a priority queue. so it can stay seperate but imo separating chunk and normal tx sending to different job types make sense because I really had a hard time reading it, and cannot attest to its correctnes as of now :).

bin/citrea/tests/bitcoin/da_job.rs
eyusufatik, line 1: will read tests after the initial review & fixes

eyusufatik (changes requested):

crates/sovereign-sdk/rollup-interface/src/state_machine/da.rs
jfldde: Kept `ZkProof` for test and backward compatiblity with `submit_fake_proof` for now.

crates/sovereign-sdk/adapters/mock-da/src/service.rs
jfldde: We don't use the job service in mock da context

crates/sovereign-sdk/full-node/db/sov-db/src/schema/types/da_jobs.rs
jfldde: Can't do repr(u8) as `Failed` variant holds data

crates/batch-prover/src/prover.rs
jfldde: Yes, this returns the receiver. I'll change the name as it's misleading

crates/bitcoin-da/src/helpers/mod.rs
jfldde: This is shared between multiple `bitcoin-da` services

bin/citrea/src/rollup/bitcoin.rs
jfldde: Not sure, I liked doing the PR this way but the resulting `daJob` namespace might be superfluous. I'll merge them

crates/bitcoin-da/src/job/rpc.rs
jfldde: That's the `DaJobRpcProvider` but the rpc trait handles the optional :
```
#[method(name = "list")]
    async fn da_job_list(
        &self,
        status: Option<JobStatusFilter>,
        limit: Option<usize>,
        offset: Option<usize>,
    ) -> RpcResult<Vec<JobInfoResponse>>;
    ```

File: crates/sovereign-sdk/full-node/db/sov-db/src/schema/types/da_jobs.rs (hunk 21 above)
Comment (jfldde) [reply crates/sovereign-sdk/full-node/db/sov-db/src/schema/types/da_jobs.rs:22]: That's something to discuss. I agree most error should be transient but we can imagine some scenario where the tx actually fail (missing utxo on an utxo that isn't pending rebroadcasting or tx builder failure ?). Not entirely sure we have valid persistent errors, I'll think about it and this could clean up a fair bit.

File: crates/bitcoin-da/src/job/service.rs (hunk 22 above)
Comment (jfldde) [reply crates/bitcoin-da/src/job/service.rs:131]: Rocksdb holds the `DaTxRequest` while this caches the `RawTxData`, that is already splitted for proof and borshed.

File: bin/citrea/tests/bitcoin/mod.rs (hunk 27 above)
Comment (jfldde) [reply bin/citrea/tests/bitcoin/mod.rs:17]: Following this PR https://github.com/chainwayxyz/citrea/pull/2547

File: bin/citrea/tests/bitcoin/da_queue.rs (hunk 28 above)
Comment (jfldde) [reply bin/citrea/tests/bitcoin/da_queue.rs:507]: Previously we monitored `Queued` tx (`TxStatus::Queued` variant is removed in this PR) that were not yet sent to mempool. Now, we only monitor tx that were actually sent (other are now track by the job service instead of the monitoring service)

File: crates/sovereign-sdk/full-node/db/sov-db/src/schema/tables.rs [hunk 29]
```diff
@@ -144,6 +152,9 @@ pub const LEDGER_TABLES: &[&str] = &[
     CommitmentIndicesByL1::table_name(),
     CommitmentMerkleRoots::table_name(),
     CommitmentsByNumber::table_name(),
+    DaTxRequestByJobId::table_name(),
```
ercecan, line 158:
```suggestion
    DaTxRequestByJobId::table_name(),
    DaJobIdByProvingJobId::table_name(),
```

crates/bitcoin-da/src/job/rpc.rs
jfldde: It defaults to 25 in `list_jobs` and is capped at 100

crates/bitcoin-da/src/service.rs
jfldde: That is the behaviour. This prevents having multiple pending job in progress while running in `Chained` mode

jfldde: As the previous behaviour to simplify utxo management

crates/bitcoin-da/src/monitoring.rs
jfldde: You mean the monitoring service as a whole ?

crates/sovereign-sdk/full-node/db/sov-db/src/ledger_db/mod.rs
jfldde: It should but just keeping the responsability of creating the JobProgress above and keep this `ledger_db` file clean

crates/bitcoin-da/src/job/service.rs
jfldde: Actually, this is a valid pattern where a job would be `InProgress` but no tx have been sent yet. (either just toggled to InProgress or has failed on mempool policy limit on first tx pair).

crates/sovereign-sdk/full-node/db/sov-db/src/schema/types/da_jobs.rs
jfldde: Keeping this for now for future-proofing though no failure are considered permanent yet.

crates/bitcoin-da/src/service.rs
jfldde: This is panicking at the caller site but we might want to panic here if it should never be recovered from instead.

crates/batch-prover/src/rpc.rs
jfldde: yeah, don't know what happened here... good call

PR #2963, opened 2025-10-12
Remove unused current_status param from attempt_rebroadcast

VolodymyrBg:
Drop the redundant current_status argument from attempt_rebroadcast, simplify the log, and update all call sites in monitoring.rs. No behavior change; improves clarity and eliminates dead parameter.


jfldde:
NACK. I'd rather keep the status in log

kpp:
Hello! Thank you for your PR.

`current_status` is omitted in the log so it changes the current behavior. Please feel free to open other PRs or even the current one with fixes to the noted issue.

PR #2965, opened 2025-10-13
feat: Restore pending jobs at startup and reset stale jobs

rakanalh:


crates/bitcoin-da/src/service.rs
```diff
@@ -257,6 +257,34 @@ impl BitcoinService {
...
     }
 
+    /// Restore pending jobs from the database and process them.
+    #[instrument(level = "info", skip_all)]
+    pub async fn restore_pending_jobs(&self) -> Result<()> {
+        info!("Restoring pending jobs from database");
+
+        let reset_count = self
+            .job_service
+            .reset_stale_jobs(Duration::from_secs(300))?;
```
rakanalh, line 267: the duration should be configurable.

PR #2967, opened 2025-10-13
ci: Run CI checks on all PRs

rakanalh:
# Description

In the very few cases where multiple people working on a feature create sub-PRs, we need those sub-PRs to actually run the CI checks before merging them into the feature PR.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -52,7 +52,7 @@ on:
   merge_group:
     types: ["checks_requested"]
   push:
-    branches: ["nightly", "devnet-freeze", "release-v*"]
+    branches: ["nightly", "release-v*"]
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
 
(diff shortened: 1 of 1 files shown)


eyusufatik:
I think this is already doable by going to the actions tab

rakanalh:
> I think this is already doable by going to the actions tab

Yea but having it happen automatically won't hurt since we don't have sub-PRs very often apart from two or more people are working on a huge feature PR.

eyusufatik (dismissed):
okay, I agree.

while at it can you also review devnet-freeze from branches field?

rakanalh:
> okay, I agree.
> 
> while at it can you also review devnet-freeze from branches field?

We won't need to change that since we only want to run checks on PRs but not the branches (if they don't have PRs).

Anything specific you want me to check regarding the devnet-freeze branch?

jfldde (dismissed):

eyusufatik (approved):

exeokan (approved):

PR #2972, opened 2025-10-14
feat: RISC0 config

rakanalh:
# Description

### DIFF
--- a/crates/risc0/src/host/config.rs
+++ b/crates/risc0/src/host/config.rs
@@ -105,27 +105,17 @@ pub enum BoundlessStorageConfig {
 }
 
 /// Configuration for the Boundless prover
-#[derive(Debug, Clone, Deserialize, Serialize)]
+#[derive(Debug, Clone)]
 pub struct BoundlessProverConfig {
-    /// Wallet private key for signing transactions
-    pub wallet_private_key: String,
-    /// RPC URL for the boundless network
-    #[serde(with = "url_serde")]
-    pub rpc_url: Url,
-    /// Whether to use offchain mode
-    #[serde(default)]
-    pub is_offchain: bool,
+    /// Boundless configuration
+    pub boundless: BoundlessConfig,
     /// Storage configuration
     pub storage: BoundlessStorageConfig,
 }
 
 impl citrea_common::FromEnv for BoundlessProverConfig {
     fn from_env() -> anyhow::Result<Self> {
-        let wallet_private_key = read_env("BOUNDLESS_WALLET_PRIVATE_KEY")?;
-        let rpc_url = read_env("BOUNDLESS_RPC_URL")?;
-        let is_offchain = read_env("BOUNDLESS_IS_OFFCHAIN")
-            .map(|v| v == "1" || v.to_lowercase() == "true")
-            .unwrap_or(false);
+        let boundless = BoundlessConfig::from_env()?;
 
         let storage = if let Ok(config) = BoundlessS3StorageConfig::from_env() {
             BoundlessStorageConfig::S3(config)
@@ -137,12 +127,7 @@ impl citrea_common::FromEnv for BoundlessProverConfig {
             ));
         };
 
-        Ok(Self {
-            wallet_private_key,
-            rpc_url: Url::parse(&rpc_url).context("Invalid RPC URL")?,
-            is_offchain,
-            storage,
-        })
+        Ok(Self { boundless, storage })
     }
 }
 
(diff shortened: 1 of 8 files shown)


crates/risc0/src/host/config.rs
```diff
@@ -127,7 +191,105 @@ impl citrea_common::FromEnv for BoundlessConfig {
...
+    Boundless(Box<BoundlessProverConfig>),
+}
+
+impl Default for ProverConfig {
+    fn default() -> Self {
+        Self::Local(LocalProverConfig::default())
+    }
+}
+
+impl citrea_common::FromEnv for ProverConfig {
```
eyusufatik: pls call this something else. suggestion `Risc0ProverConfig`

crates/risc0/src/host/config.rs
```diff
@@ -127,7 +191,105 @@ impl citrea_common::FromEnv for BoundlessConfig {
...
+}
+
+impl citrea_common::FromEnv for Risc0HostConfig {
+    fn from_env() -> anyhow::Result<Self> {
+        let prover = ProverConfig::from_env()?;
+        let dev_mode = std::env::var("RISC0_DEV_MODE")
+            .ok()
+            .map(|x| x.to_lowercase())
+            .filter(|x| x == "1" || x == "true" || x == "yes")
+            .is_some();
```
eyusufatik: there is now a function for this

eyusufatik: `is_dev_mode_enabled_via_environment`

crates/risc0/src/host/config.rs
```diff
@@ -70,32 +94,56 @@ impl citrea_common::FromEnv for BoundlessPinataStorageConfig {
...
+    /// RPC URL for the boundless network
+    #[serde(with = "url_serde")]
+    pub rpc_url: Url,
+    /// Whether to use offchain mode
+    #[serde(default)]
+    pub is_offchain: bool,
+    /// Storage configuration
+    pub storage: BoundlessStorageConfig,
+}
+
```
eyusufatik: seems to me that this could be BoundlessConfig + BoundlessStorageConfig?

and then this struct's FromEnv would be from_env() calls for underlying structs

crates/risc0/src/host/mod.rs
```diff
@@ -36,20 +37,34 @@ pub struct Risc0Host {
...
+                Prover::Bonsai(BonsaiProver::new(ledger_db, bonsai_config))
+            }
+            ProverConfig::Local(local_config) => {
+                Prover::Local(LocalProver::new(network, local_config))
             }
         };
 
+        if config.dev_mode {
+            env::set_var("RISC0_DEV_MODE", "1");
+        }
```
eyusufatik: i don't think we need this

crates/risc0/src/host/config.rs
```diff
@@ -127,7 +191,105 @@ impl citrea_common::FromEnv for BoundlessConfig {
...
+/// Configuration for Risc0Host
+#[derive(Debug, Clone, Deserialize, Serialize, Default)]
+pub struct Risc0HostConfig {
+    /// Prover config
+    #[serde(flatten)]
+    pub prover: ProverConfig,
+    /// Enable dev mode
+    #[serde(default)]
+    pub dev_mode: bool,
```
eyusufatik: seems to me that dev_mode is only used by LocalProver. so we don't need it here, let LocalProver handle it.

crates/risc0/src/host/config.rs
```diff
@@ -127,7 +191,105 @@ impl citrea_common::FromEnv for BoundlessConfig {
...
+    /// Optional directory to backup transaction inputs
+    pub tx_backup_dir: Option<PathBuf>,
```
eyusufatik, line 245: and then this was made into a config field, but is still being used with env::var calls in mod.rs.

I think we can remove this `Risc0HostConfig` altogether, then, make `ProverConfig` into "`Risc0HostConfig`".

eyusufatik (changes requested):

crates/risc0/src/host/config.rs
eyusufatik: talking about `tx_backup_dir`

rakanalh: I opted to keep the tx_backup_dir and pass it to the Risc0Host so that we use the config value instead of using the env var.

As for moving it into ProverConfig, it wouldn't make sense in this case since it's an enum of the 3 prover config variants and the backup_dir config is related to the Risc0Host.

eyusufatik (approved):

PR #2981, opened 2025-10-16
Remove sensitive params from RPC request logs to prevent API key leakage

vtjl10:
### Description
- Summary: Stop logging JSON-RPC request params to avoid leaking sensitive data (e.g., API keys) in logs.
- Motivation: Protected methods pass API key as the last param; logging params exposes credentials in log storage and aggregators.
- Changes:
  - Remove `params` from `rpc_request` debug logs in `Logger::call`.


jfldde:
Is that needed after #2760 ?

PR #2984, opened 2025-10-16
feat: DA job rollback

rakanalh:
Since jobs are not tied to block heights, we're deleting everything.

### DIFF
--- a/crates/storage-ops/src/rollback/node/sequencer.rs
+++ b/crates/storage-ops/src/rollback/node/sequencer.rs
@@ -1,8 +1,8 @@
 use std::sync::Arc;
 
 use sov_db::schema::tables::{
-    CommitmentsByNumber, L2BlockByHash, L2BlockByNumber, L2RangeByL1Height,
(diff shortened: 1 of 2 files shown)


crates/storage-ops/src/rollback/node/sequencer.rs
```diff
@@ -112,6 +112,37 @@ impl SequencerLedgerRollback {
         self.ledger_db.write_schemas(batch)?;
         Ok(rollback_result)
     }
+
+    fn rollback_da_jobs(&self, mut rollback_result: RollbackResult) -> Result {
```
kpp, line 116: Why do we have same ^C^V code? Maybe we can unify it? Or fck it and duplicate it?

rakanalh: You're right, i thought i i will have different logic for batch prover to filter by proof height or something but apparently that's too involved. I should unify as you suggested.

kpp (approved):

jfldde (approved):

PR #2990, opened 2025-10-21
fix: Limit rocksdb fd allocation

jfldde:
# Description

- Prevent crashing on macos with `Too many open files` while running ulimit default settings.
- Allocates only half of available fds to rocksdb.


rakanalh (commented):
How were the usage percentages decided?

jfldde:
> How were the usage percentages decided?

~Rough estimation from known usage between state/ledger/native dbs.
This PR is a safeguard when running without config set and default `ulimit` which is very low on macos. Prevents crashing in this case but it is adviced against running without higher ulimit and/or config~

Doesn't make much sense looking at my local node, I'll come back to this

PR #2992, opened 2025-10-21
fix: authorization list gas calculation

### LINKED ISSUE #2989: 7702 delegation and execution in same tx gas underestimation
## Summary

When we perform both a 7702 delegation and a regular transaction in the same tx, the gas is underestimated.

Example tx: https://explorer.testnet.citrea.xyz/tx/0xb3083c96053a046f85b5990b0eaeffd386f1d79a271aa8d9e0db7ba680a041fd

rakanalh:
# Description

## Linked Issues
- Fixes #2989


crates/evm/src/query.rs
```diff
@@ -1036,7 +1036,8 @@ impl<C: sov_modules_api::Context> Evm<C> {
             create_txn_env(&block_env, request, Some(account.balance), nonce, chain_id)?;
 
         // if the request is a simple transfer we can optimize
-        if tx_env.data.is_empty() {
+        // but not if there's an authorization list as it adds significant gas costs
+        if tx_env.data.is_empty() && tx_env.authorization_list.is_empty() {
```
eyusufatik: i don't fully understand how this solves the issue. I mean the situation described by @selinisik would fail to go in the `if let Ok((res, tx_info)) = res {`, resultin in the code runinng the rest of the function. so it seems this doesn't really change anything 🤔.

if it really does, then maybe, there is something wrong with how inspect_with_citrea_handler handles the tx here.

rakanalh: In the absence of data and authorization list:
```
if tx_env.data.is_empty() && tx_env.authorization_list.is_empty() {
```
This is considered to be a simple transfer.

Otherwise, we don't enter the whole of the block

Outside that if block, what happens is go into the block after where we call inspect_with_citrea_handler and calculate the proper gas for that transaction rather than returning MIN_TRANSACTION_GAS
```
running 1 test
GAS WITHOUT AUTH: 21681
GAS_WITH_AUTH: 46571
test tests::queries::estimate_gas_tests::test_eip7702_gas_with_authorization_and_call_data ... ok
```

You can see that with auth more than doubled
```
running 1 test
GAS WITHOUT DATA NOR AUTH: 21001
GAS_WITH_AUTH: 46571
test tests::queries::estimate_gas_tests::test_eip7702_gas_with_authorization_and_call_data ... ok
```

eyusufatik: what I'm trying to say is that the inspect in that if should have reverted/halted due to the gas limit not being enough, but apparently it's not.

eyusufatik:
also in @selinisik's case, she was using the same signer for delegating and sending the transaction, the test should be rewritten that way

crates/evm/src/query.rs
eyusufatik: furthermore, the if is already skipped bc the tx has data field

eyusufatik:
fixed by #3013

eyusufatik:
fixed by #3013

PR #2994, opened 2025-10-21
refactor: Add Risc0HostConfig to prover configs

### LINKED ISSUE #2986: Add Risc0HostConfig to the ProverConfig we read from toml files
make issue for also adding the Risc0HostConfig to the ProverConfig we read from toml files

_Originally posted by @eyusufatik in https://github.com/chainwayxyz/citrea/pull/2391#discussion_r2435878733_

exeokan:
# Description
Simplifies config structs for `Risc0HostConfig` and moves them to citrea common.
Includes `Risc0HostConfig` in batch prover and light client prover configs, enabling it to be read from toml files.
TODO:
- [x] Add to  citrea-e2e

Note: After this PR, to set `Risc0HostConfig` via env, you also need to set the whole batch prover/light client config via env as well. If a config file is provided for the latter and the `Risc0HostConfig` is not there, default will be used.

## Linked Issues
- Fixes #2986 
## Testing
Extended correct prover config tests.


crates/risc0/src/host/boundless.rs
```diff
@@ -95,19 +97,30 @@ impl BoundlessProver {
             BoundlessStorageConfig::Pinata(pinata_config) => StandardStorageProvider::Pinata(
                 PinataStorageProvider::from_parts(
                     pinata_config.pinata_jwt,
-                    pinata_config.pinata_api_url.to_string(),
-                    pinata_config.ipfs_gateway_url.to_string(),
+                    pinata_config.pinata_api_url,
+                    pinata_config.ipfs_gateway_url,
```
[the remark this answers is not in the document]
exeokan: provided context is enough

crates/common/src/config/mod.rs
```diff
@@ -175,6 +181,9 @@ pub struct LightClientProverConfig {
     pub enable_recovery: bool,
     /// The starting DA block to sync from
     pub initial_da_height: u64,
+    /// Configuration for Risc0Host
+    #[serde(default)]
+    pub risc0_host_config: Risc0HostConfig,
```
rakanalh:
```suggestion
    pub risc0_host: Risc0HostConfig,
```

Parent struct already has the word config

crates/common/src/config/risc0.rs
```diff
@@ -50,54 +45,32 @@ impl citrea_common::FromEnv for BoundlessS3StorageConfig {
...
     fn from_env() -> anyhow::Result<Self> {
         let pinata_jwt = read_env("BOUNDLESS_PINATA_JWT")?;
         let pinata_api_url = read_env("BOUNDLESS_PINATA_API_URL")?;
         let ipfs_gateway_url = read_env("BOUNDLESS_IPFS_GATEWAY_URL")?;
 
         Ok(Self {
             pinata_jwt,
-            pinata_api_url: Url::parse(&pinata_api_url).expect("Invalid Pinata API URL"),
-            ipfs_gateway_url: Url::parse(&ipfs_gateway_url).expect("Invalid IPFS Gateway URL"),
+            pinata_api_url,
```
rakanalh, line 66: Why are we not parsing Url anymore?

crates/common/src/config/risc0.rs
```diff
@@ -133,39 +106,32 @@ impl citrea_common::FromEnv for BoundlessProverConfig {
...
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
 /// Configuration for the Boundless Market client
 pub struct BoundlessConfig {
-    pub(crate) wallet_private_key: LocalSigner<SigningKey>,
-    pub(crate) rpc_url: Url,
-    pub(crate) deployment: Deployment,
+    pub wallet_private_key: String,
```
rakanalh, line 112: Why did this become a String?

crates/common/src/config/risc0.rs
exeokan: `PinataStorageProvider::from_parts` accepts strings, I don't think we earn much with parsing it here. Also I wanted to minimize imports in the `citrea-common`.

crates/common/src/config/risc0.rs
exeokan: I didn't want to import `boundless-market` in `citrea-common`. Is there any downsides to using String here?

crates/common/src/config/mod.rs
```diff
@@ -162,6 +165,9 @@ pub struct BatchProverConfig {
     pub enable_recovery: bool,
     /// Maximum number of commitments per proof partition
     pub max_commitments_per_proof: Option<usize>,
+    /// Configuration for Risc0Host
+    #[serde(default)]
```
eyusufatik, line 169: I am not sure about having default here, shouldn't we just error out?

or should the default be local proving with dev mode?

eyusufatik (changes requested):
looking good. update prover_config.toml s with the risc0 host config please

crates/common/src/config/mod.rs
exeokan: Defaulting to: `LocalProverConfig::default() ` was convenient because I could leave the test config files untouched, and local proving in dev mode happens.
However,  `LocalProverConfig::default() ` has `dev_mode = false`, which is a bit confusing (proofs are fake because proving mode is execute). It wouldn't change anything functionally but I can set it to `true` in regtest/mock configs if you want.

rakanalh (approved):

jfldde (approved):

crates/common/src/config/mod.rs
exeokan: I tried adding `dev_mode=true` to configs, but there is a "weird" behavior.
When `RISC0_DEV_MODE` is not set, `proving_mode=execute`:
`dev_mode=false` -> does fake proofs
`dev_mode=true` -> does real proofs

I think we should create an issue for this and address it elsewhere.

eyusufatik (approved):

PR #3013, opened 2025-11-10
feat: State block overrides

rakanalh:
# Description

Based on Eth's RPC documentation, found that we're missing state overrides for the following RPCs:
- eth_estimateGas
- eth_createAccessList


crates/evm/src/query.rs
```diff
@@ -892,28 +920,42 @@ impl<C: sov_modules_api::Context> Evm<C> {
         &self,
         request: TransactionRequest,
         block_number: Option<BlockNumberOrTag>,
+        state_overrides: Option<StateOverride>,
+        block_overrides: Option<BlockOverrides>,
```
eyusufatik: I didn't see any client that has block_overrides on estimateGas and createAccessList.

remove them

eyusufatik (changes requested):

crates/evm/src/query.rs
```diff
@@ -1088,6 +1154,15 @@ impl<C: sov_modules_api::Context> Evm<C> {
             }
         }
 
+        // Recreate evm_db with overrides if it was consumed by the early return optimization
```
eyusufatik, line 1137: there is many more get_db calls below, you'll need to handle those also.

rakanalh: Done

crates/evm/src/query.rs
rakanalh: Removed

eyusufatik (approved):

ercecan (approved):

PR #3034, opened 2025-11-18
chore: Update rpc config toml files

ercecan:


ercecan:
@eyusufatik maybe I can do all default configs like in testnet config, all commented out and look exactly the same, wdyt?

rakanalh (dismissed):
LGTM

resources/configs/mock/rollup_config.toml
```diff
@@ -20,6 +20,12 @@ enable_subscriptions = true
 max_subscriptions_per_connection = 100
 stale_filter_ttl = 300
 
+# enables JS tracer in debug endpoints, true by default
+# enable_js_tracer = true
+
+# enables filter rpc endpoints, enabled by default
```
jfldde: Might be better to choose either `true by default` or `enabled by default` for both

jfldde (dismissed):

eyusufatik (approved):

jfldde (approved):

PR #3043, opened 2025-11-21
fix: Align `eth_estimateGas` With Target Block Gas Limits

maradini77:
### PR Description
- Track the actual block gas limit in `EstimatedTxExpenses` and propagate it through gas estimation.
- Use that stored limit when returning `eth_estimateGas`/`eth_createAccessList` results, eliminating reliance on the latest head block.
- Update `gas_limit_to_return` and its tests to work with the new data flow, ensuring historical and pending estimates respect their own block limits.


jfldde (approved):

maradini77:
@jfldde merge?

eyusufatik (approved):

PR #3056, opened 2025-12-02
fix: debug_traceCall with pending block

exeokan:
# Description
`debug_traceCall` with pending block param was giving `"pending block not supported"`
Cause of the error was in `crates/evm/src/query.rs:1541`
```
 let sealed_block = self
    .get_sealed_block_by_number(Some(block_number), working_set, ledger_db)?
    .ok_or_else(|| EthApiError::HeaderNotFound(block_id.unwrap()))?;
 let l1_fee_rate = sealed_block.l1_fee_rate;
```
When `block_number == Pending`, `get_sealed_block_by_number` returns this error.

Fix: For pending blocks, use latest block for l1 fee rate.

## Testing
Added to call tracer test.

### DIFF
--- a/bin/citrea/tests/mock/evm/tracing.rs
+++ b/bin/citrea/tests/mock/evm/tracing.rs
@@ -153,7 +153,11 @@ async fn test_call_tracer() -> Result<(), Box<dyn std::error::Error>> {
 
     let call_frame_call_trace = test_client
         // test tracing in pending block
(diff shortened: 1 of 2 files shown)


eyusufatik (approved):
good catch!

jfldde (approved):

rakanalh (approved):

PR #3057, opened 2025-12-02
fix: Use reth's default max_depth for mempool maintenance task

rakanalh:


jfldde (commented):
Isn't the current behaviour better so that we don't silently skip a newly added field if ever ?

rakanalh:
> Isn't the current behaviour better so that we don't silently skip a newly added field if ever ?

Problem is:
```
max_update_depth: Default::default()
```

sets the value to 0, while what we want is to use the default value from reth
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=a77b1a0c7cc0adc61bbbb3221a08092a

Also, having the value set as 0 would basically cause the mempool maintenance task to work more because a Drifted mempool state would need to reload ALL sender accounts in the blocks produced rather than the ones actually changed. This inefficiency can be dealt with by setting this value properly.

jfldde (approved):

kpp (approved):

eyusufatik (approved):

PR #3062, opened 2025-12-04
chore: Use custom infra similar prometheus image

ercecan:
# Description
Using a script similar to one in our production infra for generating prometheus.yml file we can now directly import locally created dashboards into production grafana


docker/docker-compose.telemetry.yaml
```diff
@@ -1,6 +1,7 @@
 services:
   prometheus:
-    image: prom/prometheus
+    container_name: prometheus.citrea
+    image: local-prometheus
```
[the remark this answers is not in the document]
ercecan: makes sense

docker/prometheus/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+RUN apt update && apt -y upgrade && apt -y install python3 python3-pip && apt -y autoremove && apt clean && rm -rf /var/lib/apt/lists/*
+
+RUN pip3 install --break-system-packages pyyaml
+
+COPY --from=source /bin/prometheus /bin/prometheus
+COPY --from=source /bin/promtool /bin/promtool
+
+RUN mkdir -p /mnt/task/prometheus-data && mkdir -p /etc/prometheus
+
+RUN apt update && apt install telnet -y
```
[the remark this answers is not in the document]
ercecan: This is not actual production tho

kpp (dismissed):

docker/README.md
```diff
@@ -0,0 +1,28 @@
+# Guide to run prometheus and grafana
+
+## Guide for external users who want to see grafana dashboards next to their nodes
```
eyusufatik: I don't like this language "external users"

I think we can call them "node runners"

eyusufatik: this sentence is a bit odd, let's run this through chat gpt or smth

docker/README.md
```diff
@@ -0,0 +1,28 @@
...
+1) go into `<project_dir>/docker directory
```
eyusufatik: inconsistent uppercase / lowercase

resources/grafana/user/batch-prover.dashboard.json
eyusufatik, line 1: on "user" dir no need for batch prover, light client prover and sequencer dashboards.

if the reason you put them is we are gonna use those dashboards on our local envs (regtest) let's set up a different directory for them

docker/docker-compose.telemetry.yaml
```diff
@@ -1,6 +1,10 @@
 services:
   prometheus:
-    image: prom/prometheus
+    container_name: prometheus.citrea
+    build:
+      context: ./prometheus
+      dockerfile: Dockerfile
+    image: local-prometheus
```
eyusufatik, line 7: is this true? is it bc we are building a container?

eyusufatik (changes requested):

docker/docker-compose.telemetry.yaml
ercecan: yes, we are building that prometheus docker image and and the prometheus config in it, this will automatically do it for us instead of us calling `docker build -t local-prometheus .`

resources/grafana/user/batch-prover.dashboard.json
ercecan: us developers are not going to use them, only external node runners will use them, so I will just keep full node then

ercecan: however I believe people like auditors or someone who wanna dive deep and run the stack locally can still use them?

PR #3066, opened 2025-12-08
feat: Use floating point fee rate

rakanalh:
# Description

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - feat: Store proving session info of LCP ([#3050](https://github.com/chainwayxyz/citrea/pull/3050))\
   `lightClientProver_getLightClientProofByL1Height` endpoint now returns information about the proving session, using the same structure as the batch prover responses.
 - ci: Run citrea-e2e tests against bitcoin v30
+- feat: Use floating point precision in calculating bitcoin fee rate. ([#3066](https://github.com/chainwayxyz/citrea/pull/3066))
 
 ### Changed
 - chore: renamed `BOUNDLESS_S3_NO_PRESIGNED` to `BOUNDLESS_S3_USE_PRESIGNED`. ([#3046](https://github.com/chainwayxyz/citrea/pull/3046))\
--- a/crates/bitcoin-da/src/fee.rs
+++ b/crates/bitcoin-da/src/fee.rs
@@ -235,14 +235,6 @@ pub(crate) async fn get_fee_rate_from_mempool_space(
     network: bitcoin::Network,
     mempool_space_url: &str,
 ) -> Result<Option<f64>> {
-    match network {
-        bitcoin::Network::Bitcoin | bitcoin::Network::Testnet => {}
-        _ => {
-            trace!("Unsupported network for mempool space fee estimation");
-            return Ok(None);
-        }
-    }
-
     let url = match network {
         bitcoin::Network::Bitcoin => {
             format!("{mempool_space_url}{MEMPOOL_SPACE_PRECISE_FEE_ENDPOINT}")
@@ -250,7 +242,10 @@ pub(crate) async fn get_fee_rate_from_mempool_space(
         bitcoin::Network::Testnet => {
             format!("{mempool_space_url}testnet4/{MEMPOOL_SPACE_PRECISE_FEE_ENDPOINT}")
         }
-        _ => unreachable!(), // Already checked above
+        _ => {
+            trace!("Unsupported network for mempool space fee estimation");
+            return Ok(None);
+        }
     };
 
     let response = get_with_timeout(url.clone(), MEMPOOL_SPACE_TIMEOUT)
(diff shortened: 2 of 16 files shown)


crates/bitcoin-da/src/fee.rs
```diff
@@ -232,7 +240,7 @@ impl FeeService {
 pub(crate) async fn get_fee_rate_from_mempool_space(
```
jfldde, line 234: Should we switch to `/precise` fee rate endpoint here in this PR ?

eyusufatik: yep without that change nothing changes basically, unless @rakanalh is planning to make a second pr to activate the floating point fee rate txs

rakanalh: Done

rakanalh:
This commit: [7c933eb](https://github.com/chainwayxyz/citrea/pull/3066/commits/7c933eb21c3e08d9b1d3d4dc36241e247a7458ce)
- Forces regtest to use mainnet fees
- Replaces use `send_transaction_with_fee_rate` with `send_transaction` to force usage of `get_fee_rate` so that the fees are returned from mempool space.

The tests:
- `test_queue_da_transactions` 
- `test_queue_da_transactions_oldest_mode`

Are passing. The failure in `test_mempool_space_fee_rate` is expected since i have not modified the test to assert regtest fees from mempool space (Asserts for `None`, because we should not support this)

jfldde (dismissed):

crates/bitcoin-da/src/fee.rs
```diff
@@ -232,35 +241,58 @@ impl FeeService {
...
-
-    Ok(Some(fee_rate))
+    }
+
+    let endpoints = [
+        MEMPOOL_SPACE_PRECISE_FEE_ENDPOINT,
+        MEMPOOL_SPACE_RECOMMENDED_FEE_ENDPOINT,
+    ];
+
+    for endpoint in endpoints {
```
eyusufatik: I don't think we try each endpoint.
matter of fact, we can remove `MEMPOOL_SPACE_RECOMMENDED_FEE_ENDPOINT` and rely on the precise endpoint solely.

crates/bitcoin-da/src/fee.rs
```diff
@@ -126,23 +127,31 @@ impl FeeService {
...
-                Err(e) => {
-                    tracing::error!(?e, "Failed to get fee rate from mempool.space");
-                    self.client
+                Ok(Some(fee_rate)) => fee_rate,
+                Ok(None) | Err(_) => {
+                    match self
+                        .client
                         .estimate_smart_fee(1, Some(EstimateMode::Conservative))
-                        .await?
-                        .fee_rate
```
eyusufatik: please revert to this behaviour. don't return 1000.0 when estimate_smart_fee fails, propoagate the error

eyusufatik (changes requested):

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - feat: Add `batchProver_getLatestProvingSessionInfos` rpc ([#3070](https://github.com/chainwayxyz/citrea/pull/3070))
+- feat: Use floating point precision in calculating bitcoin fee rate. ([#3066](https://github.com/chainwayxyz/citrea/pull/3066))
```
jfldde, line 9: Maybe add that we're switching to `/precise` endpoint here as well

crates/bitcoin-da/src/fee.rs
```diff
@@ -232,35 +234,48 @@ impl FeeService {
...
+
+    let url = match network {
+        bitcoin::Network::Bitcoin => {
+            format!("{mempool_space_url}{MEMPOOL_SPACE_PRECISE_FEE_ENDPOINT}")
+        }
+        bitcoin::Network::Testnet => {
+            format!("{mempool_space_url}testnet4/{MEMPOOL_SPACE_PRECISE_FEE_ENDPOINT}")
+        }
+        _ => unreachable!(), // Already checked above
     };
```
jfldde: We could merge these two match statements

crates/bitcoin-da/src/fee.rs
```diff
@@ -137,12 +137,14 @@ impl FeeService {
...
                         .fee_rate
+                        .map(|rate| rate.to_sat() as f64)
                 }
             };
-        let sat_vkb = smart_fee.map_or(1000, |rate| rate.to_sat());
 
-        tracing::debug!("Fee rate: {} sat/vb", sat_vkb / 1000);
-        Ok(sat_vkb / 1000)
+        let sat_vkb = smart_fee.unwrap_or(1000.0);
+        let sat_vb = sat_vkb / 1000.0;
```
jfldde, line 145: We could default to 1 and remove the `/1000` here and `* 1000` from `get_fee_rate_from_mempool_space` it seems.

rakanalh: `estimatesmartfee` returns in vkb while mempool space apparently returns in vb... so this conversion unifies both.

jfldde (approved):

eyusufatik (approved):

CHANGELOG.md
eyusufatik: would be nice but I'm merging :)

PR #3075, opened 2025-12-11
test: Prioritize e2e tests

jfldde:
# Description

- Speed up test suite by running longer running tests first
- Seeing 30% improvements locally, waiting on CI to have a clearer picture


rakanalh:
How would you explain 30% improvement if we're using the same test suite and  the same `-j15`? I think regardless of whether you start the longer ones first and or the shorter ones, it would still take the same time. That or, i can't wrap my head around this.

ercecan (dismissed):

rakanalh:
Comparing this PRs nextest build:
https://github.com/chainwayxyz/citrea/actions/runs/20137654860/job/57795486936?pr=3075

and latest nightly:
https://github.com/chainwayxyz/citrea/actions/runs/20132959545/job/57794605165

It seems like this PR's nextest build is 5 minutes faster. I don't understand how though :sweat_smile:

rakanalh (dismissed):

jfldde:
> It seems like this PR's nextest build is 5 minutes faster. I don't understand how though 😅

Shouldn't impact build time, could be cache related. Test suite itself finished in 2min less

jfldde:
>How would you explain 30% improvement if we're using the same test suite and  the same `-j15`? I think regardless of whether you start the longer ones first and or the shorter ones, it would still take the same time. That or, i can't wrap my head around this.

Longer running test will start first and finish last, while the shorter test can run in the meantime. Longest running test should always start first ideally so that we don't end up waiting on a single test and lose some parallelism benefits

.github/workflows/checks.yml
```diff
@@ -549,7 +549,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: /tmp/docker
-          key: ${{ runner.os }}-docker-${{ env.TEST_BITCOIN_DOCKER }}
+          key: ${{ runner.os }}-docker-${{ env.BITCOIN_DOCKER_IMAGE }}
```
eyusufatik, line 552: what's the reason behind this change?

eyusufatik (approved):

.github/workflows/checks.yml
jfldde: docker-setup step was silently failing, related commit: https://github.com/chainwayxyz/citrea/pull/3075/commits/ba438f0ee37b0f3a2fb42220474b876de0c75713

jfldde: Wasn't pulling the correct bitcoin docker image

kpp (approved):

PR #3080, opened 2025-12-15
fix(batch-prover): L2Syncer saturates backoff at max value instead of panicking

### LINKED ISSUE #3012: Batch prover should not shutdown on "Failed to process L2 block"

jfldde:
# Description

- Keep using max expontential backoff value instead of panicking 

## Linked Issues
- Fixes #3012


eyusufatik (approved):
nice solution

kpp (approved):

PR #3087, opened 2025-12-18
fix: `bitcoin::network::Testnet` vs `bitcoin::network::Testnet4` confusion in mempool.space fee retrieval

eyusufatik:
# Description
we pass around testnet4 as the bitcoin network, however, in `get_fee_rate_from_mempool_space` we apparently compared for testnet instead of testnet4.

I also thought instead of constructing the testnet4/ part of the url, we can set it directly from env vars.

plus some simple refactoring.


crates/bitcoin-da/src/fee.rs
```diff
@@ -108,41 +108,31 @@ impl FeeService {
     }
 
     /// Get the fee rate in sat/vB from the mempool space or via the Bitcoin Core client.
-    /// If the network is regtest or testnet, it returns a default value of 1 sat/vB.
     #[instrument(level = "trace", skip_all, ret)]
     pub async fn get_fee_rate(&self) -> Result<f64> {
```
jfldde, line 112: Can get rid of this wrapper now it seems

jfldde (dismissed):

rakanalh (dismissed):

kpp (dismissed):

jfldde (approved):

PR #3090, opened 2025-12-19
perf: RecoveredPubkeyProvider

jfldde:
# Description

- precompute recovered pubkey on native


crates/ecrecover-address-provider/src/native.rs
```diff
@@ -0,0 +1,45 @@
+use parking_lot::Mutex;
+
+use super::{EcrecoverAddressProvider, EcrecoverProviderError};
+
+/// Collects ecrecover addresses during native execution in deterministic order
+pub struct NativeEcrecoverAddressCollector {
+    addresses: Mutex<Vec<[u8; 20]>>,
```
kpp: Why mutex?

crates/ecrecover-address-provider/src/zk.rs
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /// Clear all recorded addresses
+    fn clear(&self) {
+        unimplemented!("clear is not implemented for ZkEcrecoverAddressProvider");
+    }
+
+    /// Take addresses
+    fn take_addresses(&self) -> Result<Vec<[u8; 20]>, EcrecoverProviderError> {
+        unimplemented!("take_addresses is not implemented for ZkEcrecoverAddressProvider");
```
kpp: While I understand why you did it like this, is there a way to use two different traits, one for zk and the other is for native so we have a compile time guarantees?

crates/ecrecover-address-provider/src/zk.rs
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /// Get the next address from the witness
+    fn get_next(&self) -> Result<[u8; 20], EcrecoverProviderError> {
+        self.addresses
```
kpp: It may look like its inner struct is VecDeque, but it should be https://doc.rust-lang.org/std/slice/struct.Iter.html instead

kpp (dismissed):
Amazing

crates/ecrecover-address-provider/src/zk.rs
jfldde: I agree, I followed short-header-proof pattern but I don't like it either. I'll clean it up

crates/ecrecover-address-provider/src/zk.rs
kpp: ping?

jfldde: Agree, that would be better. This is not really a `VecDequeue` as we only really care about having the next element and we don't care about appending or anything available to `VecDequeue` really. We just want single-pass/in-order consumption. Iter might be tedious mut wise though, what about a slice + cursor instead ?

kpp: Here you are:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=85dd680245c53f9fc14bbcefdc5559bd

```
use std::cell::RefCell;

type Secp256k1Pubkey = u32;

pub struct RecoveredPubkeyProvider {
    pubkeys: RefCell<std::vec::IntoIter<Secp256k1Pubkey>>,
}

impl RecoveredPubkeyProvider {
    pub fn new(pubkeys: Vec<Secp256k1Pubkey>) -> Self {
        Self {
            pubkeys: RefCell::new(pubkeys.into_iter()),
        }
    }

    /// Get the next address
    pub fn get_next(&self) -> Result<Secp256k1Pubkey, ()> {
        self.pubkeys
            .borrow_mut()
            .next()
            .ok_or(())
    }
}

fn main() {
    let pubkeys = vec![1,2];
    let prov = RecoveredPubkeyProvider::new(pubkeys);
    let _ = dbg!(prov.get_next());
    let _ = dbg!(prov.get_next());
    let _ = dbg!(prov.get_next());
}
```

kpp (approved):

crates/evm/src/lib.rs
```diff
@@ -1,7 +1,8 @@
 #![deny(missing_docs)]
 #![doc = include_str!("../README.md")]
 mod call;
-mod evm;
+/// EVM handler and transaction processing
+pub mod evm;
```
exeokan: it seems like this is not necessary

crates/citrea-stf/Cargo.toml
```diff
@@ -31,6 +31,7 @@ sov-state = { path = "../sovereign-sdk/module-system/sov-state" }
 
 citrea-evm = { path = "../evm" }
 l2-block-rule-enforcer = { path = "../l2-block-rule-enforcer" }
+recovered-pubkey-provider = { path = "../recovered-pubkey-provider", default-features = false }
```
exeokan: it has no features anymore

crates/evm/Cargo.toml
```diff
@@ -12,6 +12,7 @@ publish = false
 readme = "README.md"
 
 [dependencies]
+recovered-pubkey-provider = { path = "../recovered-pubkey-provider", default-features = false }
```
exeokan: same here

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/batch_proof/input/v4.rs
```diff
@@ -59,27 +95,32 @@ pub struct BatchProofCircuitInputV3 {
...
-    /// Into Tangerine expected inputs
-    pub fn into_v3_parts(self) -> (BatchProofCircuitInputV3Part1, BatchProofCircuitInputV3Part2) {
+impl BatchProofCircuitInputV4 {
+    fn into_shared_parts(
+        self,
+    ) -> (
+        BatchProofCircuitInputV4Part1,
+        BatchProofCircuitInputV4Part2,
+        BatchProofCircuitInputV4Part3,
+    ) {
```
exeokan: `BatchProofCircuitInputV4Parts` can be used here

bin/citrea/tests/bitcoin/batch_prover_test.rs
```diff
@@ -575,241 +581,176 @@ async fn parallel_proving_test() -> Result<()> {
...
-//         Ok(())
-//     }
-// }
+        Ok(())
+    }
+}
 
-// // ignoring this test now as we won't be supporting backwards compatibility for proofs.
-// #[tokio::test]
-// #[ignore]
```
exeokan, line 804: we can remove the `#[ignore]`

bin/citrea/tests/bitcoin/batch_prover_test.rs
```diff
@@ -575,241 +581,176 @@ async fn parallel_proving_test() -> Result<()> {
...
 
-// struct ForkElfSwitchingTest;
-
-// #[async_trait]
-// impl TestCase for ForkElfSwitchingTest {
-//     fn test_config() -> TestCaseConfig {
-//         TestCaseConfig {
-//             with_batch_prover: true,
-//             with_full_node: true,
-//             with_light_client_prover: true,
```
exeokan, line 586: Do you think we should have the light client prover in this test? I played around a bit to see if it is easy to add, but it doesn't seem so, especially with the initial l2 state root of the LCP.

crates/citrea-stf/src/verifier.rs
```diff
@@ -60,6 +62,18 @@ where
...
+        let flat_pubkeys = ecrecover_pubkey_witnesses.into_iter().flatten().collect();
+        let recovered_pubkey_provider =
+            recovered_pubkey_provider::RecoveredPubkeyProvider::new(flat_pubkeys);
+        if recovered_pubkey_provider::RECOVERED_PUBKEY_PROVIDER
+            .set(recovered_pubkey_provider)
+            .is_err()
+        {
+            panic!("Recovered pubkey provider already set");
+        }
+
```
exeokan, line 75: I don't think it will have a significant impact on cycles, but do we really need to group by commitment in the circuit input if we're gonna flatten it right away? I understand it is convinient in testing though.

bin/citrea/tests/bitcoin/batch_prover_test.rs
```diff
@@ -1346,6 +1284,296 @@ async fn batch_prover_create_input_test() -> Result<()> {
...
+
+        let tampered_input =
+            borsh::to_vec(&(input_part1, ecrecover_pubkey_witnesses, input_part3))?;
+
+        let network = Network::Nightly;
+        let risc0_config = Risc0HostConfig::from_env().expect("Failed to load risc0 config");
+        let mut risc0_host = Risc0Host::new(network, risc0_config).await;
+        risc0_host.add_hint(tampered_input);
+
+        let run_result = risc0_host
```
exeokan: for control, we can run the untampered input and assert success

exeokan (changes requested):

bin/citrea/tests/bitcoin/batch_prover_test.rs
jfldde: It used to fail on CI and I don't remember exactly why but I will try

bin/citrea/tests/bitcoin/batch_prover_test.rs
jfldde: Yes I tried but I couldn't make it work and just dropped it since it didn't add much value here

crates/sovereign-sdk/rollup-interface/src/state_machine/zk/batch_proof/input/v4.rs
```diff
@@ -42,12 +34,56 @@ pub struct BatchProofCircuitInputV3Part1 {
...
+    pub merkle_proof_bytes: Vec<u8>,
+    /// Give the start of the previous sequencer commitment as a hint
+    /// so index can be calculated
+    pub prev_sequencer_commitment_start: u64,
+}
+
+/// Pre-computed ecrecover pubkey witnesses, grouped by sequencer commitment.
+///
+/// Each inner `Vec` holds the uncompressed 65-byte pubkeys recovered from the
+/// signatures of the L2 blocks in that commitment, in circuit consumption order.
```
exeokan: doc needs to be updated after flattening the pubkeys

exeokan (commented):
LGTM

exeokan (approved):

PR #3010, opened 2025-11-06
feat: Reactivate pruning

rakanalh:

PR #3092, opened 2025-12-20
Fix: Handle pruning task failures properly

viktorking7:
Previously, `Pruner::prune` ignored errors and panics from background pruning tasks, causing silent failures where pruning appeared successful even when it failed.


jfldde:
Conflicts with and not sure it will be needed after https://github.com/chainwayxyz/citrea/pull/3010/

jfldde:
Not relevant post #3010

PR #3093, opened 2025-12-22
fix: l1 fee used in block building can use precise fees

eyusufatik:


kpp (approved):

jfldde (dismissed):

crates/bitcoin-da/src/service.rs
```diff
@@ -1419,3 +1421,12 @@ fn calculate_witness_root(txdata: &[TransactionWrapper], tx_count: usize) -> [u8
         .collect();
     BitcoinMerkleTree::new(hashes).root()
 }
+
+fn f64_to_u128_or_default(x: f64, default: u128) -> u128 {
```
jfldde: Why not `Option<u128>` as return value and use `unwrap_or` at the callsite instead ?

eyusufatik: makes sense

jfldde (dismissed):

ercecan (approved):

PR #3095, opened 2025-12-24
feat: implement eip-7966

### LINKED ISSUE #3089: implement eip-7966
https://eips.ethereum.org/EIPS/eip-7966

jfldde:
# Description

- Implements `eth_sendRawTransactionSync` as per EIP-7966

## Linked Issues
- Fixes #3089


crates/ethereum-rpc/src/lib.rs
```diff
@@ -500,6 +514,91 @@ where
...
+                )
+            })?
+            .resubscribe();
+
+        let hash: B256 = self
+            .ethereum
+            .sequencer_client
+            .as_ref()
+            .unwrap()
+            .eth_send_raw_transaction(data)
```
kpp, line 551: I would say this should be on the sequencer level with a proper `eth_send_raw_transaction_sync`, so we don't spam sequencer every `eip_7966::calculate_timeout_ms(timeout_ms)` about the tx, and on the full-node side it's gonna be a direct:

```
    async fn eth_send_raw_transaction_sync(&self, data: Bytes) -> RpcResult<Option<Transaction>> {
        self.ethereum
            .sequencer_client
            .as_ref()
            .unwrap()
            .eth_send_raw_transaction_sync(data)
            .await
            .map_err(|e| match e {
                jsonrpsee::core::client::Error::Call(e_owned) => e_owned,
                _ => to_jsonrpsee_error_object("SEQUENCER_CLIENT_ERROR", e),
            })
    }
```

jfldde: If we do this, we might get a result from the sequencer for a receipt that is not yet synced on the fullnode. Also, this doesn't spam, we just do a single `send_raw_transaction` call to sequencer and then check it against local state

ercecan: I agree that the inconsistency would be a problem, and I believe making the fullnode wait instead of the sequencer is more optimal

ercecan (dismissed):
lgtm

crates/common/src/rpc/eip_7966.rs
```diff
@@ -0,0 +1,121 @@
...
+pub const TIMEOUT_ERROR_CODE: i32 = 4;
+
+/// EIP-7966 error code 5: Node is not ready to process the transaction or the transaction is erroneous.
+pub const UNREADY_ERROR_CODE: i32 = 5;
+
+/// Default timeout in milliseconds. (2secs)
+pub const DEFAULT_TIMEOUT_MS: u64 = 2_000;
+
+/// Maximum allowed timeout in milliseconds. (1min)
+pub const MAX_TIMEOUT_MS: u64 = 60_000;
```
eyusufatik: can we make this configurable?

bin/citrea/tests/mock/evm/web3_py/test.py
```diff
@@ -250,5 +250,76 @@ def test_get_logs_false_hash(self):
...
+            'nonce': self.web3.eth.get_transaction_count("0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"),
+            'gas': 200000,
+            'gasPrice': self.web3.eth.gas_price,
+        }
+        signed_tx = self.web3.eth.account.sign_transaction(transaction, "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80")
+
+        receipt = self.web3.provider.make_request(
+            'eth_sendRawTransactionSync',
+            [signed_tx.raw_transaction.hex(), 5000]
+        )['result']
```
eyusufatik, line 267: shouldn't this function already exist in web3py? maybe we should upgrade it.

I want to make sure our implementation works with existing tooling

eyusufatik (changes requested):

crates/common/src/rpc/eip_7966.rs
eyusufatik: should be passable from RpcConfig but default value is used when env var is not found

bin/citrea/tests/mock/evm/web3_py/test.py
jfldde: Not part of `web3py` nor `ethers.js` yet. `EIP-7966` is still draft. 
Though in that case, this will just be a wrapper around the `eth_sendRawTransactionSync` RPC method and handle input/output typing only.  Should be safe to use this for now and replace it with the correct method once added to `web3py`

crates/ethereum-rpc/src/lib.rs
```diff
@@ -500,6 +517,91 @@ where
...
+                            let mut working_set = WorkingSet::new(self.ethereum.storage.clone());
+                            if let Ok(Some(receipt)) = evm.get_transaction_receipt(hash, &mut working_set) {
+                                return Ok(receipt);
+                            }
+                        }
+                        Err(tokio::sync::broadcast::error::RecvError::Closed) => {
+                            return Err(internal_rpc_error("Block subscription channel closed"));
+                        }
+                    }
+                }
```
ercecan, line 599: I think using `tokio::time::timeout` would be cleaner here

ercecan (changes requested):

ercecan (dismissed):

crates/common/src/rpc/eip_7966.rs
```diff
@@ -0,0 +1,106 @@
...
+use jsonrpsee::types::ErrorObjectOwned;
+
+/// EIP-7966 error code 4: Transaction was added to mempool but not processed within timeout.
+const TIMEOUT_ERROR_CODE: i32 = 4;
+
+/// EIP-7966 error code 5: Node is not ready to process the transaction or the transaction is erroneous.
+const UNREADY_ERROR_CODE: i32 = 5;
+
+/// Default timeout in milliseconds. (2secs)
+const DEFAULT_TIMEOUT_MS: u64 = 2_000;
```
eyusufatik: isn't this too low?

CHANGELOG.md
```diff
@@ -1,9 +1,9 @@
 # Changelog
-## [unreleased]
+## [Unreleased]
 ### Added
 - Add `linux/arm64` release binary and support for multi-arch (arm64/amd64) docker image.
+- feat: implement `eth_sendRawTransactionSync` RPC as per EIP-7966. ([#3095](https://github.com/chainwayxyz/citrea/pull/3095))
```
eyusufatik, line 5: please add new env vars

eyusufatik (changes requested):

crates/common/src/rpc/eip_7966.rs
jfldde: Bumpped it to 5s (2x2s block + 1s overhead)

eyusufatik (approved):

PR #3104, opened 2026-01-05
ci: Cache riscv-guest

jfldde:
# Description

- Cache riscv-guest in `coverage`, `build` and `nextest` jobs. Sheds 1min30 each

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -222,7 +222,7 @@ jobs:
       - name: Cache riscv-guest
         uses: actions/cache@v4
         with:
-          key: build-guests-v0-${{ runner.os }}-${{ hashFiles('**/Cargo.lock') }}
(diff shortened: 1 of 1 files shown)


.github/workflows/checks.yml
```diff
@@ -100,6 +100,14 @@ jobs:
         with:
           key: "eth-tests-1c23e3c"
           path: crates/evm/ethereum-tests
+
+      - name: Cache riscv-guest
+        uses: actions/cache@v4
+        with:
```
eyusufatik: maybe use ubicloud cache here?

ubicloud github cache download is really slow

eyusufatik (changes requested):

.github/workflows/checks.yml
jfldde: Didn't notice it since actions/cache is downloading at `215.3 MBs/sec` here. Wonder how. I'll amend

eyusufatik (approved):

PR #3108, opened 2026-01-07
feat: Use address instead of pubkeys for security council

ercecan:
# Description
Initial pr for switching to eip712 structured message types for security council and lcp
Uses addresses instead of pubkeys to verify signatures, basically recovers pubkeys from signatures calculates the addresses and compares them


ercecan:
Superseded by #3166

PR #3117, opened 2026-01-23
feat: Citrea with txsender - testing

atacann:
Plugged txsender to citrea in a hacky way just to test


jfldde:
Closing in favour of #3206

PR #3119, opened 2026-01-26
feat: Unified fullnode docker image

jfldde:


docker/docker-compose.yml
```diff
@@ -47,15 +47,14 @@ services:
...
-      - SEQUENCER_CLIENT_URL=https://rpc.testnet.citrea.xyz
       - INCLUDE_TX_BODY=false
-      - SCAN_L1_START_HEIGHT=45496
       - SYNC_BLOCKS_COUNT=10
       - RUST_LOG=info
       - JSON_LOGS=1
     ports:
       - "8080:8080"
     volumes:
+      - ./genesis:/app/genesis:ro
```
eyusufatik: what is ro

docs/run-mainnet.md
eyusufatik, line 1: can you add normal binary running just like run-testnet.md docs?

.github/workflows/release.yml
```diff
@@ -160,13 +160,7 @@ jobs:
           BINARY_URL="https://github.com/chainwayxyz/citrea/releases/download/${RELEASE}/citrea-${RELEASE}-linux-amd64"
           echo "Downloading binary from ${BINARY_URL}"
           # Download the binary using wget
-          wget -O docker/build-push/testnet/citrea "${BINARY_URL}"
-
-          # Make the downloaded file executable
```
eyusufatik, line 165: are you sure about removing these?

btw we can move genesis files into the docker image itself. so the user doesn't need to download genesis, wdyt?

eyusufatik (changes requested):

docker/docker-compose.yml
jfldde: read only

docs/run-mainnet.md
jfldde: Sure, will do

.github/workflows/release.yml
jfldde: Ok I though we went the volume way but will make it much easier in the image itself, I'll switch it up.

cetin:
adding devnet configuration isnt necessary for a public repo/images, since its only used by internal services and isnt externally accessible

jfldde:
> adding devnet configuration isnt necessary for a public repo/images, since its only used by internal services and isnt externally accessible

Added as convenience for the team but I can remove

Chinacolt (approved):
LGTM

ercecan (dismissed):

eyusufatik (approved):

ercecan (approved):

PR #3121, opened 2026-01-27
feat: boundless uses claim digest match

eyusufatik:


crates/risc0/src/host/boundless.rs
```diff
@@ -299,48 +306,44 @@ impl BoundlessProver {
...
-                        .with_timeout(timeout as u32)
-                        .with_ramp_up_period(ramp_up_period as u32)
-                        .with_lock_collateral(U256::from(lock_stake))
-                        .with_ramp_up_start(bidding_start),
-                )
-                .with_cycles(total_cycles_approx);
-
-        if let Some(journal) = journal {
-            request_params = request_params.with_journal(journal);
-        }
```
ercecan, line 335: If we keep this, on the preflight layer of boundless request submission the proof will not be executed again
But we gotta make sure adding this will not automatically add journal on-chain again

ercecan (changes requested):

ercecan (approved):

PR #3133, opened 2026-02-10
fix: Update boundless max price calculation

ercecan:
# Description
Updates boundless max price calculation, now max price takes gas cost into account.


crates/risc0/src/host/boundless.rs
```diff
@@ -293,40 +307,101 @@ impl BoundlessProver {
...
+                "Failed to get gas price from provider, using fallback gas price: {} wei. err={}",
+                FALLBACK_BASE_GAS_PRICE,
+                e
+            );
+            FALLBACK_BASE_GAS_PRICE
+        });
+
+        let offer_layer = OfferLayer::new(provider, offer_layer_config);
+
+        let requirements = Requirements::new(Predicate::claim_digest_match(receipt_claim_digest))
```
eyusufatik, line 368: we call with_groth16_proof() twice now, I wonder if that's a problem 🤔

ercecan: that is okay, but unnecessary anyways so I'll just remove the second one

eyusufatik (approved):

PR #3143, opened 2026-02-17
fix: Misplaced log in boundless

### LINKED ISSUE #3124: False log bug in boundless
discovered by @exeokan 

https://github.com/chainwayxyz/citrea/blob/f0497402ee89afd3ec96582ebfcba6b8df51b331/crates/risc0/src/host/boundless.rs#L460-L474

In these lines `Resubmit boundless proving session Failed with job id` log is printed even if the resubmit is a success that log should be captured in an else block

also no need for continue in that block either

ercecan:
# Description

## Linked Issues
- Fixes #3124

### DIFF
--- a/crates/risc0/src/host/boundless.rs
+++ b/crates/risc0/src/host/boundless.rs
@@ -57,7 +57,7 @@ const LOCKTIME_INCREASE_RATIO: u32 = 2; // 2x
 const FALLBACK_BASE_GAS_PRICE: u128 = 1_000_000_000; // 1 gwei
 
 /// Duration to sleep before retrying a failed proof request in seconds
-const RETRY_RESUBMISSION_DELAY_SECS: u32 = 10;
+const RETRY_RESUBMISSION_DELAY_SECS: Duration = Duration::from_secs(10);
 
 enum ResubmitResult {
     Retry,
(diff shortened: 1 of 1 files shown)


crates/risc0/src/host/boundless.rs
```diff
@@ -542,19 +542,27 @@ impl BoundlessProver {
...
-                                tracing::info!(
-                                    "Resubmit boundless proving session Failed with job id: {}, and boundless request id: {} retrying...",
-                                    job_id,
-                                    request_id
-                                );
-                                continue;
+                                match res {
+                                    ResubmitResult::Retry => {
+                                        // Retry resubmission after a delay
+                                        let delay_duration = Duration::from_secs(10);
```
exeokan: `delay_duration` could be a constant in this file

exeokan (changes requested):
rest lgtm

crates/risc0/src/host/boundless.rs
eyusufatik: config var that defaults to 10 secs

eyusufatik (changes requested):

crates/risc0/src/host/boundless.rs
ercecan: agreed with @eyusufatik to use constant

eyusufatik (dismissed):

eyusufatik (approved):

exeokan (approved):

PR #3144, opened 2026-02-17
perf: Skip re-execution in preflight layer

### LINKED ISSUE #3122: boundless try skipping reexecution in preflight
If we keep this, on the preflight layer of boundless request submission the proof will not be executed again
But we gotta make sure adding this will not automatically add journal on-chain again

_Originally posted by @ercecan in https://github.com/chainwayxyz/citrea/pull/3121#discussion_r2734308374_

ercecan:
# Description
This enables us to skip preflight layer in boundless request builder where, if the journal is not provided the image and input is fetched from remote and re executed for journal to be re-attached to the request.
I've checked if this results in putting the journal to the fulfillment data that is written on the contract and realized that this is function `with_journal` is called eventually in preflight layer anyways. So we do not need to worry about that. The fulfillment data is determined in broker.rs

```rust
let (claim_digest, fulfillment_data, fulfillment_data_type) = match predicate_type {
      PredicateType::ClaimDigestMatch => (
          order_request.requirements.predicate.data.0.as_ref().try_into().unwrap(),
          vec![],
          FulfillmentDataType::None,
      ),
      PredicateType::PrefixMatch | PredicateType::DigestMatch => (
          order_claim_digest,
          FulfillmentDataImageIdAndJournal {
              imageId: <[u8; 32]>::from(order_img_id).into(),
              journal: order_journal.into(),
          }
          .abi_encode(),
          FulfillmentDataType::ImageIdAndJournal,
      ),
      _ => {
          return Err(anyhow!("Invalid predicate type: {predicate_type:?}"));
      }
  };
```

Since we use ClaimDigestMatch we're good on that front as well.

## Linked Issues
- Fixes #3122

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
   `RPC_MAX_SYNC_SEND_TIMEOUT_MS`: Maximum timeout in milliseconds for `eth_sendRawTransactionSync` (EIP-7966) (default: 20 seconds).\
 
 ### Changed
+- perf: Skip re-execution of proof request in boundless
 - feat: Separate l1 fee rate from block update
 **New env var:**\
   `L1_FEE_RATE_UPDATE_INTERVAL_MS`: L1 fee rate update interval in milliseconds (default: 30 seconds)\
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,16 @@
 # Changelog
 ## [Unreleased]
 ### Added
-- Add `linux/arm64` release binary and support for multi-arch (arm64/amd64) docker image.
-- feat: implement `eth_sendRawTransactionSync` RPC as per EIP-7966. ([#3095](https://github.com/chainwayxyz/citrea/pull/3095))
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
   `RPC_MAX_SYNC_SEND_TIMEOUT_MS`: Maximum timeout in milliseconds for `eth_sendRawTransactionSync` (EIP-7966) (default: 20 seconds).\
 
 ### Changed
+- perf: Skip re-execution of proof request in boundless
 - feat: Separate l1 fee rate from block update
 **New env var:**\
   `L1_FEE_RATE_UPDATE_INTERVAL_MS`: L1 fee rate update interval in milliseconds (default: 30 seconds)\
```
eyusufatik: can you add pull request info like other entries. update other added & changed ones that are missing this also pls

eyusufatik (changes requested):

eyusufatik (approved):

PR #3145, opened 2026-02-17
refactor: Use reth task manager instead of tokio spawn in sequencer services

### LINKED ISSUE #3134: Use reth task manager consistently
Some `tokio::spawn` are still lingering around and should be spawned as non-critical (or re-assessed as critical)

ercecan:
# Description
Went over other places to do the switch as well but felt like this is the only place that needs it atm.

## Linked Issues
- Fixes #3134

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### Added
 
 ### Changed
+- refactor: Use reth task manager instead of tokio spawn in sequencer services ([#3145](https://github.com/chainwayxyz/citrea/pull/3145))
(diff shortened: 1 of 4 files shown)


eyusufatik:
if we are not utilizing it in other tokio::spawn'ed services, let's also use the manager on them as well

ercecan:
> if we are not utilizing it in other tokio::spawn'ed services, let's also use the manager on them as well

I could not find other meaningful places to utilize, let me know if u got some on your mind @jfldde @exeokan

jfldde (dismissed):

exeokan (dismissed):

jfldde (dismissed):

eyusufatik (approved):

Issue #3153, opened 2026-02-18
flaky test: `test_eth_subscriptions`


eyusufatik:

jfldde:
Fixed in #3156

PR #3156, opened 2026-02-19
fix: subscription server unordered log delivery

### LINKED ISSUE #3155: Log subscription via websocket may return logs in random order
## Summary
Currently the websocket server implementation will spawn a new task for each websocket message sending process, since these tasks are executed randomly by the runtime, sometimes the order of the logs received by subscribers are not the same as they are in the blocks, which is against Ethereum spec `logIndex` semantics

In order to mitigate this we should open up a new task for each subscriber as they subscribe and pass it a broadcast stream, inside the task the subscribers will keep receiving the data from stream and instead of sending the data in a separate task, they will now send them sequentially to preserve `logIndex` ordering.

ercecan:
# Description
Because the messages are sent concurrently, this was the main reason behind flaky #3153  

## Linked Issues
- Fixes #3155 #3153

### DIFF
--- a/crates/ethereum-rpc/src/subscription.rs
+++ b/crates/ethereum-rpc/src/subscription.rs
@@ -27,8 +27,8 @@ impl SubscriptionManager {
         ledger_db: LedgerDB,
         l2_block_rx: broadcast::Receiver<u64>,
     ) -> Self {
-        let (heads_tx, _) = broadcast::channel(16);
-        let (logs_tx, _) = broadcast::channel(16);
+        let (heads_tx, _) = broadcast::channel(256);
+        let (logs_tx, _) = broadcast::channel(256);
 
         let l2_block_handle = tokio::spawn(l2_block_event_handler::<C>(
             storage,
--- a/crates/ethereum-rpc/src/subscription.rs
+++ b/crates/ethereum-rpc/src/subscription.rs
@@ -8,26 +8,17 @@ use jsonrpsee::{SubscriptionMessage, SubscriptionSink};
 use reth_rpc_eth_types::logs_utils::log_matches_filter;
 use sov_db::ledger_db::LedgerDB;
 use sov_modules_api::WorkingSet;
+use tokio::sync::broadcast;
 use tokio::sync::broadcast::error::RecvError;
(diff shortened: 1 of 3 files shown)


crates/ethereum-rpc/src/subscription.rs
```diff
@@ -37,150 +27,138 @@ impl SubscriptionManager {
         ledger_db: LedgerDB,
         l2_block_rx: broadcast::Receiver<u64>,
     ) -> Self {
-        let (new_heads_tx, new_heads_rx) = mpsc::channel(16);
-        let (logs_tx, logs_rx) = mpsc::channel(16);
-
-        let head_subscriptions = Arc::new(RwLock::new(vec![]));
-        let logs_subscriptions = Arc::new(RwLock::new(vec![]));
+        let (heads_tx, _) = broadcast::channel(16);
+        let (logs_tx, _) = broadcast::channel(16);
```
jfldde: Should bump that to higher values

crates/ethereum-rpc/src/subscription.rs
```diff
@@ -37,150 +27,138 @@ impl SubscriptionManager {
...
                 }
             }
         }
-        // Drop lock to release the read lock.
-        drop(subscriptions);
     }
 }
 
-pub async fn l2_block_event_handler<C: sov_modules_api::Context>(
+async fn l2_block_event_handler<C: sov_modules_api::Context>(
```
eyusufatik, line 150: if we put a shutdown handler here, subscription tasks auto close. but need to drop txs

eyusufatik (changes requested):
add changelog pls

crates/ethereum-rpc/src/subscription.rs
ercecan: When SubscriptionManager is dropped, heads_tx and logs_tx are dropped with it. This causes every subscriber task's rx.recv() to return RecvError::Closed on the next iteration, and the task exits. So the tasks are self-terminating.

jfldde (approved):

eyusufatik (approved):

PR #3160, opened 2026-02-20
fix: Prover dev mode config confusion

### LINKED ISSUE #3009: Unexpected real/fake proving behavior with local proving configs
I tried adding `dev_mode=true` to configs, but there is a "weird" behavior.
When `RISC0_DEV_MODE` is not set, `proving_mode=execute`:
`dev_mode=false` -> does fake proofs
`dev_mode=true` -> does real proofs

I think we should create an issue for this and address it elsewhere.

_Originally posted by @exeokan in https://github.com/chainwayxyz/citrea/pull/2994#discussion_r2486412296_

ercecan:
# Description
Currently if the config is read from env var there is no problem because
```rust
impl FromEnv for LocalProverConfig {
        let r0vm_path = read_env("RISC0_SERVER_PATH").ok().map(PathBuf::from);
        let dev_mode = is_dev_mode_enabled_via_environment();

        Ok(Self {
            r0vm_path,
            dev_mode,
        })
    }
}
```
The problem arises when the following conditions occur at the same time:
- The batch prover config is read from toml
- The `RISC0_DEV_MODE` env var is not set or set to false.
- `dev_mode` in prover config is set to true even though the  `RISC0_DEV_MODE` is not set or set to false
- and proving mode is execute which causes `with_prove` variable to be set to false.

In the current implementation with the above conditions, the first if will be entered and assertion will pass, after that 
`ProverOpts::groth16()` or `ProverOpts::succinct()` calls will call `is_dev_mode_enabled_via_environment()` which will return false even though `dev_mode` is set to false in our batch prover config toml file
```rust
if self.dev_mode {
            assert!(
                !with_prove,
                "Prove should not be called with prove in dev mode"
            );
        } else if with_prove {
            env::remove_var("RISC0_DEV_MODE");
        } else {
            env::set_var("RISC0_DEV_MODE", "1");
        }

        // std::fs::write("kumquat-input.bin", &input).unwrap();

        let prover_opts = match receipt_type {
            ReceiptType::Groth16 => ProverOpts::groth16(),
            ReceiptType::Succinct => ProverOpts::succinct(),
        };
```

by adding 
```rust
env::set_var("RISC0_DEV_MODE", "1");
```
to the end of the first if branch we can fix this

but per @exeokan suggestion we can also remove dev mode and get rid of setting risc0_dev_mode at all

```rust
// and just update prover opts
let prover_opts = prover_opts.with_dev_mode(!with_prove); // pass dev mode here
```
## Linked Issues
- Fixes #3009

### DIFF
--- a/crates/common/src/config/risc0.rs
+++ b/crates/common/src/config/risc0.rs
@@ -2,7 +2,7 @@ use std::path::PathBuf;
 
 use serde::{Deserialize, Serialize};
 
-use crate::utils::{is_dev_mode_enabled_via_environment, read_env};
+use crate::utils::read_env;
 use crate::FromEnv;
 
 /// Boundless storage configuration for S3
@@ -170,20 +170,13 @@ impl FromEnv for PricingServiceConfig {
 pub struct LocalProverConfig {
     /// Optional path to the r0vm binary
     pub r0vm_path: Option<PathBuf>,
-    /// Enable dev mode
-    #[serde(default)]
-    pub dev_mode: bool,
 }
 
 impl FromEnv for LocalProverConfig {
     fn from_env() -> anyhow::Result<Self> {
         let r0vm_path = read_env("RISC0_SERVER_PATH").ok().map(PathBuf::from);
-        let dev_mode = is_dev_mode_enabled_via_environment();
 
-        Ok(Self {
-            r0vm_path,
-            dev_mode,
-        })
+        Ok(Self { r0vm_path })
     }
 }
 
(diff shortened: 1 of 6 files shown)


crates/common/src/config/risc0.rs
```diff
@@ -170,20 +170,13 @@ impl FromEnv for PricingServiceConfig {
...
     pub r0vm_path: Option<PathBuf>,
-    /// Enable dev mode
-    #[serde(default)]
-    pub dev_mode: bool,
 }
 
 impl FromEnv for LocalProverConfig {
     fn from_env() -> anyhow::Result<Self> {
         let r0vm_path = read_env("RISC0_SERVER_PATH").ok().map(PathBuf::from);
-        let dev_mode = is_dev_mode_enabled_via_environment();
```
exeokan, line 181: I think we should also remove `is_dev_mode_enabled_via_environment` function, and `RISC0_DEV_MODE` related checks where the function is used

exeokan (changes requested):

crates/common/src/config/risc0.rs
ercecan: only used in boundless, bonsai and main to make sure it is not active on mainnet no harm to be safe I believe

exeokan (dismissed):

PR #3161, opened 2026-02-23
chore: Bump alloy, revm and reth

jfldde:
# Description

- Bump alloy, revm and reth


Cargo.toml
```diff
@@ -101,7 +101,7 @@ rustc_version_runtime = { version = "0.3.0", default-features = false }
 rs_merkle = "1.4.2"
 reqwest = { version = "0.12.12", features = ["rustls-tls", "json", "http2"], default-features = false }
 rocksdb = { version = "0.22.0", features = ["lz4"], default-features = false }
-serde = { version = "=1.0.218", default-features = false, features = ["alloc", "derive"] }
```
ercecan, line 104: oh that's a sight for sore eyes

bin/citrea/src/rollup/mod.rs
```diff
@@ -78,12 +78,12 @@ pub trait CitreaRollupBlueprint: RollupBlueprint {
         require_da_wallet: bool,
         network: Network,
     ) -> Result<Dependencies<Self>> {
-        let task_manager = TaskManager::current();
+        let task_manager = TaskExecutor::with_existing_handle(tokio::runtime::Handle::current())?;
```
ercecan: I think rename to task executor

ercecan: for all instances that are now executor but named manager

bin/citrea/src/main.rs
```diff
@@ -412,19 +412,34 @@ where
...
     let wait_duration = Duration::from_secs(5);
-
-    tokio::select! {
-        _ = signal::ctrl_c() => {}
-        _ = term_signal.recv() => {},
-        _ = interrupt_signal.recv() => {}
-        _= &mut task_manager => {}
+    let mut task_manager_handle = task_manager.take_task_manager_handle();
+
+    if let Some(handle) = &mut task_manager_handle {
```
ercecan, line 445: Is it possible for the handle to be None?

ercecan (changes requested):

bin/citrea/src/main.rs
jfldde: Could be racing if requesting shutdown at the same time as a a panic on a critical task. Probability is low though

PR #3162, opened 2026-02-23
fix(rpc): Disable base fee/eip-3607 on debug_traceCall

exeokan:
# Description
When no gas price/ base fee is provided in `debug_traceCall` transaction params, gas price is set to 0, and the trace fails because gas price < block base fee.

With this PR, we mirror reth behaviour and disable the base fee check on `debug_traceCall`.
Also disabled to eip-3607 and block gas limit to match our `eth_call` impl.

Example call:
```
{
    "id": 1,
    "jsonrpc": "2.0",
    "method": "debug_traceCall",
    "params": [
        {
            "from": "0x0f820f428ae436c1000b27577bf5bbf09bfec8f2",
            "value": "0x0"
        }
    ]
}
```
Nightly:
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32003,
        "message": "max fee per gas less than block base fee"
    }
}

```
Head:
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "failed": false,
        "gas": 53000,
        "returnValue": "",
        "structLogs": [
            {
                "pc": 0,
                "op": "STOP",
                "gas": 29947000,
                "gasCost": 0,
                "depth": 1,
                "stack": []
            }
        ]
    }
}
```

### DIFF
--- a/crates/evm/src/query.rs
+++ b/crates/evm/src/query.rs
@@ -1534,10 +1534,10 @@ impl<C: sov_modules_api::Context> Evm<C> {
             .expect("EVM chain config should be set");
 
         let mut cfg_env = get_cfg_env(cfg, evm_spec_id);
-
-        // Ignore the base fee check for trace calls
-        // This allows requests with a gas price lower than block base fee to succeed,
-        // Reference: <https://github.com/paradigmxyz/reth/blob/ed7da87da4de340a437bf46f39a7e1397ac82065/crates/rpc/rpc-eth-api/src/helpers/call.rs#L734>
+        // Match the CfgEnv of eth_call here
+        // Also see: <https://github.com/paradigmxyz/reth/blob/564ffa586845fa4a8bb066f0c7b015ff36b26c08/crates/rpc/rpc-eth-api/src/helpers/call.rs#L855>
+        cfg_env.disable_block_gas_limit = true;
+        cfg_env.disable_eip3607 = true;
         cfg_env.disable_base_fee = true;
 
         let l1_fee_block_num = match block_number {
--- a/crates/evm/src/query.rs
+++ b/crates/evm/src/query.rs
@@ -1533,7 +1533,10 @@ impl<C: sov_modules_api::Context> Evm<C> {
             .get(working_set)
             .expect("EVM chain config should be set");
 
-        let cfg_env = get_cfg_env(cfg, evm_spec_id);
+        let mut cfg_env = get_cfg_env(cfg, evm_spec_id);
+        // See:
+        // <https://github.com/paradigmxyz/reth/blob/ed7da87da4de340a437bf46f39a7e1397ac82065/crates/rpc/rpc-eth-api/src/helpers/call.rs#L734>
+        cfg_env.disable_base_fee = true;
 
         let l1_fee_block_num = match block_number {
             // use l1 fee rate of latest block for pending block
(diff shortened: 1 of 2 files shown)


crates/evm/src/query.rs
```diff
@@ -1533,7 +1533,12 @@ impl<C: sov_modules_api::Context> Evm<C> {
             .get(working_set)
             .expect("EVM chain config should be set");
 
-        let cfg_env = get_cfg_env(cfg, evm_spec_id);
+        let mut cfg_env = get_cfg_env(cfg, evm_spec_id);
+
+        // Ignore the base fee check for trace calls
+        // This allows requests with a gas price lower than block base fee to succeed,
+        // Reference: <https://github.com/paradigmxyz/reth/blob/ed7da87da4de340a437bf46f39a7e1397ac82065/crates/rpc/rpc-eth-api/src/helpers/call.rs#L734>
```
jfldde: Are we also doing this : 
```

        // Disabled because eth_call is sometimes used with eoa senders
        // See <https://github.com/paradigmxyz/reth/issues/1959>
        evm_env.cfg_env.disable_eip3607 = true;
```

In line with reth ?

exeokan: added 
```
cfg_env.disable_block_gas_limit = true;
cfg_env.disable_eip3607 = true;
```
to match our `CfgEnv` in `eth_call`, also left a reference to reth v1.11.0.

exeokan:
I want to note that even though I added `disable_block_gas_limit = true` to match our `eth_call` implementation, we cap the gas limit by block gas limit in `create_txn_env`, so it has no effect. This is also true for `eth_call`.

jfldde (dismissed):

crates/evm/src/query.rs
```diff
@@ -1533,7 +1533,12 @@ impl<C: sov_modules_api::Context> Evm<C> {
             .get(working_set)
             .expect("EVM chain config should be set");
 
-        let cfg_env = get_cfg_env(cfg, evm_spec_id);
+        let mut cfg_env = get_cfg_env(cfg, evm_spec_id);
+        // Match the CfgEnv of eth_call here
+        // Also see: <https://github.com/paradigmxyz/reth/blob/564ffa586845fa4a8bb066f0c7b015ff36b26c08/crates/rpc/rpc-eth-api/src/helpers/call.rs#L855>
+        cfg_env.disable_block_gas_limit = true;
+        cfg_env.disable_eip3607 = true;
+        cfg_env.disable_base_fee = true;
```
ercecan: Is there a reason why we are not using `prepare_call_env` that we use in `eth_call`?

ercecan (changes requested):

crates/evm/src/query.rs
exeokan: yeah that makes sense, made changes

eyusufatik:
> Also disabled to eip-3607 and block gas limit to match our eth_call impl.
is this the behaviour in geth/reht

exeokan:
> > Also disabled to eip-3607 and block gas limit to match our eth_call impl.
> > is this the behaviour in geth/reht

Yes, on reth: [call.rs](https://github.com/paradigmxyz/reth/blob/564ffa586845fa4a8bb066f0c7b015ff36b26c08/crates/rpc/rpc-eth-api/src/helpers/call.rs#L855)

on geth:
NoBaseFee is set here: [api.go](https://github.com/ethereum/go-ethereum/blob/8a4345611dc64e9c10414cb23dd0d5d38f8e532c/eth/tracers/api.go#L1034)
SkipTransactionChecks is set here: [transaction_args.go](https://github.com/ethereum/go-ethereum/blob/8a4345611dc64e9c10414cb23dd0d5d38f8e532c/internal/ethapi/transaction_args.go#L495)

eyusufatik (approved):

ercecan (approved):

jfldde (approved):

PR #3167, opened 2026-02-27
fix: L1 sync start height

jfldde:
# Description

- Prevent L1 syncer from re-processing last scanned block on startup

### DIFF
--- a/bin/citrea/src/main.rs
+++ b/bin/citrea/src/main.rs
@@ -371,17 +371,16 @@ where
 
             start_rpc_server(rollup_config.rpc.clone(), &task_executor, rpc_module, None);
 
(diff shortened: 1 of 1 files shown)


ercecan (approved):

ercecan (approved):

exeokan (dismissed):
Seems like the same problem also happens with the batch prover: https://github.com/chainwayxyz/citrea/blob/ceb04b6e2178184bd62c031598fd065632e7b5a5/crates/batch-prover/src/l1_syncer.rs#L105

We can also move `citrea_light_client_prover::da_block_handler::StartVariant` to `citrea_common` and use it for full node/batch prover, I think it would be more robust.

eyusufatik (approved):

PR #3169, opened 2026-02-27
fix: set L1 fee rate to zero in eth_estimateGas if sender has no balance

exeokan:
# Description
When the from address is not specified or has zero balance (a simulation scenario), `eth_estimateGas` failed because the transaction was inspected using a non-zero L1 fee rate.
This PR resolves the issue by executing with a zero L1 fee rate in those cases, and computing the L1 fee afterward in the handler using the L1 diff from execution.

## Testing
Added `test_estimate_gas_no_balance`.


jfldde (dismissed):

ercecan (dismissed):

jfldde (dismissed):

crates/evm/src/query.rs
```diff
@@ -1066,6 +1068,12 @@ impl<C: sov_modules_api::Context> Evm<C> {
         let block_gas_limit = U64::from(block_env_gas_limit);
         let block_env_base_fee = U256::from(block_env.basefee);
 
+        let inspect_l1_fee_rate = if account.balance > 0 {
```
exeokan: Upon discussion with Esad, we decided to run with zero L1 fee only if "from" account balance is zero(this is the case in simulations).
This is to keep the functionality where we can learn that a regular tx will fail due to L1 fee using `eth_estimateGas.`

crates/evm/src/query.rs
```diff
@@ -1036,15 +1036,17 @@ impl<C: sov_modules_api::Context> Evm<C> {
...
-            .account_info(&request.from.unwrap_or_default(), working_set)
-            .unwrap_or_default();
-
         let mut evm_db = self.get_db(working_set, citrea_spec_id);
 
         if let Some(ref state_overrides) = state_overrides {
             apply_state_overrides(state_overrides.clone(), &mut evm_db)?;
         }
+
+        let account: crate::AccountInfo = evm_db
```
eyusufatik, line 1045: from is Nonce case is covered with address 0x000..00 but in many networks, this address has a balance so the assumption is broken (wrt simulations) so please handle this differently

crates/evm/src/query.rs
```diff
@@ -2112,7 +2111,7 @@ fn map_out_of_gas_err<C: sov_modules_api::Context>(
         cfg_env,
         block_env,
         tx_env,
-        l1_fee_rate,
+        /* l1_fee_rate */ 0,
```
eyusufatik: forgot to change this maybe?

eyusufatik (changes requested):

crates/evm/src/query.rs
exeokan: made changes

eyusufatik (dismissed):

eyusufatik (approved):

PR #3165, opened 2026-02-26
feat: Stop at height fullnode argument

jfldde:
# Description

- Hidden args to stop the node at a target L1/L2 height

## Linked Issues
- Fixes #3063

PR #2983, opened 2025-10-16
chore: bump rocksdb and use jemalloc feature

eyusufatik:
# Description

## Linked Issues
- Fixes # (issue, if applicable)
- Related to # (issue) 

## Testing
Describe how these changes were tested. If you've added new features, have you added unit tests?

## Docs
Describe where this code is documented. If it changes a documented interface, have the docs been updated?

PR #3170, opened 2026-03-02
feat: Rocksdb diff script

jfldde:
# Description

- Script used for diffing rocksdb post upgrade.
- Used for ensuring #2983 upgrade doesn't introduce any regression
- Will be used in conjunction with #3165 to stop at L1/L2 height, stop both nodes, and run this script


exeokan (commented):
Not sure if it is the purpose of the script, but we may also want to assert that column families match between two DBs. Example:
Db1 column families: X, Y,
Db2 column families: X, Y, Z

eyusufatik (dismissed):

eyusufatik (approved):

exeokan (approved):

PR #3185, opened 2026-03-11
refactor: Use StartVariant in L1 processing

### LINKED ISSUE #3181: Use StartVariant consistently
Seems like the same problem also happens with the batch prover: https://github.com/chainwayxyz/citrea/blob/ceb04b6e2178184bd62c031598fd065632e7b5a5/crates/batch-prover/src/l1_syncer.rs#L105

We can also move `citrea_light_client_prover::da_block_handler::StartVariant` to `citrea_common` and use it for full node/batch prover, I think it would be more robust.

_Originally posted by @exeokan in https://github.com/chainwayxyz/citrea/pull/3167#pullrequestreview-3889795327_

exeokan:
# Description
Also fixes the batch prover bug where it re-processes the last scanned L1 block on start.

## Linked Issues
- Fixes #3181

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
   **New hidden node args:**\
   `--stop-at-l1-height`: Stop the full node L1 sync after reaching this L1 height
   `--stop-at-l2-height`: Stop the full node L2 sync after reaching this L2 height
+- refactor: Use `StartVariant` in L1 processing ([#3185](https://github.com/chainwayxyz/citrea/pull/3185))
 
 ## [v2.2.0](2026-03-02)
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@
   **New hidden node args:**\
   `--stop-at-l1-height`: Stop the full node L1 sync after reaching this L1 height
   `--stop-at-l2-height`: Stop the full node L2 sync after reaching this L2 height
-- refactor: Use `StartVariant` in L1 processing ([#3185](https://github.com/chainwayxyz/citrea/pull/3185))
 
 ## [v2.2.0](2026-03-02)
 ### Added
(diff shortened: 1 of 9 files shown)


bin/citrea/tests/bitcoin/backup.rs
```diff
@@ -817,7 +817,7 @@ impl TestCase for BackupBatchProverTest {
         let rollback_target_l1 = commitment_l1_height;
         let proof_output = first_proof.proof_output;
 
-        let rollback_target_commitment_index = 0;
+        let rollback_target_commitment_index = 1;
```
exeokan, line 820: I think rolling commitments back to zero was a mistake here; the test succeeds on nightly because of the bug where the batch prover re-processes the last scanned l1 block on start (so it recovers the commitment 1) @jfldde

jfldde:
Need to handle rollback as we reset `ProverLastScannedSlot` and we would miss a block on restart

exeokan:
> Need to handle rollback as we reset `ProverLastScannedSlot` and we would miss a block on restart

Isn't it fine already? The DA data on the L1 target height survives the rollback, and we set the last scanned height to the L1 target. Node then starts from the height: l1 target + 1

jfldde:
> > Need to handle rollback as we reset `ProverLastScannedSlot` and we would miss a block on restart
> 
> Isn't it fine already? The DA data on the L1 target height survives the rollback, and we set the last scanned height to the L1 target. Node then starts from the height: l1 target + 1

Oh right, we don't actually rollback to `target_height` but `target_height + 1`.
`for i in l1_target + 1..=last_scanned_l1_height.0`

Wasn't clear but this should be fine then

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
   **New hidden node args:**\
   `--stop-at-l1-height`: Stop the full node L1 sync after reaching this L1 height
   `--stop-at-l2-height`: Stop the full node L2 sync after reaching this L2 height
+- refactor: Use `StartVariant` in L1 processing ([#3185](https://github.com/chainwayxyz/citrea/pull/3185))
```
jfldde: I wouldn't add refactors to changelog

jfldde (dismissed):

eyusufatik (approved):

jfldde (approved):

PR #3192, opened 2026-03-20
chore: Simplify tx prefix mining

jfldde:
# Description

- Rely on randomized Schnorr signing from `sign_schnorr` in `update_witness` to mine tx prefix instead of tweaking `lock_time`

### DIFF
--- a/crates/bitcoin-da/src/helpers/builders/body_builders.rs
+++ b/crates/bitcoin-da/src/helpers/builders/body_builders.rs
@@ -3,7 +3,6 @@
 use core::result::Result::Ok;
 use std::time::Instant;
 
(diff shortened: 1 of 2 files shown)


crates/bitcoin-da/src/helpers/builders/body_builders.rs
```diff
@@ -78,10 +77,61 @@ pub enum DaTxs {
     },
 }
 
+/// Mine the reveal transaction's prefix
+fn mine_reveal_prefix(
```
ercecan, line 81: Maybe also comment that we Rely on randomized Schnorr signing from sign_schnorr

ercecan (changes requested):

eyusufatik (approved):

PR #3201, opened 2026-04-02
feat: Get raw transaction rpcs

ercecan:
# Description

## Linked Issues
- Related to # (issue) 

## Testing

## Docs

### DIFF
--- a/bin/citrea/tests/mock/evm/mod.rs
+++ b/bin/citrea/tests/mock/evm/mod.rs
@@ -868,9 +868,7 @@ async fn execute(client: &Box<TestClient>) -> Result<(), Box<dyn std::error::Err
     );
 
     // Assert getRawTransactionByHash
-    let raw_tx_by_hash = client
-        .eth_get_raw_transaction_by_hash(tx_hash, None)
-        .await;
+    let raw_tx_by_hash = client.eth_get_raw_transaction_by_hash(tx_hash, None).await;
     assert!(raw_tx_by_hash.is_some());
 
     // Encode the transaction locally and verify it matches the RPC response
--- a/bin/citrea/tests/mock/evm/mod.rs
+++ b/bin/citrea/tests/mock/evm/mod.rs
@@ -4,6 +4,7 @@ use std::time::Duration;
 
 use alloy::consensus::constants::KECCAK_EMPTY;
 use alloy::hex::FromHex;
+use alloy::network::eip2718::Encodable2718;
 use alloy::network::TransactionResponse;
 use alloy::signers::local::PrivateKeySigner;
 use alloy::signers::SignerSync;
@@ -861,10 +862,41 @@ async fn execute(client: &Box<TestClient>) -> Result<(), Box<dyn std::error::Err
 
     // Verify both block-based raw transaction methods return the same bytes
     assert_eq!(
-        raw_tx_by_block_hash.unwrap(),
-        raw_tx_by_number.unwrap(),
+        raw_tx_by_block_hash.as_ref().unwrap(),
+        raw_tx_by_number.as_ref().unwrap(),
         "Raw tx by block hash and by block number should be identical"
     );
+
+    // Assert getRawTransactionByHash
+    let raw_tx_by_hash = client
+        .eth_get_raw_transaction_by_hash(tx_hash, None)
+        .await;
+    assert!(raw_tx_by_hash.is_some());
+
+    // Encode the transaction locally and verify it matches the RPC response
+    let tx_obj = client
+        .eth_get_transaction_by_hash(tx_hash, None)
+        .await
+        .expect("Transaction should exist");
+    let locally_encoded: Bytes = tx_obj.inner.encoded_2718().into();
+    assert_eq!(
+        raw_tx_by_hash.as_ref().unwrap(),
+        &locally_encoded,
+        "Raw tx by hash should match locally encoded transaction bytes"
+    );
+
+    // Also verify consistency with block-based raw transaction methods
+    assert_eq!(
+        raw_tx_by_hash.as_ref().unwrap(),
+        raw_tx_by_block_hash.as_ref().unwrap(),
+        "Raw tx by hash and by block hash should be identical"
+    );
+
+    // Assert getRawTransactionByHash with non-existent hash returns None
+    let raw_tx_nonexistent = client
+        .eth_get_raw_transaction_by_hash(B256::ZERO, None)
+        .await;
+    assert!(raw_tx_nonexistent.is_none());
 
     let get_arg: U256 = client
         .contract_call(contract_address, contract.get_call_data(), None)
--- a/bin/citrea/tests/mock/evm/mod.rs
+++ b/bin/citrea/tests/mock/evm/mod.rs
@@ -8,7 +8,7 @@ use alloy::network::TransactionResponse;
 use alloy::signers::local::PrivateKeySigner;
 use alloy::signers::SignerSync;
 // use citrea::initialize_logging;
-use alloy_primitives::{Address, Bytes, U256, U64};
+use alloy_primitives::{Address, Bytes, B256, U256, U64};
 use alloy_rpc_types::{
     Authorization, BlockId, BlockNumberOrTag, EIP1186AccountProofResponse, Filter,
     TransactionRequest, ValueOrArray,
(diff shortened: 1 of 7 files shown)


crates/ethereum-rpc/src/lib.rs
```diff
@@ -859,6 +884,94 @@ where
...
             .await?)
     }
+
+    async fn eth_get_raw_transaction_by_hash(
+        &self,
+        hash: B256,
+        mempool_only: Option<bool>,
+    ) -> RpcResult<Option<Bytes>> {
+        match mempool_only {
+            Some(true) => {
```
jfldde, line 878: Might be better to try locally first and only go through sequencer if not found

crates/ethereum-rpc/src/lib.rs
```diff
@@ -859,6 +884,94 @@ where
...
+                let mut working_set = WorkingSet::new(self.ethereum.storage.clone());
+                match evm.get_transaction_by_hash(hash, &mut working_set) {
+                    Ok(Some(tx)) => Ok(Some(tx.as_recovered().encoded_2718().into())),
+                    Ok(None) => {
+                        match self
+                            .ethereum
+                            .sequencer_client
+                            .as_ref()
+                            .unwrap()
+                            .eth_get_transaction_by_hash(hash, Some(true))
```
jfldde:
```suggestion
                            .eth_get_transaction_by_hash(hash, mempool_only)
```

crates/ethereum-rpc/src/lib.rs
ercecan: that's what we're doing if `mempool_only` is not set to `true`, if you mean we should check locally even if `mempool_only` is set I do not think it makes sense

crates/ethereum-rpc/src/lib.rs
ercecan: why? We check the db and ask sequencer mempool

jfldde: That's the branch where `mempool_only` is either `Some(false)` or `None`. Why do we query the sequencer with `Some(true)` instead ?

crates/ethereum-rpc/src/lib.rs
jfldde: Idea is to reduce load on the sequencer. The probability that the tx is in mempool with 2s block time is very low. 
Btw, I couldn't find other implementation using `mempool_only` as second param for `eth_get_raw_transaction_by_hash`. Can you shed some context on which implementation we are following for this one ?

jfldde: Ok this is custom, just found it in sequencer RPC file.

crates/ethereum-rpc/src/lib.rs
ercecan: because if `mempool_only` is not set it should check both the db and the mempool afterwards if it is not in a block, if `mempool_only` is set it should only check mempool. 

`mempool_only` is only set if user wants to check only the mempool

crates/ethereum-rpc/src/lib.rs
ercecan: >Idea is to reduce load on the sequencer.

Of course, but the main idea of the parameter is to check only the mempool

ercecan: and we do the same behavior in eth_getTransactionByHash as well, following convention from that

crates/ethereum-rpc/src/lib.rs
```diff
@@ -219,6 +220,30 @@ pub trait EthereumRpc {
...
+    ) -> RpcResult<Option<Bytes>>;
+
+    /// Get raw transaction by block number and index
+    #[method(name = "eth_getRawTransactionByBlockNumberAndIndex")]
+    async fn eth_get_raw_transaction_by_block_number_and_index(
+        &self,
+        block_number: BlockNumberOrTag,
+        index: U64,
+    ) -> RpcResult<Option<Bytes>>;
 }
```
eyusufatik: since these don't need to query the sequencer, maybe it's better to put them in the evm crate

crates/ethereum-rpc/src/lib.rs
```diff
@@ -912,6 +1025,7 @@ where
         module.remove_method("eth_sendRawTransaction");
         module.remove_method("eth_sendRawTransactionSync");
         module.remove_method("eth_getTransactionByHash");
+        module.remove_method("eth_getRawTransactionByHash");
```
eyusufatik, line 973: then add it in the sequencer's rpc file

crates/ethereum-rpc/src/lib.rs
```diff
@@ -859,6 +884,94 @@ where
...
+                        jsonrpsee::core::client::Error::Call(e_owned) => Err(e_owned),
+                        _ => Err(to_jsonrpsee_error_object("SEQUENCER_CLIENT_ERROR", e)),
+                    },
+                }
+            }
+            _ => {
+                let evm = Evm::<C>::default();
+                let mut working_set = WorkingSet::new(self.ethereum.storage.clone());
+                match evm.get_transaction_by_hash(hash, &mut working_set) {
+                    Ok(Some(tx)) => Ok(Some(tx.as_recovered().encoded_2718().into())),
```
eyusufatik, line 898: please cross check with reth that this is the expected encoding here

crates/ethereum-rpc/src/lib.rs
```diff
@@ -859,6 +884,94 @@ where
...
+        index: U64,
+    ) -> RpcResult<Option<Bytes>> {
+        let evm = Evm::<C>::default();
+        let mut working_set = WorkingSet::new(self.ethereum.storage.clone());
+        match evm.get_transaction_by_block_hash_and_index(block_hash, index, &mut working_set) {
+            Ok(Some(tx)) => Ok(Some(tx.as_recovered().encoded_2718().into())),
+            Ok(None) => Ok(None),
+            Err(e) => Err(e),
+        }
+    }
```
eyusufatik: yep, would look much better in evm rpc impl

bin/citrea/tests/mock/evm/mod.rs
```diff
@@ -807,6 +807,65 @@ async fn execute(client: &Box<TestClient>) -> Result<(), Box<dyn std::error::Err
     assert_eq!(tx_by_number.tx_hash(), tx_hash);
     assert_eq!(tx_by_number_tag.tx_hash(), tx_hash);
```
eyusufatik, line 810: somewhere in this file, encode tx locally and assert rpc returns the same bytes

eyusufatik (changes requested):

eyusufatik (dismissed):

crates/ethereum-rpc/src/lib.rs
exeokan: When calling the sequencer with `mempool_only = Some(true)`, I think we miss the tx in this scenario:

1. eth_get_transaction_by_hash request with `mempool_only=None` comes
2. full node checks db-> not found
3. calls sequencer
4. A new block is produced with the target tx in it
5. sequencer checks mempool -> not found
6. sequencer returns None

ercecan: Okay I am convinced, we can get rid of the race condition with it, makes sense

exeokan (dismissed):

crates/ethereum-rpc/src/lib.rs
```diff
@@ -658,7 +658,7 @@ where
                             .sequencer_client
                             .as_ref()
                             .unwrap()
-                            .eth_get_transaction_by_hash(hash, Some(true))
+                            .eth_get_transaction_by_hash(hash, mempool_only)
```
eyusufatik: node already checked for local, why would you not ask the sequencer for mempool_only? saving the sequencer state read  is the only reason we did the mempool only flag

crates/ethereum-rpc/src/lib.rs
```diff
@@ -905,13 +902,10 @@ where
                             .sequencer_client
                             .as_ref()
                             .unwrap()
-                            .eth_get_transaction_by_hash(hash, Some(true))
+                            .eth_get_raw_transaction_by_hash(hash, mempool_only)
```
eyusufatik: same here

eyusufatik (changes requested):

crates/ethereum-rpc/src/lib.rs
jfldde: cc/ @eyusufatik. Check @exeokan comment.

> node already checked for local, why would you not ask the sequencer for mempool_only? saving the sequencer state read is the only reason we did the mempool only flag

Sequencer goes through mempool first, so we skip state read if it's in mempool but we don't miss it if block is produced including tx within RPC call

crates/ethereum-rpc/src/lib.rs
jfldde: Since this second argument is not standard, and is there to reduce sequencer state read.

> that's what we're doing if mempool_only is not set to true, if you mean we should check locally even if mempool_only is set I do not think it makes sense

If it could be found in fullnode state, why query the sequencer at all ? This non-standard argument is there to reduce sequencer load so it seems natural to query as much as possible fullnode side.

jfldde: I mean, if we really want `mempool_only`, we could return `None` when found in fullnode state

crates/ethereum-rpc/src/lib.rs
ercecan: that's what I went with first but this [comment](https://github.com/chainwayxyz/citrea/pull/3201/#discussion_r3032805747) changed my mind

crates/ethereum-rpc/src/lib.rs
ercecan: I get your point in reducing sequencer load, but we do the exact same thing in `eth_getTransactionByHash`, I feel like this one could stay the same as well, @eyusufatik what do you think?

crates/ethereum-rpc/src/lib.rs
eyusufatik: I think it's fair for somebody to query a few seconds later for a second time in this case. 
With this change, it becomes a bit of a DOSing problem for the sequencer unfortunately.

crates/ethereum-rpc/src/lib.rs
ercecan: He said he agrees so resolving the issue

eyusufatik (approved):

PR #3205, opened 2026-04-06
ci(proving-stats): simplify cycle reading 

### LINKED ISSUE #3022: CI: simplify proving-stats cycle reading
After #3011, we can directly query batch prover RPC instead of reading its logs.
Modify `proving-stats/get-proving-stats.py`

exeokan:
# Description
Uses the new batch prover API that allows reading the proving session info(including cycles) in `getProvingJob*` RPCs.

Fixes #3022


eyusufatik (dismissed):
nice

PR #3210, opened 2026-04-08
test(bitcoin): enable mempool_accept regression coverage

umeeSthein:
## Summary

- enable the dormant `mempool_accept` bitcoin e2e test
- replace the existing TODO with assertions around commit/reveal atomicity
- verify that restrictive ancestor policy never leaves the sequencer in a partial-broadcast state
- cover both immediate acceptance and retry-on-next-block behavior


eyusufatik:
won't be relevant once we merge txsender

PR #3214, opened 2026-04-21
ci: Reproducible builds

jfldde:
# Description

- Use nix for reproducible builds
- Use `actions/attest@v4` for attestations : https://github.com/chainwayxyz/citrea/attestations

## Testing

- Reproduce steps in `docs/reproducible-builds.md`


.github/workflows/release.yml
```diff
@@ -1,187 +1,152 @@
...
+  contents: write
+  id-token: write
+  attestations: write
 
 on:
   push:
     tags:
       - "v*.*.*"
+    branches:
+      - feat/repr-build
```
ercecan, line 13: is this for testing?

.github/workflows/release.yml
```diff
@@ -1,187 +1,152 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: linux-amd64
+            runner: ubicloud-standard-30-ubuntu-2204
+          - platform: linux-arm64
+            runner: ubicloud-standard-30-arm-ubuntu-2204
+          - platform: osx-arm64
+            runner: macos-latest
```
ercecan, line 26: why are we not using self-hosted-citrea-osx-arm64 now

jfldde: This is part of [458d2b9](https://github.com/chainwayxyz/citrea/pull/3214/commits/458d2b984ca1dea73f306a07dcfe83722119f440), so it can be signed

.github/workflows/release.yml
jfldde: Yes, this will be removed, keeping it for now to showcase the attestation and for us to be able to compare the hashes on latest run, see here https://github.com/chainwayxyz/citrea/actions/runs/25039795460

ercecan (approved):

exeokan (approved):

eyusufatik (approved):

PR #3221, opened 2026-04-27
chore: Fix actionlint issues

jfldde:
# Description

- Fix `actionlint` issues in CI workflows

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -151,10 +151,9 @@ jobs:
       - uses: actions/checkout@v5
       - uses: rui314/setup-mold@v1
       - name: Install nightly toolchain
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@nightly
         with:
           toolchain: nightly-2025-09-24
-          override: true
       - name: Run cargo-udeps
         env:
           RUSTFLAGS: -A warnings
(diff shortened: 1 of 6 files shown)


.github/workflows/checks.yml
```diff
@@ -597,15 +596,16 @@ jobs:
...
 
       - name: Get commit hash of the base branch
         id: base-commit
         run: |
-            BASE_REF="${{ github.event.pull_request.base.ref }}"
-            git fetch origin "$BASE_REF"
-            echo "base-hash=$(git rev-parse origin/"$BASE_REF")" >> $GITHUB_OUTPUT
+          BASE_REF="${{ github.event.pull_request.base.ref }}"
+          git fetch origin "$BASE_REF"
+          base_hash=$(git rev-parse "origin/$BASE_REF")
```
eyusufatik: "origin/$BASE_REF" might work differently then what we previously had

jfldde: compared to `origin/"$BASE_REF"` ?

eyusufatik: yeah. just wondering

eyusufatik:
@exeokan why didn't the cycle count ci work?

exeokan:
@eyusufatik you need to push a new commit after adding the guest code label
(you can also add the label while opening the pr)

.github/workflows/checks.yml
jfldde: I don't think so but I can remove it to err on the safe side

PR #3225, opened 2026-04-30
fix(bitcoin-da): use fee_rate in validate_txs_fee_rate comparisons

amathxbt:
## Summary

`validate_txs_fee_rate` receives a `fee_rate` parameter (in sat/vB) but the two fee-sufficiency checks compare `(input_amount - output_amount)` against just `Amount::from_sat(tx.vsize() as u64)`, which implicitly assumes 1 sat/vB regardless of the configured fee rate.

## Bug

```rust
// Before – always validates against 1 sat/vB:
if (input_amount - output_amount) < Amount::from_sat(commit_tx.vsize() as u64) {
    return Err(BitcoinServiceError::FeeCalculation(fee_rate));
}
```

If the system is operating at, say, 5 sat/vB, this check still passes a transaction that only carries a 1 sat/vB fee, so `FeeCalculation` errors are never surfaced for under-paying transactions at the actual fee rate.

## Fix

Multiply `vsize` by `fee_rate` (rounded up) in both the commit-tx and reveal-tx checks:

```rust
// After – validates against the actual fee rate:
if (input_amount - output_amount) < Amount::from_sat((commit_tx.vsize() as f64 * fee_rate).ceil() as u64) {
    return Err(BitcoinServiceError::FeeCalculation(fee_rate));
}
```

## Files changed
- `crates/bitcoin-da/src/fee.rs`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Changed
 - perf: Mine reveal prefix relying on `sign_schnorr` internal randomness.([#3192](https://github.com/chainwayxyz/citrea/pull/3192))
 - fix: use deterministic boundless patch and unpin ubuntu ci.([#3216](https://github.com/chainwayxyz/citrea/pull/3216))
+- fix(bitcoin-da): use `fee_rate` in `validate_txs_fee_rate` fee comparisons. ([#3225](https://github.com/chainwayxyz/citrea/pull/3225))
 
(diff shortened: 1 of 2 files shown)


jfldde:
Closing as fee calculation will be handled by tx-sender after #3206

amathxbt:
> Closing as fee calculation will be handled by tx-sender after #3206 

Ok boss

PR #3229, opened 2026-04-30
fix(bitcoin-da): replace debug_assert with anyhow::ensure for sequencer commitment body size limit

amathxbt:
## Summary

`create_inscription_type_4` uses `debug_assert!` to enforce the 520-byte limit on serialized sequencer commitment bodies. `debug_assert!` is compiled away in release builds, which means the check is silently inactive in production.

## Bug

```rust
// Only active in debug builds – stripped in release:
debug_assert!(
    body.len() < 520,
    "The body of a serialized sequencer commitment exceeds 520 bytes"
);
```

If a commitment body exceeds 520 bytes in a release build, the code proceeds to:

```rust
.push_slice(PushBytesBuf::try_from(body).expect("Cannot push sequencer commitment"))
```

`PushBytesBuf::try_from` returns `Err` for slices > 520 bytes, so `.expect(...)` **panics** the process. The caller never receives a proper `Err`.

## Fix

Replace `debug_assert!` with `anyhow::ensure!` so the guard is always active and a clean `Err` is returned instead of panicking:

```rust
anyhow::ensure!(
    body.len() < 520,
    "The body of a serialized sequencer commitment exceeds 520 bytes"
);
```

## Files changed
- `crates/bitcoin-da/src/helpers/builders/body_builders.rs`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Changed
 - perf: Mine reveal prefix relying on `sign_schnorr` internal randomness.([#3192](https://github.com/chainwayxyz/citrea/pull/3192))
 - fix: use deterministic boundless patch and unpin ubuntu ci.([#3216](https://github.com/chainwayxyz/citrea/pull/3216))
+- fix(bitcoin-da): replace `debug_assert!` with `anyhow::ensure!` for sequencer commitment body size. ([#3229](https://github.com/chainwayxyz/citrea/pull/3229))
 
 ## [v2.3.1](2026-03-31)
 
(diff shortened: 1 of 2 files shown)


jfldde:
Closing as transaction building will be handled by tx-sender post #3206

amathxbt:
> Closing as transaction building will be handled by tx-sender post #3206 

Ok sir thanks !

Still finding some I'll keep them updating here 

Thanks again !

amathxbt:
> > Closing as transaction building will be handled by tx-sender post #3206
> 
> Ok sir thanks !
> 
> Still finding some I'll keep them updating here
> 
> Thanks again !

Also sir seems there's a lot of CL keep failing on #3206 
If never mind I'll fix this too ! @jfldde  asap

PR #3230, opened 2026-04-30
fix(bitcoin-da): replace zip_eq with explicit length check in verify_transactions

amathxbt:
## Summary

`itertools::zip_eq` in `verify_transactions` will **panic** (not return an error) if `completeness_proof.len()` does not equal the number of relevant wtxids (those matching the reveal-tx prefix). Any caller ncluding the ZK guest circuit  that passes a mismatched completeness proof causes an unrecoverable panic instead of a clean `Err(ValidationError::RelevantTxNotInProof)`.

### Root cause

```rust
// crates/bitcoin-da/src/verifier.rs
let relevant_wtxid_iter = inclusion_proof
    .wtxids
    .iter()
    .filter(|wtxid| wtxid.starts_with(prefix));
for (wtxid, tx) in relevant_wtxid_iter.zip_eq(&completeness_proof) {  // ← panics on length mismatch
```

From the [itertools docs](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.zip_eq):
> *Panics if the iterators are not of equal length.*

A verifier must **never** panic — it must propagate errors. A malformed or adversarially crafted proof that contains a different number of transactions than relevant wtxids will crash any node running verification.

### Fix

Collect the filtered wtxids first, check the lengths explicitly, then use plain `Iterator::zip`:

```rust
let relevant_wtxids: Vec<_> = inclusion_proof
    .wtxids
    .iter()
    .filter(|wtxid| wtxid.starts_with(prefix))
    .collect();

if relevant_wtxids.len() != completeness_proof.len() {
    return Err(ValidationError::RelevantTxNotInProof);
}

for (wtxid, tx) in relevant_wtxids.iter().zip(&completeness_proof) {
```

The `use itertools::Itertools;` import is also removed since it is no longer needed.

### Impact

- **Severity: Critical** any proof with a length-mismatched completeness proof causes a panic in the verifier, affecting both the host node and the ZK guest.
- No behaviour change for well-formed proofs.

cc @jfldde @eyusufatik @ercecan


eyusufatik:
panicking inside the guest is usually the expected behaviour, if the proof run panicks, no ZK proofs can be produced (unless your system accepts zk proofs which doesn't end with successful state)

PR #3231, opened 2026-04-30
fix(bitcoin-da): guard against empty tx.input in get_script to prevent index panic

amathxbt:
## Summary

`get_script` in `crates/bitcoin-da/src/helpers/parsers.rs` accesses `tx.input[0]` without first checking whether the transaction has any inputs. A transaction with zero inputs causes an **index-out-of-bounds panic** rather than a graceful `Err`.

### Root cause

```rust
// crates/bitcoin-da/src/helpers/parsers.rs
fn get_script(tx: &Transaction) -> Result<&Script, ParserError> {
    tx.input[0]          // ← panics if tx.input is empty
        .witness
        .tapscript()
        .ok_or(ParserError::NonTapscriptWitness)
}
```

`parse_relevant_transaction` is called inside `verify_transactions` on every transaction in the completeness proof. While Bitcoin Core rejects zero-input transactions on the P2P layer, the verifier processes data passed directly from the inclusion proof, which is untrusted external input. A crafted or corrupted proof containing a zero-input transaction panics the verifier unconditionally.

### Fix

Replace the direct index access with `.first().ok_or(...)?`:

```rust
fn get_script(tx: &Transaction) -> Result<&Script, ParserError> {
    tx.input
        .first()
        .ok_or(ParserError::NonTapscriptWitness)?
        .witness
        .tapscript()
        .ok_or(ParserError::NonTapscriptWitness)
}
```

An empty-input transaction now returns `Err(ParserError::NonTapscriptWitness)`, which `verify_transactions` already handles via the `if let Ok(parsed_tx) = parse_relevant_transaction(tx)` guard — the blob is simply skipped.

### Impact

- **Severity: Critical**  any completeness proof containing a zero-input transaction panics the verifier (both host and ZK guest).
- No behaviour change for well-formed transactions.

cc @jfldde @eyusufatik @ercecan


eyusufatik:
there is no such thing as 0-input bitcoin transaction

PR #3253, opened 2026-05-02
fix(bitcoin-da): avoid monitored_txs write lock across RPC in monitor_transaction

### LINKED ISSUE #3252: bug(bitcoin-da): monitor_transaction holds monitored_txs write lock across async RPC call
## Description

`monitor_transaction` acquires a write lock on `monitored_txs` and then calls `get_block_count()` while holding the lock, blocking all readers/writers for the full RPC round-trip. Same anti-pattern fixed for `handle_evicted` in PR #3234.

## Expected behaviour
The RPC call should happen before lock acquisition.

amathxbt:
Fixes #3252

## Problem

`monitor_transaction` holds the `monitored_txs` write lock while calling `get_block_count()`, blocking all concurrent readers/writers for a full network round-trip — same pattern as #3234.

## Fix

Move `self.client.get_block_count().await?` to before the write lock is acquired.

### DIFF
--- a/crates/bitcoin-da/src/monitoring.rs
+++ b/crates/bitcoin-da/src/monitoring.rs
@@ -481,9 +481,6 @@ impl MonitoringService {
         kind: MonitoredTxKind,
     ) -> Result<()> {
         let txid = tx.id;
-
-        // Fetch block height before acquiring the write lock to avoid
-        // holding the lock across async network I/O.
         let current_height = self.client.get_block_count().await?;
 
         let mut monitored_txs = self.monitored_txs.write().await;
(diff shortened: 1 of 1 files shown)


crates/bitcoin-da/src/monitoring.rs
```diff
@@ -482,6 +482,10 @@ impl MonitoringService {
     ) -> Result<()> {
         let txid = tx.id;
 
+        // Fetch block height before acquiring the write lock to avoid
```
jfldde: Comment will have little value after the change, we can remove

jfldde (dismissed):

crates/bitcoin-da/src/monitoring.rs
amathxbt: Ok sir

amathxbt:
Done @jfldde @eyusufatik removed the low-value comment at line 485 as requested. Commit: 760dffb95cf2802595a10c3b8207470346396622

jfldde (approved):

jfldde:
Superseded by #3306

PR #3255, opened 2026-05-02
fix(bitcoin-da): avoid monitored_txs write lock across RPC calls in check_transactions

### LINKED ISSUE #3254: bug(bitcoin-da): check_transactions holds monitored_txs write lock across many async RPC calls
## Description

`check_transactions` acquires a write lock at the start and then, per entry, calls `get_transaction`, `get_mempool_entry`/`get_block_count` (via `determine_tx_status`), and `attempt_rebroadcast` while holding the lock — many RPC round-trips per monitoring cycle. Same anti-pattern as PR #3234.

## Expected behaviour
Snapshot txid/status pairs under a read lock, perform all RPCs outside any lock, then apply updates under a single write lock.

amathxbt:
Fixes #3254

## Problem

`check_transactions` holds the write lock for its entire body including multiple RPC calls per entry. This is the same issue fixed for `handle_evicted` in PR #3234.

## Fix

1. Snapshot `(txid, status)` pairs with a read lock.
2. Make all RPC calls without any lock.
3. Apply updates under a single write lock.

### DIFF
--- a/crates/bitcoin-da/src/monitoring.rs
+++ b/crates/bitcoin-da/src/monitoring.rs
@@ -643,18 +643,27 @@ impl MonitoringService {
 
     #[instrument(skip(self))]
     async fn check_transactions(&self) -> Result<()> {
-        let mut txs = self.monitored_txs.write().await;
+        // Snapshot txid→status pairs with a read lock so we don't hold the
+        // write lock across async RPC calls — same pattern as handle_evicted.
+        let entries: Vec<(Txid, TxStatus)> = self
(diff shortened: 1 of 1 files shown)


amathxbt:
Done @jfldde @eyusufatik — ran `cargo +nightly fmt --all`. The method chain in `check_transactions` was collapsed to a single line as rustfmt prefers. Commit: ebf709dbece6aac022a48d7e088be8e4bbf1975c

jfldde:
Superseded by #3306

PR #3261, opened 2026-05-02
fix(bitcoin-da): use saturating_sub(1) to prevent underflow in regtest height fallback

### LINKED ISSUE #3260: bug(bitcoin-da): verifier.rs regtest height fallback underflows when block height is 0
## Description

In `verifier.rs`, the regtest/nightly fallback constructs a default `LatestDaState` with `block_height: block_header.height() - 1`. When `height() == 0` and no prior state exists this underflows: panic in debug builds, `u64::MAX` in release.

## Expected behaviour
Use `saturating_sub(1)` to safely produce `0` at genesis.

amathxbt:
Fixes #3260

## Problem

The regtest height fallback in `verifier.rs` computes `block_header.height() - 1`. At genesis (`height == 0`) this underflows: panic in debug, `u64::MAX` in release.

## Fix

Change to `block_header.height().saturating_sub(1)`.

### DIFF
--- a/crates/bitcoin-da/src/verifier.rs
+++ b/crates/bitcoin-da/src/verifier.rs
@@ -293,7 +293,7 @@ impl DaVerifier for BitcoinVerifier {
                 self.verify_header_chain_regtest(
                     latest_da_state.unwrap_or(&LatestDaState {
                         block_hash: block_header.prev_hash().to_byte_array(),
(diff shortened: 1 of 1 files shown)


jfldde:
Not relevant, `checked_sub` would have made more sense as well

PR #3265, opened 2026-05-02
fix(bitcoin-da): use checked_sub in validate_txs_fee_rate to prevent Amount underflow

### LINKED ISSUE #3264: bug(bitcoin-da): validate_txs_fee_rate panics on Amount subtraction underflow with malformed inputs
## Description

`validate_txs_fee_rate` in `fee.rs` uses the `-` operator on `bitcoin::Amount`. If `output_amount > input_amount` (malformed input), this **panics** in debug builds and wraps silently in release, bypassing the `FeeCalculation` error return.

## Expected behaviour
Use `checked_sub` so an underflow is treated as a fee-validation failure instead of a panic.

amathxbt:
Fixes #3264

## Problem

`bitcoin::Amount` subtraction panics (debug) or wraps (release) when `output_amount > input_amount`.

## Fix

Replace `(input_amount - output_amount) < threshold` with:
```rust
input_amount.checked_sub(output_amount).map_or(true, |diff| diff < threshold)
```
for both the commit and reveal checks.

### DIFF
--- a/crates/bitcoin-da/src/fee.rs
+++ b/crates/bitcoin-da/src/fee.rs
@@ -278,7 +278,7 @@ pub(crate) fn validate_txs_fee_rate(
             .sum();
         let output_amount = commit_tx.output.iter().map(|tx| tx.value).sum();
 
-        if (input_amount - output_amount) < Amount::from_sat(commit_tx.vsize() as u64) {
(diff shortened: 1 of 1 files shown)


jfldde:
Closing as fee handling will be handled by tx-sender in #3206

PR #3269, opened 2026-05-03
fix(verifier): replace coinbase_tx.input[0] direct index with .first().ok_or()

### LINKED ISSUE #3266: fix(verifier): replace coinbase_tx.input[0] direct index with .first().ok_or()
## Summary

In `verify_transactions`, the code directly indexes `coinbase_tx.input[0]` to read the witness data needed for SegWit commitment verification. If an `InclusionMultiProof` carries a malformed coinbase transaction with an empty `input` array, this indexing panics with an index-out-of-bounds error, crashing the verifier process (denial-of-service).

A coinbase transaction is always expected to have exactly one input, but this invariant is never validated before the access, making the verifier susceptible to crafted proofs.

## Fix

Replace `coinbase_tx.input[0]` with `coinbase_tx.input.first().ok_or(ValidationError::InvalidBlock)?`, turning the potential panic into a structured `Err` that the caller can handle gracefully.

**File:** `crates/bitcoin-da/src/verifier.rs`

amathxbt:
Closes #3266

## Problem

In `verify_transactions`, the code directly indexes `coinbase_tx.input[0]` to access the witness data needed for SegWit commitment verification. A crafted `InclusionMultiProof` whose coinbase transaction carries an empty `input` array causes an index-out-of-bounds **panic**, crashing the verifier (denial of service).

## Fix

Replace the direct index with `.first().ok_or(ValidationError::InvalidBlock)?` so that a malformed coinbase is rejected with a structured error instead of a panic.

**File changed:** `crates/bitcoin-da/src/verifier.rs`

### DIFF
--- a/crates/bitcoin-da/src/verifier.rs
+++ b/crates/bitcoin-da/src/verifier.rs
@@ -218,7 +218,10 @@ impl DaVerifier for BitcoinVerifier {
                 let merkle_root =
                     merkle_tree::BitcoinMerkleTree::new(inclusion_proof.wtxids).root();
 
-                let input_witness_value = coinbase_tx.input[0]
+                let input_witness_value = coinbase_tx
+                    .input
+                    .first()
+                    .ok_or(ValidationError::InvalidBlock)?
                     .witness
                     .iter()
                     .next()


jfldde:
Malformed coinbase would be rejected by bitcoin consensus first. Not relevant here as we only work with finalized bitcoin blocks

PR #3271, opened 2026-05-03
fix(monitoring): release chain_state write lock before RPC loop in check_chain_state

### LINKED ISSUE #3268: fix(monitoring): check_chain_state holds chain_state write lock across async RPC loop
## Summary

`check_chain_state` acquires `self.chain_state.write()` before checking whether the tip changed, then—inside the `if tip_changed` branch—performs up to `finality_depth` sequential async `get_block_hash` RPC calls while holding that write lock. This is the same write-lock-across-async-RPC anti-pattern previously fixed in `handle_evicted`, `monitor_transaction`, `check_transactions`, and `handle_reorg`.

Holding the write lock for the full duration of the RPC loop:
1. Blocks every concurrent reader of `chain_state` (latency spikes).
2. Risks deadlock if any of those readers are on the same task driving the monitor loop.

## Fix

1. Snapshot `chain_state.current_tip` and `recent_blocks` under a **short read lock**.
2. Release the lock before the RPC loop.
3. Perform all `get_block_hash` calls against the snapshot **without any lock**.
4. Re-acquire the **write lock only** for the final `current_height / current_tip / recent_blocks` update.

**File:** `crates/bitcoin-da/src/monitoring.rs`

amathxbt:
Closes #3268

## Problem

`check_chain_state` acquires `self.chain_state.write()` before checking whether the tip has changed, then—inside the `if tip_changed` branch—performs up to `finality_depth` sequential async `get_block_hash` RPC calls **while still holding the write lock**. This is the same anti-pattern previously fixed in `handle_evicted`, `monitor_transaction`, `check_transactions`, and `handle_reorg`.

Holding the write lock for the full duration of the RPC loop:
- Blocks all concurrent readers of `chain_state` for the entire round-trip window (latency spikes).
- Risks deadlock when concurrent tasks that need a read lock are waiting on the same executor.

## Fix

1. Snapshot `current_tip` and `recent_blocks` under a **short read lock**, then release immediately.
2. Perform all `get_block_hash` calls lock-free against the snapshot.
3. Acquire the **write lock only once**, at the end, for the `current_height / current_tip / recent_blocks` update.

**File changed:** `crates/bitcoin-da/src/monitoring.rs`

### DIFF
--- a/crates/bitcoin-da/src/monitoring.rs
+++ b/crates/bitcoin-da/src/monitoring.rs
@@ -579,9 +579,22 @@ impl MonitoringService {
         let new_height = self.client.get_block_count().await?;
         let new_tip = self.client.get_best_block_hash().await?;
 
-        let mut chain_state = self.chain_state.write().await;
+        // Take a short read lock only to check whether the tip changed and
+        // snapshot the recent-blocks list needed for reorg detection.
+        // Releasing the lock before the RPC loop prevents blocking all readers
+        // for the duration of up to `finality_depth` round-trips.
+        let (tip_changed, recent_blocks_snapshot) = {
+            let chain_state = self.chain_state.read().await;
+            let changed = new_tip != chain_state.current_tip;
+            let snapshot = if changed {
+                chain_state.recent_blocks.clone()
+            } else {
+                vec![]
+            };
+            (changed, snapshot)
+        };
 
-        if new_tip != chain_state.current_tip {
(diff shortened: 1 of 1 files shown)


amathxbt:
@jfldde @eyusufatik — reviewed this PR. No lint issues found. Ready for your review.

jfldde:
Superseded by #3306

PR #3295, opened 2026-06-05
feat: Toggle rpc listen mode to producer sequencer

ercecan:
# Description
Rpc method to toggle the switch to producer sequencer from listen mode sequencer

### DIFF
--- a/crates/sequencer/src/rpc.rs
+++ b/crates/sequencer/src/rpc.rs
@@ -1,6 +1,6 @@
-use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::Arc;
-use std::time::Instant;
+use std::time::{Duration, Instant};
 
 use alloy_eips::eip2718::Encodable2718;
 use alloy_eips::BlockId;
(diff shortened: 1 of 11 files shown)


crates/sequencer/src/rpc.rs
```diff
@@ -43,6 +45,61 @@ enum BlockReceiveResult {
...
+    pub convert_tx: UnboundedSender<SequencerRpcMessage>,
+    /// Guards against concurrent / repeated conversion attempts.
+    pub started: Arc<AtomicBool>,
+}
+
+/// Number of times the main sequencer is probed before it is considered unreachable.
+const MAIN_SEQUENCER_PROBE_ATTEMPTS: u32 = 5;
+/// Delay between consecutive main sequencer probes.
+const MAIN_SEQUENCER_PROBE_INTERVAL: Duration = Duration::from_secs(2);
+/// Per-probe request timeout when checking the main sequencer.
```
jfldde, line 65: 5 * 2s seems a bit high

jfldde: As in, if we know for a fact that the sequencer is down/crashed, we still need to wait on these 10seconds ?

crates/sequencer/src/rpc.rs
```diff
@@ -466,6 +536,60 @@ impl SequencerRpcServer for SequencerRpcServerImpl {
...
+                debug!("Sequencer: conversion to producer accepted");
+                Ok(())
+            }
+            Ok(Err(reason)) => {
+                conversion.started.store(false, Ordering::SeqCst);
+                Err(internal_rpc_error(format!(
+                    "Conversion to producer failed: {reason}"
+                )))
+            }
+            Err(e) => Err(internal_rpc_error(format!(
```
jfldde, line 587: Is that the path on listen mode sequencer shutdown ? If not, we should reset started to false as well

jfldde: Maybe a semaphore of 1 would have been less error prone instead of having to reset started

jfldde (approved):

PR #3317, opened 2026-07-31
fix(fullnode): pending commitment overwrites & failing pending proofs

exeokan:
# Description
Fixes for the full node DA block handling:

1. Commitments that were pending due to missing predecessors could be overwritten with a conflicting commitment if the sequencer posts them. This creates a divergence between the full node and the light client prover, as the LCP discards a second commitment for the same index.

Added a guard at the top of `process_sequencer_commitment` that checks the pending commitments table as well, and discards a conflicting commitment for the same index.

2. When processing pending proofs, we break the loop on all `ProcessingError`s and don't remove the failing proof from the table. This means unrecoverable proof errors like `PreStateRootMismatch` can starve the remaining pending proofs.

Propagated `HaltingError`s back to the caller (currently not thrown for proofs), removed proofs and continued on `SkippableError`s, break'd the loop in case of `Other`.

## Testing
Added three e2e tests: `PendingCommitmentNotOverwrittenTest`, `ConflictingPendingCommitmentTest`, and `PendingProofDroppedOnPermanentErrorTest`

---

> [!NOTE]
> **Medium Risk**
> Touches full node L1 commitment and proof processing, which affects committed/proven height progression; fixes reduce divergence from LCP but incorrect handling could still stall or mis-order sync.
> 
> **Overview**
> Fixes **full node** DA handling so pending sequencer commitments and batch proofs behave consistently with the light client prover.
> 
> **Pending commitments:** A conflicting commitment at the same index is now **discarded** if another commitment is already pending at that index—the first one seen stays canonical. Pending inserts go through `insert_pending_commitment_if_not_exists` so a pending slot is not overwritten.
> 
> **Pending proofs:** When retrying pending proofs, **skippable** proof errors (e.g. `PreStateRootMismatch`) **remove** the proof from the pending table and continue; **halting** proof errors propagate; other errors still stop the loop.
> 
> Adds e2e coverage: conflicting pending commitments, pending commitment not overwritten by a later conflict, and dropping permanently invalid pending proofs so later proofs can advance proven height.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c50eb8421a71a94e02a99de882c0fe98562092d7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
   **New env vars:**\
   `SEQUENCER_MEMPOOL_ADDITIONAL_VALIDATION_TASKS`: Number of additional parallel transaction validation tasks in the sequencer mempool (default: 4)\
   `SEQUENCER_DRY_RUN_TIME_LIMIT_MS`: Time budget in milliseconds for the sequencer's per-block transaction-selection (dry run) loop (default: 500)
+- fix(fullnode): pending commitment overwrites & failing pending proofs ([#3317](https://github.com/chainwayxyz/citrea/pull/3317))
 ### Added
 - feat: full node state diff rpc ([#3310](https://github.com/chainwayxyz/citrea/pull/3310))
 
--- a/bin/citrea/tests/bitcoin/full_node.rs
+++ b/bin/citrea/tests/bitcoin/full_node.rs
@@ -4257,49 +4257,26 @@ impl TestCase for PendingCommitmentNotOverwrittenTest {
 
         let sequencer_client = sequencer.client.http_client();
 
-        let merkle_root_1 = calculate_merkle_root(
-            &sequencer_client
-                .get_l2_block_range(U64::from(1), U64::from(10))
-                .await?,
-        );
-        let merkle_root_2 = calculate_merkle_root(
-            &sequencer_client
-                .get_l2_block_range(U64::from(11), U64::from(20))
-                .await?,
-        );
-        let merkle_root_3a = calculate_merkle_root(
-            &sequencer_client
-                .get_l2_block_range(U64::from(21), U64::from(30))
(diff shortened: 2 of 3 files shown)


jfldde:
1 Looks good. I think the previous 2 behaviour is in-line with LCP, need to make sure we don't introduce a divergence here

eyusufatik (dismissed):

exeokan:
> 1 Looks good. I think the previous 2 behaviour is in line with LCP; we need to make sure we don't introduce a divergence here

For 2, the changes (propagating `HaltingError::Proof` and removing pending proofs that got `SkippableError::Proof`) are actually on dead paths:

1. We never throw an error of type`HaltingError::Proof`
2. It shouldn't be possible for a proof to be accepted into the pending proof queue and throw `SkippableError` on retry:
- `UnknownL1Hash` is checked before accepting into the proof queue.
- `SequencerCommitmentHashMismatch` and `SequencerCommitmentMissingForProof` are also checked before accepting into the queue. Before this PR, pending commitments could be overwritten, so there was a scenario where a proof passed these checks but failed later; that case shouldn't be possible anymore.
- `PreStateRootMismatch` is the only check we do after we accept it into the pending proof queue. However, it is redundant and shouldn't be reachable with valid proofs. I was only able to trigger this error in tests because we use fake proofs.

Therefore, I don't think this can create a divergence with the LCP. We can keep the changes because they are in line with the semantics of `ProcessingError`. I am also okay with diverting all `HaltingError`/`SkippableError`s to the unreachable branch, or reverting the changes.

jfldde (approved):

PR #3323, opened 2026-08-18
fix(rpc): with log config in debug trace cache

exeokan:
# Description
The `withLog` option in the call tracer config was ignored by `debug_traceTransaction` and `debug_traceBlock*` endpoints. Two issues in the `debug_trace_by_block_number` function that are fixed by this PR:

1. The tracer options we used to fill the cache were serialized with snake_case keys instead of camelCase, so cached traces never included logs.
2. When applying the call config, the frame was not actually mutated, so any logs would be left in place when `withLog` was false.

## Testing
Added `test_call_tracer_with_log` mock e2e test

---

> [!NOTE]
> **Low Risk**
> RPC debug tracing only; fixes incorrect cache keys and log stripping with no auth, state, or consensus impact.
> 
> **Overview**
> Fixes **call tracer `withLog`** handling for `debug_traceTransaction` and block trace paths that reuse the trace cache.
> 
> **Cache fill:** `create_trace_cache_opts` now builds options via `GethDebugTracingOptions::call_tracer(CallConfig::default().with_log())` instead of a hand-built JSON map with snake_case keys, so cached traces actually include logs.
> 
> **Applying config:** `remove_logs_from_call_frame` mutates a single `CallFrame` tree (including nested `calls`) instead of cloning into a temporary vec, so `withLog: false` correctly strips logs from the frame returned to the client.
> 
> Adds mock e2e **`test_call_tracer_with_log`** covering cache hits, `withLog` true/false, and default config.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dda45404f9d670f14d95c799589d775f264a0fa4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/ethereum-rpc/src/trace.rs
+++ b/crates/ethereum-rpc/src/trace.rs
@@ -492,17 +492,6 @@ fn convert_call_trace_into_4byte_map(
 }
 
 fn create_trace_cache_opts() -> GethDebugTracingOptions {
-    // Get the traces with call tracer onlytopcall false and withlog true and always cache this way
-    let call_config = serde_json::to_value(CallConfig {
-        only_top_call: Some(false),
-        with_log: Some(true),
-    })
-    .expect("CallConfig must be serializable");
-    GethDebugTracingOptions {
-        tracer: Some(GethDebugTracerType::BuiltInTracer(
-            GethDebugBuiltInTracerType::CallTracer,
-        )),
-        tracer_config: GethDebugTracerConfig(call_config),
-        ..Default::default()
-    }
+    // Get the traces with call tracer onlytopcall None (false) and withlog true and always cache this way
+    GethDebugTracingOptions::call_tracer(CallConfig::default().with_log())
 }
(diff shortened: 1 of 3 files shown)


crates/ethereum-rpc/src/trace.rs
```diff
@@ -493,10 +493,11 @@ fn convert_call_trace_into_4byte_map(
 
 fn create_trace_cache_opts() -> GethDebugTracingOptions {
```
jfldde, line 494:
```
fn create_trace_cache_opts() -> GethDebugTracingOptions {
    GethDebugTracingOptions::call_tracer(CallConfig::default().with_log())
}
```

exeokan: 👌🏻

jfldde (approved):

eyusufatik (approved):

PR #3324, opened 2026-08-19
fix(sequencer): Keep producing L2 blocks on stale DA tip

### LINKED ISSUE #3320: IBD guard in l1 syncers

jfldde:
# Description

- Fix issue where sequencer was failing to produce L2 block erroring in `setBlockInfo`
- Prevent `SetBlockInfo` while DA is behind after a restart

Fixes #3320 

---

> [!NOTE]
> **Medium Risk**
> Changes core sequencer L1 sync and when system `setBlockInfo` runs; incorrect acceptance logic could stall L1 updates or accept bad tips, but behavior is covered by unit and E2E tests.
> 
> **Overview**
> When the Bitcoin DA node reports a **finalized L1 height below** what the sequencer already folded into the Bitcoin light client (e.g. after restart or while the node is catching up), the sequencer no longer treats that as a fork or feeds those blocks into `setBlockInfo`. It **ignores the regressed tip**, keeps **L2 block production going** without new L1 updates, and **resumes L1 catch-up** once the DA node reports heights at or above the light-client view.
> 
> **`should_accept_finalized_l1_height`** gates DA monitor updates; rejected tips clear the cached finalized block so `produce_l2_block` runs with **no DA payload** (no `setBlockInfo`). Startup uses the same rule when the initial DA tip is stale. **Metrics** add `da_reported_l1_block`, `da_behind`, and `l2_blocks_on_current_l1`; warnings repeat on a 60s interval while behind.
> 
> **Bitcoin E2E tests** cover stale tips **while running** and **on startup** (invalidate/reconsider blocks), asserting L2 continues, scanned L1 height is frozen, and recovery emits ordered `setBlockInfo` calls.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 168d82ee99813aa4758fac74e3e6f7d18cb8b0eb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/sequencer/src/runner.rs
+++ b/crates/sequencer/src/runner.rs
@@ -74,6 +74,9 @@ use crate::utils::recover_raw_transaction;
 /// Maximum number of DA blocks that can be missed per L2 block
 pub const MAX_MISSED_DA_BLOCKS_PER_L2_BLOCK: u64 = 10;
 
+/// How often the warning about a DA node that is behind is repeated
+const DA_BEHIND_WARN_INTERVAL: Duration = Duration::from_secs(60);
+
 /// Returns whether a finalized L1 height should be accepted.
 ///
 /// Rejects anything below our own view of L1, which is the higher of the last height we
@@ -136,6 +139,10 @@ where
     canon_state_tx: mpsc::UnboundedSender<CanonStateNotification>,
     /// Executor for spawning async tasks
     task_executor: TaskExecutor,
+    /// How many L2 blocks were produced on the L1 block currently held by the light
+    /// client. Counted from this process' first block, so after a restart it trails the
+    /// rule enforcer's own counter until the next L1 block arrives.
+    l2_blocks_on_current_l1: u64,
 }
 
 impl<Da> CitreaSequencer<Da>
@@ -200,6 +207,7 @@ where
             backup_manager,
             canon_state_tx,
             task_executor,
+            l2_blocks_on_current_l1: 0,
         })
     }
 
@@ -466,6 +474,7 @@ where
     ) -> anyhow::Result<()> {
         let start: Instant = Instant::now();
         let l2_height = self.ledger_db.get_head_l2_block_height()?.unwrap_or(0) + 1;
+        let previous_used_l1_height = *last_used_l1_height;
         self.fork_manager.register_block(l2_height)?;
         let result = {
             if da_blocks.len() == 1 && da_blocks[0].header().height() == *last_used_l1_height {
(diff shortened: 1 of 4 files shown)


crates/sequencer/src/runner.rs
```diff
@@ -1220,20 +1241,46 @@ where
...
+                        let new_finalized_l1_height = block.header().height();
+                        if !should_accept_finalized_l1_height(
+                            last_finalized_l1_height,
+                            last_used_l1_height,
+                            new_finalized_l1_height,
+                        ) {
+                            missed_da_blocks_count = 0;
+                            last_finalized_block = None;
+                            if !da_behind {
+                                da_behind = true;
```
ercecan, line 1262: I say we add some metrics about this

```rust
#[metric(describe = "The finalized L1 height that the DA node reports")]
pub da_reported_l1_block: Gauge,
#[metric(describe = "1 when the sequencer ignores the DA node, 0 otherwise")]
pub da_behind: Gauge,
#[metric(describe = "The number of L2 blocks made on the current L1 block")]
pub l2_blocks_on_current_l1: Gauge,
```
Keep current_l1_block as the accepted view, as the PR does. Set da_reported_l1_block on both paths. A dashboard then shows the gap directly.

crates/sequencer/src/runner.rs
```diff
@@ -1220,20 +1241,46 @@ where
...
+                                warn!(
+                                    "Ignoring L1 height below our own view of L1, the DA node is \
+                                     likely behind. L2 blocks keep being produced, L1 stays put. \
+                                     New finalized height: {new_finalized_l1_height}, last finalized \
+                                     height: {last_finalized_l1_height}, last used height {last_used_l1_height}"
+                                );
```
ercecan, line 1268: nice little trick to repeat the warning so logs are more visible
```rust
let mut last_warn: Option<Instant> = None;
// on the reject path:
if !da_behind || last_warn.is_none_or(|t| t.elapsed() >= Duration::from_secs(60)) {
    da_behind = true;
    last_warn = Some(Instant::now());
    warn!(...);
}
```

crates/sequencer/src/runner.rs
ercecan, line 1:
```rust
// after produce_l2_block returns
if *last_used_l1_height != previous_used_l1_height {
    l2_blocks_on_current_l1 = 0;
} else {
    l2_blocks_on_current_l1 += 1;
}
SM.l2_blocks_on_current_l1.set(l2_blocks_on_current_l1 as f64);
```
Then alert at 80 % of max_l2_blocks_per_l1

ercecan (changes requested):
Also what do you think about stopping commitment inscribing during `da_behind = true` because if the da node is in reindex or IBD some UTXOs might be missing

jfldde:
> Also what do you think about stopping commitment inscribing during da_behind = true because if the da node is in reindex or IBD some UTXOs might be missing

In that case we can rely on existing retry behaviour as to not introduce too much extra complexity

PR #3328, opened 2026-08-19
fix(config): reject invalid optional env values

### LINKED ISSUE #2727: Most from_env impls do not enforce correct parsing on Some cases
## Summary
On environment config reads for optional configs, we don't enforce the correct-parsability of the env value if it exists, instead we set it to None.

This is different than the toml behaviour.

ayaanoncrypto:
# Description

This change fixes issue #2727 for `RunnerConfig::from_env`. When `SYNC_BLOCKS_COUNT` exists but contains an invalid integer, configuration loading now returns the parse error. Missing `SYNC_BLOCKS_COUNT` still uses the documented default.

## Linked Issues

- Fixes #2727 
## Testing

- `git diff --check` passed.
- - Added a regression test for an invalid present value.
- - Cargo was unavailable in the sandbox, so the focused Rust test could not run locally.
## Docs

No public interface changed. The missing-variable default remains unchanged.

---

> [!NOTE]
> **Low Risk**
> Small change to optional env parsing for one runner field; startup fails fast on bad values instead of running with a hidden default.
> 
> **Overview**
> **Runner env loading** now treats a present but invalid `SYNC_BLOCKS_COUNT` as a configuration error instead of silently using the default (`10`). Missing `SYNC_BLOCKS_COUNT` still falls back to that default.
> 
> A regression test sets `SYNC_BLOCKS_COUNT=not-a-number` and asserts `RunnerConfig::from_env()` fails.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1ed302f2939f4ab0fb8b851838afb9eedf3102b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 # Changelog
 ## [Unreleased]
 ### Changed
-- fix(bitcoin-da): Harden DA monitoring reveal restore ([#3329](https://github.com/chainwayxyz/citrea/pull/3329))
+- fix(config): reject invalid optional environment values ([#3328](https://github.com/chainwayxyz/citrea/pull/3328))
+- fix(config): reject invalid optional environment values ([#3328](https://github.com/chainwayxyz/citrea/pull/3328))
 - fix(rpc): Set archival state in estimate tx expenses ([#3321](https://github.com/chainwayxyz/citrea/pull/3321))
-- fix(rpc): with log config in debug trace cache ([#3323](https://github.com/chainwayxyz/citrea/pull/3323))
 
 ## [v2.7.0](2026-07-29)
 ### Changed
@@ -11,7 +11,6 @@
   **New env vars:**\
   `SEQUENCER_MEMPOOL_ADDITIONAL_VALIDATION_TASKS`: Number of additional parallel transaction validation tasks in the sequencer mempool (default: 4)\
   `SEQUENCER_DRY_RUN_TIME_LIMIT_MS`: Time budget in milliseconds for the sequencer's per-block transaction-selection (dry run) loop (default: 500)
-- fix(fullnode): pending commitment overwrites & failing pending proofs ([#3317](https://github.com/chainwayxyz/citrea/pull/3317))
 ### Added
 - feat: full node state diff rpc ([#3310](https://github.com/chainwayxyz/citrea/pull/3310))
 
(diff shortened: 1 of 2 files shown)


eyusufatik:
are you sure it's only sync_blocks_count?

is there any other config field that this could be happening in?

ayaanoncrypto:
Thanks for checking. The submitted fix targets SYNC_BLOCKS_COUNT because issue #2727 reports that field specifically, and the current diff changes its present-but-invalid value from silent defaulting to an error.

The same silent fallback pattern exists in other optional fields in the current config module, including STORAGE_BACKUP_PATH, DB_MAX_OPEN_FILES, MAX_COMMITMENTS_PER_PROOF, several fee-rate and dry-run settings, mempool reload/lifetime settings, and SEQUENCER_MEMPOOL_ADDITIONAL_VALIDATION_TASKS. Those fields need separate decisions about whether invalid values should fail or retain their optional defaults.

I kept this PR scoped to SYNC_BLOCKS_COUNT to avoid changing unrelated configuration behavior. I can open a follow-up for the remaining fields if the project wants the same fail-fast policy applied consistently.

eyusufatik:
> Thanks for checking. The submitted fix targets SYNC_BLOCKS_COUNT because issue #2727 reports that field specifically, and the current diff changes its present-but-invalid value from silent defaulting to an error.
> 
> The same silent fallback pattern exists in other optional fields in the current config module, including STORAGE_BACKUP_PATH, DB_MAX_OPEN_FILES, MAX_COMMITMENTS_PER_PROOF, several fee-rate and dry-run settings, mempool reload/lifetime settings, and SEQUENCER_MEMPOOL_ADDITIONAL_VALIDATION_TASKS. Those fields need separate decisions about whether invalid values should fail or retain their optional defaults.
> 
> I kept this PR scoped to SYNC_BLOCKS_COUNT to avoid changing unrelated configuration behavior. I can open a follow-up for the remaining fields if the project wants the same fail-fast policy applied consistently.

please also handle others

ayaanoncrypto:
Implemented the requested follow-up. The branch now rejects present-but-invalid values for STORAGE_BACKUP_PATH, DB_MAX_OPEN_FILES, MAX_COMMITMENTS_PER_PROOF, the L1 fee-rate and dry-run settings, mempool reload/lifetime settings, and SEQUENCER_MEMPOOL_ADDITIONAL_VALIDATION_TASKS. Missing variables still use their existing defaults.

Added regression coverage for storage, prover, and sequencer configuration paths, plus a CHANGELOG.md entry. I rebased the branch onto the current nightly base and resolved the CHANGELOG.md conflict. The updated head is 59d55c1.

Local diff checks pass. The workflow has not started yet because the fork workflow requires maintainer approval.

